PR #9, opened 2020-09-23
Configure Logs with Structlog

ghickman:
This uses structlog to get structured logging and middleware-based request logging.  It configures gunicorn's logs to match the application's log format to make spelunking logs easier.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: gunicorn jobserver.wsgi
+web: gunicorn jobserver.wsgi --config=gunicorn.conf.py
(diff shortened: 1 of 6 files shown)


ghickman:
@sebbacon – the old formatter was inserting timestamps, where do the logs go in production?  Some destinations (eg systemd) handle timestamping for us, hence the need for `timestamper`, but I wasn't sure if that was required here.

sebbacon:
They don't go anywhere at the moment; or, rather, they go to stdout and are collected by dokku (so you can see them in `dokku log`; so check timestamps there I guess)

ghickman:
Dokku timestamps logs so these changes are correct since we don't want to double up timestamps in production.

PR #40, opened 2020-09-25
Pin Python

### LINKED ISSUE #35: Pin Python Version
Switch to a pinned Python image so we're pinning our deploy version like our library dependencies.

This will also re-enable caching since various Dockerfile lines (apt-get update, pyenv, etc) break docker caching.

ghickman:
This pins the version of Python we run in Production (and dev via docker-compose) using a Python base image.  We gain better layer caching, fewer layers, and a simpler file by building the image this way.  This also contians changes I made to ensure the docker-compose flow works.  I've also set some options to save on image size (turning off pip caching), avoid caching edge cases (turn off pycs), and avoig buffering issues (turn off buffering).

Note: this removes, what appear to be, the unused dependencies `file` and `docker.io`.  It's been done as an atomic commit to make it easy to revert if that's incorrect.

Fixes #35

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+.direnv
+.flake8
+.github
+.pre-commit-config.yaml
+.venv
+pyproject.toml
+tests
+venv
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -101,3 +101,10 @@ jobs:
         run: |
           source ${{ github.workspace }}/venv/bin/activate
           make test
+
+  lint-dockerfile:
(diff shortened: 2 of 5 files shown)


sebbacon:
Is there a `hadolint` pre-commit-hook we should consider adding too?

sebbacon (approved):

ghickman:
There is, it's a docker image since hadolint is a haskell app.  I'll give it a go and see how fast it is!

ghickman:
hadolint via pre-commit seems good, it's a small image and only installs when the hook is necessary 👍

PR #42, opened 2020-09-28
Remove Apps

ghickman:
This drops the `api` and `frontend` apps.

`frontend` was unused since it wasn't listed in `INSTALLED_APPS` and only contained an `admin.py`.

`api` made sense before we had an HTML interface but is now a little odd!

This PR doesn't included migrations to move the existing data to the new tables (Django would naively drop and recreate those tables without a lot of work to separate the database state and model state changes).  Instead this copies the models to their new home, adds a migration to create them in the database and stops there.

My planned deployment order-of-operations is:
1. Put the site into maintenance mode (using [this](https://github.com/dokku/dokku-maintenance) plugin, which I tested earlier).
1. Run `dumpdata` to get the contents of the `api.*` models.
1. Rename their model paths to `jobserver.*`
1. Use loaddata to populate the moved model tables.

The old tables can be removed in a later PR when we've confirmed everything worked as expected.


sebbacon:
You can make your life easier by just dropping all the data. At the moment, because no-one else is using apart from me and dave from the command line, as long as there's not a whole dependency tree in the middle of running, historic job data is redundant.

sebbacon (approved):

PR #43, opened 2020-09-28
Install and Configure Sentry

### LINKED ISSUE #41: Set up Sentry
Configure in the app and set up a Project in our account.

Release tracking is also useful.  There's a [dokku plugin](https://github.com/jezdez/dokku-sentry-webhook) to handle notifications.

ghickman:
This installs the Sentry SDK and configures it, pulling relevant config from the env.

I've pulled the configuration of Sentry itself into a function to avoid polluting the global settings with random variable names.

Fixes #41

### DIFF
--- a/jobserver/settings.py
+++ b/jobserver/settings.py
@@ -15,6 +15,7 @@ import dj_database_url
 from django.urls import reverse_lazy
 
 from services.logging import logging_config_dict
+from services.sentry import initialise_sentry
 
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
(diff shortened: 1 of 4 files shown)


inglesp (approved):
Code looks fine.

Do we need to think about whether we're risking sending anything sensitive to Sentry?

ghickman:
Sentry's [PII](https://docs.sentry.io/platforms/python/guides/django/data-management/sensitive-data/#personally-identifiable-information-pii) docs suggest this is only User data (eg the Django User model).  For job-server that should be fine since any sensitive information shouldn't be tied to the web/API users.

PR #44, opened 2020-09-28
Paginate List Pages

### LINKED ISSUE #13: Paginate Jobs
We're expecting to have lots of jobs to display so we should paginate them with buttons below the Jobs table.

Some care will need to be taken with how filtering and pagination interact.  Consider what happens if:
* user is on unfiltered page 7, filters to something with <7 pages.
* user is on filtered page 1, clicks next page

ghickman:
This adds pagination to the Job and Workspace list pages.
 
Since we use query args for filtering (via the `url_with[out]_querystring` templatetags) and Pagination also makes use of them I needed to make the templatetags aware of the `page` argument.  This addresses adding/removing a filter while on a page greater than one.  Either use case can result in an erroneously empty page.  Instead I've opted to reset the page arg whenenver a filter is changed.  Note: pagination while using a filter still works as expected.

Fixes #13

### DIFF
--- a/jobserver/templates/job_list.html
+++ b/jobserver/templates/job_list.html
@@ -35,7 +35,7 @@
           <th scope="col">Action</th>
           <th scope="col">Description</th>
           <th scope="col">Started At</th>
-          <th scope="col">Elpased Time</th>
+          <th scope="col">Elapsed Time</th>
           <th scope="col"></th>
         </tr>
       </thead>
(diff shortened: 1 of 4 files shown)


jobserver/templates/job_list.html
```diff
@@ -33,6 +33,7 @@ <h2>Jobs</h2>
           <th scope="col">Action</th>
           <th scope="col">Description</th>
           <th scope="col">Started At</th>
+          <th scope="col">Elpased Time</th>
```
sebbacon, line 36: typo

sebbacon (approved):

PR #45, opened 2020-09-28
Add command to generate dummy data

ghickman:
This adds a simple command to generate some dummy data for local testing.


sebbacon (approved):
You'll pretty soon be wanting to add `needed_by` and  specific status codes

PR #47, opened 2020-09-28
Remove Signal

ghickman:
This removes the post-save signal for Job.

Signals can be tricky to debug.  Since this signal was only for one object it makes sense to make it generally availabe to the model and call it from the save.


inglesp (approved):
Have reviewed 46169892d726c94a04b0b269ab01ee11110e24f7 as requested.  Looks good.

sebbacon (approved):

PR #74, opened 2020-10-02
Add search to Job list page

### LINKED ISSUE #69: Add a simple search
For now, MVP is action_id and job number.

ghickman:
This adds a search box to the Job list page:

Behind the scenes it's searching for either action and ID or just action (if we can't convert the query to an int).

Fixes #69

### DIFF
--- a/jobserver/templates/job_list.html
+++ b/jobserver/templates/job_list.html
@@ -13,10 +13,24 @@
   <h2>Jobs</h2>
 
   {% if user.is_authenticated %}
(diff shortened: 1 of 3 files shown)


sebbacon (changes requested):
Could we change placeholder to "search action_id or job_id" (or add help text next to it to the same effect)?

ghickman:
Yep, done:

sebbacon (approved):

PR #77, opened 2020-10-02
Switch to GitHub Auth

### LINKED ISSUE #62: Set up GitHub OAuth

ghickman:
This switches the HTML portion of the site to use GitHub OAuth for authentication.  It removes the Login form in favour of directly linking to Social Auth's OAuth begin flow View.

Fixes #62


sebbacon (changes requested):
Great. We need to add the new environment variables to the sample environment.

Not sure why we have `environment-sample` and `evironment-sample-docker` mind!

ghickman:
@sebbacon – I've added the env vars to both for now, I'll dig into the docker vs non-docker setup another time!

sebbacon (approved):

PR #81, opened 2020-10-05
Group Jobs

### LINKED ISSUE #60: Group Dependant Jobs
When creating a Job there can be 0 or more dependency Jobs spawned from the user request.  Given a Job tree like so:

`build` -> `process` -> `a_third_thing`

A User creating a Job with `action_id="a_third_thing"` will trigger the `build` and `process` Jobs.

The Job list page currently shows these are 3 separate jobs with nothing to denote they are linked.  However `a_third_thing` will have to wait for the previous two Jobs to finish and a simple status icon doesn't expose _why_ it's waiting very well.

Lets try and group Jobs based on the initially requested Job, ideally showing which Jobs were created by a Human and which were automatically created dependencies.  We might need to create an extra model which has 1-many related Jobs and models "A Human Requested a Job".

ghickman:
This adds the `JobRequest` model to group `Jobs`, it models the interaction of a User requesting a Job, letting the job-runner work out which further jobs are needed.  It also handles exposing the status of a group of Jobs.

The Job list view has been rebuilt around JobRequest objects, including a workaround for filtering by status (this should be simplied back to QuerySet filters once the work on statuses in job-runner is done), and moving from `<table>` to flexbox-based `<div>`s.

Fixes #60


ghickman:
@sebbacon – side note to this PR.  The introduction of the JobRequest model is a little breaking.  There are subsequent PRs coming for rebuilding the Add Job form, and the necessary API changes (which I'll co-ordinate with job-runner).

jobserver/models.py
```diff
@@ -5,6 +5,7 @@
 from django.contrib.auth.models import AbstractUser
 from django.db import models
 from django.urls import reverse
+from first import first
```
sebbacon, line 9: I'm open to being convinced but I'm not especially in favour of requiring a whole new package whose implementation is at most 3 lines of python.  I mean, I agree it's a missing function, it's just I also have a preference for trivial code being completely under our control

jobserver/models.py
```diff
@@ -196,6 +197,10 @@ class JobRequest(models.Model):
 
     created_at = models.DateTimeField(auto_now_add=True)
 
+    @property
+    def completed_at(self):
+        return first(reversed(self.ordered_jobs)).completed_at
```
sebbacon, line 210: I think this could do with a check for `is_complete` first, to be sure no leaves are still running?

jobserver/models.py
```diff
@@ -145,5 +183,110 @@ class JobOutput(models.Model):
...
+
+    @property
+    def is_complete(self):
+        return all(j.is_complete for j in self.jobs.all())
+
+    @property
+    def is_failed(self):
+        return any(j.is_failed for j in self.jobs.all())
```
sebbacon, line 218: This could be confusing as it is possible for a group to have failed items while it is still in progress, but "failed" implies "completed".

We do want action graphs to continue on failure where possible, so I think the fix is for this property to be called something like `has_any_failures`, and the state shown the user to distinguish between "in progress (no failures)" and "in progress (with failures)"

jobserver/views.py
```diff
@@ -26,39 +26,59 @@ class JobDetail(DetailView):
...
+        the job-runner
+        """
+        status = self.request.GET.get("status")
+
+        if not status:
+            return self.object_list
+
+        status_lut = {
+            "completed": lambda r: r.is_complete,
+            "failed": lambda r: r.is_failed,
```
sebbacon, line 52: The same point applies here about possibly wanting to distinguish between finished-with-failures and in-progress-with-failures.

jobserver/views.py
```diff
@@ -26,39 +26,59 @@ class JobDetail(DetailView):
...
+            "pending": lambda r: r.is_pending,
+        }
+        func = status_lut[status]
+        return list(filter(func, self.object_list))
+
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context["statuses"] = ["completed", "in-progress", "pending"]
+
+        # FIXME: HACKY FIX
```
sebbacon, line 62: Can we provide a clearer hint what the fix will be and when?

sebbacon (changes requested):
Great, looking forward to this! Some small suggestions in the code comments. 

Some points about the visual side:

* Could you visually indent jobs so it's clear what group they relate to>
* Could we change the word "toggle" to "show individual jobs / hide individual jobs"
* I think the requesed job's action_id needs to appear in the layout
* Is there a missing header? The job with id `6` has a right-aligned `4 hours ago` and it's not clear what it means

I think we probably need to special-case where a JobRequest has exactly one job because no toggling will be necessary?

jobserver/models.py
ghickman: [10](https://github.com/hynek/first/blob/master/first.py#L69-L78) if you ignore the whitespace and include the declaration!

There was talk about adding it to the stdlib, but it's not implmenented in C so didn't fit in itertools.  As a project it's very stable, the code hasn't had any updates since 2012.

I've always treated it as a default package, I end up reaching for it in everything I work on because it's handles a bit of boiler plate (try/catch on IndexError), and its `key` kwarg makes it useful to filter and index in a single function call.

jobserver/models.py
```diff
@@ -74,6 +74,35 @@ def __str__(self):
...
+    def is_complete(self):
+        return self.status_code is None and self.completed_at
+
+    @property
+    def is_dependency_failed(self):
+        return self.status_code == 7
+
+    @property
+    def is_failed(self):
+        return self.status_code and self.status_code not in (6, 7, 8)
```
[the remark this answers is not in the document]
ghickman: Agreed, but my expectation is this property will disappear with job-runner owning more status information!  How do you feel about leaving this as it is?  Are we inviting future tech-debt?

jobserver/views.py
ghickman: Fixed in 533ebef.

jobserver/models.py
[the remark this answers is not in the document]
sebbacon: Even if they're defined as variables in the function I think it's better. I agree we'll be replacing it, but might be 2 - 4 weeks or more, and someone reading this won't know from the numbers what they mean.

jobserver/models.py
sebbacon: `¯\_(ツ)_/¯`

jobserver/models.py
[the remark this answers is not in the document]
ghickman: Fixed and rebased into 98ca67e

jobserver/models.py
ghickman: Fixed and rebased into 34fcc16.  I also caught that there will be a window when `JobRequest` has no `Job`s so checking that too.

jobserver/models.py
ghickman: Fixed and rebased 76bf7e0.  I've kept it simple for now by only defining failed as "all jobs have completed with 1 or more failures".  I'll revisit this after this is merged to add the "in progress with failures" state.

ghickman:
> * Could you visually indent jobs so it's clear what group they relate to>
> * Could we change the word "toggle" to "show individual jobs / hide individual jobs"
> * I think the requesed job's action_id needs to appear in the layout

I think all of these are addressed now:


> Is there a missing header? The job with id 6 has a right-aligned 4 hours ago and it's not clear what it means

The sub-list of Jobs had no headers before, headers makes things a little easier to read!

sebbacon (approved):

PR #85, opened 2020-10-06
New add job

ghickman:
⚠️ Blocked by #81 ⚠️ 

This rebuilds the Add Job flow into a two step process:
1. Pick or create a Workspace
2. Create a JobRequest
2a. Create a Job under the JobRequest

#### Pick or create a Workspace
The Workspace list has a max height with automate overflow (so scrollbars when it overflows) and some border shenanigans to cover:
* no Workspaces

* only a few Workspaces

* lots of Workspaces

#### Create a JobRequest
Since a JobRequest models "A Human creating a Job" we show Users the Job form and build up extra bits behind the scenes:

Visiting this page with an invalid Workspace will redirect the user to the first with an error message:


tests/jobserver/test_views.py
```diff
@@ -76,26 +88,70 @@ def test_joblist_filter_by_workspace(rf):
...
     request = rf.get("/?q=99")
     response = JobList.as_view()(request)
 
     assert len(response.context_data["object_list"]) == 1
-    assert response.context_data["object_list"][0] == job2
+    assert response.context_data["object_list"][0] == request2
+
+
+@pytest.mark.django_db
+def test_jobrequestcreate_redirects_to_new_job(client):
```
sebbacon, line 119: this test name is wrong?

tests/jobserver/test_views.py
```diff
@@ -76,26 +88,70 @@ def test_joblist_filter_by_workspace(rf):
...
+    assert response.status_code == 200
+    assert response.redirect_chain == [(reverse("job-select-workspace"), 302)]
+
+    messages = list(response.context["messages"])
+    assert len(messages) == 1
+    assert str(messages[0]) == "Unknown Workspace, please pick a valid one"
+
+
+@pytest.mark.django_db
+def test_jobrequestcreate_success(rf):
```
sebbacon, line 132: Could we add a hint about `rf` that's somehow discoverable somewhere with search by a confused future me?

tests/jobserver/test_views.py
```diff
@@ -76,26 +88,70 @@ def test_joblist_filter_by_workspace(rf):
...
+    user = UserFactory()
+    workspace = WorkspaceFactory()
+
+    data = {
+        "action_id": "twiddle",
+        "backend": "test",
+        "callback_url": "test",
+    }
+
+    request = rf.post("/", data)
```
sebbacon, line 142: Can we do something here to indicate that the slash is meaningless? e.g. a `MEANINGLESS_URL` constant or better name?

jobserver/models.py
```diff
@@ -145,5 +183,123 @@ class JobOutput(models.Model):
...
+        "User",
+        null=True,
+        blank=True,
+        on_delete=models.CASCADE,
+    )
+    workspace = models.ForeignKey(
+        "Workspace", on_delete=models.CASCADE, related_name="job_requests"
+    )
+
+    created_at = models.DateTimeField(auto_now_add=True)
```
sebbacon, line 206: Not sure where this comment should be filed but we've agreed *all* `Jobs` should be the responsibility of the `job-runner`, so we should model action_id and backend within `JobRequest`.

jobserver/forms.py
```diff
@@ -9,14 +9,18 @@ class JobCreateForm(forms.ModelForm):
...
             "force_run",
             "force_run_dependencies",
             "action_id",
             "backend",
             "callback_url",
         ]
         model = Job
+        widgets = {
+            "action_id": forms.TextInput(),
+            "backend": forms.TextInput(),
```
sebbacon, line 21: Can we make this a choice field? (`tpp`, `emis`, `all` -- it can default to `all` and things will work as-is)

sebbacon (approved):
Great, some comments that we've discussed - approving this PR though as I think you said you'll address these ones elsewhere

jobserver/forms.py
ghickman: Fixed.  `JobRequest` now has a choices driven text field with the model introducing the `all` key (and cleaning it such that `backends` is always a list), so the view can iterate over `backends` building up the required number of `JobRequests`.

tests/jobserver/test_views.py
ghickman: yep! 🤦‍♂️

tests/jobserver/test_views.py
ghickman: Done.  Although haivng done it, it looks a bit weird to be commenting on a core Django object?  Lets try it out for a bit and see if it continues to be weird 🙂

jobserver/models.py
ghickman: Done.  The UI form view no longer creates a `Job` instance.  There's some UI tidy up to do around no jobs existing in the drop down of the list page but I'll clean that up in a subsequent PR.

ghickman:
`JobRequest` create page using a dropdown for backends:

PR #87, opened 2020-10-08
Add JobRequest Tests

ghickman:
This adds tests for JobRequets custom methods.

### DIFF
--- a/tests/jobserver/test_models.py
+++ b/tests/jobserver/test_models.py
@@ -197,6 +197,319 @@ def test_job_status_unstarted_is_pending():
 
 
 @pytest.mark.django_db
+def test_jobrequest_completed_at_no_jobs():
+    assert not JobRequestFactory().completed_at
+
+
+@pytest.mark.django_db
+def test_jobrequest_completed_at_success():
+    request = JobRequestFactory()
+
+    job1 = JobFactory(request=request, completed_at=timezone.now())
+    job2 = JobFactory(request=request, completed_at=timezone.now())
+
+    job1.needed_by = job2
+    job1.save()
+
+    assert request.completed_at == job2.completed_at
+
+
+@pytest.mark.django_db
+def test_jobrequest_completed_at_while_incomplete():
+    request = JobRequestFactory()
+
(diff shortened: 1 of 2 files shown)


tests/jobserver/test_models.py
```diff
@@ -196,6 +196,335 @@ def test_job_status_unstarted_is_pending():
...
+    job1 = JobFactory(request=request, completed_at=timezone.now())
+    job2 = JobFactory(request=request, started_at=timezone.now())
+
+    job1.needed_by = job2
+    job1.save()
+
+    assert request.status == "In Progress"
+
+
+@pytest.mark.django_db
```
sebbacon, line 523: dupe of `test_jobrequest_status_fall_through`

tests/jobserver/test_models.py
```diff
@@ -196,6 +196,335 @@ def test_job_status_unstarted_is_pending():
...
+    job1.needed_by = job2
+    job1.save()
+
+    job2.needed_by = job4
+    job2.save()
+
+    job3.needed_by = job4
+    job3.save()
+
+    assert request.runtime
```
sebbacon, line 458: Seems a pity not to check the runtime value in some sense here!

sebbacon (approved):
Bravo! Couple of nits...

tests/jobserver/test_models.py
ghickman: I did consider it!  `Runtime` has it's own tests which check construction and output are as expected so testing that here felt like a duplication of effort.

sebbacon: Then all the setting of started_at and completed_at seem redundant? Might as well lose a few lines...

ghickman: I could definitely cut it down a bit, but we're only generating a `Runtime` once a `JobRequest` has both a started_at and completed_at.

tests/jobserver/test_models.py
ghickman: Yep, the catch all in `JobRequest.status` returns `"Pending"` (rather than an implicit `None`) just in case we ever hit that point.

PR #90, opened 2020-10-09
List Actions

### LINKED ISSUE #61: Provide Options for Adding Jobs
When adding a Job we currently provide free text fields for some options which should be derived from the Workspace's `project.yml` config.

The view will need to find, load, and parse the relevant `project.yml` to do this.

Two example configs:
* https://github.com/opensafely/households-research/blob/master/project.yaml
* https://github.com/opensafely/carehomes-research/blob/master/project.yaml

Job Runner's YAML parsing function is [here](https://github.com/opensafely/job-runner/blob/5893b48c275afd06cb016ddd52f308fbc9523668/jobrunner/project.py#L270-L345).

ghickman:
This provides available Actions when creating a `JobRequest`, sourced from the configured repo and branches `project.yaml`.

I haven't pulled in `project.py` from job-runner for this yet, so there's no validation because I'd like to stop a User reaching this form when a `project.yaml` is invalid instead.

EDIT: updated screenshot for new UI.

Fixes #61


ghickman:
Given the discussion earlier, I'd rather not have extra scopes creeping into our User's authorizations so I'll hold off on this PR until we've got #91 sorted.

It can still be reivewed since the only thing which should change is where the `token` is retrieved from.

sebbacon (approved):

ghickman:
@sebbacon – this is good to re-review now.

sebbacon (approved):

PR #94, opened 2020-10-09
Link Workspace to User

ghickman:
This links a Workspace to a User via an FK now that #77 is done.  There's some work to migrate Production's [currently] 2 Workspaces over but I think that can be done manually once the deploy is done.


sebbacon (approved):
Great. Just to note we also want to default to showing the current user's jobs when logged in, at some point

ghickman:
@sebbacon – captured in #98

Issue #91, opened 2020-10-09
Set up PAT


ghickman:
Rather than get access to every repo a User has access to via the OAuth `repo` scope or make researchers jump through the select-each-repo-you-care-about flow of GitHub Apps lets configure a single PAT with the right scopes.

We'll need the `repo` scope to read code within a repo so the user which owns this token should probably be a new one, and behind 2FA.

The `repo` scope also gives access to list repos of a Team so we should be able to keep the token limited to just that scope.

bloodearnest:
So, I think where we are with this as that a PAT won't work, as it has the same wide-open access that the oauth app approach did - the PAT would grant access to all of the PAT's user's orgs.

George was going to file a support ticket to ask how to get a credential that allows OpenSAFELY org only read access to private repos.

A final alternative would be to create a new user in OpenSAFELY, and give them read-access to all the study repos, and use that account to generate a PAT or an oauth token to use in job-server.

ghickman:
I'm chatting with support on what they think the best approach is for this.  My gut feeling is we'll end up with the final alternative @bloodearnest suggested, but I'll let the support request play out to see if there's something we've missed.

ghickman:
Support are saying we can get what we want with a GitHub App:

>> To answer both of these: we'd like an admin of the OpenSAFELY Org to be able to generate a secret token (much like a PAT but for the Org, rather than a User), which we'd use as the Auth token to access the API from our App.

> This is currently possible with GitHub Apps. After you install a GitHub App on the OpenSAFELY Org (this step of installing needs to be performed by a human, but happens only once), you can create an token for that installation:

> https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps/#authenticating-as-an-installation

> If at the installation step the admin installed the App on all repositories owned by the organization, then the App will have access to all current and future repositories of the organization.

> When you create an installation token, you can specify which subset of repositories the token will have access to.

Which suggests I should tackle #89 soon.

bloodearnest:
Hmm, so if our options are a) specific repos or b) all repos, that is not great either. If we do a), and reading the link they provided, it looks like we would need to re-install the app each time we added a new study repo? If we do b), it gives it access to the private infra repos.

Can the App be restricted to read-only on the repos? Is that a separate set of permission we grant at install time? If so, b) might be acceptable. But also means it could read any secrets in any repos too (which we are talking about doing, e.g. generic docker RO login and stata license stuff).

[...]

Issue #88, opened 2020-10-08
Suggestion: JobRequests should include specific git-sha


evansd:
Now that we have github auth, the job-server can determine the sha of the branch head which the user wants to run. This can then be included on the JobRequest.

I think the advantages of making it clear and deterministic which version of the code will run outweigh the convenience (in some cases) of automatically picking up code changes pushed after the job is requested.

We can still support modifying the commit for dependant jobs which have been scheduled but not yet run if we want to. (It should just be enough to modify the commit field on the JobRequest, which will then get synced back to the job-runner.) But this would be something explicitly requested by the user rather than implicit behaviour.

ghickman:
Agreed.

I think we might want to track both the branch and SHA though?  Then give the user a button to bump a JobRequest's SHA to HEAD for the Workspace's Branch?

ghickman:
Relatedly, should we move `branch|sha` onto `JobRequest` and leave `Workspace` to model a Repo?

evansd:
I think that would make sense. The way I see it, each JobRequest has a `sha` which is set when its first created but may change later. When the job-runner kicks off a job it checks the current `sha` on (its copy of) the JobRequest and uses that, storing that `sha` on the Job.

Does that make sense?

ghickman:
Definitely, and it tracks well with how we're splitting up the responsiblities of `JobRequest` and `Job` too.  I'm in the process of moving `branch` from `Workspace` to `JobRequest` (#95), which we can use to define `JobRequest`s `sha`.

Shall we say, at least for now, that once a `JobRequest` has been acked (ie has dependant `Job`s), we'll consider its `branch` (and thus `sha`) locked?  Keeps things simple, etc.

sebbacon:
This all makes sense.

Bearing in mind our priority use-case at the moment is to support iterative development, I think it's also correct that a Workspace doesn't care about `sha` and the default behaviour for a new `JobRequest` should be for it to be the current tip of the ref.

Later, as discussed, we will want to think about "production" mode at some point (when the job-runner takes into account the `sha` when deciding whether an action needs (re)running) and we could perhaps require a tag refspec rather than a branch in such cases.

PR #95, opened 2020-10-09
Move Branch to JobRequest

ghickman:
In #88 we've identified the need for a specific SHA to work from in job-runner which means the branch for a JobRequest is more specific to it than a `Workspace` (which is modelling "a work area on the server").  As a start on moving towards achieving that this moves `branch` from `Workspace` onto `JobRequest` with the expectation that once #91 is done we can list branches of the `Workspace`'s Repo on the create `JobRequest` page, saving the branch name and SHA.


sebbacon (commented):
I'm not sure this is right. Having a `sha` on `JobRequest` doesn't necessarily imply the branch name should also belong there.

My original reason for putting it on the Workspace is for iterative development on a dev branch.

If I am tweaking and running, I want to re-use dependent steps related to a conceptually coherent thing viz "the stuff I'm working on right now".  If I switch branches, I may still want to refer to outputs from old branches; the idea was further down the line there would be tooling to copy outputs between workspaces (so you don't necessarily have to run them again).

If a Workspace is no longer tied to a specific branch this workflow would no longer make sense.

I'm not 100% tied to that model but I've not put the time into thinking through the alternatives - so a discussion needed!



 that two people might be iteratively developing different branches, and discussing outputs with each other.

ghickman:
Closing after discussion, `sha` still needs to exist on `JobRequest` but `branch` should stay on `Workspace`.

PR #101, opened 2020-10-13
JobRequest Shim

ghickman:
This adds the two parts we need to allow job-runner to keep working with job-server:
* create an initial `Job` when a `JobRequest` is created.
* an API shim to hide the existance of `JobRequest`s.

#### Initial Job Creation
The Create `JobRequest` form now creates the initial `Job` for a request.

#### API Shim
job-runner doesn't know about `JobRequest`s so we need the API to continue working externally as it currently does.

This set of changes updates the `JobSerializer` to pull fields from their new locations.  It handles creating `Job`s when no `job_request_id` has been set.  It makes use of the `needed_by_id` field to walk the dependency tree to find the original `Job` for the request which we know will have a `JobRequest` and assigns that to each new `Job`.

---

It's expected that this set of changes will temporarily remove the need for changes to be made to job-runner and unstick deploys to job-servers production.


jobserver/serializers.py
```diff
@@ -22,57 +23,79 @@ class Meta:
...
         ]
         read_only_fields = ["created_at"]
 
 
-class JobSerializer(serializers.HyperlinkedModelSerializer):
+class JobSerializer(serializers.Serializer):
+    """
+    Serializer for a Job as job-runner expects it to look
+
+    Now that various fields have moved off Job and Jobs are grouped by
```
sebbacon, line 36: Perhaps the serializer should be called `JobShimSerializer`?

The comment could be a bit more explicit, e.g. "a shim to hide our current implementation of JobRequests from the job server, until it is implemented there"?

jobserver/serializers.py
```diff
@@ -22,57 +23,79 @@ class Meta:
...
+        # to attach new Jobs to.
+        try:
+            dependency = Job.objects.get(pk=validated_data["needed_by_id"])
+        except Job.DoesNotExist:
+            raise ValidationError("needed_by_id must be a valid Job ID")
+
+        while dependency.needed_by is not None:
+            dependency = dependency.needed_by
+
+        job_request = dependency.request
```
sebbacon, line 76: Reading this, I got confused, and it makes me realise `request` is perhaps a bad name for this relation (`request` being a particularly overloaded term in Django...)

I wonder if the relation should actually be `job_request`, for clarity... (sorry, should have thought of this in an earlier PR!)

sebbacon (approved):
Lovely to see all the tests being filled out!

Looks great, couple of nits

jobserver/serializers.py
ghickman: Arg, I know, I keep flip flopping on this.  I really dislike overly specific model or relation names (job.job_foo.job_bar, etc), but you're right about `request` here, just reading the snippet of code I wrote yesterday is already confusing, a good sign it needs to change!

ghickman:
Updated with better docs and more tests to cover missing lines, particularly those around walking the `Job` dependency tree.

PR #103, opened 2020-10-15
Ignore job_request Missing

ghickman:
We ignore the job_request key as a defensive measure.  It shouldn't exist in the first place but we want to make sure job-runner can't update it (at the moment).

### DIFF
--- a/jobserver/serializers.py
+++ b/jobserver/serializers.py
@@ -78,9 +78,11 @@ class JobShimSerializer(serializers.Serializer):
 
         job_request = dependency.job_request
 
-        # remove the job_request key since job_request is going to replace the
-        # values here
-        validated_data.pop("job_request")
+        # remove the job_request key since the local job_request value is
+        # going to replace any possible values here.
+        # Note: This field is added by DRF travelling field sources which point
+        # to the related job_request instance (eg source=job_request.backend)
+        validated_data.pop("job_request", None)
 
         outputs = validated_data.pop("outputs", [])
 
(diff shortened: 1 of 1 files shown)


sebbacon (commented):
Wouldn't an assertion or other exception-raising behaviour be better?

ghickman:
Both would raise an exception which isn't the desired behaviour here.  It's very unlikely the job-runner will have a `job_request` key _currently_ (I know for a fact it doesn't on master as of writing this!).

However, it will, at a later date, as we move towards the desired API (job-runner saying to job-server "please create/update `Job`s for this `JobRequest`").  We want to control that migration on both sides.  This change makes sure job-server is in control of `JobRequest`s (for the time being!).  It also decouples the move to job-runner-knowing-about-job-requests from job-server, ie a new version of job-runner can be deployed without requiring job-server be updated.

Once job-runner knows about and is ready to use `JobRequest`s, this and the dependency walking parts of these methods will go away.

sebbacon (commented):
I suppose what I don't get is that this is a shim for the old implementation. We're going to have to do a little more work in the jobserver when the jobrunner is finally ready to start submitting request ids, i.e. remove this shim. If the jobrunner starts submitting job request ids before the server is completely configured to handle them, shouldn't that be an error?

ghickman:
My plan with this had been to ignore this key so job-runner could get closer to the intended API end goal.  But you're right, particularly the point about having to do more work on this side of things before we can swtich anyway.  I'll switch this to an assert.

ghickman:
@sebbacon – updated to use asserts.

sebbacon (approved):

PR #119, opened 2020-10-19
Grid Job List

ghickman:
This rebuilds the per-JobRequest list of Jobs using CSS grid so we can get table-like functionality without the problems using `<table>` brings (namely fixed-width tables force equal-width columns).  It also drops some columns on smaller devices to make the UI more readable there.

The JobRequest data is still not great on phone sized screens (small/medium in Bootstrap terms):

But I need to have a think on how to make that more responsive since I can't apply a normal Grid to it.

### DIFF
--- a/jobserver/templates/job_list.html
+++ b/jobserver/templates/job_list.html
@@ -2,6 +2,7 @@
 
 {% load humanize %}
 {% load querystring_tools %}
+{% load runtime %}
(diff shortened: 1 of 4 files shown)


sebbacon (approved):
Great! Maybe open a new issue for the remaining mobile issues?

PR #125, opened 2020-10-20
Add a Job directly from a Workspace

### LINKED ISSUE #118: Adding a jobrequest from a workspace view should preselect that workspace
As an adjunct to #117:

When I select a workspace, I want to feel like I am doing stuff "there": this means that, when on a Workspace view, "add a job" preselects the current workspace, rather than having to select it each time.

ghickman:
This adds a button to the Workspace detail page to take the User to the Add Job page, skipping the select Workspace page.

Fixes #118

### DIFF
--- a/jobserver/templates/workspace_detail.html
+++ b/jobserver/templates/workspace_detail.html
@@ -21,6 +21,10 @@
 
     <h3 class="text-center">Tools</h3>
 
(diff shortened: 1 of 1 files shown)


sebbacon (approved):
Well that was easy

PR #136, opened 2020-10-22
More Tests

ghickman:
This is a grab bag of tests after I noticed coverage had dipped after some previous PRs failed to test code they added (oops!).

Each commit addresses a roughly single feature but I suspect viewing the entire diff should be fine since shouldn't be any overlapping chunks between commits.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,7 @@ exclude = '''
 branch = true
 omit = [
   "jobserver/asgi.py",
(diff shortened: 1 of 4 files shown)


bloodearnest (approved):
LGTM.

It'd be interesting to have a discussion around test fixtures, I'd like to understand these Factories are doing, and your thoughts on pytests fixtures.

PR #137, opened 2020-10-22
Fix JobDetail

ghickman:
[Sentry](https://sentry.io/organizations/ebm-datalab/issues/1952687026/?project=5443358&query=is%3Aunresolved)

This fixes JobDetail so it can render older style Jobs which are linked directly to Workspaces.


bloodearnest (approved):
MEANINGLESS_APPROVAL

PR #143, opened 2020-10-26
List Repos and Branches

### LINKED ISSUE #92: List Repos when Creating a Workspace
Similar to #61 we can now list the Repos available when a User creates a workspace using the API.

ghickman:
This enhances the select-or-create Workspace page's Repo and Branch fields to be driven by live GitHub data.  To avoid O(N) requests to GitHub's API I've used their GraphQL (v4) API to create a list of repo-with-branches dictionaries from which the form is generated.

The branch selector is driven by some JavaScript so we don't end up with a giant list or the set of all branches.  The data driving this is loaded into the template to avoid any further API lookups.

Fixes #92

### DIFF
--- a/jobserver/github.py
+++ b/jobserver/github.py
@@ -5,19 +5,8 @@ from furl import furl
 
 
 BASE_URL = "https://api.github.com"
-
 TOKEN = os.environ["GITHUB_TOKEN"]
-
-
-def _build_headers(accept=None):
-    if accept is None:
-        accept = "application/vnd.github.v3+json"
-
-    return {
-        "Accept": accept,
-        "Authorization": f"token {TOKEN}",
-        "User-Agent": "OpenSAFELY Jobs",
-    }
+USER_AGENT = "OpenSAFELY Jobs"
 
 
 def get_file(repo, branch):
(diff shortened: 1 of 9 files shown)


jobserver/github.py
```diff
@@ -31,8 +20,58 @@ def get_file(repo, branch):
...
+    Get Repos with their branches from the OpenSafely Researchers Team
+
+    This uses the GraphQL API to avoid making O(N) calls to GitHub's (v3) REST
+    API.
+    """
+    query = """
+    query {
+      organization(login: "opensafely") {
+        team(slug: "researchers") {
+          repositories(first: 100) {
```
sebbacon, line 45: Isn't this `100` restriction going to get old very quickly?

sebbacon (approved):

jobserver/github.py
ghickman: Yes, I think this gives us ~50 repos (IIRC 54) currently and I've yet to work out pagination with GraphQL.  But my hope was there would be enough breathing room to tackle the problem before it was an issue.

Thank you for the reminder though, I meant to make a ticket about it!

ghickman: Ah, GitHub says 58 [here](https://github.com/orgs/opensafely/teams/researchers/repositories).

PR #165, opened 2020-10-29
Add Site Status

### LINKED ISSUE #146: Add server health checks
To give people comfort that the server is running, perhaps in a little status bar / box at the top, we could add:

* queue size stats (unacked, started)
* date of most recent ack

We could add an alert icon there which appears if the unacked queue size > 0 and the date of most recent ack > 5 mins ago?

There are various other ideas we've had where the server can include health status in its posts, but I think this is the lowest-ceremony useful thing?

ghickman:
This adds a site status banner detailing the last time something touched the API and the current Queue details:

Note: API usage tracking is currently only request path based, we need to add and filter to the job-runner user-agent.

Note2: We can't currently display running jobs as distinct from Acked since `Job.save()` sets `started_at` when saving a Job with `started=True`.  We can add this in once the job-runner is driving the value of `Job.started_at`.

Fixes #146


jobserver/templates/base.html
```diff
@@ -50,6 +50,23 @@
       </div>
     </nav>
 
+    <div class="d-flex justify-content-around align-items-center border-bottom mb-2">
+      <small>
+        Runner Last Seen: {{ site_stats.last_seen }}
+
+        {% if site_stats.show_warning %}
+        <span class="text-warning">⚠️</span>
```
sebbacon, line 58: Can we add a "tool tip" here saying something like "job runner not seen for at least 5 mins" or similar?

sebbacon (approved):

Issue #175, opened 2020-11-02
Job Detail Page


ghickman:
I added the Job detail page a while ago but forgot to modify the URL name used by the `View` button in the main list views (Dashboard & Job List) so we've been using the HTML API view.

Last week I realised how much we rely on that page so now I'd like to make sure the Job detail page is at minimum a complete replacement, ideally a superset.

@sebbacon & @bloodearnest (as the biggest debug-via-the-web users currently) – is there anything else you'd like to see on a Job detail page?  Any particular gropuing of information?

PR #177, opened 2020-11-02
Job detail 2.0

ghickman:
This adds some structure and group to the Job detail page ready for further improvments as we find them.

There are no tools yet and dependencies are broken but this was mostly around rebuilding the HTML document structure rather than adding in more functionality.

#### Small

Refs #175


sebbacon (approved):
Looks good. If dependencies are currently broken, can we say so, to make this clear to the user?

ghickman:
@sebbacon – I've removed the dependencies section for now.  We're not saving to `Job.needed_by` which I need to dig into so I'll revist that section later.

PR #180, opened 2020-11-03
Fix Statuses

ghickman:
This rebuilds the `Job.is_*` and `Job.status` properties to correctly report the state of a Job.  A new `.is_finished` status has been added, with `.is_completed` and `.is_failed` building on top of it.

The JobRequest.is_*` and `JobRequst.status` properties have been updated to reflect changes in their relevant Job counterparts.

### DIFF
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -16,6 +16,13 @@ from .runtime import Runtime
 logger = structlog.get_logger(__name__)
 
 
+# TODO: remove when job-runner is driving state updates
+STATE_SUCCESS = 0
+STATE_DEPENDENCY_NOT_FINISHED = 6
+STATE_DEPENDENCY_FAILED = 7
+STATE_DEPENDENCY_RUNNING = 8
+
+
 class Workspace(models.Model):
     created_by = models.ForeignKey("User", null=True, on_delete=models.CASCADE)
 
@@ -75,10 +82,17 @@ class JobQuerySet(models.QuerySet):
         return self.exclude(started=True, completed_at=None)
 
     def failed(self):
+        ignored_states = [
+            STATE_DEPENDENCY_FAILED,
+            STATE_DEPENDENCY_NOT_FINISHED,
+            STATE_DEPENDENCY_RUNNING,
+            STATE_SUCCESS,
+        ]
+
         return (
             self.exclude(completed_at=None)
             .exclude(status_code=None)
-            .exclude(status_code__in=[6, 7, 8])
+            .exclude(status_code__in=ignored_states)
         )
 
     def pending(self):
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -5,6 +5,7 @@ import requests
 import structlog
 from django.contrib.auth.models import AbstractUser
 from django.db import models
+from django.db.models import Q
 from django.urls import reverse
 from django.utils import timezone
 from furl import furl
@@ -64,13 +65,28 @@ class Workspace(models.Model):
 
 class JobQuerySet(models.QuerySet):
     def completed(self):
-        return self.filter(completed_at__isnull=False)
+        return (
+            self.filter(started=True)
+            .exclude(completed_at=None)
+            .exclude(status_code=None)
+        )
 
-    def running(self):
-        return self.filter(started_at__isnull=False, completed_at=None)
+    def finished(self):
+        return self.exclude(started=True, completed_at=None)
+
+    def failed(self):
+        return (
+            self.exclude(completed_at=None)
+            .exclude(status_code=None)
+            .exclude(status_code__in=[6, 7, 8])
+        )
 
     def pending(self):
-        return self.filter(started_at=None, completed_at=None)
+        qwargs = Q(status_code=None) | Q(status_code__in=[6, 8])
+        return self.filter(qwargs, started=False)
+
+    def running(self):
+        return self.filter(started=True, completed_at=None)
 
 
 class Job(models.Model):
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -141,10 +141,11 @@ class Job(models.Model):
 
     @property
     def is_running(self):
-        if self.status_code is not None:
+        # can we rely on status_code being null to define running?
+        if self.status_code and self.status_code not in [6, 8]:
             return False
 
-        return self.started_at and not self.completed_at
+        return self.started and self.completed_at is None
 
     def notify_callback_url(self):
         if not self.job_request.callback_url:
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -197,7 +197,7 @@ class Job(models.Model):
             return "Failed"
 
         if self.is_running:
-            return "In Progress"
+            return "Running"
 
         if self.is_pending:
             return "Pending"
@@ -358,7 +358,7 @@ class JobRequest(models.Model):
             return "Failed"
 
         if self.is_running:
-            return "In Progress"
+            return "Running"
 
         if self.is_pending:
             return "Pending"
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -108,7 +108,10 @@ class Job(models.Model):
         return reverse("job-detail", kwargs={"pk": self.pk})
 
     @property
-    def is_complete(self):
+    def is_completed(self):
+        if not self.is_finished:
+            return False
+
         return self.status_code is None and self.completed_at
 
     @property
@@ -117,17 +120,27 @@ class Job(models.Model):
 
     @property
     def is_failed(self):
+        if not self.is_finished:
+            return False
+
+        if self.status_code is None:
+            return False
+
         dependency_not_finished = 6
         dependency_failed = 7
         dependency_running = 8
 
-        non_failure_status = self.status_code in [
+        non_failure_statuses = [
             dependency_not_finished,
             dependency_failed,
             dependency_running,
         ]
 
-        return self.status_code and not non_failure_status
+        return self.status_code not in non_failure_statuses
+
+    @property
+    def is_finished(self):
+        return self.started and self.completed_at is not None
 
     @property
     def is_pending(self):
@@ -181,7 +194,7 @@ class Job(models.Model):
 
     @property
     def status(self):
-        if self.is_complete:
+        if self.is_completed:
             return "Completed"
 
         if self.is_dependency_failed:
@@ -261,7 +274,7 @@ class JobRequest(models.Model):
         if not last_job:
             return
 
-        if not self.is_complete:
+        if not self.is_completed:
             return
 
         return last_job.completed_at
@@ -290,11 +303,11 @@ class JobRequest(models.Model):
         return f.url
 
     @property
-    def is_complete(self):
+    def is_completed(self):
         if not self.jobs.exists():
             return False
 
-        return all(j.is_complete for j in self.jobs.all())
+        return all(j.is_completed for j in self.jobs.all())
 
     @property
     def is_failed(self):
@@ -304,7 +317,7 @@ class JobRequest(models.Model):
         We don't consider a JobRequest failed until all Jobs are either
         Completed or Failed.
         """
-        if not all(j.is_failed or j.is_complete for j in self.jobs.all()):
+        if not all(j.is_failed or j.is_completed for j in self.jobs.all()):
             return False
 
         return any(j.is_failed for j in self.jobs.all())
@@ -319,7 +332,7 @@ class JobRequest(models.Model):
 
     @property
     def num_completed(self):
-        return len([j for j in self.jobs.all() if j.is_complete])
+        return len([j for j in self.jobs.all() if j.is_completed])
 
     @property
     def runtime(self):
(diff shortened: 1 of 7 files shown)


jobserver/models.py
```diff
@@ -64,13 +64,13 @@ def repo_name(self):
...
+        return self.filter(started=True, completed_at__isnull=False)
 
-    def in_progress(self):
-        return self.filter(started_at__isnull=False, completed_at=None)
+    def running(self):
+        return self.filter(started=True, completed_at=None)
 
     def pending(self):
-        return self.filter(started_at=None, completed_at=None)
+        return self.filter(started=False)
```
sebbacon, line 73: Shouldn't we combine this with the `status_code not in [6, 8]` logic?

jobserver/models.py
```diff
@@ -108,44 +108,54 @@ def get_absolute_url(self):
...
-        return self.status_code and not non_failure_status
+        return self.status_code not in non_failure_statuses
 
     @property
-    def is_in_progress(self):
-        if self.status_code is not None:
-            return False
-
-        return self.started_at and not self.completed_at
+    def is_finished(self):
```
sebbacon, line 138: We call this `completed` (rather than `finished`) in the queryset version.

jobserver/models.py
```diff
@@ -108,44 +108,54 @@ def get_absolute_url(self):
...
-    def is_complete(self):
-        return self.status_code is None and self.completed_at
+    def is_completed(self):
+        if not self.is_finished:
+            return False
 
-    @property
-    def is_dependency_failed(self):
-        return self.status_code == 7
+        return self.status_code is None and self.completed_at
```
sebbacon, line 115: Is there an argument for the implementations of these properties being via the queryset, e.g. `Job.objects.completed.filter(pk=self.pk).count() > 0` or similar?

jobserver/models.py
ghickman: Yep, I missed that, thanks!

jobserver/models.py
ghickman: Oops, that's a bug in the  QuerySet version!

jobserver/models.py
ghickman: There's definitely an argument for it… which I've had many times before 😀 

I find the QuerySet version trickier to read myself and on a list page one would have to have knowledge of the property implementation to know that it will create O(N) queries so I try to avoid that style of lookup for something like this.

ghickman:
@sebbacon – fixed up as per comments.  I found some extra bugs along the way:
* `.is_finished` didn't check for `started=True`, this is probably overkill but I'm keen to be strict with status parsing until job-runner takes over that
* The `Unknown` status is now tested since we have at least 50 Jobs in prod which produce it.

You probably want to re-review all the status parsing methods just to make sure they're still good.

jobserver/models.py
```diff
@@ -64,13 +65,28 @@ def repo_name(self):
...
+        )
+
+    def finished(self):
+        return self.exclude(started=True, completed_at=None)
+
+    def failed(self):
+        return (
+            self.exclude(completed_at=None)
+            .exclude(status_code=None)
+            .exclude(status_code__in=[6, 7, 8])
```
sebbacon, line 81: Don't we want to exclude `0` too?

jobserver/models.py
```diff
@@ -64,13 +65,28 @@ def repo_name(self):
...
-
-    def in_progress(self):
-        return self.filter(started_at__isnull=False, completed_at=None)
+        return (
+            self.filter(started=True)
+            .exclude(completed_at=None)
+            .exclude(status_code=None)
+        )
+
+    def finished(self):
```
sebbacon, line 74: Not understanding the intent of differentiating between `finished` and `completed` here.

Also, the logic for `completed` makes sense but the logic for `finished` would include things that are pending, no?

(As an aside, personally I find the use of `exclude` confusing and would find `filter` with negative assertions like `completed_at__isnull` easier to parse)

jobserver/models.py
```diff
@@ -64,13 +65,28 @@ def repo_name(self):
...
+    def failed(self):
+        return (
+            self.exclude(completed_at=None)
+            .exclude(status_code=None)
+            .exclude(status_code__in=[6, 7, 8])
+        )
 
     def pending(self):
-        return self.filter(started_at=None, completed_at=None)
+        qwargs = Q(status_code=None) | Q(status_code__in=[6, 8])
```
sebbacon, line 85: If we're using `Q` filters can we combine the `started` into the same thing, rather than mixing keyword filter argments and `Q` filters?

jobserver/models.py
```diff
@@ -108,44 +124,54 @@ def get_absolute_url(self):
         return reverse("job-detail", kwargs={"pk": self.pk})
 
     @property
-    def is_complete(self):
-        return self.status_code is None and self.completed_at
+    def is_completed(self):
+        if not self.is_finished:
+            return False
```
sebbacon, line 129: Still confused by the co-occurrence of `completed` and `finished`.

How does it make sense to call a job `completed` if it has a `status_code` of `None`?

jobserver/models.py
```diff
@@ -180,22 +206,19 @@ def runtime(self):
...
     @property
     def status(self):
-        if self.is_complete:
+        if self.is_completed:
             return "Completed"
 
-        if self.is_dependency_failed:
-            return "Dependency Failed"
-
         if self.is_failed:
```
sebbacon, line 212: As each state *should be* mutually exclusive, would it make sense to make these `elifs`?

jobserver/models.py
```diff
@@ -108,44 +124,54 @@ def get_absolute_url(self):
...
 
-        non_failure_status = self.status_code in [
+        non_failure_statuses = [
             dependency_not_finished,
             dependency_failed,
             dependency_running,
         ]
 
-        return self.status_code and not non_failure_status
+        return self.status_code not in non_failure_statuses
```
sebbacon, line 151: If a dependency fails, there's an argument for calling `dependency_failed` a failed state; we certainly don't go on to run the job even if requested.

jobserver/models.py
```diff
@@ -108,44 +124,54 @@ def get_absolute_url(self):
...
+        return self.started and self.completed_at is not None
 
     @property
     def is_pending(self):
-        if self.status_code in [6, 8]:
+        if not self.started:
             return True
 
-        if self.status_code is None and not (self.started_at and self.completed_at):
+        if self.status_code in [6, 8]:
```
sebbacon, line 162: can we use tokens again instead of integers? i.e. move the variables to module scope?

sebbacon (commented):
I'm afraid I'm thoroughly confused by `completed` versus `finished` - I suspect I'm misunderstanding the intent, in which case we need better variable names!

jobserver/models.py
ghickman: I'd rather not.  We'd have to duplicated it into each side of the Q instances.

jobserver/models.py
ghickman: Is `0` the successful status?

jobserver/models.py
ghickman: * `finished`: the job has run and is no longer running
* `completed`: the job has finished with a success mode
* `failed`: the job has finished with a failure mode

I'm thinking `succeeded` would be the better name here?  It's a stronger signal to the positive nature of the state?

We need the `finished` state to make aggregating state at the JobRequest level much easier.

---

> the logic for finished would include things that are pending

I don't think so?  It's checking a Job has been started and has `completed_at` set.  While pending short-circuits when `started=False`.

---

I often flip flop on `.exclude()`, `__isnull` can be hard to parse for some folks (including me at times) so I reach for `.exclude()` when I think it's going to be tricky to read those.  I'll see how `__isnull` works for these.

jobserver/models.py
ghickman: Hopefully [this](https://github.com/opensafely/job-server/pull/180#discussion_r517295111) comment clears up the confusion!

I was under the impression that a successful Job had no `status_code`?  Your question makes me think I'm wrong there!  Is that what `status_code = 0` is for?

jobserver/models.py
ghickman: We have Jobs which don't fit any of these states in Production, hence the fall-through to `Unknown`.

jobserver/models.py
sebbacon: See

https://github.com/opensafely/job-runner/blob/06271b23443966c4db5e1bc9a81504d39b9c88c8/jobrunner/main.py#L33-L121

A null status_code is evidence the job has not completed.

jobserver/models.py
sebbacon: That can still happen with `elif`?

jobserver/models.py
sebbacon: I prefer `succeeded` to `completed`.

jobserver/models.py
ghickman: Good point, I'll add them with a comment so we (hopefully!) remember to remove them once job-runner is driving this data.

jobserver/models.py
ghickman: Switched all the JobQuerySet methods over to the `__isnull` style.  I'm dubious on the readabily improvements but I did find a bug in `.finished()` 😬

jobserver/models.py
ghickman: That's a very good point.  From a UI perspective… if my Job isn't running because something up the tree failed and my Job _isn't going to run_… that's a failure I'd say.

jobserver/models.py
ghickman: True.  Unless you've got strong opinions I think I'd like to leave this as is.  They're semantically identical (I believe), but I find the whitespace aids my ability to scan through the method far easier than a `elif` block.

ghickman:
@sebbacon – third times the charm!

jobserver/models.py
```diff
@@ -63,14 +70,41 @@ def repo_name(self):
...
+        return self.filter(qwargs, started=False)
+
+    def running(self):
+        return self.filter(started=True, completed_at=None)
+
+    def succeeded(self):
+        return self.filter(
+            started=True,
+            completed_at__isnull=False,
+            status_code__isnull=False,
```
sebbacon, line 106: This potentially needs to exclude `pending_states` too.  It's possibly that pending states never have a completed_at, but I see no downside to a belt-and-braces here,

jobserver/models.py
```diff
@@ -107,45 +141,58 @@ def __str__(self):
...
+        ]
+
+        if self.status_code in pending_states:
             return True
 
         return False
 
+    @property
+    def is_running(self):
+        # can we rely on status_code being null to define running?
```
sebbacon, line 183: I think we can remove that comment?

jobserver/models.py
```diff
@@ -107,45 +141,58 @@ def __str__(self):
...
+        if self.status_code and self.status_code not in [6, 8]:
```
sebbacon, line 184: Can we use the module-scoped variables here?

sebbacon (commented):
Nearly there!

jobserver/models.py
ghickman: Oops, missed those, yes!

jobserver/models.py
ghickman: In `Job.is_succeeded` we're checking that the status_code is `0`, should we mirror that here instead?

sebbacon: Yes... and I'd have said having two implementations of the logic for a single desired outcome is asking for trouble anyway...?

ghickman: I agree on two implementations being bad, but calling QuerySet methods in the properties is going to result in an explosion of queries.

The JobRequest list pages are paginated to 25 items.  Each of those items has 1+ related Jobs.  We render the state for each Job on the page using these properties via the `.status` property.  Worse case scenario we're doing to execute 4 queries per Job.

The current first page of Prods Jobs page has:
* 29 failed Jobs
* 19 pending Jobs
* 1 running Job

Based on the ordering of conditionals in `Job.status` that's:
* `29 * 2 = 58`
* `18 * 4 = 72`
* `1 * 3 = 3`

For a total of 133 queries 😬

inglesp: Seagull code review, please tell me to get stuffed if I'm not being helpful.

1) 133 isn't a necessarily disaster.
2) Can this be solved with a combination of `prefetch_related` and property functions that ensure that things are loaded from `_prefetched_objects_cache`?

ghickman: I might well be wrong here (please say so!), but the suggestion [here](https://github.com/opensafely/job-server/pull/180#discussion_r516842266) was to use `Job.objects.completed.filter(pk=self.pk).count() > 0`, which I don't think we could handle with `prefetch_related` but we could potentially aggregate the counts onto a QuerySet.

However, to me, that's adding indirection.  As someone querying multiple Jobs I have to remember to aggregate the correct things onto a QuerySet or get an implicit query explosion.

---

It turns out we're not actually using the QuerySet anywhere… I have a feeling this was added (and maybe used briefly?) by the JobList page filters before `JobRequest` was added and that page rebuilt 😬  On the plus side, this discussion has been great because I did sit down and interrogate all my previous decisions around this 👏

ghickman:
@sebbacon – this will surely be the last time… 😬

sebbacon (approved):

ghickman:
🎉

PR #189, opened 2020-11-06
More Tests

ghickman:
This adds more tests!  Some functions got missed in recent PRs and along with the current in-flight PRs this should bring us up to (or at least very very close to) 100% coverage at which point I'd like to enforce that number is maintained.


sebbacon (approved):
:heart:

ghickman:
Snuck the enforced 100% coverage in there too, fingers crossed I don't missed this stuff going forward now!

PR #193, opened 2020-11-06
Zombify Jobs

### LINKED ISSUE #191: Zombify Jobs
Previously we've ended up with Jobs on job-server stuck in a state which isn't reflected in job-runner, eg SIGINTing job-runner and state for a Job being lost there, leaving it in the `Running` state on job-server.

We should add a `Zombify` button to set a failure `status_code` so we can manually clear out Jobs.

ghickman:
This adds the ability to "Zombify" a single `Job` or all `Job`s for a `JobRequest`.  There are buttons on both the `Job` and `JobRequest` detail pages, both set the appropriate `Job`s status_code to `10` and status_message to `Job manually zombified`.

Fixes #191


sebbacon (commented):
I don't think we should expose this to all logged-in users. It'll just beg loads of confusing questions for them. Can we make use of `is_admin` or whatever for this?

ghickman:
@sebbacon – updated to required `is_superuser` for a User to zombify a given Job.

sebbacon (approved):

PR #196, opened 2020-11-09
Ensure Admins are Configured on Boot

ghickman:
This gives us an easy way to configure admins at application boot via env vars.

I originally had the code in `get_admins` as a one-liner in settings.py but realised it would be better to 1) test it, 2) document what it's doing and 3) pull it out of the global namespace.


sebbacon (approved):
nice

PR #215, opened 2020-11-17
API v2

### LINKED ISSUE #152: Filter JobRequests by Backend
JobRequests have a `backend` field, we should be able to filter by that.

### LINKED ISSUE #186: API 2.0
To match job-runner v2 and #84 we need a new, JobRequest-aware API, with at least:

* `GET /api/job-requests/` -> list JobRequests with some filtering options
* `POST /api/jobs/` -> update an arbitrary list of Jobs

ghickman:
This adds API v2 alongside v1 (shuffling that under the v1 prefix).  It includes the addition of some new fields to track data from job-runner and groups the fields which we'll remove later.

There's a small amount of code added so v1 can support v2 and vice versa.

Fixes #152 
Fixes #186


jobserver/api.py
```diff
@@ -1,10 +1,56 @@
...
+        model = Workspace
+
+
+class JobRequestAPIList(ListAPIView):
+    filterset_fields = ["backend"]
+    permission_classes = []
+    queryset = (
+        JobRequest.objects.filter(
+            jobs__completed_at__isnull=True,
+        )
```
evansd, line 104: Just to double-check the logic on this as I can't remember off the top of my head what SQL Django generates here. We want two sorts of JobRequest: ones which have at least one Job where `completed_at` is null; and ones which have no associated Jobs at all. Will this match both conditions?

evansd (approved):
A++ PR

Just one clarification on the filtering logic.

jobserver/api.py
ghickman: Not even close 😂

I think I had the condition completely wrong in my head, will rewrite!

ghickman: After a lot of head scratching this actually works (!).  I cut it down to `JobRequest.objects.filter(jobs__completed_at__isnull=True).distinct()` which produces this SQL:

```
SELECT DISTINCT
                "jobserver_jobrequest"."id",
                "jobserver_jobrequest"."created_by_id",
                "jobserver_jobrequest"."workspace_id",
                "jobserver_jobrequest"."backend",
                "jobserver_jobrequest"."force_run_dependencies",
                "jobserver_jobrequest"."requested_actions",
                "jobserver_jobrequest"."callback_url",
                "jobserver_jobrequest"."sha",
                "jobserver_jobrequest"."identifier",
                "jobserver_jobrequest"."created_at"
           FROM "jobserver_jobrequest"
LEFT OUTER JOIN "jobserver_job" ON ("jobserver_jobrequest"."id" = "jobserver_job"."job_request_id")
          WHERE "jobserver_job"."completed_at" IS NULL
```

I've expanded the test (`test_jobrequestapilist_success` in `test_api.py`) to cover the four possible JobRequest states (all jobs completed, some jobs completed, no jobs completed, no jobs) and confirm we should list the latter 3 of those.

evansd: Excellent!

PR #219, opened 2020-11-19
Catch invalid YAML

ghickman:
[Sentry](https://sentry.io/organizations/ebm-datalab/issues/2042192762/?project=5443358&query=is%3Aunresolved)

When parsing invalid YAML we get an error and return a 500 to the user.  This catches that error and updates the message we show users when we can't find or parse their `project.yaml`.


bloodearnest (approved):
NINJA

PR #223, opened 2020-11-20
Better Runner Error Reporting

ghickman:
This reworks how we display error reports when we haven't heard from job-runner in a while.  It removes the Alpha banner (yay, Beta stage!), pulls the stats bar out of the site header to a dedicated page (but otherwise leaving them as-is), and makes use of the now empty banner space to notify when a job-runner hasn't communicated in a while.

**Banner**

**Status Page**

I'm aware that a second job-runner instance is on the horizon and have some first pass changes to prepare for that, but nothing major yet.


jobserver/backends.py
```diff
@@ -0,0 +1,21 @@
+from datetime import timedelta
+
+from django.utils import timezone
+
+
+def show_warning(unacked, last_seen, minutes=5):
+    print(f"unacked: {unacked}")
```
sebbacon, line 7: You prob don't want to keep this

sebbacon (approved):

jobserver/backends.py
ghickman: 🤦‍♂️ thank you!

PR #231, opened 2020-11-25
Explain Action Run Conditions

### LINKED ISSUE #198: Explain Run Conditions
When running Jobs it's not clear to a user what the conditions are for a Job being run.

Lets add some explanatory text before the Submit button of the CreateJobRequest form which explains:

* run_all will run all dependencies
* dependencies which previous failed you will have to run them explicitly

ghickman:
This add help text blocks to both the Actions list and the Force Run Dependencies toggle to explain how a Users selections affect how their chose Actions are run.

> run_all will run all dependencies

Since `run_all` is defined by a User in their project.yaml (ie not something we do for them currently), should we still include this part?

Fixes #198

### DIFF
--- a/jobserver/templates/workspace_detail.html
+++ b/jobserver/templates/workspace_detail.html
@@ -101,7 +101,7 @@
           <strong style="margin-right:80px;width:52px;">Status</strong>
         </div>
 
-        <div class="accordion" id="actions">
+        <div class="accordion" id="actions" aria-describedby="actions_help_block">
           {% for action in actions %}
           <div class="card">
 
@@ -159,6 +159,10 @@
           {% endfor %}
         </div>
 
+        <small id="actions_help_block" class="form-text text-muted">
+          Selecting an Action will run it, regardless of its previous state or outputs.
+        </small>
+
         {% if form.requested_actions.errors %}
         <ul>
           {% for error in form.requested_actions.errors %}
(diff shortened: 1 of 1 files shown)


evansd:
V2 handles run_all and ignores any user defined action of that name. Eventually we should bump the project.yaml version and make it an error to define a run_all action

jobserver/templates/workspace_detail.html
```diff
@@ -159,6 +159,10 @@ <h3>{{ workspace.name }}</h3>
           {% endfor %}
         </div>
 
+        <small id="actions_help_block" class="form-text text-muted">
+          Selecting an Action will run it, regardless of it's previous state or outputs.
```
sebbacon, line 163: `its`

jobserver/templates/workspace_detail.html
```diff
@@ -177,11 +181,19 @@ <h3>{{ workspace.name }}</h3>
...
+          <label
+            for="id_force_run_dependencies"
+            class="custom-control-label"
+            aria-describedby="force_run_dependencies_help_block">
             Force run dependencies of your selected actions?
           </label>
         </div>
 
+        <small id="force_run_dependencies_help_block" class="form-text text-muted">
+          Selecting this will run each Action in your selected Actions
```
sebbacon, line 193: `Action's`

sebbacon (changes requested):
So in v2 we'll need to special-case `run_all` because we still want it to be selectable even though it doesn't appear in the `project.yaml`?

jobserver/templates/workspace_detail.html
ghickman: Fixed and rebased in 1706032

jobserver/templates/workspace_detail.html
ghickman: Fixed and rebased in 1706032

ghickman:
@evansd – will having `run_all` in the `requested_actions` field continue to work as expected?  I can easily modify the UI if that's the case.

evansd:
Yes, absolutely. You can basically pretend run_all is a normal action like any other. It just always exists whether you've defined it or not.

ghickman:
Captured in #234

ghickman:
@sebbacon – this one is good to go now I think?

sebbacon (approved):

Issue #228, opened 2020-11-24
Set up a dummy-data-only job server as a user sandbox


ghickman:
As part of our planned onboarding story we want a dummy deployment users can kick off their first Job on to get a feel for the system without talking to a real database or adding noise to Production.

We'll need:
* a new dokku deployment
* configurable databases
* docs on devs deploying to multiple dokkus

sebbacon:
The dokku deployment could be a jobrunner-and-jobserver-in one.

sebbacon:
When deployed as a sandbox, the job-server should make this very clear in a banner. "This is a sandbox for testing opensafely-compliant repositories against dummy data. To run against real data, <visit the live job server>" or similar

ghickman:
@bloodearnest – does the Test backend remove the need for this ticket?

bloodearnest:
Yeah, I think so

PR #235, opened 2020-11-25
Make Backends Configurable

### LINKED ISSUE #229: Make backends configurable
As part of #228 we want to make `JobRequest.BACKEND_CHOICES` configurable so we can support a dummy deploy (with `expectations`) and a production deploy (without `expectations`).

ghickman:
This makes the available backends configurable from the environment so we can have an onboarding deployment (#228).

I've also taken the opportunity to add in [environs](https://github.com/sloria/environs) to make pulling non-string values from the environment easier (catching a not-quite-bug in the logging config).

Fixes #229


sebbacon (approved):
Do we need to add a key to `environment-sample`?

ghickman:
@sebbacon – oh good point.  On thinking on this more, maybe we want to default to `expectations`?  That's sort of our "dummy" backend IIRC?

Issue #244, opened 2020-11-30
For jobs, use the identifier supplied by job-runner rather than an internal id


ghickman:
We're still using Job.id around the UI, including in URLs.  Job.identifier is a more stable string with the current method of updating Jobs via the API, let's use that instead.

Should we swap out Job.id for Job.identifier?

sebbacon:
Makes sense to me

ghickman:
We're now using this around the UI everywhere.

PR #249, opened 2020-12-02
Replace Rather than Recreate Jobs

ghickman:
This changes the way we create Jobs in the v2 API.  Currently we delete all Jobs for a given JobRequest, then create a brand new set based on the incoming payload.  This has the advantage of being very simple.  However, it has made our PKs spiral upwards for no particularly good reason.  This change instead groups the incoming Jobs by their JobRequest identifier and then processes them to delete/create/update as appropriate.

The longer term fix to this is to replace `Job.pk` with `Job.identifier`, but in order to that we need to remove all Jobs created via the v1 API (and from before even that!).  Since we only deployed v2 earlier this week I'd like to let it get a bit more use before I start deleting our historical Jobs (or modify them with unique identifiers).

Ref #244

### DIFF
--- a/jobserver/api.py
+++ b/jobserver/api.py
@@ -1,3 +1,6 @@
+import itertools
+import operator
+
 from django.db import transaction
 from django_filters import rest_framework as filters
 from rest_framework import serializers, viewsets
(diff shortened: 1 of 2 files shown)


jobserver/api.py
```diff
@@ -36,46 +39,67 @@ def post(self, request, *args, **kwargs):
...
+            job_request = job_request_lut[jr_identifier]
+
+            # get the current Jobs for the JobRequest, keyed on their identifier
+            jobs_by_identifier = {j.identifier: j for j in job_request.jobs.all()}
+
+            # delete local jobs not in the payload
+            job_request.jobs.exclude(identifier__in=jobs_by_identifier.keys()).delete()
+
+            for job_data in jobs:
+                # this is already set by working through the JobRequest instance
```
sebbacon, line 71: This sentence is hard to parse! Better wording possible?

jobserver/api.py
```diff
@@ -36,46 +39,67 @@ def post(self, request, *args, **kwargs):
...
+                job_data.pop("job_request_id")
+
+                # V1 SHIM
+                # we need to translate this runner_status to avoid the .status
```
sebbacon, line 75: ditto

jobserver/api.py
```diff
@@ -36,46 +39,67 @@ def post(self, request, *args, **kwargs):
...
+                # property
+                status = job_data.pop("status")
+
+                # we set this based on started_at being set
+                started = bool(job_data["started_at"])
+                # END
+
+                # create new job
```
sebbacon, line 83: why not `update_or_create()` here?

sebbacon (approved):

jobserver/api.py
ghickman: so it is!  I think it would be hard to make the wording _worse_ at this point 😂

jobserver/api.py
ghickman: Fixed and rebased in 73cf7a4

jobserver/api.py
ghickman: Completely forgot that was a thing, that's signficantly nicer than my version.

Fixed and rebased in 73cf7a4

PR #252, opened 2020-12-02
Fuzzy Search for Workspace Selection

### LINKED ISSUE #250: Implement scaleable UI component for Workspace selection
At the moment, it's a very long list.

The minimal improvement is pagination - not as easy as we'd like - [notes about pagination using GraphQL here](https://github.com/opensafely/job-server/issues/149).

I think a typeahead find is a much more useful UI. In that case we can probably live without pagination of the results

ghickman:
This adds a search box above Workspace selection UI (on the homepage) to fuzzy search the list of Workspaces.  You can search on workspace name, repo "short name" (the bit displayed in the UI), and the branch name.

Fix #250


sebbacon (commented):
Great. Should we be using bower or something for managing javascript dependencies?

ghickman:
For the small number (2) of libraries we're currently using I'm not sold on the benefits.  When we start adding larger JS dependencies, yes probably.

I can see a point where some of the pages (eg the main Workspace one with actions and their statuses etc) get moved to React at which point it would make sense since we'll need a build pipeline.

But for now I'm happy with `wget` and keeping the version in the served name.

sebbacon:
I don't really care about *what* tooling, but I think making it easy for someone to upgrade it *somehow* is a good idea. A Makefile with wget in it; or even just a bit in the README.

ghickman:
@sebbacon – are you happy with the docs in the Readme for this?

README.md
```diff
@@ -36,3 +36,17 @@ To test the integration of job-runner and job-server follow these steps:
...
+## Add/Updating Static Assets
+We don't currently use any kind of static asset management tool (eg npm, yarn,
+etc) for this project.
+
+Each static asset has been downloaded from the appropriate site and added to
+the project with their version number.
+
+To add or update an asset:
+
+1. Download the asset from it's respective site (with `wget` or your favourite tool)
```
sebbacon, line 50: `its` :)

sebbacon (approved):

README.md
ghickman: ARG

Issue #254, opened 2020-12-03
Remove v1 Codepaths


ghickman:
Now that job-runner v2 is deployed and running we can remove all the v1 parts of this codebase.

Issue #126, opened 2020-10-20
Think about long term archiving of basic job history


evansd:
The history of which jobs were run when, by whom, and against what commit is going to become an important thing to preserve. I'm wondering it, as well as storing it in its local database, the job-server should log the key information in some plain text format which can get committed to git and pushed to github. This also means that the entire history gets cryptographically verified for us, which is nice.

evansd:
A CSV file per week would probably do the trick.

ghickman:
This could even be a scheduled Action on the destination repo which hits an endpoint on job-server?  That way we leave the scheduling to GitHub and get logging for free.

sebbacon:
I like this idea as a public backup. It's worth being clear this isn't a substitute for getting our own backups in order, which we should do at some point. Getting very close to needing that roadmap / release schedule I think!

CSVs feels maybe too arbitrary in terms of storage granularity - how about one file per Job + JobRequest?

evansd:
Yes, I agree: it's a different thing from db backups, but also important I think.

> how about one file per Job + JobRequest?
I think this will result in too many small files, which comes with it's own set of problems. Particularly in git.

I like the idea that CSVs would be readable directly on github, but if we're not bothered about that then I think my favourite storage format is gzipped, newline-separated JSON. Flexible, compact (after gzip), but easy to read in any language, and easy to work with in the shell.

sebbacon:
Github pros:
* Covenient location
* CSV rendering
* Search
* Free hosting

Github cons:
* 10Gb limit
* Slight abuse of system as we won't be using any git features
* Requires us to divide logs into daily chunks which is a bit arbitrary

The cons are pretty minor really. Back of an envelope calculation says we're unlikely to hit more than 100Mb per year of logs. So it'll do as an archive for the foreseeable future.

I'm in favour of a django command that dumps the smallest useful subset of job log data to CSV and commits to a repo (perhaps over github api?)

evansd:
For me the other pro of GitHub is that it makes it trivially easy for people to grab a copy of the history and confirm that we don't alter it.

As for getting it into github, I wonder if we could expose the logs over HTTP and then have a GitHub action which pulls them in and commits any changes. That way it becomes a standalone thing that the job-server doesn't even need to know about.

ghickman:
[...]

PR #255, opened 2020-12-04
Start v1 Removal

ghickman:
This makes a start at removing the "v1" codepaths, more specifically: the creation of `Job` instances by job-server and the API.

Deployment of these changes will require removing the v2 prefix from the API path in job-runners config.

There are some other bits of cruft removal included since I was already there making changes.

There are still more bits to remove after this but they're blocked on archiving older Jobs so we don't have to work around those Jobs (tracked in #126)

Ref #254


jobserver/urls.py
```diff
@@ -34,21 +33,11 @@
...
-    path("v2/job-requests/", JobRequestAPIList.as_view()),
-    path("v2/jobs/", JobAPIUpdate.as_view()),
-]
-
 urlpatterns = [
     path("", Index.as_view()),
     path("", include("social_django.urls", namespace="social")),
-    path("api/", include(api_patterns)),
+    path("api/job-requests/", JobRequestAPIList.as_view()),
+    path("api/jobs/", JobAPIUpdate.as_view()),
```
evansd, line 40: Is it worth temporarily add these paths under the `v2` prefix as well so there are no coordination issues around deployment? Particularly today when I'm supposed to be concentrating on OP stuff it would be good if I didn't block getting this merged and deployed.

evansd (approved):
All looks good. Just the one question below.

jobserver/urls.py
ghickman: Good call!

PR #258, opened 2020-12-04
Always Use Job Identifiers

ghickman:
This sets identifiers for Jobs created by job-server (ie "pre v2) and switches to using those identifiers everywhere.  Identifiers set this way have the suffix `-manually-set` so we can roll back if necessary.

I picked `-manually-set` because while I don't expect these older jobs to get much/any traffic I still wanted something that 1) made it clear we set them and 2) didn't suggest they were buggy somehow (eg `dummy` or `test`).


jobserver/models.py
```diff
@@ -47,7 +47,7 @@ class Job(models.Model):
 
     # The unique identifier created by job-runner to reference this Job.  We
     # trust whatever job-runner sets this to.
-    identifier = models.TextField()
+    identifier = models.TextField(unique=True)
```
sebbacon, line 50: Why not make it the primary key?

sebbacon: And if not for some reason, it should be `indexed` at least?

ghickman: > Why not make it the primary key?

I don't want to couple job-servers implementation of Job storage to that of job-runner.  Experience has shown it's a pain to decouple later on!

> should be indexed at least?

I already is! "unique implies the creation of an index" (from [the docs](https://docs.djangoproject.com/en/3.1/ref/models/fields/#django.db.models.Field.unique)).

PR #261, opened 2020-12-04
Remove Custom Job.save Paths

ghickman:
This removes the field modifying parts of Job's custom `.save()`.  Now that job-runner is setting these fields we can rely on them being in the correct state.

Ref #254

### DIFF
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -1,8 +1,6 @@
 import base64
-import datetime
 import secrets
 
-import pytz
 import requests
 import structlog
 from django.contrib.auth.models import AbstractUser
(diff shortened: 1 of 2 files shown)


jobserver/models.py
```diff
@@ -110,13 +108,6 @@ def runtime(self):
         return Runtime(int(hours), int(minutes), int(seconds))
 
     def save(self, *args, **kwargs):
-        if self.started and not self.started_at:
-            self.started_at = datetime.datetime.now(tz=pytz.UTC)
-
-        if self.status_code is not None and not self.completed_at:
-            if self.started:
-                self.completed_at = datetime.datetime.now(tz=pytz.UTC)
-
```
evansd, line 200: Any reason to not just remove this method altogether and let the superclass handle it?

evansd (approved):

jobserver/models.py
ghickman: There's still a call to callback_url below the `super()` call, removed in #264

PR #273, opened 2020-12-09
Bump sqlite3

ghickman:
This bumps the version of sqlite3 in our Docker image, since the previous version (`3.27.2-3`) no longer exists.

@bloodearnest – am I doing this right?  I've not pinned apt packages before (hadolint told me to), but I assume something changed with `python:3.9-buster` to make this happen?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ COPY requirements.txt .
 RUN pip install --requirement requirements.txt
 
 RUN apt-get update && \
-    apt-get install --no-install-recommends -y sqlite3=3.27.2-3 && \
+    apt-get install --no-install-recommends -y sqlite3=3.27.2-3+deb10u1 && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 


evansd (approved):
This looks fine to me (though like you apt-pinning is not something I have a lot of experience with).

An alternative is to do what we do on OpenCodelists and use the `pysqlite3-binary` Python package which comes bundled with a pre-compiled version of the latest SQLite and gives you isolation from the system version.

ghickman:
The macOS (and I assume windows?) story for `pysqlite3-binary` are not great 😬

evansd:
Ah, well I wouldn't want to feel I was enabling other people's poor operating system choices, but I bet we could compile macOS and Windows wheels now using Github Actions. The maintainer has previously been very responsive to PRs:
https://github.com/coleifer/pysqlite3

ghickman:
> other people's poor operating system choices

each to their own 🤷‍♂️ 

> compile macOS and Windows wheels now using Github Actions

Almost certainly, but I'm not sure how much it would benefit us given I've got a compiled version locally for OC!

PR #274, opened 2020-12-09
Display YAML Parsing Errors

### LINKED ISSUE #270: Show project.yaml Errors
When viewing the Workspace detail page, where a Workspace's actions should be listed, a broken project.yaml will leave a user with just the `run_all` action and no feedback on what the problem is, or how to fix it.

Example fix here: https://github.com/opensafely/vaccine-eligibility/pull/2

Lets hide the actions list entirely when there's a parsing issue and show some kind of error, to point them in the right direction.

ghickman:
This reworks the `get_actions()` function to raise exceptions on errors so we can bubble the `YamlScanner` and fetching-from-github exceptions up to the UI and display them to the User.

This required building a lookup table of action -> status to pass into `get_actions()` so that replaced `Workspace.get_latest_status_for_action()`.

Fixes #270

### DIFF
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -258,19 +258,27 @@ class Workspace(models.Model):
     def get_absolute_url(self):
         return reverse("workspace-detail", kwargs={"name": self.name})
 
-    def get_latest_status_for_action(self, action):
+    def get_action_status_lut(self):
         """
-        Get the latest status for an action in this Workspace
-        """
+        Build a lookup table of action -> status
 
-        try:
-            job = Job.objects.filter(action=action, job_request__workspace=self).latest(
-                "created_at"
+        We need to get the latest status for each action run inside this
+        Workspace.
+        """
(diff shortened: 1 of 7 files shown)


evansd (approved):
Nice work!

There's a question related to this which I've been meaning to ask. When we hit an error condition handling a JobRequest, the job-runner creates a failed job to communicate the error back to the server. See:
https://github.com/opensafely/job-runner/blob/4285f5274fb06c2b202caa9401a21c32b76ce972/jobrunner/create_or_update_jobs.py#L234-L267

This code arbitrarily chooses the first requested action as the action for this "error job". I think that's mostly OK but I can imagine cases where that might be confusing. I wonder if we need to be cleverer about this. Or if we need some convention for an action name which indicates an error with the JobRequest as a whole, rather than one specific action.

ghickman:
@evansd – good point!  Moved the discussion to #275

Issue #282, opened 2020-12-14
Default Workspace Names to repo-branch-db


ghickman:
Set a default value in the Workspace.name field of `{repo}-{branch}-{db}` to over the common case.

PR #288, opened 2020-12-15
Consume GraphQL Pages

### LINKED ISSUE #149: Use Pagination with GitHub's GraphQL API
When adding #143 I couldn't see how to easily paginate the `repositories` part of the call.  GitHub limits us to 100 repositories in that call and the Researchers Team has 58 repos currently.

[Pagination in GraphQL](https://graphql.github.io/learn/pagination/) is done with cursors.

Some initial questions, mostly based on GitHub's REST API:
* How do we know when we've exhausted a resources pages?
* How to track a cursor?
* Can we create a generator function which yields single Repo, Branches pairs?

ghickman:
This enables us to consume more than 100 repos from the Researchers team (which has grown to 79 since the issue was filed!) using GraphQL's cursor-based pagination to walk the pages once we have over 100 repos in that team.

Tested with `first: 10` to make sure it does actually pull more than 10 repos!

Fixes #149

### DIFF
--- a/jobserver/github.py
+++ b/jobserver/github.py
@@ -60,18 +60,22 @@ def get_file(repo, branch):
     return r.text
 
 
-def get_repos_with_branches():
+def _get_page(session, cursor):
     """
-    Get Repos with their branches from the OpenSafely Researchers Team
+    Get a page of Repos (with branches) from the OpenSAFELY Researchers Team
 
     This uses the GraphQL API to avoid making O(N) calls to GitHub's (v3) REST
-    API.
+    API.  The passed cursor is a GraphQL cursor [1] allowing us to call this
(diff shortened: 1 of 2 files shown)


sebbacon (approved):
LGTM

Question: when if ever would we change this, because getting a list of _all_ repos is too expensive? What would we do instead?  What is the impact of a "too" large list?

ghickman:
> What is the impact of a "too" large list?

Slow page loads.

> What would we do instead?

I think we probably move to a React (or similar) component for this page.  I've already been thinking about it for #282 but punted to prioritise other bits.  It would allow us to render the list asynchronously, however there would be some decisions to make around how it spoke to the GitHub API since we don't want to just throw our PAT into the JS.

PR #289, opened 2020-12-16
Update Status Icons

### LINKED ISSUE #248: Live-update job statuses on workspace detail page
WorkspaceDetail is expected to be our core page for a given Workspace, however it doesn't update so the statuses on there can get out of date.

A fairly simple solution is some javascript which polls an API endpoint to get statuses for the given workspace and updates the frontend.

ghickman:
This switches the construction of our status icons to be CSS based by modifying the SVGs with hardcoded colours, allowing icons to be updated via Javascript.  Since we're only using one colour for each currently (and I don't see that changing any time soon) so I don't think this will be a problem.

Note: I've avoided jQuery in this new script to scratch the itch of seeing how modern JS replaces many of the functions I would normally reach for.  It worked out pretty well!

Fixes #248


static/js/workspace_detail.js
```diff
@@ -0,0 +1,27 @@
+const updateStatuses = (url) => {
```
sebbacon, line 1: I am well out-of-date with current browser support for various constructions used here. Can you enlighten me?

sebbacon (approved):

static/js/workspace_detail.js
ghickman: Sure!

[`const`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const): block scoped variable which can't be changed.

[arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions): a new way to write functions in ES6.  I prefer how concise they are.  The declaration here is equivalent to `updateStatuses(url) { // do stuff };`.

[`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API): browser native, promise-based, function to fetch resources.  Syntax is similar to jQuery's `ajax()` function.

The for loop is the same as Python's `for k, v in some_dict.items():`.  Specifically the assignment is a [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) in JS terminology.

Does that cover everything?

sebbacon: Thanks, that is useful, but when I said *current browser support* I meant, what will this break in?!

ghickman: ah, misunderstood!

Nothing 🙂

Everything I've used here has been well supported on all major browsers for years.

The only outlier is IE11 but IIRC that's been superseeded by Edge for some time now?

sebbacon: IE10 is still common in the NHS...

PR #290, opened 2020-12-16
Fix JobRequestList Status Filtering

ghickman:
Django's pagination (provided by MultipleObjectMixin [1]) constructs the
Paginator instance as part of it's get_context_data so the filtered
object_list needed to be passed to that.

[1]: https://github.com/django/django/blob/3.1/django/views/generic/list.py#L113-L136


bloodearnest (approved):
lgtm

PR #294, opened 2020-12-17
Limit authentications to OpenSAFELY GitHub Org

ghickman:
This limits successful authentications to the OpenSAFELY GitHub Org.  As the custom backend class explains I had to reimplement the provided GitHub backend classes from social-core to let us use our PAT instead of the User's OAuth token.

### DIFF
--- a/jobserver/github.py
+++ b/jobserver/github.py
@@ -1,6 +1,8 @@
 import requests
 from environs import Env
 from furl import furl
+from social_core.backends.github import GithubOAuth2
+from social_core.exceptions import AuthFailed
 
 
 env = Env()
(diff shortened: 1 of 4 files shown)


evansd (approved):
Ah, the old "trivial" change that somehow involves writing a new auth backend :smile:

jobserver/github.py
```diff
@@ -102,3 +104,46 @@ def get_repos_with_branches():
...
+
+        # Use our "service-level" PAT instead of the User's OAuth token to
+        # avoid asking for the extra admin:org scope just for this check
+        headers = {"Authorization": f"token {TOKEN}"}
+
+        try:
+            self.request(f.url, headers=headers)
+        except requests.HTTPError as err:
+            # if the user is a member of the organization, response code
+            # will be 204, see http://bit.ly/ZS6vFl
```
sebbacon, line 145: I think this should be a link to https://docs.github.com/en/free-pro-team@latest/rest/reference/orgs#check-organization-membership-for-a-user ?

I'd find the link & comment (perhaps with "Github API call to check if a user is, publicly or privately, a member of the organization") more useful a few lines above next to `f.path.segments`; when reading the code I was like "what's this thing mean"?!

tests/jobserver/test_github.py
```diff
@@ -101,3 +109,63 @@ def test_get_repos_with_branches():
...
+@responses.activate
+def test_githuborganizationoauth2_user_data_302():
+    expected_url = "https://api.github.com/orgs/opensafely/members/test-username"
+    responses.add(responses.GET, url=expected_url, status=302)
+
+    class DummyBackend(GithubOrganizationOAuth2):
+        def _user_data(*args, **kwargs):
+            return {"email": "test-email", "login": "test-username"}
+
+    DummyBackend().user_data("access-token")
```
sebbacon, line 136: Can you add a comment to the effect we just want to be sure it doesn't raise anything? Just for readability.

sebbacon (commented):
Hey, that was pretty quick!

jobserver/github.py
ghickman: I probably should have marked this section in a more obvious way!  This method was taken verbatim taken from social-core's `GithubMemberOAuth2` class with only the URL and token parts changed.  I seriously considered entirely rewriting the whole thing (there's a branch which can never happen in the exception handler!) but opted for miniminal changes from their code to support fixing "wtf is going on here" later down the line.

ghickman: This comment is specifically calling out that conditional below which checks for non-204 responses, I'm not sure why they chose to link the docs there instead of the two different subclasses which construct the URL!

sebbacon: Sorry, I think there's something I'm not understanding...!

For me, at least, that bit.ly URL resolves to https://docs.github.com/en/free-pro-team@latest/rest/reference/orgs#members, which is not the endpoint being called; the URL I suggested above _is_ for the endpoint being called and does describe the various status codes that can be returned?

ghickman: Ah, yes, the new GitHub docs deep linking is a bit trash.  It's linked to the Members section, not any particular endpoint but the offset dumps you part way into the first endpoint 🤪 

On pondering this more last night I realised keeping it tied to the social-core implementation without more specific comments is going to (and already has as evidenced by this thread!) cause more confusion so I'll rewrite this to be clearer and link to the specific endpoint!

PR #296, opened 2020-12-17
Fix Action Statuses

### LINKED ISSUE #295: Action Statuses are Incorrect
Status icons for actions on the Workspace detail page are showing the incorrect status for an action.

Seen on https://jobs.opensafely.org/service_restoration_observatory_full/

ghickman:
This filters the jobs we look up when getting most recent status for a given action.

Fixes #295

### DIFF
--- a/tests/jobserver/test_models.py
+++ b/tests/jobserver/test_models.py
@@ -352,29 +352,33 @@ def test_workspace_get_action_status_lut_no_jobs():
 
 @pytest.mark.django_db
 def test_workspace_get_action_status_lut_success():
-    workspace = WorkspaceFactory()
-    job_request = JobRequestFactory(workspace=workspace)
+    workspace1 = WorkspaceFactory()
+    job_request1 = JobRequestFactory(workspace=workspace1)
+    JobFactory(job_request=job_request1, action="other-test", status="succeeded")
+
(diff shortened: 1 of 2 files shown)


tests/jobserver/test_models.py
```diff
@@ -352,29 +352,33 @@ def test_workspace_get_action_status_lut_no_jobs():
...
     JobFactory(
-        job_request=job_request,
+        job_request=job_request2,
         action="test",
         status="succeeded",
         created_at=timezone.now() - timedelta(minutes=1),
     )
 
-    output = workspace.get_action_status_lut()
+    output = workspace2.get_action_status_lut()
```
sebbacon, line 381: Wouldn't this test be a bit better if it checked an attribute of `output` that was not the same as an attribute in the workspace we're not interested in? e.g. make the `status` of the first job request `failed`, or something?

sebbacon (approved):

tests/jobserver/test_models.py
ghickman: Ooh yes.  We'll need to keep the status the same to confirm we're filtering to workspace correctly (the bug this fixes!), but changing the action name would confirm that the output shows the correct statuses for `workspace2`.

PR #297, opened 2020-12-18
Improve GitHub Org Check

ghickman:
This documents the endpoint we use to make the Organization check and groups all the status code handling into a single conditional.

### DIFF
--- a/tests/jobserver/test_github.py
+++ b/tests/jobserver/test_github.py
@@ -2,12 +2,7 @@ import pytest
 import responses
 from social_core.exceptions import AuthFailed
 
-from jobserver.github import (
-    GithubOrganizationOAuth2,
-    get_branch_sha,
-    get_file,
-    get_repos_with_branches,
-)
+from jobserver.github import get_branch_sha, get_file, get_repos_with_branches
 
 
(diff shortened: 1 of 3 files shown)


tests/jobserver/test_github.py
```diff
@@ -112,16 +111,18 @@ def test_get_repos_with_branches():
 
 
 @responses.activate
-def test_githuborganizationoauth2_user_data_404():
+def test_githuborganizationoauth2_user_data_204():
     expected_url = "https://api.github.com/orgs/opensafely/members/test-username"
-    responses.add(responses.GET, url=expected_url, status=404)
+    responses.add(responses.GET, url=expected_url, status=204)
 
     class DummyBackend(GithubOrganizationOAuth2):
```
inglesp, line 118: While you're here, could you pull `DummyBackend` into the module, and document what it's for?

inglesp (approved):

tests/jobserver/test_github.py
ghickman: Ooh, good idea.  I'll make it a fixture 👍

PR #298, opened 2020-12-18
Change Social Provider Name

ghickman:
This changes the social auth provider's name to match our previous provider's name (`github`), allowing us to avoid trying to work out how to migrate the existing database rows.

Ref #294 
Ref #297


sebbacon (approved):

ghickman:
Having to rethink this because "arg migrations" ಠ_ಠ

ghickman:
Took an entirely different approach to this after realising that this fix is more urgent than most and diving into migration resolution probably wasn't worth it for this!

inglesp (approved):
practicality >>> purity

PR #302, opened 2020-12-21
Make Job.job_request Required

ghickman:
This removes the ability to have JobRequest-less Jobs, removing the legacy Jobs which required this in the process.

Legacy jobs have been archived to the [jobs-archive](https://github.com/opensafely/jobs-archive) repo.

Ref #126 
Ref #255


evansd (approved):
This looks good, but "on delete cascade" always makes me a bit nervous unless there's a good reason we need it. I generally prefer the "throw an error" option (can't remember exactly what is called).

Issue #222, opened 2020-11-20
Each backend should use token-based auth


ghickman:
We currently use HTTP Basic Auth to authenticate job-runner for API usage.  This is tied to a User instance.  This works fine as an authentication method, but has some other downsides.

We have to remember to exclude that User when making changes to Human Users.  We will likely add successive Users for each backend.

I propose we switch to Token based auth, using the standard `Authorization` HTTP header, and tying the Token to a new `Backend` instance.  This would add a few benefits:
* API responses are now tied to the authenticating token's Backend.
* Stats gathering can be tied to the Backend (it's currently global to the API prefix).

@sebbacon @evansd @bloodearnest thoughts?

evansd:
As long as it's just "stick this magic string in a header" from the job-runner's point of view then it's fine by me :)

ghickman:
@evansd it would be exactly that 👍

ghickman:
Backend-based auth and per-backend stats are now deployed.

PR #303, opened 2020-12-22
Backend-based Auth

ghickman:
This adds two, heavily coupled, features:
* Backends as a Model/Table and the project-level filtering of them via an env var
* Token based auth, tied to Backends

There are some views to make working with Backends and their tokens (particularly the rotation of tokens) simpler.

#### Backends
Backends are now stored in a table with a unique name and a formatted display name (to build Form choices from later).  They're added via migrations.  An environment variable (`BACKENDS`) and a custom Manager are used to filter the available backends for a given installation.  When configuration is not set the filter returns all backends.  This lets Production work with just `TPP`, the upcoming onboarding server can have `Expectations`, and locally we can have all of them.  It has unintended benefit of a new developer not needing to set up backends before using the project.

#### Token Auth
Tokens are tied to a Backend.  They move Authentication away from DRF's user-centric model to that of a role/service (ie: how we talk to the API with job-runner).

The `/jobs/` endpoint will continue to work as before.

The `/job-requests/` endpoint will continue to work as before _for unauthenticated requests_.  Token-authed requests no longer need to use query-args for filtering and can rely on their authentication to filter for them but query args will be respected when used to ease switching over to this (that is, the hierarchy when filtering for authenticated requests is: query-arg, token's backend).

#### Transition
I had planned to run basic and token auths side by side but after investigating how DRF Authentication works, I remembered by why the token auth bits aren't in a backend.  The framework is tied to the idea of authentication populating `request.user` via DRF's own Request class.

Instead I'll rollback changes to job-server in production if we hit issues.

Ref #222


evansd (approved):
Very easy to follow commit series! I guess one future enhancement might be to split tokens into yet another model to allow multiple tokens per-backend, which would allow for seamless token rotation. But given that the whole system is designed to be robust against temporary disconnections it's probably not worth doing this.

PR #304, opened 2020-12-22
Remove auto_now_add

### LINKED ISSUE #245: Fix job timestamps on JobDetail page
The JobDetail page is showing created_at timestamps happening before updated_at and completed_at:

Could this be related to `Job.save()` or an oddity with job-runner vs job-server clocks?

ghickman:
Use a default for all `created_at` fields instead of `auto_now_add` so we don't overwrite any values which might be set.

I _think_ this fixes #245 which, in theory, was caused by job-server overriding `created_at` on incoming Jobs.

Ref #245


sebbacon (commented):
Code all looks fine but I don't understand why it would fix #245 ?!

ghickman:
@sebbacon – with `auto_now_add` set (eg `models.DateTimeField(auto_now_add=True)` the field is _always_ populated by Django, so say you create a model instance with `Job.objects.create(created_at=three_minutes_ago)` the value of `3_minutes_ago` will be ignored in favour of [effectively] `datetime.now()`.  Whereas setting a default for a field is only used if the field isn't already populated when saving.

For us this means job-server is setting `Job.created_at` when job-runner first tells it about a Job.  That breaks our ownership model and seems a likely reason why we're seeing created_at timestamps be later than started_at, updated_at, & completed_at timestamps.

sebbacon (approved):
Got it, thanks!

PR #305, opened 2020-12-23
Archive Workspaces

### LINKED ISSUE #283: Archive Workspace
Workspaces can become done/finished over time, particularly those tied to a merged branch.  The concept of archiving them would be useful to mitigate this situation.

It has some nice benefits:
* declutters the workspace selection page
* avoids any issues with deleting related JobRequests & Jobs (or setting the FK to NULL)

h/t @HelenCEBM for the idea and request!

I think we can keep this fairly simple as a first pass:
* Button on the workspace page for admins
* An `is_archived` field on the model
* A banner on the detail and logs pages when it's archived
* Remove it from homepage

ghickman:
This adds the ability to archive and unarchive workspaces.  An archived workspace isn't listed on the homepage, has a banner on it's detail and logs pages explaining what archival means, and can't have new JobRequests created for it.

Fixes #283


jobserver/templates/workspace_detail.html
```diff
@@ -5,16 +5,44 @@
 
 {% block content %}
 
+{% if workspace.is_archived %}
+<div class="alert alert-warning text-center mb-3" role="warning">
+  This Workspace has been archived.  Logs are still availble but new Jobs can
```
sebbacon, line 10: available

jobserver/templates/workspace_log.html
```diff
@@ -2,12 +2,20 @@
 
 {% block content %}
 
+{% if workspace.is_archived %}
+<div class="alert alert-warning text-center mb-3" role="warning">
+  This Workspace has been archived.  Logs are still availble but new Jobs can
```
sebbacon, line 7: available

sebbacon (approved):
Looks good. My only question is: why only superusers?

We let anyone create a workspace, and anyone run a job in anyone else's workspace, so it feels to me like we might as well let anyone archive a workspace.

Longer term, we want to introduce finer-grained permissions based on project (origin github repo) membership, but until then we're all about trusting everyone to do anything as long as they're in the opensafely org, right?

ghickman:
> we're all about trusting everyone to do anything as long as they're in the opensafely org, right?

Good point!  Still recalibrating to "all users can do everything" here…

PR #306, opened 2020-12-23
Explain run_all Semantics

### LINKED ISSUE #276: Explain `run_all` behaviour in words on the workspace page
I'm sure we'll be changing the behaviour to make it more intuitive at some point, but until then, we need text explaining

* clicking `run` on a previously-successful, previously-failed or never-run action will always run it
* `run_all` is a special action that runs everything *unless it's successfully run before* in which case it skips re-running that *unless it failed* in which case `run_all` refuses to run. Unless you click "force run all dependencies"

I leave the challenge of finding a better wording for bullet point 2 to you :P

ghickman:
This adds some more help text under the actions list to explain how the `run_all` actions.

Fixes #276

### DIFF
--- a/jobserver/templates/workspace_detail.html
+++ b/jobserver/templates/workspace_detail.html
@@ -163,7 +163,14 @@
         </div>
 
         <small id="actions_help_block" class="form-text text-muted">
-          Selecting an Action will run it, regardless of its previous state or outputs.
+          <p>Selecting an Action will run it, regardless of its previous state or outputs.</p>
+
+          <p>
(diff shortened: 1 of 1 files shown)


ghickman:
The help text is now very wordy!

Two possible enhancements I'd like some feedback on:
* Split up the sentences with more paragraph tags
* Encode some of the text into the UI with run_all being disabled if there are failed previous actions and force run isn't selected.  Selecting force run would enable the run_all action.

evansd (approved):
I think the wordiness is a result of the semantics being tricky to explain, rather than because you've missed a simple way of expressing it.

PR #308, opened 2020-12-23
Local Dev Tidy Ups

ghickman:
This is a collection of small changes mostly related to running the project from scratch and normalising what and how we configure various bits of the project.


docker-compose.yml
```diff
@@ -1,13 +1,20 @@
 # docker configuration for running a development server
 version: "3.6"
+
 services:
   job-server:
     build: .
     environment:
+      - ADMIN_USERS=${ADMIN_USERS}
```
evansd, line 8: Where the names match you can just [pass these through](https://docs.docker.com/compose/environment-variables/#pass-environment-variables-to-containers) using:
```
environment:
  - ADMIN_USERS
  - DATABASE_URL
  ...
```

evansd (approved):
Nice. Made a small comment which is just a FYI for later really, rather than a change that needs making now.

docker-compose.yml
ghickman: 🤯 That is incredibly useful to know, thank you!

PR #313, opened 2021-01-05
Fix Duplicate Migration Names

ghickman:
This fixes the migration name of one of the duplicate migration files which came about because I made #305 and #307 in parallel.


evansd (approved):
Arggh

ghickman:
Yes 😫 

Thank you!

PR #316, opened 2021-01-06
Expose Missing Branch as an Error State

### LINKED ISSUE #314: Expose "Missing Branch"
When the branch for a Workspace is deleted on GitHub we show the user "Could not find project.yaml" as the error message.  While technically correct we can probably expose a _bit_ more information as to why that might be the case.

Let's try and expose the fact that a branch no longer exists.  I would expect the GitHub API to tell us this in some way, but it might require an extra API call to do so.

ghickman:
This surfaces a Workspace's branch missing on GitHub as a first class error state to the UI.  When we can't get a project.yaml from GitHub we check for the presence of the branch (so the extra request is only triggered when there's already no actions to show) and if it's missing bubble that up to the UI.

Fixes #314

### DIFF
--- a/jobserver/templates/workspace_detail.html
+++ b/jobserver/templates/workspace_detail.html
@@ -90,7 +90,23 @@
 
     <pre class="text-left">{{ actions_error }}</pre>
 
-    <p>Fix the error above and reload this page to continue.</p>
+    <p>
+      To run actions for this Workspace please fix the error above and reload
+      this page to continue.
+    </p>
+
+    <p>
+      If the branch no longer exists you will need to set up a new workspace.
+      Previous jobs in this workspace are still available via the Logs button.
+    </p>
+
+    {% if not workspace.is_archived %}
+    <p>
+      If you are done with this Workspace's branch you can also Archive it to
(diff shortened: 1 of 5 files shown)


jobserver/templates/workspace_detail.html
```diff
@@ -90,7 +90,16 @@ <h3>{{ workspace.name }}</h3>
...
 
-    <p>Fix the error above and reload this page to continue.</p>
+    <p>
+      To run actions for this Workspace please fix the error above and reload
+      this page to continue.
+    </p>
+
+    <p>
+      If the branch no longer exists you will need to set up a new workspace.
+      Previous jobs in this workspace are still available via the Logs button.
```
sebbacon, line 100: If the workspace is unarchived, should we also suggest archiving it?

sebbacon (approved):

jobserver/templates/workspace_detail.html
ghickman: Yeah, good point, this _was_ the impetus to add that feature…

PR #317, opened 2021-01-08
Correctly Remove Jobs

ghickman:
This adds some useful logging inside our only write endpoint, which allowed me to find that we weren't actually deleting Jobs so everything worked by accident 🤒 

Updated the tests to capture the scenario we saw today too.

### DIFF
--- a/jobserver/api.py
+++ b/jobserver/api.py
@@ -99,7 +99,7 @@ class JobAPIUpdate(APIView):
             # delete local jobs not in the payload
             logger.info(
                 f"About to delete jobs with identifiers: {','.join(jobs_by_identifier.keys())}",
-                job_request=jr_identifier,
+                job_request=job_request.id,
             )
             job_request.jobs.exclude(identifier__in=jobs_by_identifier.keys()).delete()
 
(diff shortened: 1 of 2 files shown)


jobserver/api.py
```diff
@@ -114,7 +114,7 @@ def post(self, request, *args, **kwargs):
             log.info(
                 f"About to delete jobs with identifiers: {','.join(identifiers_to_delete)}",
             )
-            job_request.jobs.exclude(identifier__in=identifiers_to_delete).delete()
+            job_request.jobs.filter(identifier__in=identifiers_to_delete).delete()
```
evansd: Is it worth wrapping this block with `if identifiers_to_delete` so we avoid some empty log lines and no-op database calls?

evansd (approved):
:+1: Looks good

The perils of commit-by-commit reviewing meant that I had to go back and delete my "this is the wrong way round!" comment after you fixed it :)

jobserver/api.py
ghickman: Yes, good point, the last thing we need is red herring logs!

PR #321, opened 2021-01-12
Job & JobRequest UI Improvements

ghickman:
This tries to improve the Job & JobRequest detail pages in terms of answering "where am I".  This was born from a researcher sending us links to both pages and it not being clear where in the app one had landed.

Of particular note, I've added breadcrumbs to both pages and promoted the action of a Job to be its title.

**Job Request Detail**

**Job Detail**

### DIFF
--- a/jobserver/templates/job_detail.html
+++ b/jobserver/templates/job_detail.html
@@ -32,7 +32,7 @@
   <div class="col-md-9 col-lg-6 offset-lg-2">
 
     <h2><code>{{ job.action }}</code></h2>
-    <p class="small muted">
+    <p class="small text-muted">
       ID: {{ job.identifier }}
     </p>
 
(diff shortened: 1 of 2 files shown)


jobserver/templates/job_detail.html
```diff
@@ -1,13 +1,40 @@
...
+
+    <li class="breadcrumb-item">
+      <a href="{{ job.job_request.workspace.get_absolute_url }}">
+        {{ job.job_request.workspace.name }}
+      </a>
+    </li>
+
+    <li class="breadcrumb-item">
+      <a href="{{ job.job_request.get_absolute_url }}">
+        Request
```
sebbacon, line 21: I feel it would probably be useful to put the Job Request id in here too?

sebbacon (approved):

PR #322, opened 2021-01-14
Job Finished Notifications

### LINKED ISSUE #214: Notify users by email when a job has finished
as a default=no option

ghickman:
This adds email notifications for Jobs moving into the `failed|succeeded` state.

Users can now set a specific email address for notifications since their email address is populated with the email address set in their GitHub profile which might not be ideal.  By default this is their GitHub email address though.  I've added a settings page to set this in, and converted the `logout` link to a dropdown menu, copying the common UI pattern of using the User's avatar as the click target.

**Settings Page**

**New Menu**

Notifications are toggled at the Workspace level (and default to off), and the creator of a JobRequest will get the notifications email since we don't track teams for a Workspace yet.  The Notifications and Archive toggles have moved into the collapsible Details section since there wasn't enough space to add the Notifications button with the others and they're both actions so should stay together.

**Workspace Details**

---

Some implementation details:
* I've made use of incuna-mail to simplify the use of an email template.
* Updating Jobs needed to change so we can compare the state of a Job before it's updated.  I've added lots of comments to make the intent clear.

---

Fixes #214


jobserver/views.py
```diff
@@ -411,6 +411,17 @@ def get_queryset(self):
         return qs
 
 
+@method_decorator(login_required, name="dispatch")
+class WorkspaceNotificationsToggle(View):
+    def post(self, request, *args, **kwargs):
+        workspace = get_object_or_404(Workspace, name=self.kwargs["name"])
+
+        workspace.will_notify = not workspace.will_notify
```
evansd, line 419: This is fine and I doubt will cause us any problems in real life, but I think I'd prefer doing the toggling logic to generate a hidden input in the form so we could guarantee that form's effect would always match the text on the button. (No need to make any changes now, just wanted to mention it.)

jobserver/api.py
```diff
@@ -137,13 +138,50 @@ def post(self, request, *args, **kwargs):
...
                     defaults={**job_data},
                 )
-
                 log.info("Created or updated Job", job=job.id, created=created)
 
+                if created:
+                    # this is the first time job-server has heard about this
+                    # Job so we can move on to further Jobs.  We're knowingly
+                    # skipping potential notifications here to avoid creating
+                    # false positives.
```
evansd, line 151: The case this will miss is where there was a problem handling the job request and the job-runner immediately creates a failed job as a way of passing the error back. We can probably live with that as an edge case for now but worth handling at some point I think.

jobserver/api.py
```diff
@@ -137,13 +138,50 @@ def post(self, request, *args, **kwargs):
...
+                    continue
+
+                # check to see if the Job is about to transition to finished
+                # (failed or succeeded) so we can notify after the update
+                finished = ["failed", "succeeded"]
+                should_notify = (
+                    job.status not in finished and job_data["status"] in finished
+                )
```
evansd, line 159: I wonder if an alternative approach would be to add a `notified` flag to Job itself. Given that the existing view logic just updates the fields from job-runner and leaves existing fields alone I don't think this would need any special handling. That would handle the case of jobs which start in the failed state.

jobserver/models.py
```diff
@@ -318,6 +318,7 @@ class Workspace(models.Model):
     repo = models.TextField(db_index=True)
     branch = models.TextField()
     is_archived = models.BooleanField(default=False)
+    will_notify = models.BooleanField(default=False)
```
evansd, line 321: Given that the notification goes to the user who created the JobRequest, it seems a bit odd to me to have this as a property of the workspace. Wouldn't it make more sense to have it on either the JobRequest (so a flag that was set as part of the job request form) or as a per-user setting?

evansd (approved):
This looks great. I've added a few thoughts below but I think the feature exactly as you've written it will still be very useful so I don't want to make those a blocker.

jobserver/api.py
ghickman: As in track that we've sent a notification for the given Job?

My one concern, and it's a pretty weak one, would be changing the scope of job-runner's ownership.  Currently it's "all of Job" (let's just handwave away the PK for now…), so adding fields to it which job-runner didn't control could add complexity/loaded footguns for later.

ghickman: Second thought!  We'll still need to make this check since we need to know when to notify, separate from "have we notified about this before?".

jobserver/models.py
ghickman: This came from @sebbacon in #214.  For a first pass I think this tracks, clicking the "turn on notifications" every time you submit a JobRequest sounds a bit tedious and getting back to a JobRequest is not something I've promoted in the UI very much.

I can see the per-user argument but I don't have a strong opinion on either.

@sebbacon – thoughts?

jobserver/api.py
ghickman: I'd like to surface that to the UI in a different way… I've just not decided how yet!  More pondering required…

evansd:
All good points in reply. These were all mostly "things which occurred to me when reading the code" rather than specific issues or blockers.

jobserver/models.py
sebbacon: In real-life use people are very unlikely ever to do anything other than subscribe to a workspace or not, but I think we can assume we'll be refining this for various override-type situations in the future.

I'd have thought the most future-proof data model would be for a notification to be per-user-per-jobrequest (i.e. a M2M "notification" model twixt jobrequest and user.  Notifications could then have extra metadata like "notification method" or "max rate" or "frequency" etc; the fact it's typically subscribed to on a workspace level would be handled separately.

But this is all fine for now

PR #325, opened 2021-01-15
Improve Workspace's Toggle Views

ghickman:
This converts both the Archive and Notifications toggling view(s) to use hidden form inputs with the desired value.

There's no handling of or tests for an invalid form because I couldn't make one…  Any string value is considered truthy, while the lack of a value or the key is considered falsey.


evansd (approved):
:+1: Glad it all worked out in the end!

PR #341, opened 2021-01-19
Explain Notification Settings

### LINKED ISSUE #333: Add Notification Explainer Text to JobRequestCreate
When creating a JobRequest add some text to show that the current user will get notified, why, and to what email address.

ghickman:
This exposes the current state of a Workspace's notification settings below the JobRequestCreate form and provides some explainer text for what that means when notifications are on or off.

**Notifications are on**

**Notifications are off**

Fixes #333

### DIFF
--- a/jobserver/templates/workspace_detail.html
+++ b/jobserver/templates/workspace_detail.html
@@ -273,6 +273,24 @@
 
       </div>
 
+      <div class="mb-5">
+        <p>
+          Notifications for {{ workspace.name }} are:
+          <strong>{{ workspace.will_notify|yesno:"On,Off" }}</strong>
+
+          <br />
+
+          <small class="text-muted">
+            {% if workspace.will_notify %}
+            You will receive notifications for all Jobs created from this request to <code>{{ request.user.notifications_email }}</code>.
+            <br />
+            You can change this email address in your <a href="{% url 'settings' %}">settings</a>.
+            {% else %}
+            You can enable notifications for this workspace by clicking Details above.
+            {% endif %}
+          </small>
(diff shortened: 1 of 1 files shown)


ghickman:
@HelenCEBM – is there anything else you'd like to see here to explain if notifications will fire and where they will go?

HelenCEBM:
Thanks @ghickman - just to clarify what happens if say *you* have turned notifications on, and *I* submit a job in that workspace? From the above it looks like you will get notifications for jobs I submit, so can I turn on notifications for the same workspace for myself as well? (As an example in case helpful: @brianmackenna is often submitting the cohort extraction step in a workspace and I'm submitting the generate_notebook jobs in same workspace - both of us would probably like the cohort extractor notifications but the notebook ones might be a little noisy (and also these notebooks run within a few mins so notifications aren't so useful to either of us anyway...))

ghickman:
@HelenCEBM – that's a very useful example!

Currently the notifications button flips a flag on the Workspace, then whoever creates a JobRequest gets the notifications for the Jobs which result from it.  So in your example you'll get notifications for the JobRequests you're making for `generate_notebook` while Brian will get notifications for the JobRequests he's making for `cohort_extraction`.

We're almost certainly going to add some sort of subscription control later on but that's a lot more involved and for a first pass I opted for "you can only spam yourself"!

ghickman:
@HelenCEBM – do you think the changes in this PR provide enough information around how notifications work?

HelenCEBM:
Yes I think so, thanks!

evansd (approved):

PR #345, opened 2021-01-20
Refresh Job Instance Before Sending Emails

ghickman:
This refreshes the `job` variable to a Job instance as retrieved from the database.  Without this change the datetime fields (eg `started_at` and `completed_at`) used in calculating Runtime were strings as contained in the payload.

[Related Sentry issue](https://sentry.io/organizations/ebm-datalab/issues/2164587438/?project=5443358&query=is%3Aunresolved)


evansd (approved):
Ah ha!

Issue #301, opened 2020-12-21
Find a Way to Protect Against Missing Auth Settings


ghickman:
This morning I broke auth by deploying #294 & #298.  They were subsequently fixed by #300.

Social Auth is built on the concept layering up the pieces required for a concrete class, backends and strategies are no different.  In our case we're using the DjangoStrategy (because we use the Django-specific implemention: `social_django`) and our custom `GithubOrganisationOAuth2` backend.

Settings for a backend are looked up via the configured strategy using the `strategy.setting()` method such that passing `"KEY"` will result in the strategy attempting to look up these 3 keys in the project's settings: `['SOCIAL_AUTH_GITHUB_KEY', 'SOCIAL_AUTH_KEY', 'KEY']` (because our backend has the name `github`).

The issue #300 fixed is that I had added the `ORG_` substring to the _settings_ variables (note: not the _env vars_!).  This was silently ignored and requests to GitHub returned a 404 since they passed `?client_id=None` so GitHub couldn't route us to the correct app.

So, how can we protect against this?

Some not-mutually-exclusive options:
* Override our backend's inherrited (from `BaseAuth`) `.setting()` method to raise an error when calling will return `None`.
* Some form of pre-flight check
* Tests!
  * @inglesp has already suggested using responses to check the correct URL is emitted
  * Could we instantiate the backend and inspect something to confirm settings are set?

Thoughts welcome!

cc @evansd @inglesp @sebbacon @bloodearnest @madwort @CarolineMorton

madwort:
> Tests!
>
 >   @inglesp has already suggested using responses to check the correct URL is emitted

that would be my first thought, too, to verify that we're interfacing to the outside world in the way that we expect. The follow-up thought would be, is there a "poor man's staging env / smoke test / etc" we can do that would give us a live auth to real systems, so that we find out if/when the outside world changes what it expects from us before deploying to prod. Not sure if it's the right place to put effort right now, just thinking out loud. 

For example, at my previous place, it would be something like:

* code requires settings/env vars
* code does some tests

madwort:
> Tests!
>
 >   @inglesp has already suggested using responses to check the correct URL is emitted

[...]

PR #354, opened 2021-01-22
Require Auth Settings

ghickman:
This makes the GitHub auth settings required, so if we've not set them we get an error when booting the app.  It doesn't cover what was fixed in #300 but it at least makes them required (as they should have been before!).

Refs #301


inglesp (approved):
Nice test failure :)

PR #358, opened 2021-01-25
Add Stale Data Warning

### LINKED ISSUE #344: Surface job `updated_at` timestamp somehow in UI
The job-runner sets the `updated_at` timestamp each time it changes the status of a job, and will update it roughly once a minute afterwards to confirm that the job is still in the same state.

This gives us a way of confirming whether a job is actually still running (or pending) or whether something has gone wrong and we're looking at stale information.

At the most basic level we could include this timestamp alongside the other timestamps we show.

We could also have some kind of visual indication if this is too far in the past.

Note that we need to be a bit careful about what "too far" means here because while the job-runner is busy copying big files about it won't update the timestamps. This is less of a problem now that we're running natively and things are faster, but I think we'd want to leave at least five minutes before displaying any kind of warning.

ghickman:
This adds a warning to the JobDetail page if the Job has yet to complete AND we haven't seen an update in >30 minutes.  I've phrased the warning to try and avoid specifically saying "this job has died" or "sync is busted".

Image 2021-01-25 at 12 47 52 pm

Fixes #344

### DIFF
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -2,6 +2,7 @@ import base64
 import binascii
 import os
 import secrets
+from datetime import timedelta
 
 import structlog
 from django.contrib.auth.models import AbstractUser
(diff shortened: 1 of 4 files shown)


evansd (approved):
Great

PR #362, opened 2021-01-26
Be More Resilient Around Timestamps & State

ghickman:
This handles some timestamp related edgecases:
* Abstract the finished Job state (`failed` or `succeeded`) into a property
* Only look up Jobs via state (I _think_ I caught all of those)
* Add checks for timestamp fields when doing maths with them

We've agreed on:
* `created_at`: always set
* `updated_at`: always set
* `started_at`: mabye set
* `completed_at`: mabye set

However I've added the check on timestamps involved in maths to ensure we further edge cases don't 500 on us.

### DIFF
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -8,7 +8,7 @@ import structlog
 from django.contrib.auth.models import AbstractUser
 from django.core.validators import validate_slug
 from django.db import models
-from django.db.models import Count
+from django.db.models import Count, Q
 from django.urls import reverse
 from django.utils import timezone
 from environs import Env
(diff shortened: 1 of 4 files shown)


jobserver/models.py
```diff
@@ -120,6 +120,10 @@ def get_absolute_url(self):
     def get_zombify_url(self):
         return reverse("job-zombify", kwargs={"identifier": self.identifier})
 
+    @property
+    def is_finished(self):
```
evansd, line 124: This is fine, but we've been using the term `completed` elsewhere in the job-runner so it might make sense to use that for consistency

evansd (approved):
Looks good, minus minor quibble about "finished" vs "completed" which doesn't really need addressing now.

jobserver/models.py
ghickman: We're using `finished` for that consistently through job-server!  But that's becuase `completed_at` used to signify "successfully completed" and I wanted something that didn't signify a success state.  Wrote up a ticket (#363) to cover it so I don't forget.

PR #364, opened 2021-01-26
Configure Domain

### LINKED ISSUE #357: Specify the Domain as a Config Var
Currently emails are sent within the request/response cycle and get the current domain from the request.  At some point emails will be pulled into a queue to stop blocking this cycle, at which point we lose the ability to to use `request.build_absolute_uri()`.

A common pattern to handle this is to set the domain in config and use that.  It also allows the setting of `ALLOWED_HOSTS` to something more restrictive than "literally anything".

ghickman:
This configures the sites domain via the env allowing us to decouple sending emails from the request and tighten up the `ALLOWED_HOSTS` setting.

Fixes #357

### DIFF
--- a/jobserver/api.py
+++ b/jobserver/api.py
@@ -175,11 +175,9 @@ class JobAPIUpdate(APIView):
                     # already been there though)
                     continue
 
-                job_url = request.build_absolute_uri(job.get_absolute_url())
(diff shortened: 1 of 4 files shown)


inglesp (approved):
furl looks nice

PR #368, opened 2021-01-27
Display instructions on how to address status codes for a given backend

### LINKED ISSUE #259: Hint what to do next when errors are logged
Specifically, "job exited with an error code" should somehow prompt that for security reasons the error can't be published & the user needs to log into the L4 server and inspect the `metadata/` folder.

This message would be identical (or could be constructed from an identical template) for every such error. Perhaps a tooltip?

ghickman:
This shows backend-specific CTAs when Jobs have a known status code, to help the User investigate the failure, allowing us to keep error specifics in the secure backends.

Fixes #259


jobserver/templatetags/status_tools.py
```diff
@@ -0,0 +1,38 @@
...
+        "waiting_on_dependencies": "",
+        "dependency_failed": "",
+        "waiting_on_workers": "",
+        "nonzero_exit": "",
+    },
+}
+
+
+@register.simple_tag
+def status_hint(backend, status_code):
```
evansd, line 21: I actually think this should take the entire job as an argument rather than just the `status_code` as there are things from the job (e..g the action name) we can use to help format a more useful message.

ghickman:
@evansd – updated as suggested to pass the `job` into the templatetag.  I've also added a per-backend directory so we can build that path for users.  We can add more codes/explainers as and when it's useful.

evansd (approved):
Nice!

PR #369, opened 2021-01-28
Per-JobRequest Notification Overrides

### LINKED ISSUE #334: Override Notifications on JobRequest
Now that we have notifications for Jobs finishing we should add a little more control around them for a given JobRequest.

Add a collapsed-by-default section to the JobRequest form, with an always-visible title/icon to show that notifications are either on/off for the Workspace.  Inside the collapsed section the user can then override the Workspace's setting for this particular request.

This should tie into the explainer text in #333.

ghickman:
This replaces the current notifications explainer section of the JobRequest create page with a collapsible section allowing a User to override the Workspace-level setting.  Under the hood we're always saving whether to send notifications a JobRequest now, with the form being populated by the Workspace setting when created.

Fixes #334


ghickman:
@sebbacon – I don't particluarly love "click the button" but I'm struggling to come up with better words, any ideas?

jobserver/templates/workspace_detail.html
```diff
@@ -301,21 +301,101 @@ <h3>{{ workspace.name }}</h3>
...
+                        value="False"
+                        {% if not form.will_notify.value %}
+                        checked
+                        {% endif %}
+                      />
+                      Off
+                    </label>
+                  </div>
+                </div>
+
```
sebbacon, line 391: And could this just be 

> Send me notifications for this action {on|off}
> By default, you will receive notifications for all actions in this workspace. ([Change workspace settings]).

tests/jobserver/test_views.py
```diff
@@ -752,6 +752,78 @@ def test_workspacedetail_post_success(rf, monkeypatch):
...
+    assert response.url == reverse("workspace-logs", kwargs={"name": workspace.name})
+
+    job_request = JobRequest.objects.first()
+    assert job_request.created_by == user
+    assert job_request.workspace == workspace
+    assert job_request.backend.name == "tpp"
+    assert job_request.requested_actions == ["twiddle"]
+    assert job_request.sha == "abc123"
+    assert job_request.will_notify
+    assert not job_request.jobs.exists()
```
sebbacon, line 788: Why's this assertion in here?

sebbacon (approved):
Looks good, some rewording suggestions for your consideration

jobserver/api.py
```diff
@@ -166,8 +166,8 @@ def post(self, request, *args, **kwargs):
                 job.save()
                 job.refresh_from_db()
 
-                if not job_request.workspace.will_notify:
-                    # Workspace has notificaitons turned off
+                if not job_request.will_notify:
+                    # Workspace has notifications turned off
```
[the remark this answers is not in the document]
ghickman: Yep, removed and rebased

ghickman:
@sebbacon – I looked into an icon control, even going as far to mock it up (see below), but decided against it because it's much harder to build and generally SVG icons don't have great accessibility.




But I think I've hit upon a better version, taking the theme of conciseness from your feedback I've combined various parts of the control into a single UI card:



Instead of using a collapsing card control (which the `change` link controlled before) now a User can select On/Off directly (the default of which is still driven by the Workspace setting) before submitting the form as usual.

What do you think?

tests/jobserver/test_views.py
ghickman: I'm going to go with copypasta 😬

jobserver/templates/workspace_detail.html
```diff
@@ -301,21 +301,101 @@ <h3>{{ workspace.name }}</h3>
...
+        <div class="accordion" id="notifications">
+          <div class="card">
+
+            <div class="card-header d-flex justify-content-between align-items-center" id="notificationsLabel">
+              <span>
+                Notifications are <strong>{{ workspace.should_notify|yesno:"On,Off" }}</strong>
+                {% if workspace.should_notify %}
+                and will deliver to <code>{{ request.user.notifications_email }}</code>
+                {% endif %}
+              </span>
```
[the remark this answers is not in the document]
ghickman: Yes, but by default the card is collapsed so with notifications off you see:



And with them on you see:



More concise version below!

sebbacon:
I like!

PR #374, opened 2021-02-01
Duplicate Pagination Controls Above List

ghickman:
This duplicates the pagination controls for a list of JobRequests (used on the Event Log & Workspace Logs pages) to the top of that list, saving a User the need to scroll to the bottom of the list each time they want to go to the next page.


inglesp (approved):
Maybe worth pulling into its own template?

ghickman:
@inglesp – pulled into a template and removed the if around the top pagination (not sure what that was for!).

PR #376, opened 2021-02-01
JobRequest List Improvements

ghickman:
This is a list of small improvements I fixed while implementing #375.  Each commit is self contained so should be easy to review in isolation.


inglesp (approved):
Lots of nice small improvements

PR #380, opened 2021-02-02
Expose JobRequest Errors

### LINKED ISSUE #275: Handle Malformed JobRequests
[Dave has pointed out](https://github.com/opensafely/job-server/pull/274#pullrequestreview-549062184) that the way we currently inform job-server about malformed JobRequests is via creating a failed Job using the first requested action.

This could be confusing to users since the error displayed is for an action which never ran.

ghickman:
This hides the non-standard `__error__` Jobs from Users and bubbles their status message up to the JobRequest page.

Fixes #275

### DIFF
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -242,6 +242,20 @@ class JobRequest(models.Model):
         return f.url
 
     @property
+    def is_invalid(self):
+        """
+        Is this JobRequest invalid?
+
+        JobRequests are a request for a given configuration to be run on a
+        Backend.  That configuration could be unprocessable for a variety of
+        reasons when the Backend looks at it.  We currently surface that to
+        job-server by job-runner creating a Job with the action `__error__`.
+        This property finds Jobs with that action so we can easily see if this
+        particular request was valid or not.
+        """
+        return self.jobs.filter(action="__error__").exists()
+
+    @property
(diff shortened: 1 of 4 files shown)


jobserver/models.py
```diff
@@ -241,6 +241,22 @@ def get_repo_url(self):
...
+        Is this JobRequest invalid?
+
+        JobRequests are a request for a given configuration to be run on a
+        Backend.  That configuration could be unprocessable for a variety of
+        reasons when the Backend looks at it.  We currently surface that to
+        job-server by job-runner creating a Job with the action `__error__`.
+        This property finds Jobs with that action so we can easily see if this
+        particular request was valid or not.
+        """
+        actions = self.jobs.values_list("action", flat=True)
```
evansd, line 256: Minor and pointless optimisation, but would it be better to write this as?
```py
self.jobs.filter(action="__error__").exists()
```

evansd (approved):
Looks great!

jobserver/models.py
evansd: Oh, you're using `prefetch_related`, aren't you? So my comment doesn't really apply

ghickman: I am… but who doesn't like a one-liner?!

PR #382, opened 2021-02-03
Allow Users to Request a Job is Cancelled

### LINKED ISSUE #378: Cancel Jobs
We want to maintain our API paradigm with JobRequest data flowing from job-server -> job-runner and Job data flowing from job-runner -> job-server.  We also can't push data to job-runner due to required architecture for the platform.

However, JobRequest can grow a `cancelled_actions` a unique list of action names which the User would _like_ to cancel.  It's then up to job-runner how it handles that information (eg it might ignore it if a Job has already completed).

ghickman:
This adds a Cancel control to the JobDetail page, initially hidden behind a superuser check so we can trial the feature internally first.

**Before Cancellation**

**After Cancellation**

Fixes #378


evansd (approved):
Looks good. Is this automatically returned in the JobRequest API response, or does it need another change to do that?

It might be a little while before we wire up job-runner to actually cancel the jobs, but I think it's worth merging the PR so it doesn't go stale. And it's effectively behind a feature flag as it's only available to us, right?

ghickman:
@evansd – good point about the API, I'd completely forgotten about that, added in the final commit.  It doesn't look like the extra key should be an issue for job-runner based on the content of `job_request_from_remote_format`.

evansd:
Yep, that's right job-runner should happily ignore the extra field for now.

I've just had a thought that it might get confusing for _us_ if there's a big red cancel button that doesn't actually do anything! Is it worth changing the button text to something like "Cancel (not working yet!)" so we don't expect it to do anything right now.

PR #390, opened 2021-02-09
Fix JobRequest SHA URL

ghickman:
This makes the SHA URL on JobRequest detail pages _actually_ work…


inglesp (approved):
Could the label be "Git commit hash"?  I wouldn't expect most researches to know what a SHA is.

PR #394, opened 2021-02-10
Add Simple Membership, Org, and Project Admins

ghickman:
This sets up some very basic models for Orgs, Projects, and Membership of Projects, and enables them in the Admin, setting up the Admin as part of that.

There's a migration to fix the incorrect constraint on `django_admin_log`, a legacy from switching to a custom User model and removing the admin at the same time.  SQLite doesn't support modifying constraints with the ALTER TABLE statement so I've removed/readded the table, which is currently empty so nothing lost!

`ensure_admins` is now setting `is_staff` to True so admins can also access the admin.


jobserver/models.py
```diff
@@ -321,6 +321,55 @@ def status(self):
...
+        "Org",
+        on_delete=models.CASCADE,
+        related_name="projects",
+    )
+
+    name = models.TextField(unique=True)
+    email = models.TextField()
+    project_lead = models.TextField()
+    proposed_start_date = models.DateTimeField()
+    proposed_duration = models.TextField()
```
inglesp, line 364: Don't know enough about the requirements -- but a `proposed_end_date` would allow you to do more reporting.

jobserver/models.py
```diff
@@ -321,6 +321,55 @@ def status(self):
...
+class Project(models.Model):
+    org = models.ForeignKey(
+        "Org",
+        on_delete=models.CASCADE,
+        related_name="projects",
+    )
+
+    name = models.TextField(unique=True)
+    email = models.TextField()
+    project_lead = models.TextField()
```
inglesp, line 362: Should this be an FK to a `User`?

jobserver/admins.py
```diff
@@ -33,7 +33,7 @@ def ensure_admins(usernames):
         raise Exception(f"Unknown users: {', '.join(sorted_missing)}")
 
     # reset all users permissions first
-    User.objects.update(is_superuser=False)
+    User.objects.update(is_staff=False, is_superuser=False)
 
     # update configured users to be admins
-    users.update(is_superuser=True)
+    users.update(is_staff=True, is_superuser=True)
```
inglesp, line 39: Maybe rename `users` to `admins`?

inglesp (approved):
Looks good.  A handful of suggestions.

jobserver/models.py
ghickman: Not yet, we don't have a way to onboard Users without giving them access to run Jobs.  It's possible it will be in the future but hard to say at this point.

jobserver/admins.py
ghickman: Fixed in ce3399a

jobserver/models.py
ghickman: This is all a bit loose currently, the data is very free form so I'm beinf very relaxed about fields like this until we get a better sense of what's needed from use.

Issue #331, opened 2021-01-18
Handle Auth Errors when Users are not in the OpenSAFELY GitHub Org


ghickman:
[This Sentry issue](https://sentry.io/organizations/ebm-datalab/issues/2111953164/?project=5443358&query=is%3Aunresolved) shows users who aren't part of the `opensafely` GitHub Organisation are getting a 500 when they try to log in.  We should handle that failure, redirect the user to `/` with an error message explaining they need to be part of the org/pointing them to a docs page.

[This Sentry issue](https://sentry.io/organizations/ebm-datalab/issues/2129269722/?project=5443358&query=is%3Aunresolved) is expected to go away when the first is handled.

PR #395, opened 2021-02-10
Set User.notifications_email on Signup

### LINKED ISSUE #392: Set User.notifications_email after Registering
In #322 we added `User.notifications_email` with a data migration to mirror all existing User's email addresses to their notifications email address.

Let's add another migration (now that #391 is merged this won't look quite so odd) to copy any missing email addresses to notifications_email, and a pipeline action to django_social (I think we have to copy out the default pipeline to add one) to do the same thing when a User connects for the first time.

ghickman:
This ensures Users have `notifications_email` set by adding a final step to the social auth pipeline to copy their email when first joining.

Fixes #392


tests/jobserver/test_pipeline.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@pytest.mark.django_db
+def test_set_notifications_email_new_user():
+    user = UserFactory(email="test@example.com", notifications_email="")
+
+    set_notifications_email(user=user)
+
+    user.refresh_from_db()
+    assert user.notifications_email == "test@example.com"
```
inglesp, line 27: These tests give assurance that `set_notifications_email` works, but not that `set_notifications_email` ever gets called.  As such, they'd continue to pass even we upgrade Python Social Auth in such a way that means that `SOCIAL_AUTH_PIPELINE` is renamed (for instance).  Or if we migrate away from Python Social Auth (maybe we're only using 2% of the features but can't easily implement a new feature we need) then we'll have no tests to ensure that `notifications_email` is set.

It might not be possible to test this easily, in which case I wouldn't spend any time on it -- it's not mission critical and a blank `notifications_email` will (a) fail loudly and (b) be easily fixed.

inglesp (approved):

tests/jobserver/test_pipeline.py
ghickman: Very good point!  My only thought currently would be we'd have to mock/fake out the GitHub parts of the Login flow, starting with something like `client.get("/login")` and working from there to the same assert as above.

It's a lot of work, but you're right, we don't exercise the pipeline or login flow in that way at all, I think it might even handle #301 in a meaningful way too, and would allow checking things like #331.

ghickman:
@inglesp – integration tests for the login flow added.

ghickman:
@inglesp – any thoughts on the integration tests?

tests/integration/test_auth.py
```diff
@@ -0,0 +1,83 @@
...
+                "email": "test@example.com",
+                "primary": True,
+                "verified": True,
+                "visibility": "private",
+            },
+        ]
+        rsps.add(responses.GET, emails_url, json=emails_data, status=200)
+
+        membership_url = "https://api.github.com/orgs/opensafely/members/dummy-user"
+        rsps.add(responses.GET, membership_url, status=204)
```
inglesp, line 42: Could we include some details about where these URLs and responses came from?

tests/integration/test_auth.py
```diff
@@ -0,0 +1,83 @@
...
+
+    assert response.status_code == 302
+
+    f = furl(response.url)
+    assert f.scheme == "https"
+    assert f.host == "github.com"
+    assert str(f.path) == "/login/oauth/authorize"
+    assert f.args["client_id"] == "test"  # configured in pyproject.toml
+    assert "state" in f.args  # dynamically generated by the Auth Backend
+    assert f.args["redirect_uri"] == "http://testserver/complete/github/"
```
inglesp, line 82: Where does `testserver` come from?

inglesp (approved):
Nice.  This stuff's tricky to test but looks like you're testing the right kind of thing.

tests/integration/test_auth.py
ghickman: It's the hostname the testing server runs with, so if you were to call `request.build_absolute_uri()` inside a test/view under test that's what comes out.  I can't actually find anything in the docs which says _why_, but it's always been that way!

ghickman:
@inglesp – do those comments and docstring make sense?

inglesp:
> @inglesp – do those comments and docstring make sense?

They do.  :shipit:

Issue #370, opened 2021-01-29
Add IG Signoff


ghickman:
User's must go through our IG sign-off process before they can run jobs on the platform.

We can break down IG into a number of steps, some related to a given Backend (eg signed contract with TPP).

sebbacon:
Just to note the unit of IG is actually a data source, not a back end. We should discuss this soon!

ghickman:
Closing this in favour of #672 and likely future voyages.

PR #403, opened 2021-02-16
Iterate on Onboarding via Admin

ghickman:
This updates Project's fields based on chatting with Amir.

All fields are now optional, our modified FormA documents can be linked to, and approval fields have gained paired notes fields so various docs can be linked in.

Refs #370


jobserver/admin.py
```diff
@@ -54,3 +54,31 @@ class Meta:
...
+                    "proposed_duration",
+                ],
+            },
+        ],
+        [
+            "Technical Governance",
+            {"fields": ["governance_approval_notes", "has_governance_approval"]},
+        ],
+        [
+            "Technical Governance",
```
inglesp, line 94: Should this have the same heading as the section above?

jobserver/models.py
```diff
@@ -369,6 +369,9 @@ class Project(models.Model):
     proposed_start_date = models.DateTimeField(null=True, blank=True)
     proposed_duration = models.TextField(blank=True)
 
+    # store a link to our modified FormA
+    form_url = models.TextField(blank=True)
```
inglesp, line 373: Worth calling this `form_a_url`?

inglesp (approved):

jobserver/admin.py
```diff
@@ -1,9 +1,22 @@
...
 from django import forms
 from django.contrib import admin
+from django.contrib.auth.models import Group
+from social_django.models import Association, Nonce, UserSocialAuth
 
 from .models import Org, Project
 
 
+# Remove apps we don't want in the admin
+# Their admin.py files are explicitly imported above so we can unregister them
```
sebbacon, line 12: Do you mean _implicitly_?

jobserver/models.py
```diff
@@ -362,17 +362,30 @@ class Project(models.Model):
...
+    email = models.TextField(blank=True)
+    project_lead = models.TextField(blank=True)
+    proposed_start_date = models.DateTimeField(null=True, blank=True)
+    proposed_duration = models.TextField(blank=True)
+
+    # store a link to our modified FormA
+    form_url = models.TextField(blank=True)
+
+    # Governance approval
+    governance_approval_notes = models.TextField(blank=True)
```
sebbacon, line 376: It would perhaps be a slightly premature optimisation as this is an MVP, but just to say I'm pretty sure we'll be wanting more auditability here - who approved, and datetime of approval.

jobserver/models.py
```diff
@@ -362,17 +362,30 @@ class Project(models.Model):
...
     has_governance_approval = models.BooleanField(default=False)
+
+    # Technical approval
+    technical_approval_notes = models.TextField(blank=True)
     has_technical_approval = models.BooleanField(default=False)
 
+    # project published fields
+    paper_published_at = models.DateTimeField(null=True)
```
sebbacon, line 384: It may not be a paper that's published; it could be a preprint or a notebook or a blog post, for example.

There may also be more than one associated publication (for example the Nature paper could be a preprint, a paper, and a blog post)

At the least I'd avoid the terminology `paper_` at this stage. Perhaps `primary_output_publication_url` (or `output_publication_urls` if you want to add one-to-many relationship now)?

There's also an option to ask people always to provide a DOI in this field; makes it much more likely everything is citable.

jobserver/models.py
sebbacon: It's quite likely this will become a state machine pretty soon, as it may be we'll want to audit _un_ approving too; adding a state for "project commenced"; preventing a transition to `commenced` until approvals have happened; etc. Let's chat about what that might look like soon.

sebbacon (approved):

jobserver/admin.py
ghickman: Nope, the first two lines of the file import the admin modules which registers their ModelAdmins, so we can deregister them here.

jobserver/admin.py
ghickman: It should not… copypasta gone wrong 😬

jobserver/models.py
ghickman: Not in this instance I think.  We currently ship a GDoc with a form to potential project registrations (modified from NHS E's Form A) but it's both not Form A and expected to be pulled into the app in the next iteration.

jobserver/admin.py
sebbacon: So "explicitly imported above", to me, would be `from Foo import admin` - which isn't what happens - do you see what I mean? I didn't get much sleep last night so might be making no sense!

jobserver/models.py
ghickman: Noted, this was added to try and start getting some quick and dirty metrics of projects created vs papers published.  I didn't realise multiple things could be linked though!  I'll remove for now as it's not something which was asked for.

jobserver/admin.py
ghickman: I think so… but!

`from Foo import admin` and `import Foo.admin` both import `Foo/admin.py` and are functionally equivalent to trigger the import-time side effects this block of code is undoing (e.g. [the registration of a ModelAdmin](https://github.com/django/django/blob/3.1/django/contrib/auth/admin.py#L25)).  I'd say it's explict because I have a line to import them, whereas if the `jobserver` app was listed in `INSTALLED_APPS` after both `django.contrib.admin` and `social_django` when we wouldn't need the imports.

sebbacon: Things happening behind the scenes **as a side effect** of the thing that you actually wrote is the *opposite* of explicit in my book! 

I think this is clearer:

```
# Their admin.py files are imported as a side effect of importing their models, above, so we have to unregister them
```

Anyway /shrug :rofl: not important!

PR #408, opened 2021-03-04
Remove GitHub Organisation Requirement for Registration

ghickman:
This removes the requirement to be in the OpenSAFELY GitHub Org when signing up, shifting it to the views where the difference between [un]authenticated users are displayed.

This is a preparatory step towards adding a demo of the platform (with the expectations backend), and the very beginnings (more in intent rather than implementation) of roles based permissions.

### DIFF
--- a/jobserver/github.py
+++ b/jobserver/github.py
@@ -1,8 +1,6 @@
 import requests
 from environs import Env
 from furl import furl
-from social_core.backends.github import GithubOAuth2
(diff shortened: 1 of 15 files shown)


evansd (approved):
Code-wise this looks good. My question is whether we can deploy this until we've got some kind of minimal approval process in place to stop spam registrations?

ghickman:
Good point, I'll take a look at putting Users in another table at first.

PR #413, opened 2021-03-08
Release upload

bloodearnest:


jobserver/api.py
```diff
@@ -273,3 +276,45 @@ def get(self, request, *args, **kwargs):
...
+        if "file" not in request.data:
+            raise ValidationError("No data uploaded")
+
+        upload = request.data["file"]
+        release, created = handle_release(
+            workspace, backend, backend_user, release_hash, upload
+        )
+
+        response = Response(status=201 if created else 303)
+        response["Location"] = urls.reverse(
```
ghickman, line 312: `urls.reverse` is typically imported as `reverse` directly, could you do that here for consistency?

jobserver/api.py
```diff
@@ -273,3 +276,45 @@ def get(self, request, *args, **kwargs):
...
+
+        # authenticate and get backend
+        backend = get_backend_from_token(request.headers.get("Authorization"))
+        backend_user = request.headers.get("Backend-User", "").strip()
+        if not backend_user:
+            raise NotAuthenticated("Backend-User not valid")
+
+        # TODO: validate user once we have mappings from backend username.
+
+        release_hash = request.headers.get("If-None-Match")
```
ghickman, line 299: I've not used this header before, [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) suggests a non GET/HEAD requests are related to saving files, could you document in a comment here why we're using the header in this instance.

jobserver/settings.py
```diff
@@ -215,3 +216,9 @@
 
 # Sentry
 initialise_sentry()
+
+
```
ghickman, line 220: Could you stick a comment (similar to the [third party comment](https://github.com/opensafely-core/job-server/blob/66f4f5dc403eab428268a46e5258cdd158effa0d/jobserver/settings.py)) to the effect of `# PROJECT SETTINGS`.

(testing out the suggestions button for the first time!)

```suggestion

# PROJECT SETTINGS

```

jobserver/urls.py
```diff
@@ -52,6 +53,11 @@
         WorkspaceStatusesAPI.as_view(),
         name="workspace-statuses",
     ),
+    path(
+        "api/v2/workspaces/<workspace_name>/releases/",
+        ReleaseUploadAPI.as_view(),
+        name="upload-release",
```
ghickman, line 59: Can you prefix the name with `workspace`, keeps it in line with all the other URL names.

jobserver/models.py
```diff
@@ -487,3 +488,37 @@ def repo_name(self):
         if not f.path:
             raise Exception("Repo URL not in expected format, appears to have no path")
         return f.path.segments[-1]
+
+
+class Release(models.Model):
+    id = models.TextField(primary_key=True)
```
ghickman, line 494: Can you pop a comment above this line to explain why it's not using Django's AutoField?  I'm assuming this is because you're using the hash as the PK, but future Us will thank Now You!

jobserver/models.py
```diff
@@ -487,3 +488,37 @@ def repo_name(self):
...
+    workspace = models.ForeignKey(
+        "Workspace",
+        on_delete=models.PROTECT,
+        related_name="releases",
+    )
+    backend = models.ForeignKey(
+        "Backend",
+        on_delete=models.PROTECT,
+        related_name="+",
+    )
```
ghickman, line 509: Can you move the FKs to the top of the model definition?  I find it's really useful for scanning the model definitions if they're all in one place.

(Sorry, I realise there's no way you could have known about this given there's zero docs on the format!)

jobserver/models.py
```diff
@@ -487,3 +488,37 @@ def repo_name(self):
...
+    backend_user = models.TextField()
+
+    # TODO: link this formally via backend_username once we have a mapping
+    # between User and their backend username
+    publishing_user = models.ForeignKey(
+        "User",
+        on_delete=models.PROTECT,
+        null=True,
+        related_name="+",
+    )
```
ghickman, line 521: Same thing with this one, can it be grouped with the other FKs please.

jobserver/models.py
```diff
@@ -487,3 +488,37 @@ def repo_name(self):
         if not f.path:
             raise Exception("Repo URL not in expected format, appears to have no path")
         return f.path.segments[-1]
+
+
+class Release(models.Model):
```
ghickman, line 493: Super personal nit (sorry)… can the alphabetical ordering of models be kept please?  (it really helps me scan down the list!)

jobserver/models.py
```diff
@@ -487,3 +488,37 @@ def repo_name(self):
...
+    # between User and their backend username
+    publishing_user = models.ForeignKey(
+        "User",
+        on_delete=models.PROTECT,
+        null=True,
+        related_name="+",
+    )
+
+    def file_path(self, filename):
+        return settings.RELEASE_STORAGE / self.upload_dir / filename
```
ghickman, line 524: Really realising the value of Paths with this PR!

tests/jobserver/test_releases.py
```diff
@@ -0,0 +1,106 @@
...
+        release.workspace, release.backend, "user", release_hash, upload.open("rb")
+    )
+    assert not created
+    assert release.id == release_hash
+    assert release.backend_user == "original"
+    assert release.files == ["file.txt"]
+
+
+@pytest.mark.django_db
+def test_handle_release_bash_hash(monkeypatch, tmp_path):
```
ghickman, line 63: `s/bash/bad/`? 😁

tests/jobserver/test_releases.py
```diff
@@ -0,0 +1,106 @@
...
+    monkeypatch.setattr(Release.objects, "create", raise_error)
+
+    upload = tmp_path / "release.zip"
+    release_hash = make_release_zip(upload)
+    workspace = WorkspaceFactory()
+    backend = BackendFactory()
+
+    with pytest.raises(DatabaseError):
+        handle_release(workspace, backend, "user", release_hash, upload.open("rb"))
+
```
ghickman, line 90: I can't quite get my head around what this test is confirming, that the database is used inside `handle_release` or?

Something I've been trying to do recently is to add comments for my asserts to express what they're intended to check, might be worth it here.

ghickman (changes requested):
Looking good!

A few things to change, but otherwise this is nearly there.

jobserver/models.py
```diff
@@ -487,3 +488,37 @@ def repo_name(self):
         if not f.path:
             raise Exception("Repo URL not in expected format, appears to have no path")
         return f.path.segments[-1]
+
+
+class Release(models.Model):
+    id = models.TextField(primary_key=True)
+    published_on = models.DateTimeField(default=timezone.now)
```
sebbacon, line 495: nit: I prefer `published_at` for DateTimes and `published_on` for Dates.

jobserver/releases.py
```diff
@@ -0,0 +1,66 @@
...
+@transaction.atomic
+def handle_release(workspace, backend, backend_user, release_hash, upload):
+    try:
+        # check for duplicates
+        existing_release = Release.objects.get(id=release_hash)
+        return existing_release, False
+    except Release.DoesNotExist:
+        pass
+
+    # Even though id's are globally unique, we partition the filesystem
```
sebbacon, line 21: `s/id's/ids/`

jobserver/releases.py
```diff
@@ -0,0 +1,66 @@
...
+    try:
+        # first, ensure we can extract the zip
+        with ZipFile(upload) as archive:
+            archive.extractall(actual_dir)
+
+        calculated_hash = hash_files(actual_dir)
+        if calculated_hash != release_hash:
+            raise Exception(
+                f"provided hash {release_hash} did not match files hash of {calculated_hash}"
+            )
```
sebbacon, line 39: Consider raising `rest_framework.exceptions.ValidationError` here?

jobserver/releases.py
```diff
@@ -0,0 +1,66 @@
...
+        release = Release.objects.create(
+            id=release_hash,
+            workspace=workspace,
+            backend=backend,
+            backend_user=backend_user,
+            upload_dir=upload_dir,
+            files=os.listdir(actual_dir),
+        )
+
+    except Exception:
```
sebbacon, line 50: Should this be a `finally` rather than a catch-all exception handler?

jobserver/releases.py
```diff
@@ -0,0 +1,66 @@
...
+        # something went wrong, clean up files, they will need to reupload
+        shutil.rmtree(actual_dir, ignore_errors=True)
+        raise
+
+    # TODO: push zip to Github as release.
```
sebbacon, line 55: I feel like the github release part is likely to be a completely separate process (i.e. an explicit gesture when the study developer is ready), & if you agree that makes sense, you can drop this comment

tests/jobserver/test_releases.py
bloodearnest: There's some missing asserts that somehow got dropped - it's meant to check the files have been deleted.

jobserver/releases.py
bloodearnest: arg, yes, I updated the others to ValidationError, but missed this one.

jobserver/releases.py
bloodearnest: No, we only want to delete the files if there was a problem, happy path is to keep them

jobserver/releases.py
bloodearnest: The researcher (and peer reviewer in future) has already approved these files for release, I was assuming we'd just push them straight to github.  Do you envision a separate release to Github process on the job-server?

Note that to support the idea of opensafely-cli downloading releases and unpacking locally, I was assuming we'd probably want to do that via github releases rather than downloading via job-server, so that a) authn is simpler and b) public/private repo authz is simpler.

jobserver/api.py
bloodearnest: Well, we are saving a file :smiley:  - but good idea to add docs

jobserver/releases.py
bloodearnest: I remember why! I was trying to figure out how to return a  412 Pre-condition failed

jobserver/api.py
bloodearnest: done

jobserver/api.py
bloodearnest: Putting a hash (ie etag) in If-None-Match in a POST is not an explicitly documented usage, but is arguably valid and I've done it before, but its supposed to be optional, and I was requiring it, which made it all awkward.

Then I realised I can just use PUT with the hash in the url, and it's all simple much simpler, so I did that. Only reason I didn't do that originally was because I started out with the client not knowing the id, which meant a POST made sense. Half way through, I switched to using the hash as id as a means to detect republishing the same release, and missed that PUT with a full url was better.

ghickman (commented):
Looks good!

ghickman (approved):

PR #415, opened 2021-03-10
chore: add better dev env setup via scripts.

bloodearnest:


scripts/dev-env.sh
```diff
@@ -0,0 +1,30 @@
...
+fi
+
+if test -z "$BW_SESSION"; then
+    echo "Unlocking bitwarden..."
+    BW_SESSION=$(bw unlock --raw)
+    export BW_SESSION
+fi
+
+SOCIAL_AUTH_BW_ID=d48e3e3f-07ed-4865-b348-ace700ee004
+GH_DEV_TOKEN_ID=10242708-234a-4162-982f-ace700ee5c53
```
ghickman, line 20: Small nit: one contains `BW`, one doesn't.  I'm not sure we need the `BW`, but up to you 🙂

ghickman (approved):
This is great!  One small nit but don't hold up the PR for it.

scripts/dev-env.sh
bloodearnest: updated

PR #419, opened 2021-03-11
fix: require manifest in uploaded zip

bloodearnest:
- require valid manifest.json in zip
 - fix bug in Release.files listing (via list_files())
 - refactor hash_files logic to use list_files()

And some minor fixes:
 
 - fix raising ValidationError() for consistent error formatting
 - return full urls in the Location header from release api.

### DIFF
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -425,7 +425,7 @@ class Release(models.Model):
 
     @property
     def manifest(self):
-        return json.load(self.file_path("metadata/manifest.json").open())
+        return json.loads(self.file_path("metadata/manifest.json").read_text())
 
 
 class Stats(models.Model):
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -1,5 +1,6 @@
 import base64
 import binascii
+import json
 import os
 import secrets
 from datetime import timedelta
@@ -421,6 +422,10 @@ class Release(models.Model):
 
     def file_path(self, filename):
         return settings.RELEASE_STORAGE / self.upload_dir / filename
+
+    @property
+    def manifest(self):
+        return json.load(self.file_path("metadata/manifest.json").open())
 
 
 class Stats(models.Model):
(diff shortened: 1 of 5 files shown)


jobserver/releases.py
```diff
@@ -47,23 +60,29 @@ def handle_release(workspace, backend, backend_user, release_hash, upload):
             backend=backend,
             backend_user=backend_user,
             upload_dir=upload_dir,
-            files=os.listdir(actual_dir),
```
bloodearnest, line 50: This only listed the top level files, and was a bug.

jobserver/releases.py
```diff
@@ -31,14 +32,26 @@ def handle_release(workspace, backend, backend_user, release_hash, upload):
     try:
         # first, ensure we can extract the zip
         with ZipFile(upload) as archive:
+            try:
+                json.load(archive.open("metadata/manifest.json"))
```
ghickman, line 36: Won't this leave the filehandle for the manifest open?

jobserver/models.py
```diff
@@ -422,6 +423,10 @@ class Release(models.Model):
     def file_path(self, filename):
         return settings.RELEASE_STORAGE / self.upload_dir / filename
 
+    @property
+    def manifest(self):
+        return json.load(self.file_path("metadata/manifest.json").open())
```
ghickman, line 428: Same as below, won't this leave the manifest filehandler open?

ghickman (commented):
Looking good, just a couple of questions around opening the manifest.

jobserver/releases.py
bloodearnest: No, as it's not actually opening a file - it's a file-like object that reads bytes out of the zipfile, which is the actual open file handle, and handled by the with statement

jobserver/models.py
bloodearnest: This one actually will leave a handle, which as its cpython with be closed implicitly as it's immediately dereferenced, but it is properly better to be explicit

jobserver/releases.py
ghickman: TIL!

ghickman (approved):

PR #422, opened 2021-03-16
Pull Registration Form into Application

### LINKED ISSUE #409: Pull Modified FormA into Project
Convert [this](https://docs.google.com/document/d/1_Fd124NqrkJeiprra4vSXU5vytQNUnrTfs4ib3L8vo4/edit) to a form we can have unauthenticated users fill in.

Since this isn't tied our Registration flow yet we'll need to put it behind a unique, non-indexed URL.

ghickman:
This pulls our ["Modified Form A"](https://docs.google.com/document/d/1_Fd124NqrkJeiprra4vSXU5vytQNUnrTfs4ib3L8vo4/edit) into the application and adds the various CRUD views around it.
.
This form is the first point of contact for someone wanting to register with OpenSAFELY.  Currently they fill in a copy of the linked template document, and Amir collates that along with some other information to pass on the relevant authority for signoff.  This change is part of the iteration to pull that form into the application and automate the sending of it.

I've set the requirement that they will need to be a registered User to use the Form but the researchers they list (more below) aren't required to do that.

Section 3 of the linked form has been skipped since it's not relevant to us (as per conversation with Amir).

All Views added are currently behind an is_superuser = True check while we test it internally.

Researchers are in a separate table so we don't have to require they all register with the site during an application, especially since we don't know if an application will be accepted.  There's a nullable FK to User so we can link them later if the Project is accepted.  I'm leaving that process TBD so we can focus on this iteration of the Onboarding flow first.

I've added slug fields to Project and Org in preparation for moving the page structure to `/<org>/<project>/<workspace>/` in a later iteration.

Fixes #409


jobserver/models.py
```diff
@@ -349,30 +350,67 @@ def __str__(self):
...
-    email = models.TextField(blank=True)
-    project_lead = models.TextField(blank=True)
+    name = models.TextField(unique=True)
+    slug = models.SlugField(max_length=255, unique=True)
     proposed_start_date = models.DateTimeField(null=True, blank=True)
     proposed_duration = models.TextField(blank=True)
 
-    # store a link to our modified FormA
-    form_url = models.TextField(blank=True)
+    # Section 1
```
sebbacon, line 387: Worth explaining what "section 1" refers to (i.e. NHSE Form B?)

jobserver/models.py
```diff
@@ -428,6 +477,31 @@ def manifest(self):
...
+
+    user = models.ForeignKey(
+        "User",
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name="researcher_registrations",
+    )
+
+    name = models.TextField()
+    passed_researcher_training_at = models.DateTimeField()
```
sebbacon, line 498: Could this just be a DateField?

static/js/project_create.js
```diff
@@ -0,0 +1,45 @@
...
+  if (classes.length < 1) {
+    console.log(
+      `no classes beginning with ${name} in remove button class list`
+    );
+    return;
+  }
+
+  const formId = classes[0];
+  const form = document.getElementById(formId);
+  console.log(form)
```
sebbacon, line 26: you want to keep this?

static/js/project_create.js
```diff
@@ -0,0 +1,45 @@
+const addForm = (name, event) => {
+  const formIndex = $(`#id_${name}-TOTAL_FORMS`).val();
+  const newForm = $(`#${name}-forms .template`)
+    .html()
+    .replace(/__prefix__/g, formIndex);
```
sebbacon, line 5: What is this `__prefix__` business? Comment would help me here!

static/js/project_create.js
```diff
@@ -0,0 +1,45 @@
...
+  $(`#id_${name}-TOTAL_FORMS`).val(parseInt(formIndex) + 1);
+
+  event.preventDefault();
+};
+
+const removeRow = (name, event) => {
+  const classes = [...event.target.classList].filter((c) => c.startsWith(name));
+
+  if (classes.length < 1) {
+    console.log(
```
sebbacon, line 18: `console.error`, perhaps?

tests/jobserver/test_views.py
```diff
@@ -523,6 +530,192 @@ def test_jobrequestzombify_unknown_jobrequest(rf):
...
 
 
+@pytest.mark.django_db
+def test_orgcreate_get_success(rf):
+    oxford = OrgFactory(name="University of Oxford")
+    datalab = OrgFactory(name="DataLab")
+
+    request = rf.get(MEANINGLESS_URL)
+    request.user = UserFactory(is_superuser=True)
+    response = OrgCreate.as_view()(request)
```
sebbacon, line 540: I'm sure we've been here before and you've explained this before and I'm sure it makes sense really, but I'm expecting this test to test the create *view* _creates something_, but actually we're creating something and then testing that the view does the right kind of thing assuming it's successfully created objects - right?

 I wonder if there's some more expressive way of making it clear what's going on?  This confuses me every time...!

sebbacon (approved):
Looks good, there may be usability improvements in terms of wording in the templates, but that'd be easier to test after it's deployed, which given it's protected from viewing by anyone else, is fine!

static/js/project_create.js
ghickman: I do not, thank you!

tests/jobserver/test_views.py
ghickman: No, this is testing the `GET` handler of `OrgCreate`, whose template lists existing Orgs as well as the create form.

So the test is creating some objects, calling OrgCreate with a GET request, checking the orgs we constructed beforehand are in the response's template context.

sebbacon: I see. I wish there were a clearer way to express this but I can see it's all in there on reflection, so no clever ideas I'm afraid

jobserver/models.py
ghickman: I've shifted the registration fields to the bottom of the model fields and explained what they all, linking to our Form A template, that should given enough context I think? (hope?)

static/js/project_create.js
ghickman: Commented each new variable, hopefully makes sense now!

jobserver/models.py
ghickman: It could, but I always opted to use DateTimeFields over DateFields because they so often become the latter.

PR #426, opened 2021-03-19
Enable cancel button for all users

evansd:
This has now been wired up on the job-runner side and tested:
https://github.com/opensafely-core/job-runner/pull/186

### DIFF
--- a/jobserver/templates/job_detail.html
+++ b/jobserver/templates/job_detail.html
@@ -146,7 +146,7 @@
 
   <div class="col-md-3 col-lg-2">
 
-    {% if request.user.is_superuser %}
+    {% if user_can_run_jobs %}
     <h3 class="text-center">Tools</h3>
 
     <form class="mb-3" method="POST" action="{{ job.get_cancel_url }}">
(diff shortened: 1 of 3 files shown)


jobserver/templates/job_detail.html
```diff
@@ -158,16 +158,18 @@ <h3 class="text-center">Tools</h3>
...
     </form>
 
+    {% if request.user.is_superuser %}
     <form class="mb-3" method="POST" action="{{ job.get_zombify_url }}">
       {% csrf_token %}
 
       <button class="btn btn-block btn-danger" type="submit">Zombify</button>
     </form>
     {% endif %}
+    {% endif %}
```
ghickman, line 172: We'll want to end the authenticated and authorized user check at the end of the Cancel button, Zombification is only for us superusers still.

ghickman (changes requested):
Just one small thing!

jobserver/templates/job_detail.html
evansd: I think the effect is that same either way (the superuser check is still there, just moved down) but your suggestion makes things clearer.

ghickman: Oh yes, I see what you mean now 👍

ghickman (approved):

PR #435, opened 2021-03-22
Require GH Org Membership  for Some Actions

ghickman:
This consolidates how we check for superusers and makes use of our `can_run_jobs` function to gate access to views where a User must be part of our GitHub organisation.

### DIFF
--- a/jobserver/roles.py
+++ b/jobserver/roles.py
@@ -1,3 +1,8 @@
+from functools import wraps
+
+from django.contrib import messages
+from django.shortcuts import redirect
+
 from .github import is_member_of_org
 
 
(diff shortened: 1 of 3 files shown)


evansd (approved):
Happy to approve as is, but I wonder if this could benefit from a slightly different refactoring. I'm not sure the `superuser_required` decorator belongs in the `roles` module and there's a lot of repetition of the `@method_decorator(X, name="dispatch")` pattern.

Maybe we could do something like this:
```py
requires_is_superuser = permissions_decorator(lambda user: user.is_superuser)
requires_can_run_jobs = permissions_decorator(can_run_jobs)

...

@requires_can_run_jobs
class JobCancel(View):
    def post(self, request, *args, **kwargs):
        ...

...

@requires_is_superuser
class ProjectCreate(CreateView):
    form_class = ProjectCreateForm
    ...
```

i.e. factor out the `user_passes_test` and `method_decorator` stuff into one `permissions_decorator` function that would then make the individual view definitions more readable.

ghickman:
@evansd – completely agree on all points!  I didn't think to talk about this in the body (sorry about that!), but much of the work in this PR will be superseded by the incoming roles&permissions framework, so the result is messier than I would have liked.

However, that being said, I'll definitely be using your suggestions in the API for that work 🙂

evansd:
:+1 Got you. Yes, makes sense not to spend time gold-plating it if this is all being replaced soon.

PR #445, opened 2021-04-02
Add Context-less Roles&Permissions Framework

ghickman:
This is the initial set of changes for the [Roles and Permissions Framework](https://github.com/opensafely-core/roadmap/discussions/13).

It does a variety of initial changes:
* Adds our initial Roles and their Permissions as per the OFP.
* Adds a custom `RolesField` model field, based on a JSONField which converts Role classes to their dotted path for storage in the database, and converts them back to Role classes on read
* Switches our `ensure_admin` command to set the CoreDeveloper role instead of `is_superuser`
* Adds a `require_role` decorator and a shortcut, `require_core_dev`, for views
* Switches the relevant views from the superuser decorator and checks to the new role based ones

Contextual ("local") roles and the mechanics to enforce them will come in a future PR.


jobserver/authorization/decorators.py
```diff
@@ -0,0 +1,26 @@
...
+    """Decorator for views which require a given Role"""
+
+    def decorator_require_role(f):  # sigh
+        @wraps(f)
+        def wrapper_require_role(request, *args, **kwargs):
+            if has_role(request.user, role):
+                return f(request, *args, **kwargs)
+
+            # TODO: shoudl this redirect to / with a message?
+            raise Http404()
```
bloodearnest, line 19: Hmm, feel like this should probably be a 403, unless leaking the existance of something is a problem here?

jobserver/authorization/permissions.py
```diff
@@ -0,0 +1,7 @@
+cancel_job = "cancel-job"
+check_output = "check-output"
+invite_org_members = "invite-org-members"
+invite_project_members = "invite-project-members"
+publish_output = "publish-output"
+review_project = "review-project"
+run_job = "run-job"
```
bloodearnest, line 7: Is there any reason to to the _/- difference? Wonder if it's worth keeping them identical? i.e. _ in both symbol and string?

I do wonder if an Enum would be good here, for better repr in logging and tracebackes

jobserver/authorization/utils.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def has_permission(user, permission):
+    if not user.is_authenticated:
+        return False
+
+    roles = user.roles
+
+    # Each role has N permissions, combine them all into a final set
+    permissions = set(itertools.chain.from_iterable(r.permissions for r in roles))
```
bloodearnest, line 11: I would suggest a for loop and early return if found here. Constructing a full set for the likely common case of having the required permission is unnecessarily slow. And while it might not make much difference now, where its simple looping, down the line fetching all roles/permissions could get more expensive, e.g. with the local context stuff.

Another option might be caching this operation per request? Idk how often we'd expect permissions checks in a single request, but I'd imagine it could be quite a lot (e.g. templates checking to render cancel buttons for a list of jobs).

jobserver/fields.py
```diff
@@ -0,0 +1,51 @@
+from django.db import models
+from django.utils.module_loading import import_string
+
+
+def parse_roles(paths):
+    """Convert Role dotted paths to Role objects"""
+    return [import_string(p) for p in paths]
```
bloodearnest, line 7: I wonder if some basic validation of the import path is worth doing here, given this is effectively arbitrary code execution?

E.g. check `p.startswith('jobserver.authorization.roles')`

Or even just have the base class name in `roles` field, and add (and thus enforce) `jobserver.authorization.roles` prefix here?

bloodearnest (commented):
LGTM, the consuming API is nice and simple, and the json munging was nicer than I'd anticipated.

Left a few comments, one security one that does need fixing IMO, and one performance one that I suspect does, and a couple of minor questions.

jobserver/authorization/decorators.py
ghickman: I thought it could have been but I can't come up with a good reason for it now which shows there likely wasn't one!

Fixed and rebased in.

jobserver/authorization/permissions.py
ghickman: No, just a habit, completely agree in hindsight they should be the same.

I've never had great luck with Python's Enums, they whole `.value()` part of their API has always felt clunky and opaque to me.  I'm not sure what the repr would do extra here given this a readable string?

jobserver/authorization/roles.py
```diff
@@ -0,0 +1,72 @@
...
+    cancel_job,
+    check_output,
+    invite_project_members,
+    publish_output,
+    review_project,
+    run_job,
+)
+
+
+class CoreDeveloper:
```
sebbacon, line 11: Add a docstring here explaining the meaning of each role?

This should then be included in our official documentation, I think. I realise most of the roles won't mean anything much *yet* but I think freezing our terminology is probably sensible, i.e. copying from that Google Doc, adding a highlighted note there to say it's now in the docs.

That also implies adding a new page to the docs which imports the docstrings. That part could be a relatively straightfoward job for @StevenMaude who's starting out by owning our new user experience?

jobserver/authorization/utils.py
sebbacon: I agree that early short-circuiting is probably sensible, as long as it's readable (`next(filter(pred, iterable), default)` is the way I have often seen of doing this but I don't like it!).

YAGNI on caching though, perhaps?

jobserver/fields.py
ghickman: Fixed and rebased in with a check at both the save and read sites.

jobserver/authorization/admins.py
```diff
@@ -32,8 +33,6 @@ def ensure_admins(usernames):
         sorted_missing = sorted(missing)
         raise Exception(f"Unknown users: {', '.join(sorted_missing)}")
 
-    # reset all users permissions first
-    User.objects.update(is_staff=False, is_superuser=False)
-
-    # update configured users to be admins
-    admins.update(is_staff=True, is_superuser=True)
+    for admin in admins:
+        admin.roles.append(CoreDeveloper)
```
sebbacon, line 37: Should we have a `SuperUser` role for superusers?

`CoreDeveloper` may still not be able to to *everything*. Naming the role so it's clear it's special seems sensible.

sebbacon (changes requested):

jobserver/authorization/admins.py
sebbacon: Ideally, in fact, these would all check specific permissions, rather than roles - and the `SuperUser` role automatically gets all permissions - but I thought perhaps you wanted to make the change incremental, hence the suggestion here. However, now I think about it, it might save pain in the long run to keep it all permissions-based from the start - WDYT?

jobserver/authorization/utils.py
bloodearnest: Ugh, yes, -1 on `next(filter(pred, iterable))`. I think a simple for loop with if ....: return True is the most readable.

On caching, sure, I was just suggesting an alternative.

jobserver/authorization/permissions.py
bloodearnest: Well, it could repr as `Permission(publish_output)` or similar, which is nicer. But happy to got with whatever, was just a thought.

jobserver/authorization/admins.py
ghickman: I don't think we should, no.

We've talked about this!  Having a group of users with, roughly speaking, root access to the platform is not ideal.  Which is why we set it up this way so we can always check the deployment config.  Longer term I would like us to keep the configuration of internal staff but apply more appropriate roles, and then look up permissions at the call site based on those.  However, we're not there yet, and there are quite a few changes to make between now and then.  In the interim the CoreDeveloper role is what we discussed (and you came up with it!), and also what's in the design GDoc.

jobserver/authorization/roles.py
ghickman: In the platform docs or docs for this project?

I'm not sure we should document the implementation of Roles as part of the platform docs?  Importing a module from this project (assuming that means we have to import the whole project) feels like a very heavy weight way of pulling anything into the docs (which already have to deal with the cohort-extractor's dependencies), and I'm not keen on making the docs project harder to build.

jobserver/authorization/admins.py
ghickman: > it might save pain in the long run to keep it all permissions-based from the start

You persuaded me to add `has_role` for exactly this section of the codebase!

jobserver/authorization/roles.py
sebbacon: Oh I forgot we're not in cohortextractor! Maybe that's not the way - but we do need to freeze "description of roles" somewhere and link to it from here - so people can make decisions about which roles/permissions to use for new features, etc.

ghickman: Added docstrings to the Roles, rebased changes, updated doc.

ghickman: I've not really tidied up the descriptions too much as I think they're pretty descriptive as they are but we might want to do a pass later.

jobserver/authorization/utils.py
ghickman: Caching has definitely been on my mind, but we don't need it yet (YAGNI-yet if you will) and I'm keen to avoid performance optimisations until they're necessary.  Our page loads are currently in the 30-100ms range (SQLite 🎉).

Relatedly I don't see any performance problems with building the full set.  I've tried out a for-loop with a short-circuit

```
    permissions = itertools.chain.from_iterable(r.permissions for r in roles)
    for permission in permissions:
        if permission == permission:
            return True
```

and don't find that more readable than a membership test.

bloodearnest: Well, for a start you wouldn't need the itertools, as you can just nest for loops. That is more readable to me.

What happens when there's a project context to check though? If you are issuing DB queries for those, building the full set becomes more expensive, no?

And, if we are not caching, we could be checking a permission this 10s of times per request, building the set each time?

Just seems like an easy win to me :man_shrugging:

jobserver/authorization/roles.py
```diff
@@ -0,0 +1,131 @@
...
+
+class OutputChecker:
+    """
+    Review output folders that have been proposed for release, approve and
+    publish.
+    """
+
+    permissions = [
+        check_output,
+        publish_output,
```
sebbacon, line 67: Based on how we've named the roles, I'd expect this to have only the permission `check_output`.   If the same user were to be able to publish *and* check outputs, then I'd expect them to have the two roles...?

ghickman: The description suggested it should be able to do both, do you want to leave it as is or update permissions and description?

sebbacon: So, thinking aloud, I don't think it makes sense for separate roles for output checking and output publication when the default permissions are as they are. 

I also don't think we know yet if the checker and the publisher are always going to be the same people or not. However we do know we want two independent output checks so it's feasible that we'll want to prevent *some* checkers from also doing the releases.

It's also feasible that releases will become automatic state transitions following the acquisition of a certain number of approvals.

Overall I think the problem is that the current wording "review output folders that have been proposed for release, approve and publish" is ambiguous: I'm going to guess the writer (could have been me!) meant "review output folders (that have been proposed for release)" rather than "review output folders (that have been proposed) for release".

So I think the permissions should be non-overlapping and the description should be "Review contents of output folders that have been proposed for public release" or similar.

However this is all up for debate so let me know if you think this is still unclear

ghickman: Based on our call:

* Add a SuperUserRole
* Comment that it is expected to go away, please don't use it further
* Comment that it exists as a transition so we can assign the correct roles to the relevant staff members

jobserver/authorization/admins.py
ghickman: Based on our call:

* Add a SuperUserRole
* Use it in place of CoreDeveloper to replace is_superuser
* Comment that it is expected to go away, please don't use it further
* Comment that it exists as a transition so we can assign the correct roles to the relevant staff members

sebbacon (approved):

PR #446, opened 2021-04-02
Add Contextual Roles

ghickman:
This adds a list of models (via dotted path) for each Role, as discussed in the `Local and Global Roles` section of the [Roles & Permissions Framework OFP](https://github.com/opensafely-core/roadmap/discussions/13).

It adds membership models for both Project and Org, which will both hold the User's Roles for the specific Org or Project.  The `created_by` field shows who invited that User, with the expectation that later we'll do more around inviting unregisterd users and linking that invitation to the Membership object somehow.

It ensures Roles are used with the correct model type inside RolesField, producing a list of failures when the type(s) are wrong.

When passing context to either `has_permission` or `has_role` the keys and value-types of the passed context are checked to help developers avoid unexpected failures.

An OrgCoordinator Role has been added so we can test the `org_converter` function with an Org-specific Role.


jobserver/migrations/0027_add_membership_org_and_project.py
```diff
@@ -0,0 +1,106 @@
...
+                    ),
+                ),
+            ],
+            options={
+                "unique_together": {("project", "user")},
+            },
+        ),
+        migrations.DeleteModel(
+            name="Membership",
+        ),
```
benbc, line 105: Is there no need to map this old data into the new model? Or am I missing the place where that happens?

ghickman: This table is empty!  There's some backstory to that…

When I added Orgs and Projects in #394 I realised we'd need a Membership table so added it there.  As part of the Onboarding work #394 was part of we realised Roles&Permissions needed to come first (hence [OFP#13](https://github.com/opensafely-core/roadmap/discussions/13)).  Then, when I was desiging that framework I realised Membership was a good place to hold Local Roles for a Project, but also an Org.  Since the old Membership table was empty I didn't bother trying to refactor it via migrations.

benbc: Well, if it's empty... :-)

sebbacon: @benbc, context you may be missing (apols if not) is that these migrations are code generated and consumed by the Django ORM and are generally considered a black box.

jobserver/authorization/roles.py
```diff
@@ -47,6 +56,20 @@ class OnboardingAgent:
...
 
+    models = [
+        "jobserver.models.User",
+    ]
+    permissions = []
+
+
+class OrgCoordinator:
+    """
+    The responsible party for an Org
```
inglesp, line 67: The definite article in the docstring suggests that there might only ever be one `OrgCoordinator` for a given organisation -- is this true?  If not, maybe change `The` to `A`?

jobserver/fields.py
```diff
@@ -2,6 +2,11 @@
 from django.utils.module_loading import import_string
 
 
+def dotted_path(cls):
+    """Get the dotted path for a given class [instance]"""
```
inglesp, line 6: I don't understand what `class [instance]` means here.

jobserver/models.py
```diff
@@ -371,6 +352,33 @@ def save(self, *args, **kwargs):
         return super().save(*args, **kwargs)
 
 
+class OrgMembership(models.Model):
+    created_by = models.ForeignKey(
+        "User",
+        on_delete=models.SET_NULL,
+        related_name="org_invitations",
```
inglesp, line 359: I think this is a bit ambiguous -- does `george.org_invitations.all()` return all the `OrgMembership`s that George has invited people to, or all the `OrgMembership`s that George has been invited to?  Given that we're modelling more than just an invitation here, what about something like `related_name="created_org_memberships"`?

jobserver/models.py
```diff
@@ -446,6 +454,33 @@ def save(self, *args, **kwargs):
         return super().save(*args, **kwargs)
 
 
+class ProjectMembership(models.Model):
+    created_by = models.ForeignKey(
+        "User",
+        on_delete=models.SET_NULL,
+        related_name="project_invitations",
```
inglesp, line 461: Same as above.

tests/jobserver/test_fields.py
```diff
@@ -136,3 +140,19 @@ def test_roles_field_to_python_with_empty_list():
...
+
+    class ProjectRole:
+        models = ["jobserver.models.ProjectMembership"]
+        permissions = []
+
+    msg = "Some roles could not be assigned to OrgMembership\n"
+    msg += " - ProjectRole can only be assigned to these models:\n"
+    msg += "   - jobserver.models.ProjectMembership"
+    with pytest.raises(ValueError, match=msg):
+        OrgMembershipFactory(org=org, user=user, roles=[ProjectRole])
```
inglesp, line 158: The code that's actually being tested here is obscured by the machinery of `OrgMembershipFactory`, and so it's not obvious to the casual reader what's being tested.

jobserver/authorization/converters.py
```diff
@@ -0,0 +1,15 @@
+from ..models import OrgMembership, ProjectMembership
+
+
+def org_converter(org, user):
```
inglesp, line 4: After reading the tests, I think a better name for this function would be `get_org_roles_for_user`.  Otherwise, it's not clear what the given org is being converted to.

inglesp: I'm not sure what you should call the module though.  Maybe these could live in `utils.py`, where they're used?

jobserver/authorization/utils.py
```diff
@@ -1,11 +1,102 @@
...
+    Ensure keys and types-of-values in context are correct
+
+    When passing in context to either has_permission or has_role we want to
+    help developers track down incorrect keys or incorrect values.
+
+    For example, given this call:
+
+        _validate_context(["project"], {"projects": project})
+
+    we want to tell the develoer that "projects" is an invalid key.
```
inglesp, line 17: https://jokes.one/joke/abcdefghijklmnoqrstuvwxyz

jobserver/authorization/utils.py
```diff
@@ -1,11 +1,102 @@
...
+        return
+
+    msg = "Some invalid values were used in the context:"
+    for key, value in incorrect_types.items():
+        expected = type_map[key]
+        msg += (
+            f'\n - key "{key}" got a {value.__name__}, expected a {expected.__name__}'
+        )
+
+    raise ValueError(msg)
```
inglesp, line 63: I can see how this all works, but it feels like quite a lot of work.  If instead you had `has_permission(user, permission)`, `has_permission_on_project(user, permission, project)`, and `has_permission_on_org(user, permission, org)`, I think you'd end up with simpler code.

jobserver/authorization/utils.py
```diff
@@ -1,11 +1,102 @@
...
+
+
+def _get_roles(user, **context):
+    """
+    Build up a set of Roles from the User and context
+
+    A User's Roles can come from various locations, and be local or global,
```
inglesp, line 70: local to a project or organisation?

inglesp (approved):
Have made some suggestions for how you could improve some names, and have a suggestion about changing the implementation slightly.

jobserver/authorization/utils.py
ghickman: There's section in the OFP on Local vs Global (can't deep link to it sorry 😬 https://github.com/opensafely-core/roadmap/discussions/13), does that clarify the difference between the two types?

To your specific question, local to both, and the expectation of the design is we'll have more objects which need "local" Roles as we move forwards.  It's the concept that a Role is tied to a very particular object, eg you might need the ProjectDeveloper role on ProjectA but we don't want you have that role on every Project.

tests/jobserver/test_fields.py
ghickman: Good call, added a comment to explain the test.

jobserver/authorization/utils.py
ghickman: A core goal of the OFP was to have minimal entry point functions (ideally one!) to the framework, this change would add O(N) entry points, one for each Object where local roles can be applied (Project, Org, etc) and one for each Object where global roles can be applied (User, and we're expecting Groups at some point too).

It also means a developer has to know where permissions are coming from to know which functions to call since Roles can exist in multiple places.  Take, for example George, a member of both OrgA and ProjectA.  ProjectA is part of OrgA.  A caller of `has_permission` can pass in both OrgA and ProjectA as context for the lookup, without having to know if they should look at the Org or Project specifically.

inglesp: You've clearly thought about this much more than I have, so happy to defer.

But I might have misunderstood something.  I'd assumed that when checking whether a user can do something to a project, you'd always pass that project in the context.  But are you saying that you can also pass that project's organisation in the context instead?  (Part of my problem is that I'm not looking at any real examples of how this is used.)

(Also, a general point: I'm not sure whether it's right to think about building a framework -- a framework has to be useful for lots of users across lots of applications, while here we're building a single application, with a small number of users.  As such, we don't have to offer something as general-purpose as a framework might have to.)

ghickman:
Failed test is from hadolint not being pinned.

PR #448, opened 2021-04-05
Roles & Permissions API

ghickman:
This adds an authenticated API for Users listing their roles and permissions, with context for any Local Roles (and thus Permissions) as part of the output.

The output for a given User is:

```
{
    "permissions": {
        "global": ["cancel-job", "run-job"],
        "orgs": [],
        "projects": [
            {
                "slug": "great-project",
                "permissions": [
                    "cancel-job",
                    "check-output",
                    "run-job",
                ],
            }
        ],
    },
    "roles": {
        "global": ["CoreDeveloper"],
        "orgs": [],
        "projects": [
            {
                "slug": "great-project",
                "roles": [
                    "ProjectCollaborator",
                ],
            }
        ],
    },
}
```

The output for `orgs` mirrors that of `projects` with a slug and related permissions/roles.

### DIFF
--- a/tests/jobserver/test_api.py
+++ b/tests/jobserver/test_api.py
@@ -10,10 +10,12 @@ from rest_framework.test import APIClient
 from jobserver.api import (
     JobAPIUpdate,
     JobRequestAPIList,
+    UserAPIDetail,
     WorkspaceStatusesAPI,
     get_backend_from_token,
     update_stats,
 )
+from jobserver.authorization import CoreDeveloper, OrgCoordinator, ProjectDeveloper
 from jobserver.models import Backend, Job, JobRequest, Stats
 from tests.jobserver.test_releases import make_release_zip
 
@@ -21,8 +23,13 @@ from ..factories import (
     BackendFactory,
     JobFactory,
     JobRequestFactory,
+    OrgFactory,
+    OrgMembershipFactory,
+    ProjectFactory,
+    ProjectMembershipFactory,
     ReleaseFactory,
     StatsFactory,
+    UserFactory,
     WorkspaceFactory,
 )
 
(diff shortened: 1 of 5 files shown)


ghickman:
Should this endpoint be authenticated?

It would hide the [roles&]permissions attached to a User, but is that important?

sebbacon:
Sorry, I don't have time for a full review here, but I don't quite understand the API.  Don't we need to say *what the context is* for context-specific permissions to make sense?

For example, I might have `run_job` permission on project A but not project B.

I expect the API either (a) to allow us to specify the project or org I'm interested in, or (b) if it's going to return everything, for it to include the identifiers for each project and org that local roles exist for. Or both!

ghickman:
I got the example very wrong 😂 

Fixed, TL;DR it's doing B!

tests/jobserver/test_api.py
```diff
@@ -524,6 +529,44 @@ def test_jobrequestapilist_success(api_rf):
...
+
+    request = api_rf.get("/", HTTP_AUTHORIZATION=backend.auth_token)
+    response = UserAPIDetail.as_view()(request, username=user.username)
+
+    assert response.status_code == 200
+
+    # permissions
+    permissions = response.data["permissions"]
+    assert permissions["global"] == ["cancel_job", "run_job"]
+    assert permissions["orgs"] == []
```
inglesp, line 548: Is it worth testing with an org permission?

tests/jobserver/test_models.py
```diff
@@ -565,6 +571,78 @@ def test_user_name_without_first_and_last_name():
...
+def test_user_get_all_permissions():
+    project = ProjectFactory()
+    user = UserFactory(roles=[CoreDeveloper])
+
+    ProjectMembershipFactory(project=project, user=user, roles=[ProjectDeveloper])
+
+    output = user.get_all_permissions()
+    expected = {
+        "global": ["cancel_job", "run_job"],
+        "orgs": [],
```
inglesp, line 584: Again, we should include a test for an org permission.

inglesp (approved):
Missing a couple of tests, but otherwise good.

Looks slightly funny exposing roles in CamelCase and permissions as snake_case, but maybe I should get out more.

PR #449, opened 2021-04-06
Selectively Show Cancel Button

### LINKED ISSUE #438: Selectively Show Cancel Job Button
We should hide (or maybe just disable?) the Cancel Job button when a Job is in the `failed` or `completed` states since the action is meaningless at that point.

We also need to make sure the View supporting the button has the same semantics.

ghickman:
This disables the cancel button when a Job has already finished, and makes the handler view a noop in that situation too.

Fixes #438

### DIFF
--- a/jobserver/templates/job_detail.html
+++ b/jobserver/templates/job_detail.html
@@ -154,7 +154,7 @@
 
       <button
         class="btn btn-block btn-danger"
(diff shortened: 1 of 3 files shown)


CarolineMorton (approved):
Looks fine to me.

PR #450, opened 2021-04-06
Cancel All Jobs in a JobRequest

### LINKED ISSUE #437: Cancel All Jobs in a JobRequest
Now that we can cancel single Jobs it would be useful to cancel all Jobs for a JobRequest.

This should be fairly easy since we know all the currently related Jobs and thus their actions.

We can hide the button (and disallow the action) when the JobRequest is in the `failed` or `completed` states, or has no Jobs.

ghickman:
This bring the mighty Cancel button to the JobRequestDetail page.

fixes #437

### DIFF
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -231,6 +231,9 @@ class JobRequest(models.Model):
 
         return last_job.completed_at
 
+    def get_cancel_url(self):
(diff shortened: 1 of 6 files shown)


bloodearnest (commented):
And there was much rejoicing \o/

bloodearnest (approved):

PR #462, opened 2021-04-14
Prepare for Migrating to Projects

ghickman:
This implements the changes and tools for the [Initial Roles & Permissions Roll Out OFP](https://github.com/opensafely-core/roadmap/discussions/15).

The Projects and Memberships expected by the OFP are missing.  Building them inside a migration breaks all the tests since the relevant Workspaces and Users don't exist.  Instead some commands have been created to help simplify the creation of Projects from existing Workspaces and link Users (with Roles) to those Projects.


jobserver/management/commands/create_project_from_workspace.py
```diff
@@ -0,0 +1,67 @@
...
+            workspace.name.removesuffix("-main")
+            .removesuffix("_main")
+            .removesuffix("-master")
+            .removesuffix("_master")
+        )
+
+        # generate a name from the slug
+        name = " ".join(word.capitalize() for word in slug.split("-"))
+
+        project = Project.objects.create(org=org, name=name, slug=slug)
```
rebkwok, line 30: `get_or_create`?  Is it possible that there could be two workspaces for the same org with names e.g. "my_workspace" and "my_workspace_main", and this could try to create a project with a duplicate slug?  Should those two workspaces legitimately be assigned to the same project?

rebkwok (approved):
LGTM.  I had a couple of questions which might just be due to my lack of total understanding of what workspaces/projects are, and ignorable comments on variable naming :)

jobserver/management/commands/create_project_from_workspace.py
ghickman: It's possible but I think unlikely, and I'd rather this exploded when I used it so I knew there was an issue!  I'm passing the Org slug in via the CLI so if there was some clash it would at least go to the correct Org.

PR #465, opened 2021-04-15
Hadolint fixups

### LINKED ISSUE #464: Pin Hadolint
A recent change to the hadolint GitHub Action caused our checks to fail as new tests were added to Hadolint.

Can we pin hadolint?

Either way we should fix the warnings.

ghickman:
This pins the hadolint action and fixes the warnings raised in the latest version.

Fixes #464

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,8 @@ ENV PIP_NO_CACHE_DIR=1 \
     PYTHONDONTWRITEBYTECODE=1 \
     PYTHONUNBUFFERED=1
 
+WORKDIR /app
+
 # Only requirements to cache them in docker layer so we can skip package
 # installation if they haven't changed
 COPY requirements.txt .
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -7,17 +7,18 @@ ENV PIP_NO_CACHE_DIR=1 \
     PYTHONDONTWRITEBYTECODE=1 \
     PYTHONUNBUFFERED=1
 
+WORKDIR /app
+
 # Only requirements to cache them in docker layer so we can skip package
 # installation if they haven't changed
 COPY requirements.txt .
-RUN pip install --requirement requirements.txt
+RUN pip install --no-cache-dir --requirement requirements.txt
```
bloodearnest, line 15: This is fine, it makes the a bit image smaller.

If you really want some fun, we can enable buildkit and get a persistent cache, which would speed up rebuilding the docker image locally. But probably not worth it.

bloodearnest (approved):

Dockerfile
ghickman: Definitely something to add later if we get anyone using the docker-compose flow in anger I think!

PR #474, opened 2021-04-23
Invite Users to a Project

ghickman:
This adds the pages necessary to invite existing users of job-server to a Project, cancel a pending invitation, and accept an invitation, creating a membership for the relevant User.

I've opted to move Users through an invite->accept flow to avoid potential abuse.  As an added bonus it helps signpost to invitees that the inviter would like to add them to a Project, avoiding more banner notifications.  Emails use a static subject line based on the inviter's name and the Project name to allow for easy filtering.

A new ProjectSettings page is available to Users who can manage members of a given Project.  I expect we'll change how we gate access to this page and the button to access it as Projects do more over time but for now it's only for member management.

**ProjectDetail linking to ProjectSettings**

Selecting a User to invite uses a [`select2`](https://select2.org/) control to make searching for Users easier.  This list is filtered down to Users who are not already members or have pending invites.

**Example of selecting a User**

---

The Settings page lists Members and Pending Invitations.  Invitations are Pending when they have no linked ProjectMembership.

**Pending Invitations**

(ignore the Members list UI for now, I'll address it and the inability to set/modify a Members Roles in a subsequent PR)

In the screenshot above you can see Peter and Dave have been invited to the Project.  The email they would be sent contains a link with a signed hash of the Invite PK to the Invite accept page.  Invites can only be accepted by the User they were created for, and they can only be accepted once.

A ProjectCollaborator can cancel an invite if necessary, which deletes the invite.


jobserver/views/projects.py
```diff
@@ -99,6 +101,35 @@ def post(self, request, *args, **kwargs):
...
+        can_manage_members = has_permission(
+            self.request.user,
+            "manage_project_members",
+            project=membership.project,
+        )
+        if can_manage_members:
+            membership.delete()
+        else:
+            messages.error(
+                request, "You do not have permission to remove Project members."
```
inglesp, line 175: I don't think "Project" should be capitalised here.  I understand why you have, but I think model names should only be capitalised in code (obvs) and developer docs (maybe), but not in anything user-facing.

jobserver/models.py
```diff
@@ -468,6 +468,46 @@ def save(self, *args, **kwargs):
...
+    """
+    Models an Invitation to join a Project
+
+    If accepted an Invite is linked to a ProjectMembership.
+    """
+
+    created_by = models.ForeignKey(
+        "User",
+        on_delete=models.SET_NULL,
+        related_name="created_project_invites",
```
inglesp, line 482: For consistency, can this be `created_project_invitations`?

jobserver/templates/project_settings.html
```diff
@@ -70,4 +70,25 @@ <h3>Members</h3>
...
+    <h4>Pending</h4>
+    <ul class="list-unstyled">
+      {% for invite in invitations %}
+      <li class="mb-3">
+        <strong>{{ invite.user.name }}</strong>
+      </li>
+      {% endfor %}
+    </ul>
+    {% else %}
+    <p class="mt-5">There are no pending Invitations for this Project.</p>
```
inglesp, line 142: Same point about capitalisations.

jobserver/views/projects.py
```diff
@@ -131,22 +131,43 @@ def post(self, request, *args, **kwargs):
...
-        can_manage_members = has_permission(
+        self.can_manage_members = has_permission(
             self.request.user,
             "manage_project_members",
-            project=project,
+            project=self.project,
+        )
+        return super().dispatch(request, *args, **kwargs)
+
+    def form_valid(self, form):
```
inglesp, line 201: Are we missing a check here that the user has permission to create the invitations?

tests/jobserver/views/test_projects.py
```diff
@@ -283,11 +284,57 @@ def test_projectsettings_success(rf, superuser):
...
+    request.user = superuser
+
+    response = ProjectSettings.as_view()(
+        request, org_slug=org.slug, project_slug=project.slug
+    )
+
+    assert response.status_code == 302
+    assert response.url == project.get_settings_url()
+
+    assert ProjectInvitation.objects.filter(project=project).count() == 2
```
inglesp: You might like to pinch the [`assert_difference` function from opencodelists](https://github.com/opensafely-core/opencodelists/blob/master/opencodelists/tests/assertions.py).

jobserver/views/projects.py
```diff
@@ -153,11 +154,30 @@ def dispatch(self, request, *args, **kwargs):
...
+                failed_to_invite.append(user)
+
+        if failed_to_invite:
+            # tell the user about each failed invite in a single message
+            users = [f"<li>{u.username}</li>" for u in failed_to_invite]
+            users = f"<ul>{''.join(users)}</ul>"
+
+            messages.error(
+                self.request,
+                f"<p>Failed to invite {len(failed_to_invite)} User(s):</p>{users}<p>Please try again.</p>",
```
inglesp, line 227: Is this vulnerable to XSS?

jobserver/templates/emails/project_invite.txt
```diff
@@ -0,0 +1,5 @@
+You have been invited to join the {{ project_name }} by {{ inviter_name }}.
```
inglesp, line 1: This should mention OpenSAFELY.

jobserver/emails.py
```diff
@@ -25,3 +25,22 @@ def send_finished_notification(email, job):
...
+    context = {
+        "inviter_name": invite.created_by.name,
+        "project_name": project.name,
+        "url": f.url,
+    }
+
+    send(
+        to=email,
+        sender="notifications@jobs.opensafely.org",
+        subject=f"[os] Project Invite from {invite.created_by.name} to {project.name}",
```
inglesp, line 43: Will an invited user know what [os] is?  Maybe change to [OpenSAFELY]?

jobserver/views/projects.py
```diff
@@ -153,11 +154,30 @@ def dispatch(self, request, *args, **kwargs):
...
+                with transaction.atomic():
+                    invite = ProjectInvitation.objects.create(
+                        created_by=self.request.user,
+                        project=self.project,
+                        user=user,
+                    )
+                    send_project_invite_email(
+                        user.notifications_email, self.project, invite
+                    )
+            except Exception:
```
inglesp, line 217: Can we log the exception?

inglesp (changes requested):
Very easy to review, and generally good -- but there are a couple of security concerns that need addressing before this can be merged.  (There's one permissions check that I think we're missing, and a possible XSS vulnerability.)

There's also a (non-shipstopping) Improvement to do with capitalising certain Nouns in user-visible Text.  It all feels [a bit eighteenth Century](https://ask.metafilter.com/29691/What-is-the-History-of-English-Capitalization).

jobserver/emails.py
ghickman: Oh, hmm, this was copying the notification email subject prefix… but now I think about it, that's not really the same as an invite.

jobserver/views/projects.py
ghickman: I don't _think_ so…

`users` coming from `form.cleaned_data` is a QuerySet of User instances so even if someone managed to push modified User IDs from the form the backend, it's still only using data from the database which GitHub gave us.

Unless I'm missing something else?

jobserver/views/projects.py
ghickman: Yes… 😬

jobserver/views/projects.py
inglesp: We're (a) trusting that GitHub doesn't allow usernames with funny characters in and (b) assuming that we never get usernames from anywhere else.

(a) is true now, and is likely to be true in future, but given how easy this is to protect against, I think we should do so now.

jobserver/templates/project_settings.html
```diff
@@ -0,0 +1,59 @@
...
+            {% endfor %}
+          </ul>
+        </div>
+      </li>
+      {% endfor %}
+    </ul>
+    {% else %}
+    <p class="mt-5">
+      There are no projects for this Organisation yet, click "Register Project"
+      to create one.
```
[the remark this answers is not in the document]
ghickman: copy/paste fail 😬 thanks!

jobserver/views/projects.py
ghickman: Fair point!

inglesp (approved):
VG.

PR #481, opened 2021-04-27
Update or Create ProjectMemberships in link_user_to_project

ghickman:
This switches to `update_or_create` in the `link_user_to_project` script so we can change the Roles of an existing Project member without having to delete them first.


sebbacon (approved):
I can never understand why they picked `defaults` as the argument name as opposed to something like `then_update_with=`!

ghickman:
I suspect it's to copy the `get_or_create` API where it makes sense… whereas it gets me every time with this one!

PR #482, opened 2021-04-28
Link Multiple Workspaces to Project

ghickman:
Some workspace names normalise in such a way that they end up with the same Project, during initial roll-out testing it's ideal to combine them into the same Project (famous last words probably).

### DIFF
--- a/jobserver/management/commands/create_project_from_workspace.py
+++ b/jobserver/management/commands/create_project_from_workspace.py
@@ -27,7 +27,7 @@ class Command(BaseCommand):
         # generate a name from the slug
         name = " ".join(word.capitalize() for word in slug.split("-"))
 
-        project = Project.objects.create(org=org, name=name, slug=slug)
+        project, _ = Project.objects.get_or_create(org=org, name=name, slug=slug)
 
         # tell the User what was made and where they can view it
         f = furl(settings.BASE_URL)
--- a/jobserver/management/commands/create_project_from_workspace.py
+++ b/jobserver/management/commands/create_project_from_workspace.py
@@ -32,9 +32,7 @@ class Command(BaseCommand):
         # tell the User what was made and where they can view it
         f = furl(settings.BASE_URL)
         f.path = project.get_absolute_url()
-        self.stdout.write(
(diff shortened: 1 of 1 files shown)


jobserver/management/commands/create_project_from_workspace.py
```diff
@@ -27,14 +27,12 @@ def _create_project(self, *, org, workspace):
         # generate a name from the slug
         name = " ".join(word.capitalize() for word in slug.split("-"))
 
-        project = Project.objects.create(org=org, name=name, slug=slug)
+        project = Project.objects.get_or_create(org=org, name=name, slug=slug)
```
rebkwok, line 30: `get_or_create` returns a tuple, so you need `project, _ = Project...` here
(or `project, new = ` if it's useful to log whether it created a new project or merged with an existing one)

jobserver/management/commands/create_project_from_workspace.py
```diff
@@ -27,14 +27,12 @@ def _create_project(self, *, org, workspace):
...
 
         # tell the User what was made and where they can view it
         f = furl(settings.BASE_URL)
         f.path = project.get_absolute_url()
-        self.stdout.write(
-            f"Name: {project.name}\nSlug: {project.slug}\nURL:  ({f.url})"
-        )
+        self.stdout.write(f"Name: {project.name}\nURL:  {f.url}")
```
rebkwok, line 35: Do we just not care enough about the slug?  Or is the url the same so it's repetitive?

rebkwok (changes requested):

jobserver/management/commands/create_project_from_workspace.py
ghickman: 🤦‍♂️ Thanks!

jobserver/management/commands/create_project_from_workspace.py
ghickman: Both, I'm only outputting them so I can do a quick check they look useful and get to the created Project quickly, and realised the slug is right there in the URL…

rebkwok (approved):

Issue #433, opened 2021-03-22
Make the deploy Action Blocking


ghickman:
When we merge multiple subsequent PRs, eg on Monday mornings for Dependabot, we end up with failing deploys because dokku will [correctly] only execute one deploy at a time.

[It seems we're not alone](https://github.community/t/how-to-limit-concurrent-workflow-runs/16844).

Mentioned in that thread, the [turnstyle](https://github.com/softprops/turnstyle) action looks useful.

madwort:
same here https://github.com/ebmdatalab/datalab-team-manual/issues/27 - was considering https://github.com/styfle/cancel-workflow-action but not tried

madwort:
alternative idea: deploy on schedule rather than on push https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule (as deploy is `git push`, will be virtually a no-op if nothing merged?)

StevenMaude:
Looking up something independently and spotted the new (2021-04-19) [`concurrency`](https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/) feature (albeit in beta), which I suspect might fix this.

madwort:
great news, `concurrency` seems like it'll do exactly what we need.

ghickman:
Ok so in initial testing this mostly worked.

Merging #488, #487, & #486 (in that order) resulted in the expected 3 runs ([black](https://github.com/opensafely-core/job-server/actions/runs/806670558), [responses](https://github.com/opensafely-core/job-server/actions/runs/806670740), & [pytest-mock](https://github.com/opensafely-core/job-server/actions/runs/806670998).

The run for black ran without issue, the run for responses was cancelled because of a "higher priority job", and the run for pytest-mock was queued at the `deploy` job, until the deploy of black completed.

So this worked for the happy case as expected but I've yet to find an explanation for why the responses merge got cancelled on me.

madwort:
The docs say say "[Any previously pending job or workflow in the concurrency group will be canceled](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency)" (even if you don't have `cancel-in-progress` turned on), so it probably went:

* merge 488
* start deploy for 488
* merge 487
* queue 487
* merge 486
* queue 486
* => cancel 487
* end deploy for 488
* start deploy for 486
* end deploy for 486

[...]

PR #490, opened 2021-05-03
Limit Concurrency on the Deploy Action

ghickman:
This sets up [job-level concurrency](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idconcurrency) for the `deploy` job.

I'm not quite sure how to test this does what we want without just using it…

My hope is it blocks just the `deploy` job in the CI workflow on the `main` branch, however I could see it blocking the entire CI workflow while another one ran on `main`, which I don't think would be the worst, just a bit slower than we'd like.

However if this blocks PR branches from running CI then we'll have to roll it back.

Ref #433

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,6 +119,8 @@ jobs:
 
     if: github.ref == 'refs/heads/main'
 
+    concurrency: deploy-production
+
     steps:
       - uses: actions/checkout@v2
         with:


madwort:
Agree that from reading the docs, it sounds like it will do the right thing (deploy job blocks other deploy jobs but not any other jobs) - let's try it. 

One thing that might be mildly annoying though is `When concurrency is specified at the job level, order is not guaranteed for jobs or runs that queue within 5 minutes of each other.` - we might still get some noise from failed deploy jobs if a later merge jumps the queue & gets deployed first.

madwort (approved):

PR #491, opened 2021-05-03
Set Roles When Inviting Users

### LINKED ISSUE #483: Select Roles when Inviting Users
Currently we allow Users to be invited to a Project, making them members when they accept the invitiation.  It was expected that a ProjectCoordinator would then return to the Settings page and set the Roles for that Member.

However, that requires the coordinator to return to that page for each User they add.  Instead lets let a coordinator set the Roles on an Invitation.

ghickman:
This sets up Project invitations to have Roles set when inviting a User to a Project, copying those Roles to the accepted Membership if the User accepts.

It makes use of the RolesForm from #489 and copies the HTML for the Form into the Project settings template (I'd like to abstract that at some point but this is not that point).

Fixes #483


jobserver/templates/project_settings.html
```diff
@@ -85,27 +86,64 @@ <h3>Invitations</h3>
...
           {% endfor %}
         </select>
 
+        {% for error in form.users.errors %}
+        <p class="text-danger">{{ error }}</p>
+        {% endfor %}
+      </div>
+
+      <div class="form-group mb-3">
+        <p>Select one or more Roles for the User(s) you are inviting.</p>
```
inglesp, line 110: My personal preference is not to capitalise Role and User here, but I'll shut up about this if you like.

tests/jobserver/views/test_projects.py
```diff
@@ -54,6 +54,7 @@ def test_projectacceptinvite_success(rf):
     invite.refresh_from_db()
     assert invite.membership is not None
     assert invite.membership.project == project
+    assert invite.membership.roles == invite.roles
```
inglesp, line 57: We never assert anything about what `invite.roles` is.  Is it definitely set to a non-empty list?

inglesp (approved):
One of the tests could be tightened up, but apart from that, looks good.

jobserver/templates/project_settings.html
ghickman: You're literally the only person keeping me on the straight and narrow with this, please don't stop! 😂

PR #495, opened 2021-05-04
Disable FLoC

ghickman:
This disables Google's FLoC for the site using [django-permissions-policy](https://pypi.org/project/django-permissions-policy/).

[This article](https://adamj.eu/tech/2021/05/04/disabling-floc-googles-new-advertising-technology/) gives a great overview of why we might want this change.


StevenMaude (approved):

StevenMaude:
I think there's also a question of "do we want to do this across other services we manage?" too.

PR #497, opened 2021-05-04
Use Hashes with pip[-compile]

ghickman:
This switches to generating and using hashes for the Python dependencies.

I've not scripted the generation of dependencies because I expect the `split-dependencies` branch to bring that in.


bloodearnest (approved):
+1

I think this is slightly better than currently, in that it will start failing to build if a PyPI user deletes and uploads a new wheel.

But it's only as secure as our updating process. How is a hash change failure surfaced? Are we going to fix it by just updating the requirements.txt, and thus trust the new hash?

Food for thought.

bloodearnest:
> +1
> 
> I think this is slightly better than currently, in that it will start failing to build if a PyPI user deletes and uploads a new wheel.
> 
> But it's only as secure as our updating process. How is a hash change failure surfaced? Are we going to fix it by just updating the requirements.txt, and thus trust the new hash?
> 
> Food for thought.

Also, I realised that I think this is of little benefit if we are not also pinning in requirements..n

bloodearnest:
do, sorry, meant to cancel out of a comment, but closed it instead :facepalm:

ghickman:
> deletes and uploads a new wheel

You haven't been able to replace an existing file on PyPI for many years now!  You can delete a file from a version, but you can't replace that file in the given version.

Some more info here: https://github.com/pypa/packaging-problems/issues/74#issuecomment-137744015 (linked to a GitHub issue because the mailing list link is broken for me but it might just be a transient thing)

PR #499, opened 2021-05-05
Install Python Dependencies after APT Dependencies

ghickman:
This swaps the ordering of our dependency installs, putting the less frequent APT ones first, and the more frequent Python ones second, saving us some unnecessary layer invalidation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,15 +9,15 @@ ENV PIP_NO_CACHE_DIR=1 \
 
 WORKDIR /app
 
-# Only requirements to cache them in docker layer so we can skip package
-# installation if they haven't changed
-COPY requirements.txt .
-RUN pip install --no-cache-dir --require-hashes --requirement requirements.txt
-
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


benbc (commented):
I'm not actually sure that merging the `RUN` commands in a good idea in this case. They are semantically separate things going on. Combining them makes the `Dockerfile` a bit harder to understand; it also means that the `pip install` will be re-run unnecessarily if we change the APT packages we're installing.

(Whereas combining all the APT-related operations into a single `RUN` is definitely a good thing to do.)

ghickman:
@benbc – good point.  Given we've got the apt requirements pinned I'll put that before the python requirements which will also make the RUNs non-consecutive.

ghickman:
@benbc – updated with new ordering.

benbc (approved):
Oh, that's a much better way of doing it. I'm sure the Python dependencies are going to change more often than the API package(s) aren't they?

ghickman:
Definitely!  It turns out I only added the APT section to the dependencies for `litecli`, which is probably why I added that section after the Python section initially.

PR #500, opened 2021-05-05
Install litecli in Production

ghickman:
This moves litecli back into the production requirements since it's so useful on the server.

### DIFF
--- a/requirements.dev.in
+++ b/requirements.dev.in
@@ -5,7 +5,6 @@ black
 factory_boy
 flake8
 isort
-litecli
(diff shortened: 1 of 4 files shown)


benbc (approved):
@ghickman Am I right in thinking that before your previous change to split the `requirements.in` file, the stuff noted as "dev requirements" was also being installed in production? So the fact that `litecli` appeared in that list was an error/oversight?

ghickman:
@benbc – exactly that, yes.

PR #502, opened 2021-05-07
Add Workspace -> Project Linking Command

ghickman:
Some Projects need more than one Workspace attaching to them, this simplifies my ability to do that while migrating Users to the Project lifestyle.  I expect this command to go away once the migration is done.

### DIFF
--- a/jobserver/management/commands/link_workspace_to_project.py
+++ b/jobserver/management/commands/link_workspace_to_project.py
@@ -1,0 +1,36 @@
+import sys
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from furl import furl
+
+from jobserver.models import Project, Workspace
+
+
+class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


jobserver/management/commands/link_workspace_to_project.py
```diff
@@ -0,0 +1,36 @@
...
+        parser.add_argument("project")
+        parser.add_argument("workspace")
+
+    def handle(self, *args, **options):
+        project_slug = options["project"]
+        try:
+            project = Project.objects.get(slug=project_slug)
+        except Project.DoesNotExist:
+            self.stderr.write(f"Unknown Project slug: {project_slug}")
+            sys.exit(1)
```
evansd, line 21: Very disciplined of you! I would've just let the exception bubble up here

evansd (approved):
Nice. Writing little helper scripts like this is such a good move which I always underuse

PR #515, opened 2021-05-14
Handle Release Notifications

ghickman:
This sets up the API for handling Release Notifications from osrelease, and pushing them into a Slack channel.

Example (created by running locally):

Related PR: opensafely-core/output-publisher#21

### DIFF
--- a/jobserver/api.py
+++ b/jobserver/api.py
@@ -9,10 +9,13 @@ from django.utils import timezone
 from first import first
 from rest_framework import serializers
 from rest_framework.exceptions import NotAuthenticated, ValidationError
-from rest_framework.generics import ListAPIView
+from rest_framework.generics import CreateAPIView, ListAPIView
 from rest_framework.parsers import FileUploadParser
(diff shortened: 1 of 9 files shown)


bloodearnest (commented):
This looks good.

However, I'm not convinced we need a db table of ReleaseNotifications.

I assume that its added for an audit trail.

However, we have actual Releases table, which while we're not using in production yet, would serve as our audit long terms, as it represents the actual release.

So, ReleaseNotifications would only be a temporary audit trail, until we swtich to handling releases properly in job-server. So I suspect it's not worth adding here, and we'll just deal with the fact that slack notifications may occasionally get lost until we've moved over.

If you are concerned about lost slack messages, you could return 502 if the push to slack fails, and have the client retry?

ghickman:
@bloodearnest – agreed, force of [django-]habit!  I've removed the model, and switched the API serializer to a plain one.

bloodearnest (approved):

PR #516, opened 2021-05-14
Remove Zombification

ghickman:
Now that job-runner v2 is well and truly deployed and tested we no longer need to deal with stale Job statuses.

### DIFF
--- a/tests/jobserver/views/test_job_requests.py
+++ b/tests/jobserver/views/test_job_requests.py
@@ -10,7 +10,6 @@ from jobserver.views.job_requests import (
     JobRequestCancel,
     JobRequestDetail,
     JobRequestList,
-    JobRequestZombify,
 )
 
 from ...factories import (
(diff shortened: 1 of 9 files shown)


tests/jobserver/views/test_job_requests.py
```diff
@@ -143,7 +142,6 @@ def test_jobrequestdetail_with_superuser(rf, superuser):
     response = JobRequestDetail.as_view()(request, pk=job_request.pk)
 
     assert response.status_code == 200
```
sebbacon, line 144: Do we need some other "superusery" test here to replace the one we're losing?

sebbacon (approved):
So much yummy redness!

tests/jobserver/views/test_job_requests.py
ghickman: No, even better, we can remove more code 🎉 (rebased into original commit)

PR #534, opened 2021-05-24
Add Backend Membership

ghickman:
This adds Backend "memberships", where a membership models the ability for a User to run jobs on a given Backend.

This allows us to remove the SuperUser checks for adding backends as part of the form and rendering them in the template, instead setting an initial value and hiding that part when there's only 1 backend.  Now that backends are always part of the form we can also rely on always them for validation on submission too.


sebbacon (commented):
Not a full review as I don't have time, sorry! - so just a quick observation.

In the fullness of time, a users' access to data will be via DPIA and other IG-related approvals; and a backend is currently our most granular entity to which we can attach this kind of permission-based information.

It seems likely that these permissions will be granted in the context of a Project, as this is the unit to which our IG process is (will be) applied.  ("All members of this project have access to these backends")

This it might be that a Project has many Backends, rather than a User has many Backends; I don't know, from a cursory think, if it might also make sense to use permissions in the view logic somehow as this is about what you're *allowed* to do. But that's a bit hand-wavy.

ghickman:
@sebbacon – agreed, Project is the end goal to tie all of this together, it fits with how approvals work, etc.  This is a stop-gap (but neatly fixes a problem we were seeing for existing researchers) until we can move to that model which requires some more ground work first 🙂

jobserver/forms.py
```diff
@@ -32,10 +32,15 @@ def __init__(self, actions, *args, backends=None, **kwargs):
...
         )
 
-        if backends:
-            self.fields["backend"] = forms.ChoiceField(
-                choices=backends, widget=forms.RadioSelect
-            )
+        try:
+            initial = backends[0][0]
+        except IndexError:
+            initial = None
```
inglesp, line 38: When would this `IndexError` be hit?  If a user has no backends, should they ever see this form?  And if they shouldn't see this form but do somehow, should it blow up?

inglesp (approved):

jobserver/forms.py
ghickman: > When would this IndexError be hit?

when there are no backends 😜🥁

This can probably go away once the current rush of changes is done!

PR #535, opened 2021-05-25
Add Model Docstrings

ghickman:
This adds docstrings to each model to show their intended use.

### DIFF
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -67,6 +67,8 @@ class BackendManager(models.Manager):
 
 
 class Backend(models.Model):
+    """A job-runner instance"""
+
     name = models.TextField(unique=True)
     display_name = models.TextField()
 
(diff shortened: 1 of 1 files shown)


jobserver/models.py
```diff
@@ -602,6 +628,8 @@ def get_remove_url(self):
 
 
 class Release(models.Model):
+    """Outputs released from a Workspace"""
```
ghickman, line 631: @bloodearnest – would love a quick check on this from you since you built the releases bits!

CarolineMorton (approved):

evansd (approved):

jobserver/models.py
bloodearnest: Maybe "Review and redacted outputs from a Workspace"?

ghickman: Did you mean "Reviewed" here?

ghickman: Assuming yes and updated commit with better words, thanks!

bloodearnest: :facepalm:

PR #536, opened 2021-05-26
Add pyupgrade

ghickman:
This adds pyupgrade to catch syntax changes so I don't have to.


bloodearnest (approved):
This is a really interesting tool!

PR #537, opened 2021-05-27
Handle Multiple GitHub Orgs

ghickman:
This sets up the ability for our Orgs to be tied to one or more GitHub Organisations.  It updates all the calls to GitHub to take an org name, expecting that any given GitHub Organisation will have a corresponding token in the environment derived from its name.

While this adds support for _multiple_ GitHub Organisations it does not make use of them.  We have no immediate need for Users to be able to see more than one Organisation but we do need to support more than `opensafely` as quick as possible.  I have a spike which I think will work out to handle trying N orgs for each call but it needs more work and we want this set of changes in ASAP.


tests/conftest.py
```diff
@@ -27,3 +27,20 @@ def fixture_configure_structlog(log_output):
...
+
+@pytest.fixture
+def user():
+    """
+    Generate a User instance with useful things attached
+
+    We almost always want a User to be part of an OpenSAFELY Org and have that
+    Org tied to a GitHub Organisation.
+    """
+    org = OrgFactory()
```
inglesp, line 40: Make explicit that this org is OpenSAFELY?

tests/jobserver/views/test_job_requests.py
```diff
@@ -110,14 +107,19 @@ def test_jobrequestcancel_unknown_job_request(rf):
...
 @responses.activate
-def test_jobrequestdetail_with_authenticated_user(rf):
+def test_jobrequestdetail_with_authenticated_user(rf, user):
     job_request = JobRequestFactory()
-    user = UserFactory(is_superuser=False, roles=[])
+
+    # ensure the user doesn't have any elevated privileges
+    user.is_superuser = False
+    user.roles = []
+    user.save()
```
inglesp, line 116: worth making a fixture for this?

tests/jobserver/views/test_workspaces.py
```diff
@@ -166,7 +163,8 @@ def test_workspacecreate_post_success(rf):
     request = rf.post(MEANINGLESS_URL, data)
     request.user = user
 
-    membership_url = f"https://api.github.com/orgs/opensafely/members/{user.username}"
+    gh_org = user.orgs.first().github_orgs[0]
+    membership_url = f"https://api.github.com/orgs/{gh_org}/members/{user.username}"
```
inglesp, line 167: how about a `user.url_for_gh_org(gh_org)` method?

jobserver/admin.py
```diff
@@ -8,6 +9,9 @@
 from .models import Org, Project, User
 
 
+logger = structlog.get_logger(__name__)
```
inglesp, line 12: I can't see where this is used.  Is there some magic going on here?

tests/jobserver/views/test_workspaces.py
inglesp: and maybe also a `gh_org` fixture?

inglesp (approved):

jobserver/admin.py
ghickman: Nope, just forgot to remove it 😬

tests/jobserver/views/test_job_requests.py
ghickman: Probably not, it's going away soon 🤞

PR #542, opened 2021-06-01
Handle unexpected filter values on the job listing page

ghickman:
This returns an unfiltered list of jobs when the filter value is unexpected.

### DIFF
--- a/jobserver/views/job_requests.py
+++ b/jobserver/views/job_requests.py
@@ -30,6 +30,10 @@ def filter_by_status(job_requests, status):
         "pending": lambda r: r.status.lower() == "pending",
         "succeeded": lambda r: r.status.lower() == "succeeded",
     }
+
+    if status not in status_lut:
+        return job_requests
+
     func = status_lut[status]
     return list(filter(func, job_requests))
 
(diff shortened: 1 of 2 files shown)


jobserver/views/job_requests.py
```diff
@@ -30,6 +30,10 @@ def filter_by_status(job_requests, status):
         "pending": lambda r: r.status.lower() == "pending",
         "succeeded": lambda r: r.status.lower() == "succeeded",
     }
+
+    if status not in status_lut:
+        return job_requests
+
     func = status_lut[status]
     return list(filter(func, job_requests))
```
inglesp, line 38: Is this equivalent to:

```
def filter_by_status(job_requests, status):
    if status not in ["failed", "running", "pending", "succeeded"]:
        # comment about why we might see an unexpected status
        return job_requests

    return [r for r in job_requests if r.status.lower() == status]
```

inglesp (approved):
Fine.  The dict of lambdas made me think something complicated was going on, but I think you might be able to replace it with a list comprehension.

jobserver/views/job_requests.py
ghickman: Yep, that's exactly it, and that's much nicer, thanks!

inglesp: :cartwheeling:

PR #544, opened 2021-06-01
Add Simple Deployment Checks

ghickman:
This runs Django's checks framework in deployment mode, and adds a simple check to dokku to grab `/`.

Django's checks identified a couple of settings we needed to set, but I've silenced a couple of TLS-related ones since they don't apply with our infra.

I haven't yet worked out how to run the checks with an authenticated User so I'm leaving those for another time.


jobserver/settings.py
```diff
@@ -39,7 +38,7 @@
 
 BASE_URL = env.str("BASE_URL", default="http://localhost:8000")
 
-ALLOWED_HOSTS = [furl(BASE_URL).host]
+ALLOWED_HOSTS = ["*"]
```
rebkwok, line 41: Are we assuming this is OK because nginx/cloudflare will deal with the host check?  I liked the setting to base url host :woman_shrugging:

entrypoint.sh
```diff
@@ -2,6 +2,7 @@
 
 set -euo pipefail
 
+./manage.py check --deploy
```
bloodearnest, line 5: My previous experience means I'm cautious about hardcoding this check into the *entrypoint*. If it fails, we then *cannot* start up the service. While that's technically valid and other apps do the same, practically it can be problematic. However, it's probably fine, given the low uptime situation we are in.

However, in case there's an easy way to do it, here's my reasoning:

There are are all kinds of reasons a why a check could give a false negative. There will be times when we want to just bring job-server anyway, even though we know a particular check has failed. Hardcoding it here makes that difficult, especially in a docker world.

Ideally, I'd prefer to run `manage.py check` at the same point that CHECKS is run, if dokku supports that.

jobserver/settings.py
```diff
@@ -174,6 +173,19 @@
...
+# https://docs.djangoproject.com/en/3.2/topics/checks/
+SILENCED_SYSTEM_CHECKS = [
+    "security.W004",  # (SECURE_HSTS_SECONDS) TLS is handled by Nginx
+    "security.W008",  # (SECURE_SSL_REDIRECT) HTTPS redirection is handled by CloudFlare
+]
+
+# Security
+# https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-CSRF_COOKIE_SECURE
+CSRF_COOKIE_SECURE = not DEBUG
+SESSION_COOKIE_SECURE = not DEBUG
```
bloodearnest, line 186: Hmm, so DEBUG is entirely possibly to set in prod, right?

Does `manage.py check --deploy` error if DEBUG is set?

rebkwok (approved):
just a question about ALLOWED_HOSTS

jobserver/settings.py
ghickman: > DEBUG is entirely possibly to set in prod, right?

Yes

> Does manage.py check --deploy error if DEBUG is set?

No it raises a warning, which doesn't cause `manage.py check --deploy` to return a failed error code.

entrypoint.sh
ghickman: I definitely see your point here but didn't see a better place to put it 😬

CHECKS is effectively "curl $url, optionally check for $sting in response" unfortunately.

jobserver/settings.py
ghickman: Yep!  The problem was getting correct the IP to add to this list for the CHECKS file to work correctly.  When a deploy runs the new container is spun up and checks are run against it, the IP for that new container was static in my manual tests but I couldn't find anyway to ask another system for it.  Since all external traffic to the site goes through nginx/dokku this change should be fine 🤞

entrypoint.sh
bloodearnest: Yeah. One of the thinks I've done before is to expose a url that 200s on successful check, and 500s on error, which could then be included in CHECKS

ghickman: I think I've seen that pattern talked about before too, with a `healthz` url.  Are you thinking we would effectively expose `python manage.py check --deploy` under this URL?

PR #546, opened 2021-06-03
Fix Registration Bugs

ghickman:
This fixes two bugs which stop a User from using the site after initial registration.

**No OpenSAFELY Org**
A fresh User will have no Organisation until some action is taken by us, which is expected after initial registration, so lets not break the site for that.

**No BackendMembership**
When first signing up a User will have no BackendMemberships, which provide access to 1 or more Backends when running Jobs.  This adds the ability to create them to the Django Admin.  It does _not_ teach the WorkspaceDetail view how to deal with a User with no backends, something for a subsequent PR.


inglesp (approved):
Looks good, although I'm not sure what the `BackendMembership` model is doing in this PR.

ghickman:
@inglesp – it's a short term (🤞) ~hack~ model to allow us to give filtered access to backends.  Longer term it'll be tied to a Project via Approvals, but that's months away at this point!

PR #547, opened 2021-06-03
Add an Approve Action

ghickman:
This adds an action to the User section of the admin to approve 1 or more Users.  In this context "approval" is simply connecting the Users to 1 or more Backends and Orgs.

This will be superseded later as the onboarding process is built out but for now it allows us to approve users after the changes in #534 and #537.

### DIFF
--- a/jobserver/admin.py
+++ b/jobserver/admin.py
@@ -3,6 +3,9 @@ import social_django.admin  # noqa: F401
 from django import forms
 from django.contrib import admin
 from django.contrib.auth.models import Group
+from django.shortcuts import redirect
+from django.urls import reverse
+from furl import furl
 from social_django.models import Association, Nonce, UserSocialAuth
 
 from .models import Org, Project, User
@@ -176,6 +179,7 @@ class ProjectAdmin(admin.ModelAdmin):
 
 @admin.register(User)
 class UserAdmin(admin.ModelAdmin):
+    actions = ["approve_users"]
     exclude = [
         "password",
         "groups",
@@ -191,3 +195,13 @@ class UserAdmin(admin.ModelAdmin):
         "is_approved",
     ]
     ordering = ["username"]
+
+    @admin.action(description="Approve selected users")
+    def approve_users(self, request, queryset):
+        url = reverse("approve-users")
+        f = furl(url)
+
+        for user in queryset:
+            f.add({"user": user.id})
+
+        return redirect(f.url)
--- a/jobserver/views/admin.py
+++ b/jobserver/views/admin.py
@@ -1,0 +1,64 @@
+from django import forms
+from django.contrib import messages
+from django.db import transaction
+from django.shortcuts import redirect
+from django.urls import reverse_lazy
+from django.utils.decorators import method_decorator
+from django.views.generic import FormView
+
+from ..authorization.decorators import require_superuser
+from ..models import Backend, BackendMembership, Org, OrgMembership, User
+
+
+class ApprovalForm(forms.Form):
+    def __init__(self, backends, orgs, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # bulid the backends and orgs fields based on the values passed in
+        self.fields["backends"] = forms.ModelMultipleChoiceField(queryset=backends)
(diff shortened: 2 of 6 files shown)


jobserver/admin.py
```diff
@@ -3,8 +3,12 @@
 from django import forms
 from django.contrib import admin
 from django.contrib.auth.models import Group
+from django.shortcuts import redirect
+from django.urls import reverse
+from furl import furl
 from social_django.models import Association, Nonce, UserSocialAuth
 
+# from .actions import ,c
```
inglesp, line 11: stray comment?

jobserver/views/admin.py
```diff
@@ -0,0 +1,64 @@
...
+    form_class = ApprovalForm
+    success_url = reverse_lazy("admin:jobserver_user_changelist")
+    template_name = "approve_users.html"
+
+    def dispatch(self, request, *args, **kwargs):
+        pks = self.request.GET.getlist("user")
+
+        if not pks:
+            messages.error(request, "One or more users must be selected")
+            return redirect(self.success_url)
```
inglesp, line 33: Should we define `failure_url` too?

tests/jobserver/test_admin.py
```diff
@@ -0,0 +1,28 @@
...
+def test_useradmin_approve_users(rf, superuser):
+    UserFactory.create_batch(5)
+    users = User.objects.all()
+
+    user_admin = UserAdmin(model=User, admin_site=AdminSite())
+
+    request = rf.get("/")
+    request.user = superuser
+
+    response = user_admin.approve_users(request, users)
```
inglesp, line 21: I think it's worth passing only a subset of the existing users to `.approve_users()` here.  For instance, this test wouldn't catch a bug where all users in the database were approved.

tests/jobserver/test_admin.py
```diff
@@ -0,0 +1,28 @@
...
+
+    assert response.status_code == 302
+
+    base = reverse("approve-users")
+    query = [f"user={u.pk}" for u in users]
```
inglesp, line 26: Could this be flaky?  Is it worth explicitly ordering the users to make this deterministic?

tests/jobserver/views/test_admin.py
```diff
@@ -0,0 +1,99 @@
...
+    request = rf.get(f"/?user={superuser.pk}")
+    request.user = superuser
+
+    response = ApproveUsers.as_view()(request)
+
+    assert response.status_code == 200
+
+    expected = set(User.objects.filter(pk=superuser.pk))
+    output = set(response.context_data["users"])
+    assert output == expected
```
inglesp, line 32: Could you do `assert list(response.context_data["users"]) == [superuser]`?

tests/jobserver/views/test_admin.py
```diff
@@ -0,0 +1,99 @@
...
+    assert response.status_code == 302
+    assert response.url == reverse("admin:jobserver_user_changelist")
+
+    user1.refresh_from_db()
+    assert set(user1.backends.all()) == {test, tpp}
+    assert org in user1.orgs.all()
+
+    user2.refresh_from_db()
+    assert set(user2.backends.all()) == {test, tpp}
+    assert org in user2.orgs.all()
```
inglesp, line 67: Again, you could beef up this test by:

* only approving users for one backend
* creating a second organisation, but not approving users for it

and then checking that the users are only approved for the expected backend and organisation.

inglesp (approved):
Could beef up tests, but otherwise looking good.

tests/jobserver/views/test_admin.py
ghickman: Oh! I hadn't thought about wrapping the output in a list - YES

tests/jobserver/views/test_admin.py
ghickman: Backends already exist because they get created by migrations but definitely good call on the orgs.

jobserver/views/admin.py
ghickman: It's the same URL and I'm copying the GCBV paradigm for the name, but I see your point… I think I'll leave it for now and address if I end up adding anything to this view.

inglesp: Totally fair enough!

PR #557, opened 2021-06-07
Import Remaining Projects

ghickman:
This switches the site over from being Workspace-centric to be Project-centric.  Predominantely this means:
* switch the first segment of the URL from being a Workspace name to being an Org slug
* only look up Workspaces with the previously named ProjectWorkspaceDetail
* allow DataLab & LSHTM Org members to create Projects without going through the full onboarding form (because of blanket IG approval currently in effect)

There is an import script included for turning a CSV I've generated locally into Orgs & Projects with the appropriate members and workspaces attached.

As part of deploying this I'm planning to use dokku's maintenance mode to block access to the site so I can manually run the import script with the mapping CSV, rather than bake those relationships into code.


jobserver/management/commands/map_existing_workspaces.py
```diff
@@ -0,0 +1,55 @@
...
+    ProjectMembership,
+    User,
+    Workspace,
+)
+
+
+class Command(BaseCommand):
+    @transaction.atomic()
+    def handle(self, *args, **options):
+        with open("project-mappings.csv") as f:
```
inglesp, line 20: where does this come from?

jobserver/management/commands/map_existing_workspaces.py
```diff
@@ -0,0 +1,55 @@
...
+            org_members = [coordinator, *usernames]
+            for member in org_members:
+                user = User.objects.get(username=member)
+                OrgMembership.objects.get_or_create(user=user, org=org)
+
+            names = [n for n in row["workspaces"].split(",") if n]
+            for name in names:
+                workspace = Workspace.objects.get(name=name)
+                workspace.project = project
+                workspace.save()
```
inglesp, line 55: Has this been tested manually?

jobserver/templates/workspace_detail.html
```diff
@@ -27,7 +27,7 @@
       <h3>{{ workspace.name }}</h3>
 
       <div class="d-flex">
-        <a class="btn btn-primary mr-2" href="{% url 'workspace-logs' name=workspace.name %}">
+        <a class="btn btn-primary mr-2" href="{% url 'workspace-logs' org_slug=workspace.project.org.slug project_slug=workspace.project.slug workspace_slug=workspace.name %}">
```
inglesp, line 30: Might benefit from a helper method to generate the URL?

jobserver/templates/index.html
```diff
@@ -56,11 +56,27 @@
...
     {% if user_can_run_jobs %}
-    <div class="d-flex justify-content-center">
-      <a class="btn btn-lg btn-primary" href="{% url 'workspace-create' %}">
-        Or Add a New Workspace
-      </a>
+    <div class="text-center">
+      <p><strong>OR</strong></p>
+
+      <p>
+        Pick {% if request.user.orgs.count > 1 %}an Organisation then {% endif %}a Project to create a Workspace in
```
inglesp, line 63: Should Organisation/Project/Workspace be organisation/project/workspace?

jobserver/views/orgs.py
```diff
@@ -22,16 +25,35 @@ def get_success_url(self):
...
+            org = Org.objects.get(slug=slug)
+        except Org.DoesNotExist:
+            # we used to serve up Workspaces at the root URL but have moved
+            # them under their relevant Orgs & Projects.  So attempt to
+            # redirect requests to a valid Workspace here for a while.
+
+            try:
+                workspace = Workspace.objects.get(name=slug)
+            except Workspace.DoesNotExist:
+                raise Http404
```
inglesp, line 43: What happens if you get multiple workspaces with the same URL?

jobserver/views/orgs.py
```diff
@@ -22,16 +25,35 @@ def get_success_url(self):
...
-        return context
+    def get(self, request, *args, **kwargs):
+        slug = self.kwargs["org_slug"]
+
+        try:
+            org = Org.objects.get(slug=slug)
+        except Org.DoesNotExist:
+            # we used to serve up Workspaces at the root URL but have moved
+            # them under their relevant Orgs & Projects.  So attempt to
+            # redirect requests to a valid Workspace here for a while.
```
inglesp, line 38: How long do we plan to support this?  Is it likely that any workspaces URLs are relied on somewhere?

inglesp (commented):
This is quite hard to review, but not as hard as I imagine it was to put together.  Chapeau.  Nothing major to flag code-wise.

This looks like a big bang migration.  Before I approve, can we have a chat about what the risks are, and how they could be mitigated?

inglesp (approved):

jobserver/management/commands/map_existing_workspaces.py
ghickman: Yes, regularly 😁

jobserver/templates/index.html
ghickman: Yes 😂

jobserver/views/orgs.py
ghickman: Workspace names are already unique in the model so we should't do.

jobserver/views/orgs.py
ghickman: Unsure currently, I've vaguely thought somewhere in the region of a month or two, but it's possible this could live here permanantely depending on how we feel about the far older workspace links being valid.

jobserver/templates/workspace_detail.html
ghickman: Definitely, there's a few little bits like this which I need to go and clean up in subsequent PRs.

jobserver/management/commands/map_existing_workspaces.py
ghickman: I've got it locally on my machine currently, including a command to generate it from the giant dict I've been using to collate all the information.  While the site and code is public I wasn't confortable making this block of data public in case it exposed a project which hadn't been publicised yet.

jobserver/management/commands/get_workspace_data.py
```diff
@@ -0,0 +1,50 @@
...
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        workspaces = (
+            Workspace.objects.annotate(jr_count=Count("job_requests"))
+            .exclude(jr_count=0)
+            .order_by("name")
+        )
+
+        with open("workspaces.csv", "w") as f:
+            writer = csv.DictWriter(f, ["workspace", "users with counts"])
```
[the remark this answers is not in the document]
ghickman: Possibly, but I don't think enough to rewrite this as it's expected to be a one-off migration.

jobserver/management/commands/get_workspace_data.py
```diff
@@ -0,0 +1,50 @@
...
+                        "count": user.jr_count,
+                        "name": user.name,
+                    }
+                    for user in users
+                ]
+
+                # sort users by their count
+                users_and_counts = sorted(
+                    users_and_counts, key=lambda uc: uc["count"], reverse=True
+                )
```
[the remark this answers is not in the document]
ghickman: Not a problem, this is for giving humans some intuition.

PR #558, opened 2021-06-08
Ensure values used for filters exist in the db

ghickman:
Django does some automated conversation when we use the ORM so passing
in the string "123" to a filter, eg .filter(pk="123") will find a record
with the integer pk 123.  However if this value can't be converted to an
integer this filter will throw a ValueError.  Since our filter inputs
here are user generated this checks query is using valid values before
adding the filter to the query.

Fixes [this Sentry ticket](https://sentry.io/organizations/ebm-datalab/issues/2424701368/?project=5443358&query=is%3Aunresolved).

### DIFF
--- a/jobserver/views/job_requests.py
+++ b/jobserver/views/job_requests.py
@@ -1,4 +1,5 @@
 from django.contrib.auth.decorators import user_passes_test
+from django.core.exceptions import BadRequest
 from django.db.models import Q
 from django.shortcuts import get_object_or_404, redirect
 from django.utils.decorators import method_decorator
@@ -109,8 +110,15 @@ class JobRequestList(FormMixin, ListView):
                 # it then we can look for a job number
                 qs = qs.filter(qwargs | Q(jobs__pk=q))
 
+        def raise_if_not_int(value):
+            try:
+                int(value)
+            except ValueError:
+                raise BadRequest
+
         backend = self.request.GET.get("backend")
         if backend:
+            raise_if_not_int(backend)
             qs = qs.filter(backend_id=backend)
 
         username = self.request.GET.get("username")
@@ -119,6 +127,7 @@ class JobRequestList(FormMixin, ListView):
 
         workspace = self.request.GET.get("workspace")
         if workspace:
+            raise_if_not_int(workspace)
             qs = qs.filter(workspace_id=workspace)
 
         return qs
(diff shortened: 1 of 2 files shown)


jobserver/views/job_requests.py
```diff
@@ -109,16 +109,19 @@ def get_queryset(self):
                 # it then we can look for a job number
                 qs = qs.filter(qwargs | Q(jobs__pk=q))
 
+        def available_values(Model, field):
+            return {str(f) for f in Model.objects.values_list(field, flat=True)}
+
         backend = self.request.GET.get("backend")
-        if backend:
+        if backend and backend in available_values(Backend, "pk"):
```
evansd, line 116: This approach feels a bit weird to me. If you supply an invalid value I'd expect the response to be a 400, rather than just for the value to be ignored. In fact, if you supply a syntactically valid value but one that doesn't exist in the db it will also be ignored, which seems even more problematic to me.

Could we define a function like `assert_is_int(value)` which raises an appropriate HTTP error if that value isn't an int?

evansd (changes requested):

jobserver/views/job_requests.py
```diff
@@ -109,16 +110,22 @@ def get_queryset(self):
...
 
+        def is_an_int(value):
+            try:
+                return int(value)
+            except ValueError:
+                raise BadRequest
+
         backend = self.request.GET.get("backend")
-        if backend:
+        if backend and is_an_int(backend):
```
evansd, line 120: Nice, I think this is much better. Only minor nit is that the conditional is now slightly misleading because it looks like it will just skip the block if `backend` isn't an int, rather than raising an error.

Even more minor nit is that `is_an_int("0")` will evaluate as false rather than true (because it returns 0) :smile:

Could you instead have?
```py
if backend:
    raise_if_not_int(backend)
    qs = qs.filter(backend_id=backend)
```

evansd (approved):

jobserver/views/job_requests.py
ghickman: Arg yes, clearly didn't pay attention to the if 🤦‍♂️

Updated as suggested 👍

evansd (approved):
Thanks! Sorry for the pedantry

ghickman:
@evansd – the pedantry was appreciated, and, as always, delivered delightfully 🙏

PR #560, opened 2021-06-09
Remove references to project registration when user is not an Org member

ghickman:
This removes the Register Project button, and the text reference to it when there are no Projects, for an Org when the viewing User isn't a member of that Org.

### DIFF
--- a/jobserver/templates/org_detail.html
+++ b/jobserver/templates/org_detail.html
@@ -18,9 +18,11 @@
     <div class="d-flex justify-content-between align-items-center">
       <h2>{{ org.name }}</h2>
 
+      {% if is_member %}
       <div>
         <a class="btn btn-primary" href="{% url 'project-create' org_slug=org.slug %}">Register Project</a>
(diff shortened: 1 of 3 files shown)


jobserver/templates/org_detail.html
```diff
@@ -34,8 +36,11 @@ <h3>Projects</h3>
     </ul>
     {% else %}
     <p class="mt-5">
-      There are no projects for this Organisation yet, click "Register Project"
-      to create one.
+      {% if is_member %}
+      There are no projects for this Organisation yet, click "Register Project" to create one.
+      {% else %}
+      There are no projects for this Organisation yet.
```
inglesp, line 42: s/O/o/!

inglesp (approved):

jobserver/templates/org_detail.html
ghickman: ARG

Genuinely tempted to write a pre-commit hook for this…

PR #561, opened 2021-06-09
Add breadcrumbs to WorkspaceDetail

ghickman:
This brings the WorkspaceDetail page inline with the other pages by adding some navigation via breadcrumbs.


benbc (approved):
Looks like this is duplicated on every page. Any way of generating them with a tag instead?

ghickman:
@benbc – I've not found a nice way to do it with a tag in previous attempts unfortunately.  You still have to pass in a lot of context on each page.

PR #562, opened 2021-06-09
Add ProjectDeveloper role to User who creates a Project

ghickman:
This adds the `ProjectDeveloper` role to ProjectMembership created when a User creates a Project so said User can run jobs in Workspaces for that Project.


iaindillingham (approved):
Well, auto-assigning reviewers has worked 😄

PR #569, opened 2021-06-11
Quieter Background Warnings

ghickman:
We currently have some backends in development, it's not useful to have a site-wide warning when one of them hasn't checked in for a few minutes, which is expected during development.


bloodearnest (approved):
+1, with the small suggestion that the db column might be better named `active` or something more generic.`is_production_ready` implies only one use, where as we might have many uses for making a backend active or not (e.g. going down for maintenance, etc)

ghickman:
@bloodearnest – ooh yes, much better name 👍

PR #576, opened 2021-06-14
Remove backend default when creating JobRequests

### LINKED ISSUE #549: Require manual backend selection.
Currently when selecting the backend to run jobs against on the job server the default backend is pre-selected as EMIS.  It is easy to start a job against the wrong backend.  

Forcing users to select the backend by having no default would prevent this.

ghickman:
This removes the default option for Backends when the User has access to more than one Backend so the default isn't accidentlly picked.

Fixes #549

### DIFF
--- a/jobserver/forms.py
+++ b/jobserver/forms.py
@@ -32,10 +32,9 @@ class JobRequestCreateForm(forms.ModelForm):
             },
         )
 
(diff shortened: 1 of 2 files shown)


bloodearnest (approved):
Nice

PR #577, opened 2021-06-14
Partial job id search

### LINKED ISSUE #493: Add a "search by partial job ID" feature
A non-urgent, nice-to-have.

The lack of copy-paste in the secure environment can make certain debugging tasks more difficult. I quite often make use of the fact that the first 4 or 5 characters of a job ID are generally long enough to get uniqueness but short enough that you can hold them temporarily in memory while re-typing them in the secure environment. This works for searching log files and most of the tooling I've added will accept partial job IDs for this reason.

However sometimes we want to go in the other direction by taking a job ID on the secure server and finding the corresponding job on the website. It would be useful to have some simple page which would let you type a partial ID and take you straight to the corresponding job (if there was an unambiguous match) or link to the possible matches if there are multiple.

ghickman:
This adds the ability to use a partial identifier to look up a Job when visiting the JobDetail page.  If you use a partial id the view will redirect to itself so the URL becomes the full ID.

I've also reworked the view to use TemplateResponse now that DetailView's `get_object` is no longer a good fit.

Fixes #493

### DIFF
--- a/jobserver/views/jobs.py
+++ b/jobserver/views/jobs.py
@@ -1,7 +1,10 @@
 from django.contrib.auth.decorators import user_passes_test
+from django.core.exceptions import MultipleObjectsReturned
+from django.http import Http404
 from django.shortcuts import get_object_or_404, redirect
+from django.template.response import TemplateResponse
 from django.utils.decorators import method_decorator
-from django.views.generic import DetailView, View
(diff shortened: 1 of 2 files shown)


evansd (approved):
:tada: Thanks!

PR #578, opened 2021-06-14
Add help text to the name field of WorkspaceCreateForm

### LINKED ISSUE #472: Add Help Text for Workspace Names
When creating a Workspace we don't tell Users what they need to do when creating a Workspace name.

We can add some help text to the field to explain the requirements to Users.

ghickman:
This adds some help text to the Workspace creation form to explain how it will be used.

Fixes #472

### DIFF
--- a/jobserver/forms.py
+++ b/jobserver/forms.py
@@ -134,6 +134,9 @@ class WorkspaceCreateForm(forms.ModelForm):
             "repo",
             "branch",
         ]
+        help_texts = {
+            "name": "Enter a descriptive name which makes this workspace easy to identify.  It will also be the name of the directory in which you will find results after jobs from this workspace are run.",
+        }
         model = Workspace
         widgets = {
             "name": forms.TextInput(),


jobserver/forms.py
```diff
@@ -134,6 +134,9 @@ class Meta:
             "repo",
             "branch",
         ]
+        help_texts = {
+            "name": "This will the name of the directory on the backend server(s) in which your research jobs are run.",
```
HelenCEBM, line 138: missed a word "be" :) 

Might be a little more user-friendly something like this:
"Enter a descriptive name which makes this workspace easy to identify. It will also be the name of the directory in which you will find results after jobs from this workspace are run."

Are there any requirements like "this must not contain special characters like £$%" or "you can enter spaces but they will be converted to underscores"?

HelenCEBM (commented):
See annotation

jobserver/forms.py
ghickman: Brilliant, thank you!

Yes, there's a validator on the name that ensures only letters, numbers, underscores, or hyphens are allowed.  If you use a character other than those you get an error explaining what is allowed.

ghickman:
@HelenCEBM – updated help text with example error for disallowed characters:

HelenCEBM (approved):

ghickman:
@HelenCEBM – could you approve the PR please?

HelenCEBM:
I have already - perhaps I don't have sufficient permissions and someone else needs to review?

iaindillingham (approved):
Disappointed not to see any C18 capitalization. I'll get over It.

PR #581, opened 2021-06-15
Improve Backend Filtering of JobRequests API

ghickman:
This adds a fix for the databricks job-runner filtering name and creates a warning in Sentry should we see anything similar from any backend to help us track down misconfigurations during intial backend setup.

### DIFF
--- a/jobserver/api.py
+++ b/jobserver/api.py
@@ -1,6 +1,7 @@
 import itertools
 import operator
 
+import sentry_sdk
 import structlog
 from django.db import transaction
 from django.http import Http404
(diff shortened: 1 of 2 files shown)


bloodearnest (approved):

jobserver/api.py
```diff
@@ -269,7 +270,25 @@ def get_queryset(self):
...
+
+        # short term special case
+        if query_arg_backend == "nhsd":
+            query_arg_backend = "databricks"
+
         db_backend = getattr(self.backend, "name", None)
+
+        # send a warning to Sentry if the query arg and token-linked backend
+        # names differ
+        if query_arg_backend != db_backend:
```
evansd, line 282: I think this needs an ` and db_backend is not None` clause on the end, otherwise accessing the API unauthenticated triggers a log message.

PR #582, opened 2021-06-15
Add Badges to Repo URLs

### LINKED ISSUE #442: Improve UX for unauthenticated users browsing to private Github Repos
For transparency, we currently allow unauthenticated users to see most information on the job server.

This means for any private repo, when they view [a project page like this one](https://jobs.opensafely.org/anticoagulant-sensitivity-analysis-5-check/), they are presented with a link to a repo to which they have no access; Github then gives a 404, which makes it look like all our links are broken
image

In the first instance, some explanatory text would be good on the home page, at least for unauthenticated users. Ben suggested

> some very brief light touch single para explainer at the top of the page to explain that this is a working tool for the team; shows every github repo and commit ID for every task executed against the data. Some GitHub links will show a 404 if you are not logged in, or not part of an OpenSAFELY analyst group: that means the repository is currently closed; we log all repo's and commit IDs executing code against NHS data here; we require analysts, at minimum, to make code open at the time when results are shared publicly; you can read the Principles of OpenSAFELY here

ghickman:
This adds badges to GitHub repo links showing if they are public or private.  As part of this work I realised we were duplicating repo URLs on the ProjectDetail page so this fixes that too.

Fixes #442

### DIFF
--- a/jobserver/github.py
+++ b/jobserver/github.py
@@ -73,6 +73,38 @@ def get_file(org, repo, branch):
     return r.text
 
 
+def get_repo(org, repo):
+    f = furl(BASE_URL)
+    f.path.segments += [
+        "repos",
+        org,
+        repo,
+    ]
+
+    headers = {
+        "Accept": "application/vnd.github.v3+json",
+        "Authorization": f"token {GITHUB_TOKEN}",
+        "User-Agent": USER_AGENT,
+    }
+    r = requests.get(f.url, headers=headers)
+
+    if r.status_code == 404:
+        return
+
+    r.raise_for_status()
+
+    return r.json()
+
+
+def get_repo_is_private(org, repo):
+    repo = get_repo(org, repo)
+
+    if repo is None:
+        return None
+
+    return repo["private"]
+
+
 def _get_page(session, org, cursor):
     """
     Get a page of Repos (with branches) from the OpenSAFELY Researchers Team
--- a/jobserver/views/workspaces.py
+++ b/jobserver/views/workspaces.py
@@ -15,7 +15,7 @@ from ..forms import (
     WorkspaceCreateForm,
     WorkspaceNotificationsToggleForm,
 )
-from ..github import get_branch_sha, get_repos_with_branches
+from ..github import get_branch_sha, get_repo_is_private, get_repos_with_branches
 from ..models import Backend, JobRequest, Project, Workspace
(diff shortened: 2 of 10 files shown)


jobserver/github.py
```diff
@@ -73,6 +73,35 @@ def get_file(org, repo, branch):
...
+        org,
+        repo,
+    ]
+
+    headers = {
+        "Accept": "application/vnd.github.v3+json",
+        "Authorization": f"token {GITHUB_TOKEN}",
+        "User-Agent": USER_AGENT,
+    }
+    r = requests.get(f.url, headers=headers)
```
evansd, line 89: Fine as is for now, but this might benefit from splitting out the generic "talk to github API" code from the stuff specific to repos at some point.

jobserver/github.py
```diff
@@ -73,6 +73,35 @@ def get_file(org, repo, branch):
...
+
+    r.raise_for_status()
+
+    return r.json()
+
+
+def get_repo_is_private(org, repo):
+    repo = get_repo(org, repo)
+
+    return repo["private"]
```
evansd, line 102: `get_repo` returns None on a 404, which will result in a TypeError here. Just checking that's expected/handled already?

jobserver/views/workspaces.py
```diff
@@ -180,6 +182,12 @@ def get_initial(self):
         # setting as a default for the form which the user can override.
         return {"will_notify": self.workspace.should_notify}
 
+    def get_repo_is_private(self):
+        owner = furl(self.workspace.repo).path.segments[0]
```
evansd, line 186: Should this be a method on `workspace`? `get_repo_owner` or something like that? Or even a `repo_owner` property

evansd (approved):
Looks good. I made a few bikesheddy comments but nothing blocking.

I guess at some point we'll need to think about caching our interactions with the github API, but presumably we're OK for now.

jobserver/github.py
ghickman: Yeah, I've been thinking about this.  We have a couple of small differences so far (one gets a file, one is the GraphQL paging stuff…), but something for the bog standard "get some JSON please" is definitely starting to look useful!

jobserver/github.py
ghickman: It was not 😄 thanks for catching it!

evansd (approved):

PR #585, opened 2021-06-15
Add migration missing from 3dfc084

ghickman:


inglesp (approved):
Do we actually need the `db` field?

ghickman:
@inglesp – I did think about removing it but I think that's a large discussion.  This and the changes in #584 don't remove the historical data for workspaces so taking the middle ground for now!

PR #590, opened 2021-06-16
Tidy up SuperUser Usage

ghickman:
This reduces how much we use the SuperUser role in various places, moving us further towards permissions based access.

Notably it also modifies the ensure_admins function to remove the SuperUser role from any User who isn't specified in the list passed in.


jobserver/views/projects.py
```diff
@@ -111,7 +110,6 @@ def get_context_data(self, **kwargs):
         )
 
 
-@method_decorator(require_superuser, name="dispatch")
```
inglesp, line 115: Just double-checking: is it intentional that no permissions are required anywhere in the view?

tests/jobserver/views/test_job_requests.py
```diff
@@ -385,13 +386,13 @@ def test_jobrequestlist_with_authenticated_user(rf):
...
 @pytest.mark.django_db
-def test_jobrequestlist_with_superuser(rf, superuser):
+def test_jobrequestlist_with_core_developer(rf, core_developer):
     job_request = JobRequestFactory()
     JobFactory(job_request=job_request)
     JobFactory(job_request=job_request)
 
     request = rf.get(MEANINGLESS_URL)
-    request.user = superuser
+    request.user = UserFactory(roles=[CoreDeveloper])
```
inglesp: Can you not do `request.user = core_developer` here?

jobserver/authorization/admins.py
```diff
@@ -19,17 +20,25 @@ def get_admins():
...
+@transaction.atomic()
 def ensure_admins(usernames):
     """
     Given an iterable of username strings, ensure they have the SuperUser role
     """
-    admins = User.objects.filter(username__in=usernames)
+    for user in User.objects.all():
+        try:
+            user.roles.remove(SuperUser)
+        except ValueError:
```
inglesp, line 31: Why would this raise `ValueError`?

tests/jobserver/authorization/test_admins.py
```diff
@@ -6,6 +6,16 @@
...
 
 
+@pytest.mark.django_db
+def test_ensure_admins_remove_user():
+    user = UserFactory(username="ghickman", roles=[SuperUser])
+
+    ensure_admins([])
+
+    user.refresh_from_db()
+    assert SuperUser not in user.roles
```
inglesp, line 16: Given importance of this function, could it be tested a bit more thoroughly?  For both the tests where `ensure_admins()` completes without error, there's only one user in the database, and you can imagine a bug in `ensure_admins()` so that it acts on all the existing users while still making the tests pass.

How about something like:

```
def test_ensure_admins():
    users = [
        UserFactory(username="aaa", roles=[SuperUser]),  # is superuser now, and will be afterwards
        UserFactory(username="bbb", roles=[SuperUser]),  # is superuser now, and won't be afterwards
        UserFactory(username="ccc"),  # isn't superuser now, and will be afterwards
        UserFactory(username="ddd"),  # isn't superuser now, and won't be afterwards
    ]

    ensure_admins(["aaa", "ccc"])

    assert SuperUser in User.objects.get(username="aaa").roles
    assert SuperUser not in User.objects.get(username="bbb").roles
    assert SuperUser in User.objects.get(username="ccc").roles
    assert SuperUser not in User.objects.get(username="ddd").roles
```

inglesp (approved):
Generally good; one test to tighten up.

jobserver/authorization/admins.py
ghickman: `roles` is a list and calling `remove(thing)` when `thing` isn't in the list raises that.  The idea with this was to make sure no users have `SuperUser` before adding it to the Users we want.

jobserver/views/projects.py
ghickman: Yep, this was an oversight previously!

tests/jobserver/views/test_job_requests.py
ghickman: Yep, done in the subsequent commit 😁

tests/jobserver/authorization/test_admins.py
ghickman: Very good point, thanks 🙂  Fixed and rebased.

PR #596, opened 2021-06-18
Add Some User Management Pages

ghickman:
This adds a list and detail page for Users, for use by DataLab staff, via the `manage_users` permission.  It's being added to address the lack of ability to assign Backend access and global Roles via the site.

This is the first step towards removing User modification from the Django Admin, which can't handle Role assignment (easily), so we're taking advantage of the existing RolesForm.

**User detail page**


jobserver/authorization/forms.py
```diff
@@ -13,8 +13,8 @@ class RolesForm(forms.Form):
     classes on save.
     """
 
-    def __init__(self, available_roles, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, *, available_roles, **kwargs):
```
iaindillingham, line 16: What magic is this!?

iaindillingham (approved):

jobserver/authorization/forms.py
ghickman: Keyword-only arguments! [PEP](https://www.python.org/dev/peps/pep-3102/)

It forces all args to be used as keyword arguments when called, great for making interfaces very explicit.

iaindillingham: I'm not sure I understand. I know that `*args` will be a `list` of as many arguments as are supplied and `**kwargs` will be a `dict` of as many keyword arguments as are supplied. I'm unsure about the single asterisk, though. Are we saying "Pass as many arguments as you like; only the last one matters"?

evansd: The PEP explains the various details, but in essence the unnamed single asterisk means "end of positional arguments". So it's an error to supply any positional arguments at all to this method.

iaindillingham: Wowzers, I didn't know that. I'll read the PEP. Thanks both!

PR #598, opened 2021-06-18
Updates for local development

rebkwok:
Some updates for getting a local development server running

- added a `read_env()` to settings.py so it'll read env variables from the file if it exists
- updated README to use the dev-env.sh script for native and docker setup
- a couple of updates to the dev-env.sh script to get it working for me (ubuntu 20.04)
- Changed the Testing section to document how to run tests (it was duplicating the docker-compose section)

### DIFF
--- a/scripts/dev-env.sh
+++ b/scripts/dev-env.sh
@@ -10,7 +10,9 @@ if ! command -v bw > /dev/null; then
     exit 1
 fi
 
-if test -z "$BW_SESSION"; then
+session_from_env=${BW_SESSION-"not-found"}
+
+if (test $session_from_env = "not-found"); then
     echo "Unlocking bitwarden..."
     BW_SESSION=$(bw unlock --raw)
     export BW_SESSION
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -24,23 +24,34 @@ Errors are logged to the DataLab [Sentry](https://sentry.io) account.
 
 ## Local Development Set Up
 ### Native
+
+Prerequisites:
+- **Python v3.9.x**
+- **Pip**
```
ghickman, line 30: Nit: probably virtualenv too?  I'm not sure all pythons come with `venv` installed.

README.md
```diff
@@ -70,7 +88,9 @@ The DataLab job server is deployed to our `dokku2` instance, instructions are ar
 Run `make dev-config` if you have not already. Note: you will need the
 bitwarden cli tool installed in order to access passwords.
 
-A Django development server can be started with `docker-compose up`.
+Run the tests with:
+
+    pytest tests
```
ghickman, line 93: There's a makefile target for this, `test`, which runs both coverage and collectstatic.  It's what CI uses and I think we should push for that one by default as it's what's required to get the green ticks for CI.

scripts/dev-env.sh
```diff
@@ -22,9 +24,9 @@ GH_DEV_TOKEN_BW_ID=10242708-234a-4162-982f-ace700ee5c53
 write() {
     local name="$1"
     local value="$2"
-    sed -i "s/$name=.*/$name=$value/" "$target"
+    sed -i"" -e "s/$name=.*/$name=$value/" "$target"
```
ghickman, line 27: Have you checked this is portable between macOS (which uses BSD sed) and Linux?

ghickman (approved):
Couple of nits but otherwise great 👍

scripts/dev-env.sh
rebkwok: I'm on linux now and it works

rebkwok:
Changed a couple of other commands for `make` ones (`make setup`, `make run`) as well as `make test` where they already existed

PR #600, opened 2021-06-21
Speed up Event Log

ghickman:
This modifies the JobRequest.status` property to make use of prefetch Jobs if they exist.  It avoids running duplicate queries when treating the status as a field and trying to filter on it.  The event log page was doing this and generating 2600+ (duplicate!) queries even for pages where the status filter produced no results.

### DIFF
--- a/jobserver/models.py
+++ b/jobserver/models.py
@@ -340,7 +340,18 @@ class JobRequest(models.Model):
 
     @property
     def status(self):
-        statuses = self.jobs.values_list("status", flat=True)
+        prefetched_jobs = (
+            hasattr(self, "_prefetched_objects_cache")
+            and "jobs" in self._prefetched_objects_cache
+        )
+        if not prefetched_jobs:
+            # require Jobs are prefetched to get statuses since we have to
+            # query every Job for the logic below to work
+            raise Exception("JobRequest queries must prefetch jobs.")
+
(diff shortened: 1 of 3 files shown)


jobserver/models.py
```diff
@@ -340,7 +340,16 @@ def started_at(self):
...
+        prefetched_jobs = (
+            hasattr(self, "_prefetched_objects_cache")
+            and "jobs" in self._prefetched_objects_cache
+        )
+        if prefetched_jobs:
+            # make use of prefetch Jobs if we can, this is particularly useful
+            # when filtering JobRequests on the event log page
+            statuses = [j.status for j in self.jobs.all()]
+        else:
+            statuses = self.jobs.values_list("status", flat=True)
```
evansd, line 352: Just for my own understanding, when would we expect this branch to be hit? And if it's there as a "just in case" condition should we consider make it an error to hit this branch so we catch it in testing?

evansd (approved):

jobserver/models.py
ghickman: Lots of places, anywhere which uses the status icons (there's at least one API using it).  It's definitely a "just in case" but coming from knowing this is the default option for all the places I can't think of!

evansd (approved):
:+1: Nice

PR #605, opened 2021-06-21
Add text explaining security levels to job request details

benbc:


jobserver/templates/job_request_detail.html
```diff
@@ -38,6 +38,21 @@ <h2 class="mb-4">
       <code><small class="text-muted">{{ jobrequest.identifier }}</small></code>
     </h2>
 
+    <p>
+      This page shows the technical details of what happened when authorised researcher {{ jobrequest.created_by.username }} requested one or
```
ghickman, line 42: You could use the `name` property (instead of `username`) on User here, it gets the Users full name falling back to their username if they don't have that set.

ghickman (approved):
Great explainer 👍

jobserver/templates/job_request_detail.html
benbc: Oh that's great. Thanks. Done.

benbc:
@ghickman No credit to me -- the text is Seb's. :grinning:

PR #606, opened 2021-06-22
Better User Tooling

ghickman:
This adds search, filtering by backend access, and filtering by applied roles to the UserList page.  Roles filtering is a little involved due to the db field being a JSONField and SQLite not support Django's `__contains` expression, but I think the custom function makes the API workable.

Many thanks to @Ian-Foote for their help in creating the ORM function!

### DIFF
--- a/jobserver/views/users.py
+++ b/jobserver/views/users.py
@@ -12,6 +12,7 @@ from ..authorization import roles
 from ..authorization.decorators import require_permission
 from ..forms import SettingsForm, UserForm
 from ..models import Backend, User
+from ..utils import raise_if_not_int
 
 
 @method_decorator(login_required, name="dispatch")
@@ -85,6 +86,12 @@ class UserList(ListView):
     queryset = User.objects.order_by("username")
     template_name = "user_list.html"
 
+    def get_context_data(self, **kwargs):
+        return super().get_context_data(**kwargs) | {
+            "backends": Backend.objects.order_by("name"),
+            "q": self.request.GET.get("q", ""),
+        }
+
     def get_queryset(self):
         qs = super().get_queryset()
 
@@ -95,5 +102,10 @@ class UserList(ListView):
                 | Q(first_name__icontains=q)
                 | Q(last_name__icontains=q)
             )
+
+        backend = self.request.GET.get("backend")
+        if backend:
+            raise_if_not_int(backend)
+            qs = qs.filter(backends__pk=backend)
 
         return qs
--- a/jobserver/views/users.py
+++ b/jobserver/views/users.py
@@ -3,6 +3,7 @@ import inspect
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.db import transaction
+from django.db.models import Q
 from django.shortcuts import redirect
 from django.utils.decorators import method_decorator
 from django.views.generic import ListView, UpdateView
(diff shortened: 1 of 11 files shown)


tests/jobserver/authorization/test_fields.py
```diff
@@ -1,12 +1,27 @@
...
 from ...factories import OrgFactory, OrgMembershipFactory, UserFactory
 
 
+@pytest.mark.django_db
+def test_extractroles():
+    user1 = UserFactory(roles=[OutputChecker])
+    user2 = UserFactory(roles=[CoreDeveloper])
+
+    users = User.objects.alias(extracted=ExtractRoles("roles")).filter(
+        extracted__contains="CoreDeveloper"
```
evansd, line 17: Does this want to be `'"CoreDeveloper"` i.e to match on the quotes as well? Otherwise it would match the role `NotACoreDeveloper`, which is obviously a silly example but you can imagine other cases where that might happen and it would be good to establish the right pattern here.

jobserver/views/users.py
```diff
@@ -108,4 +112,10 @@ def get_queryset(self):
             raise_if_not_int(backend)
             qs = qs.filter(backends__pk=backend)
 
+        role = self.request.GET.get("role")
+        if role:
+            qs = qs.alias(extracted=ExtractRoles("roles")).filter(
+                extracted__contains=role
```
evansd, line 118: Possibly should be `f'"{role}"' as mentioned elsewhere

jobserver/views/users.py
```diff
@@ -108,4 +112,10 @@ def get_queryset(self):
             raise_if_not_int(backend)
             qs = qs.filter(backends__pk=backend)
 
+        role = self.request.GET.get("role")
+        if role:
+            qs = qs.alias(extracted=ExtractRoles("roles")).filter(
```
evansd, line 117: nit (and fine to leave for now) but structurally I wonder if this logic belongs as a method on a custom QuerySet on the `User` model. Something like `filter_by_role`? I can imagine this is something we might want to do in various places and it's really the models responsibility to worry about the implementation details here.

evansd (approved):
Looks great. A couple of minor points about the filtering code, but nothing blocking.

jobserver/views/users.py
ghickman: I had a similar thought and opted to leave it here until it's used elsewhere.  I could see it going either way but I'm reasonably certain there's no short or medium term plans to need it outside this particular view, so my reasoning was adding an abstraction only used in one place wouldn't be overly useful.

evansd: Got you. Makes sense.

tests/jobserver/authorization/test_fields.py
ghickman: The full string for the role is the dotted path to the defined class, so in this case `jobserver.authorization.roles.CoreDeveloper`, an implementation detail I wanted to hide from a consumer of the roles APIs, but you're right that we could potentially add roles which cause problems, we already have three Project* roles already!

A better fit with the current use of Roles would be passing the class and letting the framework handle the conversion, so really this should be abstracted off to a QuerySet/Manager on the Model.

jobserver/views/users.py
ghickman: Having looked at your comment on the quoting of role names, I am entirely changing my mind and agreeing with you, go team 😁

ghickman:
@evansd – could you take another look at this (only the last two commits have changed) now that I've rolled the changes as a QuerySet?

evansd (approved):
:+1: Nice, this looks good

PR #612, opened 2021-06-24
Escape project.yaml Content

ghickman:
This escapes the content of a given project.yaml before it is rebuilt in HTML and marked_safe for template rendering.

### DIFF
--- a/jobserver/views/job_requests.py
+++ b/jobserver/views/job_requests.py
@@ -4,6 +4,7 @@ from django.db.models import Q
 from django.http import Http404
 from django.shortcuts import redirect
 from django.utils.decorators import method_decorator
+from django.utils.html import escape
 from django.utils.safestring import mark_safe
 from django.views.generic import DetailView, ListView, View
 from django.views.generic.edit import FormMixin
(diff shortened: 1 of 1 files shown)


jobserver/views/job_requests.py
```diff
@@ -64,7 +65,7 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context["project_definition"] = mark_safe(
             render_definition(
-                self.object.project_definition,
+                escape(self.object.project_definition),
```
evansd, line 68: Should the escaping be done inside the `render_definition` function as that's the thing which is adding the additional HTML?

evansd (approved):

jobserver/views/job_requests.py
ghickman: I don't have a strong opinion here.  I was trying to keep `render_definition` "pure", in that it didn't deal with sanitisation/escaping, but transformed a string -> HTML.  However I could see the argument that expecting a caller to escape input is asking for trouble.

Do you feel strongly about moving it in?

evansd: Don't feel strongly, no.

PR #616, opened 2021-06-25
Use npm for vendor assets

tomodwyer:
So that we don't need to include minified JS and CSS assets from libraries in the repo, and we can version control our third-party libraries.


ghickman (approved):
🎉

Do you want to remove the bit in the readme around adding or updating static assets too?

tomodwyer:
@ghickman Good point, I'll remove that from the docs and add a task to add documentation when we're bundling all the assets.

PR #625, opened 2021-06-28
Split Up Models

ghickman:
This breaks models module into a package since it's grown fairly large.  The new modules are a little arbitrary but here is my reasoning:

* `onboarding` because the spike into an onboarding flow showed we likely needed more than just `ResearcherRegistration` to preserve state of in-flight sign ups.
* `stats` because it didn't have a better home.
* `backends` because they're sort of a (core?) sub-system unto themselves.
* `outputs` because the project outputs voyage is adding more models around this.

No new migrations are required because of the importing in `jobserver/models/__init__.py` but unfortuately the Roles definitions needed updating.

### DIFF
--- a/jobserver/authorization/roles.py
+++ b/jobserver/authorization/roles.py
@@ -22,7 +22,7 @@ class CoreDeveloper:
         "Internal user who develops and deploys the OpenSAFELY core framework."
     )
     models = [
-        "jobserver.models.User",
+        "jobserver.models.core.User",
     ]
     permissions = [
         cancel_job,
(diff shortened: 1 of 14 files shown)


bloodearnest (approved):
LGTM, on the basis that I think the new organisation makes sense, but I just assumed the bit blocks of models and tests additions are just wholesale moves into their new modules, with no changes, as getting a sense of that was tricky.

ghickman:
@bloodearnest – yep, it's a pure refactor, all changes are either "move to another module" or "fix import paths".

PR #629, opened 2021-06-29
Fix Migrations

ghickman:
This creates a concrete manager class for the User model, inheriting from django.contrib's `UserManager` and mixing in our UserQuerySet for the roles filtering method.  This allows Django's migrations framework to serialise the manager class, fixing the migrations.

I'm unsure why CI didn't flag this up when running tests for #606 but this unblocks deploys.


evansd (approved):

inglesp (approved):
Don't understand this, but looks reasonable!

ghickman:
@inglesp – the old way was dynamic in some way that migrations exploded, I _think_ because there's no class definition for them to point to.  However, the tests not picking that up is rather odd!

PR #635, opened 2021-06-30
Remove Use of SuperUser Role

ghickman:
This removes our use of the SuperUser Role from all parts of the code base, opting to either add an appropriate new permission or fallback to `User.is_superuser`.

The Role itself can be removed when all Users no longer have it assigned.


bloodearnest (approved):
Nice

PR #636, opened 2021-06-30
Refactor Release testing machinery

bloodearnest:
1. New ReleaseUploadFactory to setup a zip stream and valid hash for
   testing upload handling

2. Re-implementend ReleaseFactory in order to handle the fact that it
   needs filesystem state set up first before we create the ORM object,
   and that some of its attribute depend on the hash of the uploaded
   content.

3. Add autouse fixture to ensure every tests uses a clean
   RELEASE_STORAGE dir

4. Refactored some of handle_release workflow to share more things in
   common with the testing setup

### DIFF
--- a/jobserver/releases.py
+++ b/jobserver/releases.py
@@ -26,8 +26,7 @@ def handle_release(workspace, backend, backend_user, release_hash, upload):
     # 2. much easier for human operators to navigate on disk (e.g. for removal
     #    of any disclosive information).
     upload_dir = os.path.join(workspace.name, release_hash)
-    actual_dir = settings.RELEASE_STORAGE / upload_dir
-    actual_dir.parent.mkdir(exist_ok=True, parents=True)
+    actual_dir = None
 
     try:
         # first, ensure we can extract the zip
@@ -43,10 +42,9 @@ def handle_release(workspace, backend, backend_user, release_hash, upload):
                     {"detail": f"metadata/manifest.json file is not valid json: {exc}"}
                 )
 
-            archive.extractall(actual_dir)
+            actual_dir = extract_upload(upload_dir, archive)
 
-        files = list_files(actual_dir)
-        calculated_hash = hash_files(actual_dir, files)
+        calculated_hash, files = hash_files(actual_dir)
         if calculated_hash != release_hash:
             raise ValidationError(
                 {
@@ -60,7 +58,7 @@ def handle_release(workspace, backend, backend_user, release_hash, upload):
             backend=backend,
             backend_user=backend_user,
             upload_dir=upload_dir,
-            files=[str(f) for f in files],
+            files=files,
         )
 
     except Exception:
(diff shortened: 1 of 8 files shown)


jobserver/releases.py
```diff
@@ -60,29 +58,35 @@ def handle_release(workspace, backend, backend_user, release_hash, upload):
...
             upload_dir=upload_dir,
-            files=[str(f) for f in files],
+            files=files,
         )
 
     except Exception:
         # something went wrong, clean up files, they will need to reupload
-        shutil.rmtree(actual_dir, ignore_errors=True)
+        if actual_dir:
+            shutil.rmtree(actual_dir, ignore_errors=True)
```
ghickman, line 67: Doesn't the ignore_errors kwarg being set to true mean we don't need the if statement here?

jobserver/releases.py
```diff
@@ -60,29 +58,35 @@ def handle_release(workspace, backend, backend_user, release_hash, upload):
...
-def list_files(directory):
-    """Utility to list files in a directory tree."""
-    return list(
-        sorted(p.relative_to(directory) for p in directory.glob("**/*") if p.is_file())
-    )
-
-
-def hash_files(directory, files):
+def hash_files(directory):
     # use md5 because its fast, and we only care about uniqueness, not security
```
ghickman, line 74: Can you pop this down by the use of md5? I wouldn't want security researchers to miss the context of that line.

jobserver/releases.py
```diff
@@ -60,29 +58,35 @@ def handle_release(workspace, backend, backend_user, release_hash, upload):
...
+    files = list(
+        sorted(
+            str(p.relative_to(directory)) for p in directory.glob("**/*") if p.is_file()
+        )
+    )
```
ghickman, line 79: Why do you generate the file list inside this function instead of passing it in?  (and can you add a comment to explain please 🙂)

tests/conftest.py
```diff
@@ -29,6 +30,11 @@ def fixture_configure_structlog(log_output):
     structlog.configure(processors=[log_output])
 
 
+@pytest.fixture(autouse=True)
+def set_release_storage(monkeypatch, tmp_path):
+    monkeypatch.setattr(settings, "RELEASE_STORAGE", tmp_path / "releases")
```
ghickman, line 35: Nice 👍

tests/factories.py
```diff
@@ -110,15 +116,83 @@ class Meta:
...
+
+    manifest = kwargs.pop("manifest", DEFAULT_MANIFEST)
+    raw_manifest = kwargs.pop("raw_manifest", None)
+
+    # create these if needed
+    kwargs.setdefault("workspace", WorkspaceFactory())
+    kwargs.setdefault("backend", BackendFactory())
+
+    # create an upload, so we know the release_hash ahead of time.
+    upload = ReleaseUploadFactory(files, manifest, raw_manifest)
```
ghickman, line 184: Handing this off to another factory is a really nice paradigm 👍

tests/jobserver/models/test_outputs.py
```diff
@@ -1,31 +1,39 @@
...
-    assert str(release.file_path("file.txt")) == "releases/workspace/release/file.txt"
-
-
-@pytest.mark.django_db
-def test_release_get_absolute_url():
-    org = OrgFactory()
-    project = ProjectFactory(org=org)
-    workspace = WorkspaceFactory(project=project)
-    release = ReleaseFactory(workspace=workspace)
+def test_get_absolute_url():
```
ghickman, line 9: Can you add the model name back into the test name please, there are lots of get_absolute_urls!

ghickman (changes requested):
Few questions to address and missing/moved comments in other places.  The new factories look great!

jobserver/releases.py
bloodearnest: It does, will remove

tests/jobserver/models/test_outputs.py
bloodearnest: ah, for sure

jobserver/releases.py
bloodearnest: hmm, just did it to avoid a needless one-line function that was only ever used in one place.

jobserver/releases.py
bloodearnest: good call

jobserver/releases.py
ghickman: Makes sense 👍

ghickman (approved):

PR #640, opened 2021-07-01
Add releases read API

### LINKED ISSUE #627: File serving API in job-server
Add a release read API, to a) list files in a release and b) serve up a specific file from a relase

Both endpoints should be authenticated by a user with the correct access to that workspace.

We plan to use X-Accel-Redirect to have nginx serve the files in prod, but will also support directly serving the bytes from django for local usage w/o nginx

bloodearnest:
One to list the files in release, another to get the file contents.

Download api supports a production optimisation for using nginx.

Drive by fix to report missing coverage lines by default.

Fixes #627

### DIFF
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -1,7 +1,9 @@
 import structlog
+from django.conf import settings
+from django.http import FileResponse
 from django.urls import reverse
 from rest_framework import serializers
-from rest_framework.exceptions import NotAuthenticated, ValidationError
+from rest_framework.exceptions import NotAuthenticated, NotFound, ValidationError
 from rest_framework.generics import CreateAPIView
 from rest_framework.parsers import FileUploadParser
 from rest_framework.response import Response
@@ -9,7 +11,7 @@ from rest_framework.views import APIView
 from slack_sdk.errors import SlackApiError
 
 from jobserver.api import get_backend_from_token
-from jobserver.models import Workspace
+from jobserver.models import Release, Workspace
 from jobserver.releases import handle_release
 from services.slack import client as slack_client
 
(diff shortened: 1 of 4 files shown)


jobserver/api/releases.py
```diff
@@ -79,3 +84,52 @@ def put(self, request, workspace_name, release_hash):
...
+            # we're behind nginx, so use X-Accel-Redirect to serve the file from nginx
+            relative_path = handle.relative_to(settings.RELEASE_STORAGE)
+            redirect_path = f"{internal_redirect}/{relative_path}"
+            response = HttpResponse()
+            response.headers["X-Accel-Redirect"] = redirect_path
+        else:
+            # serve directly from django in dev
+            mtype, _ = mimetypes.guess_type(handle)
+            # use regular django response to bypass DRFs renderer framework and just serve bytes
+            response = HttpResponse(FileWrapper(handle.open()), content_type=mtype)
```
ghickman, line 132: Did you look at the FileResponse class for this?

tests/jobserver/api/test_releases.py
```diff
@@ -187,3 +188,74 @@ def test_upload_api_redirected(api_rf):
     release.refresh_from_db()
     assert response["Release-Id"] == release.id
     assert response["Location"] == f"http://testserver{release.get_absolute_url()}"
+
+
+@pytest.mark.django_db
+def test_release_index_api(api_rf):
```
ghickman, line 194: Not need for `api_rf` if you're using client here, might as well make it a fixture if we're using it in a few places too.

Does DRF's Client give us anything over a RequestFactory though?

tests/jobserver/api/test_releases.py
```diff
@@ -187,3 +188,74 @@ def test_upload_api_redirected(api_rf):
...
+    assert (
+        response.headers["X-Accel-Redirect"]
+        == f"/releases/{release.workspace.name}/{release.id}/file.txt"
+    )
+    assert response.headers["Content-Dispostion"] == 'attachment; filename="file.txt"'
+
+    # delete file
+    release.file_path("file.txt").unlink()
+    response = client.get(file_url)
+    assert response.status_code == 404
```
ghickman, line 261: I'd prefer to break these down into separate unit tests if possible, so if say we break the nginx config that test breaks rather than the tests not using it.

ghickman (commented):
Few questions around implementation and layout.

tests/jobserver/api/test_releases.py
bloodearnest: Yeah, I prefer Client() and APIClient() over request factory by default, as it checks that url is hooked up to the view. 

If you test everything with RF and calling the view directly, its possible to ship bad url config and break production. I have the scars :)

ghickman: Fair point!  Do you want to make a fixture for it to make `api_rf`?

jobserver/api/releases.py
bloodearnest: Great, used that, thanks.

tests/jobserver/api/test_releases.py
bloodearnest: Understandable. I prefer this lighter-weight approach in general, but will refactor

ghickman (approved):

tests/jobserver/api/test_releases.py
bloodearnest: I didn't seem much point in a fixture, as there was nothing to clean up

Issue #655, opened 2021-07-06
Expose SPA at Various URLs


ghickman:
The outputs SPA is a directory and file viewer, and we can link to it from various contexts, which the view serving it needs to handle so the complexity for sorting that out (and the requisite permissions) can be handled in Django.  The contexts/URLs it needs to handle:

* `/org/project/workspace/releases/<hash>/` -> directory listing of the given release
* `/org/project/workspace/releases/<hash>/<path>` -> directory listing OR file contents (depending on the path) of the path/file in the given release
* `/org/project/workspace/releases/<path>` -> directory listing or file contents of the current version of the path/file for a Workspace

All links at the Project level will be to the relevant Workspace-based URL.

PR #658, opened 2021-07-06
List Releases for Workspaces and Projects

### LINKED ISSUE #654: List Releases
Add some simple listing views for Releases under both Project and Workspace, where the former shows all Releases across all Workspaces.

These can be fairly skeleton views at this stage.

ghickman:
This adds two new views listing Releases in their Workspace or Project contexts.  It includes a a function to get the latest version of each file, akin to a current working directory view for a Workspace, however the implementation of this is expected to potentially change as the current voyage proceeds.

**Workspace Files**

**Project Files**

Fixes #654

### DIFF
--- a/jobserver/templates/project_detail.html
+++ b/jobserver/templates/project_detail.html
@@ -113,20 +113,16 @@
           </ul>
         </div>
 
-        {% if releases %}
         <div class="mb-4">
-          <h5>Releases</h5>
+          <h5>
+            Releases
+            <small>({{ releases_count }} files)</small>
+          </h5>
 
-          <a class="btn btn-primary btn-sm" href="{% url 'project-releases' org_slug=project.org.slug project_slug=project.slug %}">
+          <a class="btn btn-primary btn-sm" href="{{ project.get_releases_url }}">
             View
           </a>
-          {# <ul class="list-unstyled"> #}
-          {#   {% for release in releases %} #}
-          {#   <li>{{ release }}</li> #}
-          {#   {% endfor %} #}
-          {# </ul> #}
         </div>
-        {% endif %}
 
       </div>
 
--- a/jobserver/templates/project_release_list.html
+++ b/jobserver/templates/project_release_list.html
@@ -1,0 +1,74 @@
+{% extends "base.html" %}
+
+{% load humanize %}
+
+{% block content %}
+
+<nav aria-label="breadcrumb">
+  <ol class="breadcrumb">
+
+    <li class="breadcrumb-item"><a href="/">Home</a></li>
+
+    <li class="breadcrumb-item">
+      <a href="{{ project.org.get_absolute_url }}">
+        {{ project.org.name }}
+      </a>
+    </li>
+
+    <li class="breadcrumb-item">
+      <a href="{{ project.get_absolute_url }}">
+        {{ project.name }}
+      </a>
+    </li>
+
+    <li class="breadcrumb-item active" aria-current="page">Releases</li>
+
+  </ol>
+</nav>
+
+
+<div class="row">
+  <div class="col-lg-8 offset-lg-2">
+
+    <p>
+      Below are all released files for the {{ project.name }} project across
+      all of its workspaces.
+    </p>
+    {% for release in object_list %}
+    <div class="border rounded p-2 pl-3 mb-3">
+
+      <div class="d-flex justify-content-between">
+        <span>
+          Files released by {{ release.created_by.name }} in the
+          <a href="{{ release.workspace.get_absolute_url }}">{{ release.workspace.name }}</a>
+          workspace {{ release.created_at|naturaltime }}
+        </span>
+
+        <button
+          class="btn btn-sm btn-primary"
+          type="button"
+          data-toggle="collapse"
+          data-target="#release-{{ release.id }}-files"
+          aria-expanded="false"
+          aria-controls="release-{{ release.id }}-files">
+          Files
+        </button>
+      </div>
+
+      <div class="collapse" id="release-{{ release.id }}-files">
+        <div class="card card-body border-0 mt-2 p-0 px-2">
+          <ul class="list-unstyled mb-0">
+            {% for file in release.files.all %}
+            <li><code>{{ file.name }}</code></li>
+            {% endfor %}
+          </ul>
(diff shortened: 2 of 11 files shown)


jobserver/releases.py
```diff
@@ -103,3 +104,15 @@ def extract_upload(upload_dir, archive):
...
+def latest_files(releases):
+    """
+    Gets the latest version of each file across the given Releases
+    """
+
+    files = ReleaseFile.objects.filter(release__in=releases).order_by(
+        "-release__created_at"
+    )
+
+    return list(unique_everseen(files, key=lambda file: file.name))
```
bloodearnest, line 118: I am assuming we do this in python because of figuring out how to do it via ORM is tricky without DISTINCT ON support?

I think that's fine - we will never have many Releases, so all good.

However, I think adding an entire dependency just for single simple iterator is too much cost for not enough benefit.

I think it's simpler and more understandable just to do a simple for loop ourselves, and drop the dependency, e.g. something like this:

    seen = dict()
    for file in workspace.files.order_by("-release__created_at"):
        if file.name not in seen:
            seen[file.name] = file
    return list(seen.values())

jobserver/templates/project_detail.html
```diff
@@ -113,20 +113,16 @@ <h5>Researchers</h5>
           </ul>
         </div>
 
-        {% if releases %}
         <div class="mb-4">
-          <h5>Releases</h5>
+          <h5>
+            Releases
+            <small>({{ releases_count }})</small>
```
bloodearnest, line 119: `({{ releases_count }} files)` perhaps?

jobserver/templates/project_release_list.html
```diff
@@ -0,0 +1,74 @@
...
+
+  </ol>
+</nav>
+
+
+<div class="row">
+  <div class="col-lg-8 offset-lg-2">
+
+    <p>
+      Below are all releases for the {{ project.name }} project across all of
```
bloodearnest, line 34: s/releases/released files/ ?

jobserver/templates/project_release_list.html
```diff
@@ -0,0 +1,74 @@
...
+      its workspaces.
+    </p>
+    {% for release in object_list %}
+    <div class="border rounded p-2 pl-3 mb-3">
+
+      <div class="d-flex justify-content-between">
+        <span>
+          Release by {{ release.created_by.name }} in the
```
bloodearnest, line 42: s/Release by/File released by/?

I think avoiding noun-ing "Release", and speaking instead of "releasing files" and "released files", is better for user facing text, as users have no real context of A Release as a Noun.

jobserver/templates/workspace_release_list.html
```diff
@@ -0,0 +1,92 @@
...
+</nav>
+
+
+<div class="row">
+  <div class="col-lg-8 offset-lg-2">
+
+    <div class="mb-4">
+      <h4>Workspace Files</h4>
+      <p>
+        These files are the latest version of each path across all releases for
```
bloodearnest, line 42: perhaps?

    These are the latest versions of all files that have been released so far for this workspace.

I think we should avoid 'path' as a term if possible.

jobserver/templates/workspace_release_list.html
```diff
@@ -0,0 +1,92 @@
...
+      </ul>
+    </div>
+
+    <h4>All Releases</h4>
+    {% for release in object_list %}
+    <div class="border rounded p-2 pl-3 mb-3">
+
+      <div class="d-flex justify-content-between">
+        <span>
+          Release by {{ release.created_by.name }} {{ release.created_at|naturaltime }}
```
bloodearnest, line 62: s/Release by/File released by/ ?

jobserver/urls.py
```diff
@@ -148,7 +153,8 @@
         name="project-membership-remove",
     ),
     path("new-workspace/", WorkspaceCreate.as_view(), name="workspace-create"),
-    path("releases/", Releases.as_view(), name="project-releases"),
+    path("releases/", ProjectReleaseList.as_view(), name="project-release-list"),
+    path("releases/spa/", Releases.as_view(), name="project-releases"),
```
bloodearnest, line 157: Hmm, not sure I like `spa` in the url. What about "releases/view"?

bloodearnest (commented):
Got some questions about whether we really need more_itertools, and a few wording suggestions.

jobserver/releases.py
```diff
@@ -103,3 +104,15 @@ def extract_upload(upload_dir, archive):
...
     archive.extractall(actual_dir)
     return actual_dir
+
+
+def latest_files(releases):
+    """
+    Gets the latest version of each file across the given Releases
+    """
+
+    files = ReleaseFile.objects.filter(release__in=releases).order_by(
```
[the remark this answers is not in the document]
ghickman: Ah, yes I completely forgot about the link between Workspace and ReleaseFile!  Putting it on `Workspace` is a great idea, I started this function before we had ReleaseFile and didn't think about moving it.

jobserver/urls.py
ghickman: This is just a placeholder to move the SPA somewhere else, #655 will move it to a more relevant place.

jobserver/releases.py
ghickman: Yes, all the ORM/SQL solutions involved some hoop jumping and since we're unlikely to have that many Releases or ReleaseFiles in a Workspace (or even Project), doing it in Python should be fine.

I don't agree on the more-itertools argument, we've added many libraries for a single function call previously (first, pytest-subtests, incuna-mail, slack-sdk, etc) when we could have built many of those ourselves!

I'll swap it out for this PR though.

bloodearnest: So, I'm not advocating a general rule here, it's just in this specific case, the cost of maintaining an extra dependency is not worth it IMO, since the implementation is trivial.

If we are wanting to use other `more_iterools` functions in future, then sure no problem. But in the context of this change, it didn't seem worth it.

For slack and pytest-subtest, the implementation is not trivial, regardless of the fact its only one function. And we'd expect to use pytest-subtests multiple times going forward.

bloodearnest (approved):

PR #659, opened 2021-07-08
Add ReleaseDetail View

### LINKED ISSUE #655: Expose SPA at Various URLs

ghickman:
This adds a more fleshed out ReleaseDetail view as the consumer of two types of URL:
* top-level directory listing using just the ID of a release
* file contents viewing via the ID of a release and a path

Inside the view these are the same thing since we hand off to the SPA (with the relevant API endpoint for the given release) and let it handle querying for a file by path.

This doesn't support a URL of the form `/org/project/workspace/releases/<path>` because that current working view of files needs more thought.

The auth process for accessing this view isn't clear yet but I thought it better to get some review on what's done, adding that in once we have a better idea, but this PR does require it to be merged.

Fixes #655


jobserver/api/releases.py
```diff
@@ -67,17 +66,7 @@ def put(self, request, workspace_name, release_hash):
...
-                "workspace-release",
-                kwargs={
-                    "org_slug": workspace.project.org.slug,
-                    "project_slug": workspace.project.slug,
-                    "workspace_slug": workspace.name,
-                    "release": release.id,
-                },
-            )
-        )
+        response["Location"] = request.build_absolute_uri(release.get_absolute_url())
```
bloodearnest, line 69: I wonder if we should actually link them to the Workspace url here, rather than the Release?

jobserver/templates/project_release_list.html
```diff
@@ -44,22 +44,29 @@
...
+        </div>
       </div>
 
       <div class="collapse" id="release-{{ release.id }}-files">
         <div class="card card-body border-0 mt-2 p-0 px-2">
           <ul class="list-unstyled mb-0">
             {% for file in release.files.all %}
-            <li><code>{{ file.name }}</code></li>
+            <li>
+              <a href="{{ file.get_absolute_url }}">
```
bloodearnest, line 66: Hmm, does this hotlink direct to the file? Or does it load the SPA, which then loads that file?

jobserver/urls.py
```diff
@@ -122,10 +121,11 @@
         WorkspaceReleaseList.as_view(),
         name="workspace-release-list",
     ),
+    path("releases/<pk>/", ReleaseDetail.as_view(), name="release-detail"),
     path(
-        "releases/<release>",
-        WorkspaceReleaseView.as_view(),
-        name="workspace-release",
+        "releases/<pk>/<path:path>",
```
bloodearnest, line 126: I'm confused by this, it seems to link directly to a file, rather than to the SPA pointing to a release?

bloodearnest (commented):
This look good, but I am confused about the direct file links, which I didn't think would be served to the use direct from job-server, but instead, the from SPA, once loaded?

ghickman:
Discussed in chat but writing up here so future readers understand the design, this is re [these](https://github.com/opensafely-core/job-server/pull/659#discussion_r666142630) [two](https://github.com/opensafely-core/job-server/pull/659#discussion_r666140816) comments:

`ReleaseFile.get_absolute_url()` builds a `release-detail-with-path` URL where a path is captured after the release hash.  Both this URL and `release-detail` are handled by the ReleaseDetail view which loads the SPA and does nothing with the captured `path`.  This allows us to construct URLs pointing to a file path in Django but send them to the SPA for rendering.

bloodearnest (approved):

PR #660, opened 2021-07-08
Add a Feature Flag for Release Pages

ghickman:
Releases are still being developed but we don't want to run a staging site or have a long running branch while developing them so this adds logic to remove links to them from the relevant pages.


bloodearnest (approved):
perfect

PR #661, opened 2021-07-08
Add a release management command for local dev:

bloodearnest:
Usage:

    python manage.py release <workspace> <backend> directory/

Involved a refactor of the ReleaseUploadFactory code to extract the zip-construction logic, including automatically generating a manifest.

Also, found and fixed a bug in the exclusion of MANIFEST_PATH from the release hash.

### DIFF
--- a/jobserver/management/commands/release.py
+++ b/jobserver/management/commands/release.py
@@ -1,0 +1,26 @@
+import sys
+from pathlib import Path
+
+from django.core.management.base import BaseCommand
+
+from jobserver.models import Backend, Workspace
+from jobserver.releases import create_upload_zip, handle_release
+
+
+class Command(BaseCommand):
+    help = "Configure admins to "
+
+    def add_arguments(self, parser):
+        parser.add_argument("workspace_name", help="worksapace name to release to")
+        parser.add_argument("backend_name", help="backend to release from")
+        parser.add_argument("directory", help="directory of files to release")
+
+    def handle(self, workspace_name, backend_name, directory, *args, **options):
+        try:
+            workspace = Workspace.objects.get(name=workspace_name)
+            backend = Backend.objects.get(name=backend_name)
+            release_hash, zipstream = create_upload_zip(Path(directory))
+            handle_release(workspace, backend, "local user", release_hash, zipstream)
+        except Exception as e:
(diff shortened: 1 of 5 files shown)


jobserver/management/commands/release.py
```diff
@@ -0,0 +1,26 @@
...
+        parser.add_argument("directory", help="directory of files to release")
+
+    def handle(self, workspace_name, backend_name, directory, *args, **options):
+        try:
+            workspace = Workspace.objects.get(name=workspace_name)
+            backend = Backend.objects.get(name=backend_name)
+            release_hash, zipstream = create_upload_zip(Path(directory))
+            handle_release(workspace, backend, "local user", release_hash, zipstream)
+        except Exception as e:
+            print(str(e), file=sys.stderr)
```
ghickman, line 25: Not really relevant for this PR but you might find this useful another time: https://docs.djangoproject.com/en/3.2/howto/custom-management-commands/#django.core.management.BaseCommand.style

ghickman (approved):
👍

jobserver/management/commands/release.py
bloodearnest: hah, I copied that verbatim from your `ensure_admins.py` command! :rofl:

jobserver/management/commands/release.py
```diff
@@ -0,0 +1,26 @@
...
+from pathlib import Path
+
+from django.core.management.base import BaseCommand
+
+from jobserver.models import Backend, Workspace
+from jobserver.releases import create_upload_zip, handle_release
+
+
+class Command(BaseCommand):
+    help = "Configure admins to "
```
sebbacon, line 11: Late seagulling but not sure this help is very useful :)

Could we describe when you might want to use it? 

It might also warrant a sentence or two in `DEVELOPERS.md`?

It's so common that I make a management command that's useful for development, everyone forgets about it, two years later someone removes it because "no-one uses it", very mild sadness ensues

ghickman: Hah! That's entirely my fault, I think the `ensure_admins` command has the same "help".

bloodearnest: Oh, sorry, didn't notice that c/p from the admin command. Will fix in next PR

PR #671, opened 2021-07-09
Add Release Permissions

### LINKED ISSUE #670: Set up Permissions for ProjectCollaborator Role
We need to give ProjectCollaborator a permission which defines if it a User with it can access outputs for a Workspace.

ghickman:
This adds a new permission to the ProjectCollaborator role, and uses that permission to gate access to view releases via the main function.

I've also enabled ProjectCollaborator to be applied to Users so we can easily test and debug the release functionality.

Fixes #670


inglesp (approved):
All fine, shipit.

But at some point we should consider whether the various permission checks are adequately visible.  They're all buried inside different parts of different CBVs, which makes it harder than it might be to audit permissions across the project.

I wonder if there's a way we could standardise how permissions are checked.  AFAICT all (or almost all) views following the same pattern:

1. load some objects from the database-
1. check the current user has some permission on those objects
1. if so, do something; otherwise bail

Given this, we could either write our own base CBV class with hooks for 1. and 2., or we could write decorators to do these (as we do, in a simpler scenario, in OpenCodelists, eg in [this view](https://github.com/opensafely-core/opencodelists/blob/main/codelists/views/version_publish.py) with [these decorators](https://github.com/opensafely-core/opencodelists/blob/main/codelists/views/decorators.py)).

Taking this further, an idea I like is treating your codebase a bit like a queryable database, which in this case means something like allowing us to ask "what can user X do with project Y?" or "which users can do Z to project Y"?  More work, obviously, but given how fundamental access control is to our story, it's something that we should think about.

ghickman:
@inglesp – agreed, I've been eyeing the concept of opencodelist's decorators.  We're definitely starting to see some patterns emerge as part of the current voyage so I think it'll be time to revist these soon.

Ideally we'd avoid a base class or mixins, I have a feeling they'll hide the checks too much from the reader!  Something I really wanted to be core in the roles and permissions framework was making the check explicit, so I'd like to try and retain that concept with any abstraction we add.

inglesp:
> Ideally we'd avoid a base class or mixins, I have a feeling they'll hide the checks too much from the reader!

Agree there's a tension here.  But as long as the checks themselves are obviously attached to the class, I don't think it's unreasonable to expect readers to understand that there is a base class.

Here's a sketch of what it could look like:

```
class BaseView:
    get_permissions = []
    post_permissions = []

    def load_objects(self):
        pass

    def dispatch(self, request):
        try:
            self.load_objects()
        except DoesNotExist:
            return self.handle_404()

        if request.is_a get():
            permissions = self.get_permissions
        elif request.is_a_post():
            permissions = self.post_permissions

        for permission in permissions:
            param_names = get_param_names_for_permission(permission)
            kwargs = {k: v for k, v in vars(self).items() if k in param_names}
            kwargs["user"] = request.user
            if not permission(**kwargs):
                return self.handle_403(request)

        if request.is_a get():
            return self.get(request)
        elif request.is_a_post():
            return self.post(request)


class WorkspaceView(BaseView):
    get_permissions = [can_view_workspace]
    post_permissions = [can_update_workspace]

    def load_objects(self, request):
        self.project = Project.objects.get(pk=request.project_id)
        self.workspace = project.workspaces.get(pk=request.workspace_id)

    def get(self, request):
        # render workspace page

    def post(self, request):
        # update workspace and redirect


def can_view_workspace(user, project):
    return user in project.members

def can_update_workspace(user, workspace):
    return workspace.owner == user
```

PR #673, opened 2021-07-12
Add <time>_ago Helper Functions

### LINKED ISSUE #669: Pull minutes_ago into Test Utils
The `minutes_ago` function is a simple wrapper around `some_dt - timedelta(minutes=n)`.  A lot of our tests make use of this pattern and minutes ago would be useful to improve their readability.

ghickman:
This pulls the `minutes_ago` function out of a test function and into a utils module, applying it everywhere we can.  It also adds `seconds_ago` since we were also doing that in various places.

Fixes #669

### DIFF
--- a/tests/jobserver/api/test_jobs.py
+++ b/tests/jobserver/api/test_jobs.py
@@ -1,5 +1,3 @@
-from datetime import timedelta
-
 import pytest
 from django.utils import timezone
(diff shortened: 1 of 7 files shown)


CarolineMorton (approved):
Nice - this is neat!

PR #685, opened 2021-07-13
Update release management command to use new models

bloodearnest:
It's also more dev friendly, as it will create default workspace and
users for you if you pass --create

### DIFF
--- a/jobserver/management/commands/release.py
+++ b/jobserver/management/commands/release.py
@@ -1,26 +1,114 @@
-import sys
+import hashlib
 from pathlib import Path
 
+from django.conf import settings
 from django.core.management.base import BaseCommand
+from furl import furl
 
-from jobserver.models import Backend, Workspace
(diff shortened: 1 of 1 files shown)


jobserver/management/commands/release.py
```diff
@@ -1,26 +1,114 @@
...
+                    slug="test-project",
+                    org=org,
+                )
+                workspace = Workspace.objects.create(
+                    name=workspace_name,
+                    project=project,
+                    repo="https://github.com/opensafely/test-repo",
+                )
+            else:
+                raise
```
ghickman, line 99: Seeing as this is such a short else block can we flip the conditional and not bother with an else? eg 

```
if not create:
    raise

…
```

ghickman (approved):
Nice 👍

Just one point about flipping an if.

jobserver/management/commands/release.py
bloodearnest: done

PR #687, opened 2021-07-13
Small tweak to release notification API to upload file details

bloodearnest:


jobserver/api/releases.py
```diff
@@ -36,9 +37,20 @@ def initial(self, request, *args, **kwargs):
...
         data = serializer.data
 
-        message = f"{data['created_by']} released outputs from {data['path']}"
+        if "files" in data:
+            files = data["files"]
+            message = [
+                f"{data['created_by']} released {len(files)} outputs from {data['path']}:"
+            ]
+            for f in files:
+                message.append(f"`{f}`")
```
ghickman, line 46: ffffff 😂

A str.join() might be slightly less confusing here? or maybe just renaming the loop variable? Not entirely sure it would improve things though!

ghickman (approved):
👍

One point about a potential variable rename, but otherwise looks great.

jobserver/api/releases.py
bloodearnest: me neither! I'll land as is and fix when we update this notification code

PR #690, opened 2021-07-14
Refactor ReleaseFile API and use ULID ids

bloodearnest:
- ignore some 3.9.6 warnings
- Move the Release and ReleaseFile ids to ULIDs
- Refactor release API again


jobserver/api/releases.py
```diff
@@ -125,21 +119,28 @@ def get(self, request, workspace_name):
...
         File must be listed in Release.requested_files, and the hash must match.
         """
         release = get_object_or_404(Release, id=release_id)
+        backend, user = validate_upload_access(request, release.workspace)
 
-        if filename not in release.requested_files:
-            raise NotFound
+        try:
+            upload = request.data["file"]
+            filename = upload.name
```
ghickman, line 136: Can this line throw a KeyError? if not can we pull it out of the try/catch to leave it focussed?

jobserver/releases.py
```diff
@@ -7,8 +7,29 @@
 from .models.outputs import absolute_file_path
 
 
+class ReleaseFileAlreadyExists(Exception):
+    pass
```
ghickman, line 11: This makes the API code really readable 👍👍

ghickman (approved):
Looks great 👍

jobserver/api/releases.py
bloodearnest: fixed

jobserver/releases.py
bloodearnest: yeah internal logic exceptions are a Good Idea, I think

PR #694, opened 2021-07-16
List Outputs for Workspace

### LINKED ISSUE #689: List Outputs
Outputs for a Workspace are now the latest version of all ReleaseFiles for the Workspace and any Snapshots.  The UI will combine these two internal concepts as "Outputs".

However, we have two different types of User viewing them:
* **Unprivileged**: Logged out, or an authenticated User without the `view_release_file` permission.
* **Privileged**: Authenticated and has the `view_release_file` permission.

Privileged users will see all Snapshots (ordered by latest first) and the latest files for the Workspace (first in the list), while unprivileged users will only see the published Snapshots.

Both Snapshots and latest files can redirect to an SPA entry point with relevant API endpoints.

For the privileged user perspective we can add badges for `latest` (the current files), and draft/published status for the Snapshots.

ghickman:
This adds a list page showing the Workspace's current files, and any Snapshots it has.  Users without the `view_release_file` permission will only see publicised Snapshots on the page.

The current files continue to route to the SPA with the Snapshots linking to a new SnapshotDetail view, which is a new SPA entrypoint (with accompanying API endpoint).

**As a user with the correct permissions:**

**As a user without the correct permissions (or logged out):**

Fixes #689


jobserver/templates/workspace_output_list.html
```diff
@@ -30,6 +32,47 @@
...
+        </div>
+      </li>
+      {% endif %}
+
+      {% for snapshot in snapshots %}
+      <li class="list-group-item d-flex justify-content-between">
+
+        <div>
+          <a class="mr-3" href="{{ snapshot.get_absolute_url }}">
+            Outputs as of {{ snapshot.created_at|naturaltime }}
```
bloodearnest, line 60: This doesn't quite read right to me. I think a date would be better than natural time, and we only need to the day I think, minutes don't matter IMO.

e.g.

View Workspace Outputs:
  * Latest (or Current?)
  * Published on 2021-07-16 by bloodearnest
  * Published on 2021-06-15 by ghickman

jobserver/templates/workspace_output_list.html
```diff
@@ -30,6 +32,47 @@
...
+
+    <p>
+      Listed below are the outputs for the {{ workspace.name }} workspace.
+    </p>
+
+    <ul class="list-group list-unstyled mb-4">
+      {% if user_can_view_all_files %}
+      <li class="list-group-item d-flex justify-content-between">
+        <div>
+          <a href="{{ workspace.get_current_outputs_url }}">Latest Files</a>
```
bloodearnest, line 47: I'd call both of these Outputs or Files, not mix an match. But see below for a more complete suggestion.

jobserver/models/outputs.py
```diff
@@ -154,3 +122,47 @@ def get_absolute_url(self):
     def get_api_url(self):
         """The API url that will serve up this file."""
         return reverse("api:release-file", kwargs={"file_id": self.id})
+
+
+class Snapshot(models.Model):
```
bloodearnest, line 127: Minor nit - I think I prefer `ReleaseSnapshot` more than plan snapshot, as we may have other kinds of snapshots, perhaps. But not a blocker.

bloodearnest (commented):
A couple non-blocking minor suggestions, but I think we should resolve the permissions logic question before landing.

jobserver/templates/workspace_output_list.html
bloodearnest: I'm not sure I like putting Latest in the same list as the snapshots. What about:
___________________
| View Current Outputs |
‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
Draft Output Publications:
* 2021-07-16 by bloodearnest (2 files)



Published Outputs:
* Published on  2021-07-16 by bloodearnest (5 files)
* Published on 2021-06-15 by ghickman (2 files)

jobserver/views/workspaces.py
```diff
@@ -271,28 +285,69 @@ def post(self, request, *args, **kwargs):
...
         workspace = get_object_or_404(
             Workspace,
             project__org__slug=self.kwargs["org_slug"],
             project__slug=self.kwargs["project_slug"],
             name=self.kwargs["workspace_slug"],
         )
 
-        # TODO: check permissions here
+        # only a privileged user can view the current files
+        if not has_permission(
```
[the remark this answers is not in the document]
ghickman: Definitely agree with this in principle but there are some stricky parts:
1) the exceptions raised are DRF specific and use API specific language
2) the anon check is repeated (for good reason in that function!) before the has_permission function

I'd like to add a few more of these calls, getting the Roles and Permissions we use for gates settled, and then see if we can extract a nice abstraction across all of them.  I think we're still working out some of those bits currently, but we should definitely make sure this is dealt with before the end of the voyage (captured in #699).

jobserver/templates/workspace_output_list.html
ghickman: I really hate writing the text for these parts, so being given better words is hugely appreciated here!

jobserver/views/workspaces.py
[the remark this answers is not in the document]
bloodearnest: Good point.

I think we should probably move into `jobserver/releases.py`, but raise custom logic Exceptions, that the views/DRF can handle as needed.

If `has_permission` checks for anon, then it can be removed for now.

bloodearnest (approved):

ghickman:
Updated as requested, with a slight tweak to the text on the list page so we don't use "Published" for draft Snapshots:

**As a user with the correct permissions:**


**As a user without the correct permissions (or logged out):**

bloodearnest:
I still think we want to separate out "Current" from the list, to a "View Current Files" button, as it's gonna be the most common path for the user, but not blocking on that, as I'm sure other will have thoughts too.

ghickman:
Agreed on the button, captured in #700

PR #695, opened 2021-07-16
Expose more metadata in release file indexes

bloodearnest:


jobserver/api/releases.py
```diff
@@ -90,8 +90,25 @@ def validate_release_access(request, workspace):
...
+        files=[
+            dict(
+                name=k,
+                url=v.get_api_url(),
+                user=v.created_by.username,
+                date=v.created_at.isoformat(),
+                sha256=v.filehash,
+                size=get_size(v),
+            )
+            for k, v in files.items()
```
ghickman, line 110: Now there's a selection of lines using them, could you swap out `k` and `v` for what they represent?  I think they're filename and ReleaseFile? (so maybe `filename` and `f`?)

ghickman (approved):
Looks great, just one comment on naming things.

jobserver/api/releases.py
bloodearnest: good idea, will do

PR #704, opened 2021-07-19
List Outputs on Project page

### LINKED ISSUE #702: List Outputs on Project Pages
Now that we have outputs listed for a workspace from #689 we need to display some outputs for a Project too.  Given we're planning to put this in the sidebar for ProjectDetail which is a slimmer space, and how hard it would be to tie outputs to their workspaces with so little space we can limit the list to the most recent published Snapshot for each Workspace, linking to the existing view for them.

ghickman:
This adds a list of outputs to the ProjectDetail page.  It only takes the most recent published version from each Workspace.  Each output links to the same entry point used on the Workspace's Outputs page.

Fixes #702

### DIFF
--- a/jobserver/views/projects.py
+++ b/jobserver/views/projects.py
@@ -2,6 +2,7 @@ from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.core import signing
 from django.db import transaction
+from django.db.models import OuterRef, Subquery
 from django.http import Http404
 from django.shortcuts import get_object_or_404, redirect
 from django.utils.decorators import method_decorator
@@ -26,7 +27,7 @@ from ..forms import (
     ResearcherFormSet,
 )
 from ..github import get_repo_is_private
-from ..models import Org, Project, ProjectInvitation, ProjectMembership, Release, User
+from ..models import Org, Project, ProjectInvitation, ProjectMembership, Snapshot, User
 
 
 @method_decorator(login_required, name="dispatch")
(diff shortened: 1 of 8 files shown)


jobserver/views/projects.py
```diff
@@ -188,17 +189,41 @@ def get_context_data(self, **kwargs):
...
+            .values("pk")[:1]
+        )
+
+        # annotate the subquery onto the Workspace QuerySet
+        workspaces = workspaces.annotate(latest_snapshot=Subquery(latest_snapshot_qs))
+
+        # build the QuerySet of Snapshots from the PKs annotated onto each Workspace
+        snapshot_pks = workspaces.exclude(latest_snapshot=None).values_list(
+            "latest_snapshot", flat=True
+        )
```
bloodearnest, line 223: man, I'd forgotten django how obtuse django's compound query build up can be sometimes. This took a while to parse, but it makes sense!

jobserver/views/projects.py
```diff
@@ -188,17 +189,41 @@ def get_context_data(self, **kwargs):
...
+        We only want to show the most recent published Snapshot for each
+        Workspace.
+        """
+        # build a Subquery to get the PK for the most recently published
+        # Snapshot for each Workspace
+        latest_snapshot_qs = (
+            Snapshot.objects.filter(workspace_id=OuterRef("id"))
+            .exclude(published_at=None)
+            .order_by("-published_at")
+            .values("pk")[:1]
```
bloodearnest, line 214: OOI, I assume this ends up issuing a LIMIT 1 on the subquery?

bloodearnest (approved):

jobserver/views/projects.py
ghickman: That's right, all slicing on a QuerySet becomes a `LIMIT` on the query.

jobserver/views/projects.py
ghickman: "sometimes" 😉

It was quite the faff to work out too!  Could I have added a better explanation anywhere to speed up parsing?

PR #716, opened 2021-07-22
Move Publish Outputs into SPA

ghickman:
This teaches the outputs SPA about the snapshot preparation URL, showing a "Publish" button if it's instantiated with one:

Since a User must already be privileged to view the current ReleaseFile for a Workspace we don't need to do any further permission checks.

The SPA will use the prepare URL to post a list of ReleaseFile IDs to the API, which creates a draft Snapshot, and returns the direct URL for an SPA.

I've not taken much (any) time to style the button because there's a confirmation flow to add on top of it, where I'll deal with that.

Refs #675

PR #718, opened 2021-07-22
Publish Snapshots with the SPA

### LINKED ISSUE #675: Add Confirmation to Outputs Publication
Building on #674 we want to make sure the files being published at the ones the User thinks they are.  We can do this by showing a list of the ReleaseFiles with data from their Release, eg who released them and when.  This will avoid accidental publication of files if one User is about to publish when a second user releases from an L4 environment.

### LINKED ISSUE #681: Create api_client Fixture
Now that we're using DRF's `Client()` wrapper in the releases tests lets create a fixture to save some lines of code in those tests.

ghickman:
This adds the publish step to the SPA to build on the prepare step added in #716, and uses the same paradigm with exposing UI features when different URLs are present.  There is a badge to show a Snapshot is in a draft state, which is only visible to viewers who can see draft Snapshots, so when they're published it goes away completely.

I've done a small refactor inside of this to extract DRF's `APIClient` into a fixture to match `api_rf` now we're using it so much.  The bulk of the changes are in the final commit.

Fixes #675
Fixes #681


jobserver/templates/snapshot_detail.html
```diff
@@ -36,6 +36,13 @@
   </ol>
 </nav>
 
+<h1 class="h3">
+  Outputs for {{ snapshot.workspace.name }}
+  <span class="badge badge-secondary">
+    {% if snapshot.is_draft %}Draft{% endif %}
```
bloodearnest, line 42: Should this entire badge be skipped if it's not draft?

bloodearnest (approved):
LGTM, allthough @tomodwyer might want to check the mutation bits.

1 minor question, but good to land.

jobserver/templates/snapshot_detail.html
ghickman: Yep, good catch!

tomodwyer (approved):

PR #729, opened 2021-07-27
Reparse content-disposition header to get full filename

bloodearnest:
DRF helpfully strips of leading paths.

### DIFF
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -1,3 +1,5 @@
+import cgi
+
 import structlog
 from django.db import transaction
 from django.http import FileResponse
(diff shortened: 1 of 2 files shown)


jobserver/api/releases.py
```diff
@@ -192,7 +194,12 @@ def post(self, request, release_id):
         except KeyError:
             raise ValidationError({"detail": "No data uploaded"})
 
-        filename = upload.name
+        # DRF validates and parses the Content-Disposition header, but sadly
+        # strips any directy paths from the filename. We need those paths, so
```
ghickman, line 198: s/directy/directory/ I think?

ghickman (approved):

PR #730, opened 2021-07-27
Track React errors using Sentry

### LINKED ISSUE #712: Add error logging to the Outputs Viewer SPA
So we can can be notified when users are having issues.

tomodwyer:
Closes #712 

## TODO

- [x] Set up site on Sentry
- [x] Only run error logging in production


assets/src/scripts/outputs-viewer/index.jsx
```diff
@@ -1,7 +1,18 @@
...
 import React from "react";
 import ReactDOM from "react-dom";
 import App from "./App";
 
+if (
+  process.env.NODE_ENV === "production" &&
+  document.location.hostname === "jobs.opensafely.org"
+) {
+  Sentry.init({
+    dsn: "https://adf34e0d9a5445fea760b43cf10ad45b@o173701.ingest.sentry.io/5881265",
```
ghickman, line 11: Can we build this from an env var?  We have that available in production as `SENTRY_DSN`, and ideally we'd not expose per-env config in the code.

ghickman (changes requested):
Couple of things I think need tweaking before we pull this in.

Is the intention to use Sentry with the react side of things in dev normally?  In Python land we set up the integration in `services/sentry.py` but the client's own init is a no-op if the DSN is empty, could you use that here too?

dotenv-sample
```diff
@@ -19,3 +19,4 @@ SOCIAL_AUTH_GITHUB_KEY=dummy
 SOCIAL_AUTH_GITHUB_SECRET=dummy
 
 DJANGO_VITE_DEV_MODE=False
+VITE_REACT_SENTRY=https://www.example.com/
```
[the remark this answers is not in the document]
tomodwyer: Comment added 👍

assets/src/scripts/outputs-viewer/index.jsx
tomodwyer: Yes, I've added it as a Vite env var which needs to be available when the assets are built. I will get it added to the GitHub Actions.

ghickman (approved):

PR #741, opened 2021-07-30
List Live Files on Backends for Workspace

### LINKED ISSUE #728: Live File Viewer
For privileged users we want a way to view the current files on disk for a Workspace in a particular L4 environment.

Each Backend needs to know where its endpoint exists for this feature so we'll add a field to Backend, and then check if that has content to derive support for the feature.

For a user to see a button for a given backend they also need to have access to that backend.

Clicking a Backends file viewer button will open the SPA, passing in a URL built from the Backend's L4 endpoint field with the suffix `/workspaces/<workspace_name>/`, and signed using the signing code from release-hatch.

We can copy in signing.py from release-hatch for now, until we have a more solid story for cross-project code reuse.

ghickman:
This adds a "Files" section to the workspaces detail page (currently just a button chucked into the mix as we know that we want to redo that whole UI) which only shows up if there are backends for the user to view.

The Files page lists the Backends that 1) a user can see via the membership M2M and 2) have a value set in their `level_4_url` field.

Each Backend displayed on the page links to a new page which loads the SPA, constructing a URL and auth token to release-hatch to serve the index and file endpoints to follow our usual pattern.  The SPA has grown knownledge of an optional auth token, and adds the `Authorization` header if it's present.

I've also added an edit page for Backends so we can easily set the `level_4_url` as we roll this out.

**Files page:**

**Backend Edit page:**

**ToDo:**
* [x] Define permissions required to see live files
* [x] Use permissions to gate access to the button, Files list page, and SPA entrypoint
* [x] Only show the Files button if the user can see valid backends

Fixes #728


jobserver/views/workspaces.py
```diff
@@ -41,6 +43,60 @@ def post(self, request, *args, **kwargs):
...
+        # TODO: handle permissions here, I think we need a new permission for
+        # this?
+
+        f = furl(backend.level_4_url)
+
+        # build the base URL for which we want the auth token to authenticate,
+        # all paths beneath this one will be valid with the token
+        f.path.segments += ["workspace", workspace.name]
+        auth_token = build_auth_token(
+            url=request.build_absolute_uri(f.url),
```
bloodearnest, line 79: huh? Shouldn't this be building an absolute_uri for release-hatch, not for job-server?

Maybe I don't understand furl/django well enough here, the furl url is already absolute, no? What's the reasone for using build_absolute_uri from *this request*?

Anyway, see my comments above regards build_auth_token, I think the furl stuff could probably move in there, and we passing in `backend` and `workspace`?

jobserver/releases.py
```diff
@@ -1,16 +1,29 @@
...
 from .models import Release, ReleaseFile
 from .models.outputs import absolute_file_path
+from .signing import AuthToken
 
 
 class ReleaseFileAlreadyExists(Exception):
     pass
 
 
+def build_auth_token(*, url, username, expiry=None, backend_token):
```
bloodearnest, line 17: maybe it's worth abstracting this to a backend and workspace, rather than a url? so sth like:
 
```
def build_auth_token(backend, workspace, user, expiry=None):
    ...
    f = furl(backend.level_4_url)
    f.path.segments += ["workspace", workspace.name]

    token = AuthToken(url=f.url, user=user.username, expiry=expiry)

    return token.sign(key=backend.backend_token, salt="hatch")
```

This means the url generation is abstracted out, and we can change it as needed without the view needing to care.

jobserver/templates/workspace_backend_files.html
```diff
@@ -0,0 +1,40 @@
...
+      </a>
+    </li>
+
+    <li class="breadcrumb-item active" aria-current="page">{{ backend.display_name }} Files</li>
+
+  </ol>
+</nav>
+
+<p class="mb-2">
+  Files currently on <strong>{{ backend.display_name }}</strong> in the
```
bloodearnest, line 40: Perhaps "Medium privacy (Level 4 ) files currently on ...."?

jobserver/templates/workspace_file_list.html
```diff
@@ -0,0 +1,56 @@
...
+    </li>
+
+    <li class="breadcrumb-item active" aria-current="page">Files</li>
+
+  </ol>
+</nav>
+
+<div class="row">
+  <div class="col-lg-8 offset-lg-2">
+    <h3>Files</h3>
```
bloodearnest, line 35: "Medium Privacy (Level 4) Files"? Files on it's own is a bit vague?

bloodearnest (commented):
I am confused by `request.build_absolute_uri`! But other than than, looks good.

Think we should refactor build_auth_token to be a bit smarter/high-level.

jobserver/views/workspaces.py
bloodearnest: Hmm I now release you need the url as to pass to SPA.

Maybe return a tuple?

`return f.url, token.sign(backend.backend_token, salt="hatch")`

ghickman: I think I got confused over what URL this was being built for when writing an early version of this.  Coming back to it I'm a little confused too!

ghickman:
@bloodearnest – updated as suggested, and added permissions checks too.

bloodearnest (approved):

PR #742, opened 2021-08-02
Various fixes for Outputs SPA

### LINKED ISSUE #723: Investigate Consolidating PrepareButton & PublishButton
Both buttons are very similar in that they:
* are displayed based on the presence of a URL
* POST to the given API
* need `csrfToken` to be able to POST
* use React Query's `useMutation`hook to do the POST

We've taken a couple of very quick runs at consolidating them but an obvious abstraction didn't present itself so parking until we have a bit more time.

My working theory is pulling the mutation outside the button component would mean we could pass it and an onClick handler in as a prop.

tomodwyer:
- Update `useWindowSize` and `useFile` hook file names
- Improve error message on Viewer component
- Update metadata styles
- Add a fallback for an unknown file size
- Add extra sorting options into `useFileList` hook
- Create a consolidated `<Button />` component
    - closes #723


ghickman (approved):
Looks great!

Issue #642, opened 2021-07-01
Remove can_run_jobs Function


ghickman:
This function gates most of the functionality of the site.  It was added when we removed the login requirement for Users to be a part of the `opensafely` GitHub org, and calls that check on demand.  It's used in various places:

| Module | User Action | Proposed Permission | Notes |
| - | - | :-: | - |
| `views/jobs.py`  | Cancelling a job  | `cancel_job` | |
|`views/job-requests.py` | Cancelling multiple jobs | `cancel_job` | |
|`views/index.py` | Create a workspace | `create_workspace` | |
|`views/workspaces.py` | Archive a workspace | `archive_workspace` | |
| | Create a workspace | `create_workspace` | |
| | Run Jobs | `run_job` | |
| | Add a Job | ? | This is a button on the Logs page, now that we have breadcrumbs we can probably just remove it. |
| | Toggle Notifications | `edit_workspace_notifications` | |

Most uses can be easily broken down into permission-based checks, but which roles they go on is not entirely clear, particularly if we're locking down the ProjectCollaboration role.

PR #745, opened 2021-08-02
Remove Some Uses of can_run_jobs

ghickman:
This switches cancelling jobs on either the Job or JobRequest detail pages to use the `cancel_job` permission instead of relying on `can_run_jobs()`.  This refines who can cancel jobs, to only those with either the CoreDeveloper role or the ProjectDeveloper role, the latter of which is tied to the Project any given Job[Request] is owned by.

Ref #642

### DIFF
--- a/jobserver/views/job_requests.py
+++ b/jobserver/views/job_requests.py
@@ -1,18 +1,15 @@
-from django.contrib.auth.decorators import user_passes_test
 from django.db.models import Q
 from django.http import Http404
 from django.shortcuts import redirect
-from django.utils.decorators import method_decorator
 from django.utils.html import escape
 from django.utils.safestring import mark_safe
 from django.views.generic import DetailView, ListView, View
 from django.views.generic.edit import FormMixin
 
-from ..authorization import CoreDeveloper, has_role
+from ..authorization import CoreDeveloper, has_permission, has_role
 from ..forms import JobRequestSearchForm
 from ..models import Backend, JobRequest, User, Workspace
 from ..project import render_definition
-from ..roles import can_run_jobs
 from ..utils import raise_if_not_int
 
 
@@ -34,7 +31,6 @@ def filter_by_status(job_requests, status):
     return [r for r in job_requests if r.status.lower() == status]
 
 
-@method_decorator(user_passes_test(can_run_jobs), name="dispatch")
 class JobRequestCancel(View):
     def post(self, request, *args, **kwargs):
         try:
@@ -42,6 +38,12 @@ class JobRequestCancel(View):
                 pk=self.kwargs["pk"]
             )
         except JobRequest.DoesNotExist:
+            raise Http404
+
+        can_cancel_jobs = job_request.created_by == request.user or has_permission(
+            request.user, "cancel_job", project=job_request.workspace.project
+        )
+        if not can_cancel_jobs:
             raise Http404
 
         if job_request.is_completed:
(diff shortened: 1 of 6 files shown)


jobserver/views/job_requests.py
```diff
@@ -62,16 +64,20 @@ class JobRequestDetail(DetailView):
     template_name = "job_request_detail.html"
 
     def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
-        context["project_definition"] = mark_safe(
-            render_definition(
-                escape(self.object.project_definition),
-                self.object.get_file_url,
-            )
+        can_cancel_jobs = has_permission(
```
bloodearnest, line 67: Can we make so a user can always cancel their own jobs? But can only cancel other peoples if they have this permission?

So sth like
```
can_cancel_jobs = job.user == request.user  or has_permission(...)?
```

bloodearnest (commented):
Looks good - can we make it so users can always cancel their own jobs?

I think that's like the 80% use case

jobserver/views/job_requests.py
sebbacon: Agreed. In terms of implementation, it'd be better to incorporate this into the framework (i.e. we've a concept of "local roles" which are non-global / pertaining to a given context) - this may or may not be straightforward right now (no time to refamiliarise myself with the code), but could we note it as a good use-case to consider in the next voyage (where we'll be reviewing and finishing all the permissions framework) @ghickman ?

ghickman: There's a few bits to unpick here!

> always cancel their own jobs

Cancelling a job is gated on the `cancel_job` permission, which is exposed via the `ProjectDeveloper` role.  This role is also how one runs a job, the idea being that running and cancelling are separate actions so they should have separate permissions but in current use it doesn't make sense to separate them so they both go on the same role.  So, with how the site is currently used I don't think the `job.user == request.user` check gains us anything?

> incorporate this into the framework

tldr: I see where you're coming from but it doesn't really gel how jobs work or (I think?) gain us anything.

To put a role on the relationship between a Job and User we'd need to create that relatonship as a thing we can target (new model with a `RolesField`, FKing both), set up a new role to encapsulate just the `cancel_job` permission, and automatically apply that role when a Job was created.  This is complicated by Jobs being owned by job-runner and job-server acting like a read-only consumer of their data.  However, while all of those points are surmountable if not ideal, I think where this falls apart is comparing it to the creator of a Job which provides us the same relationship without needing roles or permissions.

(side note: this made me think through the whole local/global context section of the framework again 👍)

sebbacon: By "incorporate into the framework" I was being deliberately vague / evasive, sorry!

So to elaborate, in my experience most permissions systems inevitably grow some _computed_ (rather than explicit) roles over time (for example is "authenticated user").  The computed role I have in mind here is "owner", which is typically computed along the lines `context.creator == logged_in_user`.

In this case, it _could_ make sense for `cancel_job` to be a permission of the `Owner` role, and `create_job` to be a permission of the `ProjectDevelopers` role. I think at the moment we're saying `ProjectDevelopers` can all cancel _each others_ jobs, in the context of _a given project_, which is also a completely valid way to look at it.

In other words, there are lots of ways to skin these kinds of decisions, a none of of them are necessarily obviously the right ones, which is why I think it's worth coming back to this as a good example in the next voyage to create a well-documented permissions model on which we can have some wide consensus (with real users as well as developers). Otherwise the can of worms risks remaining permanently opened :)

ghickman: aha, ok!

This is definitely a useful discussion to have had, but I'm unclear of the takeaway for my single boolean test…?

bloodearnest: My proposal is that `cancel_job` becomes the permission needed to cancel other people's jobs, that's all

You can always cancel your own jobs, no permissions check needed, just that the loggedin user is the user who created the job.

bloodearnest: Not every check needs to be implemented as a permission?

evansd: I think there's a big advantage to having every permissions check go through a common codepath so there's one place where everything is enforced.

While knowing absolutely nothing about the current implementation it looks to me like the permissions check ought to be job level rather than project level i.e.
```py
has_permission(request.user, "cancel_job", job=job)
```
If the job belongs to a project on which the user has "cancel_job" permission then the answer's "yes". Otherwise, we check if the user is the job's creator.

But as Seb says, I think this is more a good example to come back to later when we're thinking more about permissions stuff rather than a thing which needs solving now.

sebbacon: I don't think there's any particular takeaway for right now that I feel strongly about, sorry! 😆

ghickman: Captured the further context bits here: https://github.com/opensafely-core/job-server/issues/795

Added ownership checks and tests and rebased in, do you want to check them again @bloodearnest?

bloodearnest (approved):

PR #758, opened 2021-08-03
Set up state management for the Outputs SPA

### LINKED ISSUE #710: Add a state management library for the Outputs Viewer SPA
So that we don't need to pass props around.

tomodwyer:
Closes #710 

- Use [Zustand](https://github.com/pmndrs/zustand) as a state management tool for the Outputs SPA


assets/src/scripts/outputs-viewer/hooks/use-file.js
```diff
@@ -1,24 +1,27 @@
...
-    return fetch(file.url, config)
+    return fetch(
+      file.url,
+      authToken
+        ? {
+            headers: new Headers({
+              Authorization: authToken,
+            }),
+          }
+        : null
```
ghickman, line 23: I'm not sure we need this conditional.  I had the same idea with conditionally populating `config` originally, and decided to just just always set it with either a value or empty string in useFileList (and then forgot about useFile, oops!), and I don't see an issue with sending the extra header on the request for a bit less code!

ghickman (approved):
One comment on normalising how we add the Authorization header, but otherwise this is a huge improvement all round 👍

assets/src/scripts/outputs-viewer/hooks/use-file.js
tomodwyer: Cool - I left that in there incase you were handling it in a special way on the server-side.

I've removed it and set up headers for both fetch requests.

PR #763, opened 2021-08-04
Better Outputs Buttons

ghickman:
This tidies up the outputs buttons on a WorkspaceDetail page so they're in the order one gains access to them, and improve the logic around when they're added.

I've also removed the two buttons (with different names!) pointing to the "current" outputs (latest version of every file released in a workspace) since that's available via the Released Outputs and it felt confusing in the UI.

### DIFF
--- a/jobserver/templates/workspace_detail.html
+++ b/jobserver/templates/workspace_detail.html
@@ -79,13 +79,7 @@
       <div class="d-flex">
         {% if can_use_releases %}
         {% if user_can_view_outputs %}
-        <a class="btn btn-primary mr-2" href="{{ workspace.get_current_outputs_url }}">
-          Current Outputs
-        </a>
         <a class="btn btn-primary mr-2" href="{{ workspace.get_outputs_url }}">Outputs</a>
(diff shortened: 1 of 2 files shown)


jobserver/templates/workspace_detail.html
```diff
@@ -78,21 +78,19 @@ <h3>{{ workspace.name }}</h3>
...
-        </a>
-        <a class="btn btn-primary mr-2" href="{{ workspace.get_outputs_url }}">Outputs</a>
-        <a class="btn btn-primary mr-2" href="{{ workspace.get_current_outputs_url }}">
-          Publish
-        </a>
-        {% endif %}
 
         {% if user_can_view_files %}
-        <a class="btn btn-primary mr-2" href="{{ workspace.get_files_url }}">Backend Files</a>
+        <a class="btn btn-primary mr-2" href="{{ workspace.get_files_url }}">Current Outputs</a>
```
bloodearnest, line 83: I wonder if "Level 4 Outputs" is less ambiguous, as "Current" could be interpreted as current level 4, current released, currently published?

bloodearnest (approved):
LGTM, with a possible wording change

PR #769, opened 2021-08-04
Disable Viewing Files when Release is Empty

### LINKED ISSUE #751: Disable Files Button for Releases with No Files

ghickman:
This disables the view and files buttons on the ReleaseList page when a Release has no files to view (eg from a broken upload) and returns a 404 on the ReleaseDetail page if there are no files for a Release.

Fixes #751

### DIFF
--- a/jobserver/templates/workspace_release_list.html
+++ b/jobserver/templates/workspace_release_list.html
@@ -42,7 +42,12 @@
 
       <div class="d-flex justify-content-between">
         <span>
-          Files released by {{ release.created_by.name }} {{ release.created_at|naturaltime }}
+          {% if release.files.exist %}
+          Files released
+          {% else %}
+          Release
+          {% endif %}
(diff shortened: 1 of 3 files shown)


bloodearnest (approved):
So, AIUI, this will prevent launching the SPA if there are no files yet uploaded to a release. But we can still view the intended files by clicking on the 'Files' button.

So, that seems reasonable to me as a next step, but wonder if we really need the `Files` button in the long run?

An alternate UI would be that we load the SPA with the list of files, but index API entry has an empty url, and so clicking on that file shows a 'not approved/uploaded yet' page in the content pane of the SPA. We'll need to support this anyway for partially approved uploads.

ghickman:
No, the files button is also disabled since it's the control for a bootstrap-collapsed div (with a simple list of file names in it).

Agreed on potentially removing the files button though!

PR #770, opened 2021-08-04
Replace Google Fonts with local fonts

tomodwyer:
So that we can serve the correct fonts in secure environments.

This PR also removes the need for local env mixins in the Scss files.


madwort:
Do we need to bundle the font licence as well as the font files?

PR #772, opened 2021-08-04
File size fixes for the Outputs SPA

tomodwyer:
- Show file size as `title` in the file list
- Improve performance when rendering CSVs
- Stop large CSVs from being previewed


ghickman (approved):
👍

PR #776, opened 2021-08-04
Download Releases

### LINKED ISSUE #766: Download Releases in a Zip
Output checkers need to be able to download the files for a release to look through them sometimes.  Let's add a download button to the ReleaseDetail template.  The bundling of files into a zip should be a function in `jobserver/releases.py` taking an iterable of ReleaseFiles since we'll almost certainly use it in other places.

ghickman:
This adds the ability for privileged users (those with ProjectCollaborator role/`view_release_file` permission) to download a zip of a Release.

**Downlad from Release list page**

**Download from ReleaseDetail page** 

Fixes #766

### DIFF
--- a/tests/jobserver/test_releases.py
+++ b/tests/jobserver/test_releases.py
@@ -1,5 +1,6 @@
 import random
 import string
+import zipfile
 from datetime import timedelta
 
 import pytest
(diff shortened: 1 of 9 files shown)


tests/jobserver/test_releases.py
```diff
@@ -52,6 +53,18 @@ def test_build_hatch_token_and_url_with_custom_expires():
...
+@pytest.mark.django_db
+def test_build_outputs_zip():
+    release = ReleaseFactory(ReleaseUploadsFactory(["test1"]))
+
+    zf = releases.build_outputs_zip(release.files.all())
+
+    with zipfile.ZipFile(zf, "r") as zip_obj:
+        assert zip_obj.testzip() is None
+
+        assert zip_obj.namelist() == ["test1"]
```
bloodearnest, line 65: I'd be tempted to check that the extracted contents of "test1" match

bloodearnest (approved):
I like the convention of `get_$TYPE_url()`

ghickman:
> I like the convention of `get_$TYPE_url()`

Yes!  It's really been making life a lot easier in the templates with all the different pages we're adding in this recent run.

PR #779, opened 2021-08-05
Download Latest Outputs

### LINKED ISSUE #778: Download Latest Outputs as Zip
Similar to #766 we want a way to download the current files for a Workspace.

ghickman:
This adds the ability to download a zip of a Workspace's latest outputs for users with the correct permissions.  It also switches from current to latest to define these files, with code and UI updates to reflect that.  Current had become abiguous between Release, Snapshot, and Outputs.

Fixes #778


bloodearnest (approved):
LGTM

PR #781, opened 2021-08-05
Download Snapshots

### LINKED ISSUE #767: Download a Snapshot as a Zip
Similar to #766 we want the general public to be able to download published outputs, and privileged users to be able to download draft versions.

ghickman:
This lets users download the files in a Snapshot (published outputs) as a zip.  Users must be a ProjectCollaborator to download the files in a draft Snapshot.

Fixes #767

### DIFF
--- a/jobserver/templates/snapshot_detail.html
+++ b/jobserver/templates/snapshot_detail.html
@@ -36,14 +36,20 @@
   </ol>
 </nav>
 
-<h1 class="h3">
-  {% if snapshot.is_draft %}
-  Outputs for {{ snapshot.workspace.name }}
-  <span class="badge badge-secondary">Draft</span>
-  {% else %}
-  Published outputs for {{ snapshot.workspace.name }}
-  {% endif %}
-</h1>
+<div class="d-flex justify-content-between align-items-center">
+  <h1 class="h3">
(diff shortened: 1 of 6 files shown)


jobserver/templates/snapshot_detail.html
```diff
@@ -36,14 +36,20 @@
...
-  Outputs for {{ snapshot.workspace.name }}
-  <span class="badge badge-secondary">Draft</span>
-  {% else %}
-  Published outputs for {{ snapshot.workspace.name }}
-  {% endif %}
-</h1>
+<div class="d-flex justify-content-between align-items-center">
+  <h1 class="h3">
+    {% if snapshot.is_draft %}
+    Outputs for {{ snapshot.workspace.name }}
```
bloodearnest, line 42: I might be tempted to mirror this language, rather than just badge it. So either output

"Draft outputs for ..." or  "Published outputs for ..."

but just a suggestion

bloodearnest (approved):
I kinda like how all these are views are basically just permission checks.

jobserver/templates/snapshot_detail.html
ghickman: Yeah, good call 👍

PR #788, opened 2021-08-05
Delete ReleaseFiles

### LINKED ISSUE #759: Delete ReleaseFiles
We need to be able to delete ReleaseFiles from the UI.  We can mark them as deleted with a flag, but the files on disk need to be deleted.

The SPA will also need to reflect a ReleaseFile has been deleted, so the index endpoints will need to know about it, and the 404 response from a file endpoint needs to have some text in it.

ghickman:
This gives us buttons on the rolled down files of a Release (in any release list page):

ReleaseFiles have their on-disk file deleted, but the object itself is left untouched.

The SPA no longer links to files which have been deleted:

Fixes #759


bloodearnest (commented):
This seems fine, but is not quite what I expected - I thought we were also going to track timestamp/user of deletions? Is that gonna be a later thing?

ghickman:
> Is that gonna be a later thing?

It is not that you're reminded me about it 😬

jobserver/api/releases.py
```diff
@@ -135,6 +135,7 @@ def get_size(rfile):
                 date=rfile.created_at.isoformat(),
                 sha256=rfile.filehash,
                 size=get_size(rfile),
+                is_deleted=not rfile.absolute_path().exists(),
```
bloodearnest, line 138: I'd be inclined to changed this to `is_deleted=file.deleted_at or not rfile.exists()`. The reason being that the failure mode is safer - if we somehow managed to set `deleted_at` and `deleted_by`, but forgot to delete the file, then the users would still be able to see the file. Better to disable the SPA UA on either condition

But a minor corner case, so no worries.

bloodearnest (approved):

ghickman:
@bloodearnest – updated as requested.  I've also abstracted the template for the release lists into an include since it's used in three places, unfortunately that required building a structure in the views since the titles couldn't be built in the templates.

PR #789, opened 2021-08-06
Add Release Upload Permission

### LINKED ISSUE #754: Add upload_release_file Permission
OutputChecker looks like the appropriate role.  We need to use it in the API for uploading ReleaseFiles.

ghickman:
This adds a permission for uploading ReleaseFiles, continuing to use `validate_upload_access` function.

It also adds some tests for the `validate_upload_access` function rather than relying on testing via the view and validates that the calling User has access to the Backend via a BackendMembership.

Fixes #754


jobserver/api/releases.py
```diff
@@ -81,7 +81,14 @@ def validate_upload_access(request, workspace):
     except User.DoesNotExist:
         raise NotAuthenticated
 
-    # TODO: check user permissions
+    # check the user has access to this backend
+    if user not in backend.members.all():
```
bloodearnest, line 85: oh! good catch

bloodearnest (approved):
Looks great.

I'll add a ticket to look at moving the API authentication over to using a signed token rather than passing the BACKEND_TOKEN in a header, as it'd be a nice clean up.

PR #796, opened 2021-08-09
Directly Link to Files

### LINKED ISSUE #780: Directly Link to Files in the SPA
We'd like to be able to directly link to files inside the SPA.  Currently we capture various paths as entry points to the SPA, and each of those has a second URL capture for any further path components, however don't then pass those in.

We'll need to decide how the SPA will handle consuming of paths in the URL, ie from Django via the template or pull them in from the URL itself.

ghickman:
This changes the URL when a file is clicked, and opens a file if a path is in the URL so one can directly link to any SPA view.

Fixes #780


bloodearnest (approved):
python stuff makes sense, haven't got my head round the JS state stuff yet, so can't really provide feedback on that

PR #803, opened 2021-08-10
Render FileList using a virtualisation library

### LINKED ISSUE #802: Use a virtualisation library to render the FileList
So that we can improve performance of large lists in the Outputs SPA.

tomodwyer:
[React window](https://github.com/bvaughn/react-window) works by only rendering part of a large data set (just enough to fill the viewport). This helps address some common performance bottlenecks:

- It reduces the amount of work (and time) required to render the initial view and to process updates.
- It reduces the memory footprint by avoiding over-allocation of DOM nodes.

We can use this library for the FileList to help alleviate performance issues.

---

Closes #802

### DIFF
--- a/assets/src/scripts/outputs-viewer/components/FileList/FileList.jsx
+++ b/assets/src/scripts/outputs-viewer/components/FileList/FileList.jsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useRef, useState } from "react";
 import { useHistory, useLocation } from "react-router-dom";
+import { FixedSizeList } from "react-window";
 import useFileList from "../../hooks/use-file-list";
 import useWindowSize from "../../hooks/use-window-size";
 import useStore from "../../stores/use-store";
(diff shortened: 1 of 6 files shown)


assets/src/scripts/outputs-viewer/components/FileList/FileList.jsx
```diff
@@ -29,7 +30,7 @@ function FileList() {
     // If there are horizontal scrollbars, minus 17px for the scrollbar
     const fileListHeight =
       window.innerHeight -
-      listEl.current?.getBoundingClientRect().top -
+      (listEl.current?.getBoundingClientRect().top || 0) -
       30 -
       (hasScrollbarX ? 17 : 0);
```
ghickman, line 36: Might be worth pulling some of these calculations out into separate values to help with readability?  The comment is great but I find myself jumping back and forth between the magic numbers and the explanation!

ghickman (approved):
👍

assets/src/scripts/outputs-viewer/components/FileList/FileList.jsx
tomodwyer: Good point, added some more in.

PR #835, opened 2021-08-19
Add Timezone to Timings Hovers

### LINKED ISSUE #451: Handle Timezones for Timestamps
We store timestamps with timezones attached (`USE_TZ` is set to True in settings.py) but not all viewers of the site are in UTC (thanks BST).

When we use the `naturaltime` template filter we're generating a humanized time based on the UTC timestamp, but this is incorrect to a viewer in eg BST.  We should make the display of those timestamps accurate for the viewer.

ghickman:
This adds the timezone, in the format +0000, to all timestamp hovers across the site (where we use the `title` attribute) and a short explainer to why timestamps are all in UTC.

We can't reliably get the local timezone for a User without asking them to set it on their account which currently felt overkill.

Fixes #451

### DIFF
--- a/jobserver/templates/_job_requests.html
+++ b/jobserver/templates/_job_requests.html
@@ -29,7 +29,7 @@
       <div class="d-none d-md-block action text-truncate" title="{{ group.requested_actions|join:"," }}">
         {{ group.requested_actions|join:"," }}
       </div>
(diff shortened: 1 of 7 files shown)


StevenMaude (approved):
This seems reasonable over having to get the user to set a timezone.

Issue #711, opened 2021-07-20
Add tests to the Outputs Viewer SPA


tomodwyer:
So that we have confidence our code is working

PR #837, opened 2021-08-20
Various fixes for the Outputs SPA

tomodwyer:
Whilst adding tests (https://github.com/opensafely-core/job-server/issues/711) - various issues with the code have been identified. This PR covers fixes to the Outputs SPA as the first part of adding tests.

---

One issues identified was added in https://github.com/opensafely-core/job-server/pull/821:

> Replace `data` prop with query in each preview component

This meant that each child component of the `<Viewer />` required knowledge of the state of the query. This overcomplicates each child component, and therefore has been reverted to pass `data` as a prop to the child components.


ghickman (approved):
👍

PR #839, opened 2021-08-20
Stop assigning ProjectCoordinator on Project creation

### LINKED ISSUE #838: Don't Assign ProjectCoordinator to New Projects
We've moved to only allowing Amir to be a ProjectCoordinator for now so all collaborators must going through IG sign off to use the platform, so we shouldn't be handing out that role as part of Project creation anymore either.

ghickman:
Fixes #838


inglesp (approved):
The code appears to do what the commit message says!

PR #840, opened 2021-08-20
Remove Old can_run_ jobs Function

### LINKED ISSUE #642: Remove can_run_jobs Function

ghickman:
This removes the use of the `can_run_jobs` function in favour of permissions based checks.  That function checked the user was logged in, and a member of the `opensafely` GitHub Organisation on each call.  Since moving to requiring all users are member of that org on each login we can now rely on it being true for all users.

Both new permissions have ended up on ProjectDeveloper as it made the most sense for what is already possible as a user of the platform.

Fixes #642


jobserver/views/workspaces.py
```diff
@@ -122,6 +117,11 @@ def dispatch(self, request, *args, **kwargs):
         if not can_manage_workspaces:
             raise Http404
 
+        gh_org = self.request.user.orgs.first().github_orgs[0]
+        self.repos_with_branches = sorted(
+            get_repos_with_branches(gh_org), key=lambda r: r["name"].lower()
+        )
+
```
CarolineMorton, line 124: Would you be able to explain what this is doing?

CarolineMorton (approved):
Makes this much clearer. My understanding, and correct me if I am wrong, is that this groups the former can_run_jobs function / check into the existing permissions structure.Looks much simpler!

jobserver/views/workspaces.py
ghickman: Of course 🙂

`self.request`: in a Class-Based View (CBV) you can always access the current request on the view instance with `self.request`.
`self.request.user`: the current user, either a logged in one or an instance of `AnonymousUser`.
`self.request.user.orgs.first()`: the first related Org connected to a User.  `first()` gives you an instance (rather than a QuerySet) or None.
`self.request.user.orgs.first().github_orgs[0]`: Org has a `github_orgs` field, get the first item in it.

There are a couple of assumptions here:
1. the user has any Orgs, not always true currently but they shouldn't be able to reach this page in that state and we're aiming to address this in the next voyage.
2. The Org has at least one `github_org`, I manually add Orgs so they always have one of these set.

`get_repos_with_branches(gh_org)`: [this function](https://github.com/opensafely-core/job-server/blob/c6931968085aec936d4173f41896d0376cc90136/jobserver/github.py#L166-L199) gets all repos in the `opensafely` GitHub Org, in the `researchers` GitHub Team.  This returns a list of dicts containing name, url, and a list of branches.  Under the hood it uses GitHub's GraphQL API so we can do the minimum number of calls to pull a small bit of info from each repo.

The final piece is sorting that list of dicts by the name key, so the repo list when creating a workspace is ordered alphabetically.

PR #844, opened 2021-08-20
Add Staff Area

### LINKED ISSUE #592: Add way to look up workspace by repo
Given a repo URL, it would be very handy to be able to find its workspace.

### LINKED ISSUE #632: New Users Don't Have an Org
When signing up to the service Users aren't immediately added to an Organisation.  We use Org membership to drive access to GitHub Organisations for various parts of the service.

We want GitHub Orgs to be stored on the Org because of the work we're doing with Graphnet, but we also don't have a flow for adding Users to a job-server Org on sign up yet.

[Example Sentry](https://sentry.io/organizations/ebm-datalab/issues/2484512785/?project=5443358&query=is%3Aunresolved)

### LINKED ISSUE #737: Add an Inspector for Staff-only Tooling
We have a handful of pages which are only for privileged staff users.  Currently they're linked to in the main menu, with no real distinction over the rest of the menu (there's a separator).  We're expecting to add more pages like this over time as the project expands and we need more tooling to support users.

The model I've used for this before was an "inspector", which is "a tailored Django admin" where pages are built around the concepts you need in the site (eg for us Backends and Users currently).

Let's pull the pages into a new Django app, with a URL namespace of `inspector`, and URL prefix of `/inspector/`.

We can link to the inspector in the drop down menu below a Users avatar, although if that gets annoying we could also link to it above the current menu with a different colour so it's obviously a different section and doesn't interfere with screenshots.

ghickman:
This adds a staff-only set of pages designed to make common tasks easier for Datalabbers.  Each detail page is expected to have any necessary tools as buttons on that page or links to pages where buttons can provide those actions (eg adding an Org for a User).

We already had two sets of these pages, Backends and Users, which were stuffed into the main nav confusing just about everyone except George because he added them.  This breaks out those, and adds the new ones into a new app.

This would be a good one to test out locally.

Access to the staff area is done with a new avatar-menu item:

The index page provides search for most things:

A separate menu sits above the main site menu only on inspector pages:

Fixes #737 
Fixes #632
Fixes #592


staff/views/index.py
```diff
@@ -0,0 +1,80 @@
...
+    results.
+
+    """
+    if not q:
+        return []
+
+    queries = []
+    for target in configured_searches:
+        qwargs = functools.reduce(
+            operator.or_, (Q(**{f"{f}__icontains": q}) for f in target["fields"])
```
inglesp, line 50: not even a nit: You can also do `Q.__or__`, to avoid an import!

staff/views/index.py
```diff
@@ -0,0 +1,80 @@
...
+
+def get_results(q):
+    """
+    Turn a search term into queries for each configured model
+
+    This takes the given search term and for each model configured to be
+    searchable does:
+
+        1. creates a Q(field=search_term) instance
+        2. combines them with a bitwise OR, typically expressed with a | (pipe)
```
inglesp, line 35: nit: it's not technically a bitwise OR, but a logical OR (although it uses the syntax for a bitwise operation).

staff/views/index.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+# configure searchable models here, each must have get_staff_url defined
+configured_searches = [
+    {"model": Backend, "fields": ["name", "slug"], "order_by": "name"},
+    {
+        "model": User,
+        "fields": ["first_name", "last_name", "username"],
+        "order_by": "username",
+    },
```
inglesp, line 22: Personal preference: I'd find this slightly easier to read if each key was on its own line.  You can get Black to do this by adding a comma before the closing brace.

Thanks to Black, I've missed arguing about style in PRs!

staff/views/index.py
```diff
@@ -0,0 +1,80 @@
...
+class Index(View):
+    def get(self, request, *args, **kwargs):
+        q = self.request.GET.get("q")
+        results = get_results(q)
+
+        # group the results by their model's classname for display in the view
+        results = itertools.groupby(results, key=lambda m: m.__class__.__name__)
+
+        # consume the groupby generator such that Django won't butcher it
+        # https://stackoverflow.com/a/16171518/158304
```
inglesp, line 69: Aargh, TIL.

staff/views/index.py
```diff
@@ -0,0 +1,80 @@
...
+    {"model": Backend, "fields": ["name", "slug"], "order_by": "name"},
+    {
+        "model": User,
+        "fields": ["first_name", "last_name", "username"],
+        "order_by": "username",
+    },
+]
+
+
+def get_results(q):
```
inglesp, line 27: Really nice.

staff/views/users.py
```diff
@@ -0,0 +1,133 @@
...
+    def get_context_data(self, **kwargs):
+        return super().get_context_data(**kwargs) | {
+            "orgs": self.object.orgs.order_by("name"),
+            "projects": self.object.projects.order_by("name"),
+        }
+
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+
+        # UserForm isn't a ModelForm so don't pass instance to it
```
inglesp, line 50: Should we be using `UpdateView` then?

staff/views/users.py
```diff
@@ -0,0 +1,133 @@
...
+
+        role = self.request.GET.get("role")
+        if role:
+            qs = qs.filter_by_role(role)
+
+        return qs
+
+
+@method_decorator(require_permission("user_manage"), name="dispatch")
+class UserSetOrgs(FormView):
```
inglesp, line 106: Looks like this duplicates most of `UserDetail` -- do we need both?

inglesp (approved):
All good, well done.

staff/views/index.py
ghickman: ooh thank you, I almost certainly just read the `|` docs!

staff/views/index.py
ghickman: TIL 👍

staff/views/users.py
ghickman: Yes, we get loads for free (I tried it without!), it's just that we don't want UserForm to be a model form (again it _could_ be, we're right in the grey area here).

staff/views/users.py
ghickman: I didn't want Org membership to be on the same page as everything else (it was getting a little large), but there's nothing stopping it being there.

staff/views/index.py
ghickman: Fixed in https://github.com/opensafely-core/job-server/pull/941

PR #851, opened 2021-08-23
Fix Homepage for Unauthenticated Users

ghickman:
This fixes the test for checking the page without users by making sure there is a Workspace so that whole section of the template loads, and fixes the actual bug by only showing the button block when the user is authenticated.

The broken version didn't deploy thanks to the deploy checks at least.

### DIFF
--- a/jobserver/templates/index.html
+++ b/jobserver/templates/index.html
@@ -56,6 +56,7 @@
       {% endfor %}
     </div>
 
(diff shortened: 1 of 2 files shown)


iaindillingham (approved):

iaindillingham (commented):
LGTM 🚀

PR #855, opened 2021-08-24
Better Timing Labels

### LINKED ISSUE #610: Use Better Timings Labels
The current labels for Job timings have led to some confusion, particularly when a Job is cancelled.

Take this cancelled Job: https://jobs.opensafely.org/jobs/nkwdbphyc34fnc57/

The timings section states that it "started" at a particular time, "finished" at another time, and had a runtime of 27 minutes.  We know from the job-runner logs that this was not actually the case.

Due to the way job-runner works (optimised in opensafely-core/job-runner#22) the flow goes something like this:
1. a JobRequest is seen
2. a Job is created and started_at populated
3. Job data is synced up to job-server
4. job-runner moves onto another Job
5. the second Job takes a long time to manage (this is commonly seen when there are large file transfers involved)
6. User requests the cancellation of a Job on job-server
7. job-runner spends several minutes still dealing with the second Job
8. job-runner sees the request to cancel, cancels the relevant Job
9. Job data is synced up to job-server

None of the steps here are incorrect in the design we created, however how we dispaly the outcomes to the user is.

Since started_at is more accurately the acknowledgement of a Job moving to the `pending` (or `running`) state we should label it `Acknowledged`.

Runtime is also confusing but we've yet to come up with a better label, however we know a Job has been cancelled when `status_mesage` has the value `Cancelled by user`.  We could expose an `is_cancelled` check on the Job object aid ease of use in the templates.

ghickman:
This adds explanations for what users are seeing on a jobs page, particularly around the timings of jobs.

Having tried out the suggestions in #610 I opted not to use long form labels as it made the section look very crowded.  The smaller explainer paragraphs mean a regular user can get the information they need at a glance but the extra detail is still there if needed.

I've also threaded in some "live" information to the top and timing paragraph to help with the context of what a secure backend actually is.

Fixes #610

### DIFF
--- a/jobserver/templates/job_detail.html
+++ b/jobserver/templates/job_detail.html
@@ -117,8 +117,10 @@
       <h3>Timings</h3>
 
       <p>
-        The timestamps below are generated and stored using the UTC timezone on
-        the secure backend which created this job, {{ job.job_request.backend.name }}.
+        The timestamps below are generated and stored using the
+        <a href="https://en.wikipedia.org/wiki/Coordinated_Universal_Time">UTC</a>
+        timezone on the secure backend which created this job,
+        {{ job.job_request.backend.name }}.
       </p>
 
       <div class="mb-2">
--- a/jobserver/templates/job_detail.html
+++ b/jobserver/templates/job_detail.html
@@ -117,39 +117,61 @@
       <h3>Timings</h3>
 
       <p>
-        The timestamps below are generated and stored using the UTC timezone.
+        The timestamps below are generated and stored using the UTC timezone on
+        the secure backend which created this job, {{ job.job_request.backend.name }}.
       </p>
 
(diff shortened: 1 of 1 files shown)


jobserver/templates/job_detail.html
```diff
@@ -112,40 +117,62 @@ <h3>Config</h3>
       <h3>Timings</h3>
 
       <p>
-        The timestamps below are generated and stored using the UTC timezone.
+        The timestamps below are generated and stored using the UTC timezone on
+        the secure backend which created this job, {{ job.job_request.backend.name }}.
```
CarolineMorton, line 121: Link out to what UTC is. Obvious to coders but not to other folks I don't think

CarolineMorton (approved):
Looks good to me!

PR #866, opened 2021-08-25
Gracefully Handle No GitHub

### LINKED ISSUE #853: Gracefully Degrade without GitHub
Some of our pages make calls to GitHub, and sometimes GitHub goes away for various reasons.  Some of the pages/flows return semi-cryptic errors when this happens.

Specifically a broken PAT means we get 401s, and either throw an exception or in the case of the login flow we tell a User they're not a member of the OpenSAFELY GitHub org.

Lets make these more graceful, even if it's just reporting the correct issue to Sentry.

ghickman:
This adds some handling around various GitHub (API) calls for when GitHub is down so we don't throw 500s or cryptic errors to users and staff.

Fixes #853


rebkwok (approved):
LGTM

PR #879, opened 2021-08-30
Update Tests

ghickman:
This is a grab bag of changes, mostly to bring the tests inline with our testing expectations defined in TESTING.md.

The switch from Client -> RequestFactory saw a runtime drop of ~6s (~9.3s -> ~3.3s) in the release API tests, although I'm still unsure what about the use of Client (technically DRF's Client) was slowing them down.

I'd added a parametrized test to check all the URLs go where we think they do (with redirects in another test), but this doesn't currently check we have all configured URLs covered.  I have some ideas for that but wanted to get them out in another PR.

I've removed the `MEANINGLESS_URL` variable because it seemed rather meaningless (😜) and felt like an indirection when reading back through tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,13 +10,6 @@ from .factories import OrgFactory, OrgMembershipFactory, UserFactory
 
 
 @pytest.fixture
-def api_client():
-    from rest_framework.test import APIClient
-
-    return APIClient()
-
-
-@pytest.fixture
 def api_rf():
     from rest_framework.test import APIRequestFactory
 
--- a/tests/unit/jobserver/api/test_releases.py
+++ b/tests/unit/jobserver/api/test_releases.py
@@ -6,7 +6,13 @@ from rest_framework.exceptions import NotAuthenticated
 from slack_sdk.errors import SlackApiError
 
 from jobserver.api.releases import (
+    ReleaseAPI,
+    ReleaseFileAPI,
     ReleaseNotificationAPICreate,
+    ReleaseWorkspaceAPI,
+    SnapshotAPI,
+    SnapshotCreateAPI,
+    SnapshotPublishAPI,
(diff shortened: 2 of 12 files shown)


inglesp (approved):
All looks sensible and well thought through.

> a runtime drop of ~6s (~9.3s -> ~3.3s) in the release API tests

Cool

> unsure what was slowing them down

Is it worth spending a timeboxed half hour profiling?

It might also be worth spending a little of time looking at some of the common patterns in the tests that use `rf`, and writing some helper functions.

eg we've got lots of things like:

```
    request = rf.get("/")
    request.user = coordinator

    response = SomeProjectCBV.as_view()(
        request, org_slug=org.slug, project_slug=project.slug
    )
```

You could define:

```
def get_project_response(rf, cbv, project, user=None):
    request = rf.get("/")
    if user:
        request.user = user
    response = cbv.as_view()(request, org_slug=project.org.slug, project_slug=project.slug)
```

and then in your test:

```
    response = get_project_response(rf, SomeProjectCBV, project, collaborator)
```

ghickman:
I'm keen to not obscure too much of the setup and calls in tests, it's common to avoid being too DRY in them ([this is my go-to high SO on the subject](https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests)).  In particular constructing a request inside the function which returns a response means we'd be violating the arrange-act-assert principle.

> Is it worth spending a timeboxed half hour profiling?

It wasn't 😞   I tried all the usual tricks, disabling password validators, setting password hasing to MD5, disabling all middleware.  I installed pytest-profiling but it didn't show up anything obvious/usefull.  Spelunking APIClient, APIRequestFactory, Client, and RequestFactory which was interesting-to-me, didn't flag up anything obvious.

PR #880, opened 2021-08-30
Enable ProjectDeveloper role being set on User instances

ghickman:
This lets us set the ProjectDeveloper role "globally" on a User instance, so staff can have this role without having to be members of every Project.  This is more important now that running jobs is tied to the `run_job` permission provided by this role.

### DIFF
--- a/jobserver/authorization/roles.py
+++ b/jobserver/authorization/roles.py
@@ -182,6 +182,7 @@ class ProjectDeveloper:
     models = [
         "jobserver.models.core.ProjectInvitation",
         "jobserver.models.core.ProjectMembership",
+        "jobserver.models.core.User",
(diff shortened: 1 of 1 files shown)


inglesp (approved):
Are there any roles that we shouldn't allow to be set globally?

ghickman:
Not as of yet, and my gut feeling is all roles will be able to set on User (I shouldn't have used the word "globally", it's a vestige of the original OFP that is no longer a good fit!), and likely Group when we add that in (mumble mumble eventually) if only becasuse staff need to be able to administer the platform.  Currently the `models` attribute on roles are used to drive the various UIs where we edit roles by asking for roles tied to a given model.

Issue #641, opened 2021-07-01
Break Down CoreDeveloper Role


ghickman:
It's replaced the SuperUser role and `User.is_superuser` before it as the "root" role for staff.  While we're still in a better position than we used to be, in that we're now querying permissions which the role has, it's still too broad.

PR #881, opened 2021-08-30
Remove SuperUser Role

ghickman:
This removes the SuperUser role now that it's no longer in use by any users.

Follow up to #635

### DIFF
--- a/jobserver/authorization/__init__.py
+++ b/jobserver/authorization/__init__.py
@@ -9,7 +9,6 @@ from .roles import (
     ProjectCollaborator,
     ProjectCoordinator,
     ProjectDeveloper,
-    SuperUser,
     TechnicalReviewer,
 )
 from .utils import has_permission, has_role, roles_for, strings_to_roles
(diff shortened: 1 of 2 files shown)


inglesp (approved):
This PR's obviously fine, shipit.

However... I understand why we might not want superusers, but I'm slightly concerned that we'll end up with "implicit" superusers, where some users end up having almost all the permissions that a superuser would have, without it being obvious who those users are.

Has this been discussed at all?

ghickman:
@inglesp – yes!  It's come up a couple of times as part of permissions but I haven't had a good place to document the decisions which led us here (time to revisit ADRs I think).

Until recently the CoreDeveloper role mostly fulfilled this, being the implicit superuser role you mentioned.  #641 took away the duplicate permissions from that role so it didn't become the next superuser role, but the implicit cohort of superusers (by dint of having all the roles) is still possible.

Groups is the endgame for this, with some sort of staff/platform engineer group which gets ~all roles (there are likely caveats).

PR #882, opened 2021-08-30
Tidier project.yaml Rendering

ghickman:
We render project.yaml contents on JobRequest detail pages, but any reasonably sized file adds a lot of scrolling to the page.

In lieu of the upcoming parser library this makes some aesthetic changes in an attempt to make the page a little more usable.

**Before:**

**After:**


inglesp (approved):
:heart:

PR #891, opened 2021-08-30
Filter Workspace Logs by Backend

### LINKED ISSUE #527: Filtering/separating backends in workspace logs
Currently, all actions from all backends are displayed, with no way to filter for a specific backend.

You can get there via the main log view, and filter on workspace and backend, but it's awkward.

Backends are logically distinct from each other, I'm not sure it make sense to mix their logs in a single stream, and perhaps we should have a UI do select the backend we wish to look at? Or we could add a filter.

Related to #526

ghickman:
This adds a backend filter to the WorkspaceLog page, using GET args as we do with all our other filter pages:

Fixes #527

### DIFF
--- a/jobserver/views/workspaces.py
+++ b/jobserver/views/workspaces.py
@@ -360,7 +360,9 @@ class WorkspaceLog(ListView):
         qs = (
             JobRequest.objects.filter(workspace=self.workspace)
             .prefetch_related("jobs")
-            .select_related("workspace")
+            .select_related(
+                "backend", "workspace", "workspace__project", "workspace__project__org"
+            )
             .order_by("-pk")
         )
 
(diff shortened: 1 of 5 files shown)


jobserver/views/workspaces.py
```diff
@@ -353,14 +353,17 @@ def dispatch(self, request, *args, **kwargs):
 
     def get_context_data(self, **kwargs):
         return super().get_context_data(**kwargs) | {
+            "backends": Backend.objects.all(),
```
inglesp, line 356: Could we filter the backends so we only show those that have associated jobs in the workspace?  For instance if I was running something on TPP and EMIS, I wouldn't want to see Graphnet.

inglesp (approved):
Looks good; one suggested tweak.

jobserver/views/workspaces.py
ghickman: Good point, done and rebased in.

PR #894, opened 2021-08-31
Update and style workspace templates

tomodwyer:
- Update Workspace templates
  -  backend files
  - detail
  - file list
  - latest outputs detail
  - log
  - output list
  - release list
- Update release detail template
- Move alerts above the breadcrumbs
- Add `prose-width` util CSS class

PR #897, opened 2021-08-31
Misc Fixes

ghickman:
Two lots of fixes found while reviewing #894

### DIFF
--- a/jobserver/management/commands/release.py
+++ b/jobserver/management/commands/release.py
@@ -98,7 +98,7 @@ class Command(BaseCommand):
             else:
                 raise
 
-        backend = get_or_maybe_create(create, Backend, {"name": backend_name})
+        backend = get_or_maybe_create(create, Backend, {"slug": backend_name})
         user = get_or_maybe_create(create, User, {"username": username})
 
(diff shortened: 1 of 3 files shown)


jobserver/management/commands/release.py
```diff
@@ -98,7 +98,7 @@ def handle(
             else:
                 raise
 
-        backend = get_or_maybe_create(create, Backend, {"name": backend_name})
+        backend = get_or_maybe_create(create, Backend, {"slug": backend_name})
         user = get_or_maybe_create(create, User, {"username": username})
```
CarolineMorton, line 102: Why is slug better than name?

CarolineMorton (approved):

jobserver/management/commands/release.py
ghickman: In database terms it's the unique field of the two, in UX terms it avoids having to remember the correct capitalisation of a backend's name!

This is all because I changed the names around recently and missed a couple 😬

CarolineMorton: makes sense

PR #910, opened 2021-09-06
Syntax Highlight project.yaml on JobRequest Detail Pages

ghickman:
This adds syntax highlighting to the project.yaml rendering on JobRequest detail pages, while also retaining the linking of scripts.  This saves us having to load a syntax highlighter in the front end bundle.  However the way we link scripts had to change since we couldn't pass the linked scripts version of a project.yaml into pygments, it returned an HTML version with the links rendered as escaped HTML elements.

Instead we now find the contents of `run:` lines and use them as a key for new version of the same line with links in, eg

```
{
    "a_docker:latest research.py input input input": "a_docker:latest <a href="github.com/serious/research/blob/research.py">research.py</a> input input input
}
```

This allows us to essentially do find and replace on the HTML pygments generates.  We only capture the target of a `run:` line since the `run:` portion of the line and its target are split up into two different spans by pygments.


iaindillingham (approved):

jobserver/project.py
```diff
@@ -77,25 +80,67 @@ def load_yaml(content):
...
+    links_map = (
+        (line, " ".join(link_run_scripts(line, link_func))) for line in run_lines
+    )
 
-    # replace newlines with <br /> elements so the normal newlines aren't
-    # collapsed when the browser renders them.
-    definition = "\n".join(lines)
+    # compact the list to remove Nones, from lines with no appropriate script
+    # as the run target (eg a generate_cohort action)
+    links_map = (m for m in links_map if m is not None)
```
evansd, line 120: I think this line is redundant because `m` will always be a tuple now, won't it? And if a line doesn't have any scripts in it it will just get mapped to itself and so doesn't need special handling.

evansd (commented):
I just took a quick look at this as the output looked so pretty and I wanted to see how you'd done it!  Looks great.

In the process I spotted what looks to me like a redundant, but harmless line.

jobserver/project.py
ghickman: Oh yes, good catch!

PR #919, opened 2021-09-07
Notify Sentry About Jobs With an Internal Error

ghickman:
This reports jobs with the status message "Internal error" to Sentry as message events (instead of exceptions) so we can start getting notifications about them.

### DIFF
--- a/jobserver/api/jobs.py
+++ b/jobserver/api/jobs.py
@@ -119,6 +119,16 @@ class JobAPIUpdate(APIView):
                 else:
                     updated_job_ids.append(str(job.id))
 
+                if "internal error" in job.status_message.lower():
+                    # bubble internal errors encountered with a job up to
+                    # sentry so we can get notifications they've happened
+                    with sentry_sdk.push_scope() as scope:
+                        scope.set_tag("backend", job_request.backend.slug)
+                        scope.set_tag(
+                            "job", request.build_absolute_uri(job.get_absolute_url())
+                        )
+                        sentry_sdk.capture_message("Job encountered an internal error")
+
                 if created:
                     # this is the first time job-server has heard about this
                     # Job so we can move on to further Jobs.  We're knowingly
--- a/tests/unit/jobserver/api/test_jobs.py
+++ b/tests/unit/jobserver/api/test_jobs.py
@@ -443,6 +443,39 @@ def test_jobapiupdate_post_only(api_rf):
 
 
 @pytest.mark.django_db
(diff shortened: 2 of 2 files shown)


evansd (approved):
Perfect! Thanks for this.

We can do something more sophisticated than `if "internal error"` once we make better use of the "status_codes" system but this will do the trick for now.

jobserver/api/jobs.py
```diff
@@ -119,6 +119,16 @@ def post(self, request, *args, **kwargs):
                 else:
                     updated_job_ids.append(str(job.id))
 
+                if "internal error" in job.status_message.lower():
```
benbc, line 122: Am I right that this string is generated in job-runner [here](https://github.com/opensafely-core/job-runner/search?q=%22internal+error%22+filename%3Arun.py)?

This seems like a good immediate solution to a current problem, but I guess that ideally we'd want something a bit more structured so we're not depending on string matching?

tests/unit/jobserver/api/test_jobs.py
```diff
@@ -442,6 +442,37 @@ def test_jobapiupdate_post_only(api_rf):
...
+            "updated_at": now,
+            "completed_at": None,
+        },
+    ]
+    request = api_rf.post(
+        "/", HTTP_AUTHORIZATION=backend.auth_token, data=data, format="json"
+    )
+    response = JobAPIUpdate.as_view()(request)
+
+    assert response.status_code == 200, response.data
```
benbc, line 473: Is it possible to check that that we send the error to Sentry?

jobserver/api/jobs.py
benbc: I've just seen Dave's comment. :+1:

evansd: @benbc We already have the basic building blocks to do this, but just aren't actually using them. Jobs have a `status_code` field whose values are defined [here](https://github.com/opensafely-core/job-runner/blob/4ef28a9c2e64d523439536c21663447c3f28bd86/jobrunner/models.py#L26-L36) and which is pushed up to job-server along with everything else. We need to add an `internal_error` code and then set that in the appropriate place.

benbc: Great.

tests/unit/jobserver/api/test_jobs.py
ghickman: Good catch, I've added an assert to check we're calling `capture_message()`.  It doesn't look like sentry ships any of their testing tools with the SDK for checking you've called it correctly unfortunately, would be nicer than mocking.

PR #928, opened 2021-09-08
Add an application page which renders contents of snippets/application.md

inglesp:
image

All of this is up for discussion, including:

* is the approach sensible?
* are things named properly?
* is the code in the right place?

### DIFF
--- a/applications/templates/application.html
+++ b/applications/templates/application.html
@@ -1,0 +1,10 @@
+{% extends "base.html" %}
+
+{% load snippet %}
+
+{% block content %}
+<div class="container">
+  <h1>Application form</h1>
+  {% snippet "application" %}
+</div>
+{% endblock %}
--- a/applications/views.py
+++ b/applications/views.py
@@ -1,0 +1,5 @@
+from django.shortcuts import render
+
+
+def application(request):
+    return render(request, "application.html")
--- a/jobserver/urls.py
+++ b/jobserver/urls.py
@@ -255,6 +255,7 @@ urlpatterns = [
     path("admin/approve-users", ApproveUsers.as_view(), name="approve-users"),
     path("admin/", admin.site.urls),
     path("api/v2/", include((api_urls, "api"))),
+    path("applications/", include("applications.urls")),
     path("event-log/", JobRequestList.as_view(), name="job-list"),
     path("event-list/", RedirectView.as_view(url="/event-log/")),
     path("jobs/", RedirectView.as_view(query_string=True, pattern_name="job-list")),
--- a/tests/unit/applications/test_views.py
+++ b/tests/unit/applications/test_views.py
@@ -1,0 +1,13 @@
+import pytest
+
+from applications.views import application
+
+from ...factories import UserFactory
+
+
+@pytest.mark.django_db
+def test_get_application(rf):
+    request = rf.get("/")
+    request.user = UserFactory()
+    response = application(request)
+    assert response.status_code == 200
--- a/applications/urls.py
+++ b/applications/urls.py
@@ -1,0 +1,10 @@
+from django.urls import path
+
+from . import views
+
+
+app_name = "applications"
+
+urlpatterns = [
+    path("", views.application, name="application"),
+]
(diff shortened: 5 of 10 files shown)


sebbacon:
I appreciate this PR is really about a technique for including any markdown, but in passing, this reminds me of [a Slack thread](https://ebmdatalab.slack.com/archives/C02CNEF5RU1/p1631020723063900) from yesterday, where I realised there are two flavours of application: an application for a project to be approved, whose inputs are a description of the project, and whose outputs are a set of artefacts describing contractual relationships, legal basis, etc; and an application for a user to have access to a given project (and thereby secure backed environment(s)), whose inputs are the project to which they want access and some kinds of declarations, and whose output is some kind of certification that they are allowed (e.g. NHS honorary contract).

My question: what kind of _application_ is this? Should the URL reflect that?

inglesp:
> My question: what kind of application is this? Should the URL reflect that?

We've got some wireframing to do first!

jobserver/urls.py
```diff
@@ -255,6 +255,7 @@
     path("admin/approve-users", ApproveUsers.as_view(), name="approve-users"),
     path("admin/", admin.site.urls),
     path("api/v2/", include((api_urls, "api"))),
+    path("applications/", include("applications.urls")),
```
ghickman, line 258: If you copy the namespace bit of the`staff` include, we can reference urls with the namespace, eg `"applications:some_name"`, while only doing `name="some_name"` in the urls.py

applications/views.py
```diff
@@ -0,0 +1,5 @@
+from django.shortcuts import render
+
+
+def application(request):
+    return render(request, "application.html")
```
ghickman, line 5: I know this is just a spike, but something like this could be `TemplateView.as_view(template_name="application.html")` directly in the urls.py (really useful for spikes I find!)

applications/templates/application.html
```diff
@@ -0,0 +1,9 @@
+{% extends "base.html" %}
+{% load snippet %}
```
ghickman, line 2: nit: I like putting a blank line before `load`s so they're a group separate from the `extends`.

tests/unit/applications/test_views.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+
+@pytest.mark.django_db
+def test_get_application(client):
+    rsp = client.get("/applications/")
```
ghickman, line 6: RequestFactories for unit tests please!

ghickman (approved):
Looks great!

I was worried about "an app called applications", but I don't think it's as much of an issue seeing it in practice.

I can't comment on them but can you remove the models.py and migrations directory until they're needed.

Should we put snippets inside their relevant application to mirror how templates work?

applications/views.py
inglesp: Yeah but then I couldn't use a request factory!

jobserver/urls.py
inglesp: How's this different to having `app_name = "applications"` in `applications/urls.py`?

tests/unit/applications/test_views.py
inglesp: Done.

jobserver/urls.py
ghickman: I thought you had to have both… do you not??  Useful to see the name explicitly used as a namespace though I think?

ghickman:
@inglesp – merging so we can build on top of the app you created, we can address the couple of notes (if they need it!) later 🙂

jobserver/urls.py
inglesp: This is a complicated corner of Django that's new to me.

Pretty sure you don't have to have both -- [from the docs](https://docs.djangoproject.com/en/dev/topics/http/urls/#reversing-namespaced-urls) I think that the "namespace=" argument to `path()` (the "instance namespace") is only used if you have multiple instances of the app (at different URLs).

I can see the appeal of having the namespace in two places, but I'd rather not, since you might end up with a confusing situation if you ever change one of them.

ghickman: I didn't realise you could just use the app_name!  I'd have prefered the namespace to be explicit in the main urls.py but I agree on not having it in two places 👍

PR #935, opened 2021-09-09
Form Field Component

ghickman:
Initial working version of the form field "components".

Everything worked out well except for help_text, which now need to be defined in Django forms, not ideal but it's the one big con I've found here.


inglesp (approved):
All fine, no dragons.

Will be interesting to see what happens as these partials get used more widely.

PR #942, opened 2021-09-10
Notify Slack when a New User Registers

ghickman:
Some duct tape to tell us when a user first logs into the job-server while we build a better onboarding story.

### DIFF
--- a/jobserver/pipeline.py
+++ b/jobserver/pipeline.py
@@ -1,3 +1,22 @@
+from django.conf import settings
+from furl import furl
+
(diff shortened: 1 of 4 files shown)


jobserver/pipeline.py
```diff
@@ -1,3 +1,22 @@
+from django.conf import settings
+from furl import furl
```
iaindillingham, line 2: Just reading about `furl`. Nice!

iaindillingham (approved):

PR #946, opened 2021-09-10
Misc Staff Area Fixes and Changes

ghickman:
This is a grab bag of changes to the staff area:
* Fix the project detail page to show its workspace list
* Make projects editable
* Add orgs (detail, list, edit) to the staff area
* Update some links to point to pages in the staff area
* Allow members to be added to/removed from orgs

### DIFF
--- a/staff/views/orgs.py
+++ b/staff/views/orgs.py
@@ -1,21 +1,56 @@
+from django.contrib import messages
+from django.db.models.functions import Lower
+from django.shortcuts import get_object_or_404, redirect
 from django.utils.decorators import method_decorator
-from django.views.generic import DetailView, ListView, UpdateView
+from django.views.generic import FormView, ListView, UpdateView, View
 
 from jobserver.authorization import CoreDeveloper
 from jobserver.authorization.decorators import require_role
-from jobserver.models import Org
+from jobserver.models import Org, OrgMembership, User
+
+from ..forms import OrgAddMemberForm
 
 
 @method_decorator(require_role(CoreDeveloper), name="dispatch")
-class OrgDetail(DetailView):
-    model = Org
+class OrgDetail(FormView):
+    form_class = OrgAddMemberForm
     template_name = "staff/org_detail.html"
+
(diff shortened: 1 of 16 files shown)


CarolineMorton (approved):
Looks good to me. 

Thanks for clarifying re: `method_decorator` on slack. 

Nice to see such a comprehensive set of tests!!

ghickman:
> can a user be a member of more than one organisation

Yes

> This might be something that a co-pilot might want

If they have access to the staff area, they have acess to every page in it!

tests/unit/staff/views/test_orgs.py
```diff
@@ -0,0 +1,118 @@
...
+def test_orglist_success(rf, core_developer):
+    OrgFactory.create_batch(5)
+
+    request = rf.get("/")
+    request.user = core_developer
+
+    response = OrgList.as_view()(request)
+
+    assert response.status_code == 200
+    assert len(response.context_data["org_list"])
```
inglesp, line 142: Missing `== 5`?

staff/views/orgs.py
```diff
@@ -1,22 +1,57 @@
...
     template_name = "staff/org_detail.html"
 
+    def dispatch(self, request, *args, **kwargs):
+        self.object = get_object_or_404(Org, slug=self.kwargs["slug"])
+
+        return super().dispatch(request, *args, **kwargs)
+
+    def form_valid(self, form):
+        for user in form.cleaned_data["users"]:
+            self.object.memberships.create(user_id=user, created_by=self.request.user)
```
inglesp, line 26: Could we end up with multiple memberships for a single user here?

tests/unit/staff/views/test_orgs.py
ghickman: oops, yes!

staff/views/orgs.py
ghickman: No, they're unique on Org & User so we should get an error then, but I'll also trim the available orgs down so we only see ones without a membership 👍

PR #949, opened 2021-09-13
Display All User Roles

ghickman:
This displays the roles for Projects and Orgs against the relevant Projects/Orgs on the user page of the staff admin:

Note: we don't use roles on OrgMemberships currently, but this will display them when we do.

### DIFF
--- a/staff/views/users.py
+++ b/staff/views/users.py
@@ -39,9 +39,25 @@ class UserDetail(UpdateView):
         return redirect(self.object.get_staff_url())
 
     def get_context_data(self, **kwargs):
+        orgs = [
+            {
+                "name": m.org.name,
+                "roles": [r.display_name for r in m.roles],
+                "staff_url": m.org.get_staff_url(),
+            }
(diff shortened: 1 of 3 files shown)


CarolineMorton (approved):
Looks great

staff/views/users.py
```diff
@@ -39,9 +39,25 @@ def form_valid(self, form):
...
+                "name": m.org.name,
+                "roles": [r.display_name for r in m.roles],
+                "staff_url": m.org.get_staff_url(),
+            }
+            for m in self.object.org_memberships.order_by("org__name")
+        ]
+        projects = [
+            {
+                "name": m.project.name,
+                "roles": [r.display_name for r in m.roles],
```
inglesp, line 53: Is this sorted?

ghickman: It's not, will fix this and the orgs one 👍

ghickman: Fixed in #954

PR #950, opened 2021-09-13
Disable the Django Admin

ghickman:
We're going to build tooling for the new onboarding flow with bespoke views so the admin is no longer necessary.


evansd (approved):
I mean what's not to like?
github com_opensafely-core_job-server_pull_950_files

PR #954, opened 2021-09-14
Sort the roles for users org&projects lists

ghickman:
This sorts the role labels for Orgs & Projects listed on a User's detail page in the Staff Area.

### DIFF
--- a/staff/views/users.py
+++ b/staff/views/users.py
@@ -42,7 +42,7 @@ class UserDetail(UpdateView):
         orgs = [
             {
                 "name": m.org.name,
-                "roles": [r.display_name for r in m.roles],
+                "roles": sorted(r.display_name for r in m.roles),
                 "staff_url": m.org.get_staff_url(),
             }
             for m in self.object.org_memberships.order_by("org__name")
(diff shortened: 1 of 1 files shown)


staff/views/users.py
```diff
@@ -42,15 +42,15 @@ def get_context_data(self, **kwargs):
         orgs = [
             {
                 "name": m.org.name,
-                "roles": [r.display_name for r in m.roles],
+                "roles": sorted(r.display_name for r in m.roles),
```
StevenMaude, line 45: A note for myself, because I'm still to some extent trying to remember Python that I used to know.

I wasn't quite sure how this change worked in the sense:

```
>>> sorted(x for x in range(3))
[0, 1, 2]
>>> x for x in range(3)
  File "<stdin>", line 1
    x for x in range(3)
      ^
SyntaxError: invalid syntax
```

Why don't I need parentheses inside the call, but do outside of a call?

> 6.2.8. Generator expressions
> …
> The parentheses can be omitted on calls with only one argument.

https://docs.python.org/3/reference/expressions.html#generator-expressions

StevenMaude (approved):

PR #955, opened 2021-09-14
Staff Area Orgs Fixes

ghickman:
Misc fixes from feedback on #946

### DIFF
--- a/staff/views/users.py
+++ b/staff/views/users.py
@@ -139,8 +139,12 @@ class UserSetOrgs(FormView):
         return redirect(self.user.get_staff_url())
 
     def get_form_kwargs(self):
+        existing_membership_orgs = self.user.orgs.values_list("pk", flat=True)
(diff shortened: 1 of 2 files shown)


StevenMaude (approved):

staff/views/users.py
```diff
@@ -139,8 +139,12 @@ def form_valid(self, form):
         return redirect(self.user.get_staff_url())
 
     def get_form_kwargs(self):
+        existing_membership_orgs = self.user.orgs.values_list("pk", flat=True)
+        available_orgs = Org.objects.exclude(pk__in=existing_membership_orgs).order_by(
+            "name"
+        )
         return super().get_form_kwargs() | {
-            "available_orgs": Org.objects.order_by("name"),
+            "available_orgs": available_orgs,
```
inglesp, line 147: If we're only showing the organisations that a user is not a member of, how can we remove a user from an organisation?

PR #973, opened 2021-09-20
Use a function for Org.github_orgs for migrations

ghickman:
Django Migrations can't serialize lambdas so we have to use a function
as per:

https://docs.djangoproject.com/en/3.2/ref/models/fields/#default


iaindillingham (approved):
How annoying!

PR #975, opened 2021-09-20
Add Ownership Fields to Org&Project

ghickman:
This adds nullable `created_by` fields to both Org & Project so we can track (at least internally) who created both of those objects.


jobserver/models/core.py
```diff
@@ -328,6 +328,13 @@ def status(self):
 class Org(models.Model):
     """An Organisation using the platform"""
 
+    created_by = models.ForeignKey(
+        "User",
+        null=True,
+        on_delete=models.SET_NULL,
```
iaindillingham, line 334: From memory, this means that if the instance of `User` is deleted, then this value is set to null. Under what circumstances do we delete users?

iaindillingham (approved):
I've added a comment about deleting users, but it's just a comment.

jobserver/models/core.py
ghickman: That's correct.  We're unlikely to delete users I think, but as of a recent version (3 maybe?) Django requires `on_delete` to be specified and this seemed like the logical choice for this field.

iaindillingham: I'm in two minds. On the one hand, this isn't really the best place for enforcing a policy ("Users must not be deleted"). On the other hand, if we want to track which users created which organizations and projects over any length of time, then we should make it hard to delete users. Might this be worth a broader discussion? We could, for example, consider `models.PROTECT`.

ghickman: We don't currently have this data for any of the existing Projects or Orgs so I'm also mirroring the existing situation.

But you make a very valid point, we could definitely bring it up in the weekly meeting.

Note one currently has to go into the database to delete users (where this constraint isn't enforced) so it's rather moot right now!

iaindillingham: Very reasonable: we don't want to arbitrarily assign a user to existing projects or organizations, so the DB is modelling current use 🙂 If we have time, then it would be good to discuss in the weekly meeting. I'll add it to the agenda.

PR #985, opened 2021-09-21
Add Application Wizard

ghickman:
This is our initial wizard for applying to use the OpenSAFELY platform.  It contains some entry pages (under `/apply/`), a wizard driven by the high level `form_specs` structure allowing us to add cross-page prerequisites and further display details without having to misuse [Model]Forms.


inglesp (approved):
rubber stamp

Issue #795, opened 2021-08-09
Investigate More Context In Permission Checks


ghickman:
We currently have some authorization checks which look at more than the current permission of a User, a good example of this is the [`validate_snapshot_access` function](https://github.com/opensafely-core/job-server/blob/e0522251a37c2f4e41d41547854267f9bab4895b/jobserver/api/releases.py#L108-L123) which also checks that a Snapshot has been published as the first step because anyone can view a published snapshot so the permissions check further down isn't required.

This also came up in [this](https://github.com/opensafely-core/job-server/pull/745#discussion_r682422622) discussion around the concept of ownership of an object being something we could check as an internal to the `has_permission` function.

There are sort of two parts to this ticket:
* Look for types of authorization check we do around the code base (we already know about ownership)
* Investigate how the roles & permissions framework could handle those

ghickman:
Some stream-of-consciousness thoughts on this topic which is already fairly abstract:

Pushing non-permission-string checks into `has_permission` could be confusing because then it begs the question "what does the framework cover?", eg say we add ownership checks, how is a dev supposed to know it covers that and not the published state of a Snapshot?

Secondly I worry that pushing all those checks into the framework will hide useful information from important check sites, Snapshots being published is a good example here.  Uploading ReleaseFiles is another one, it's a slightly complex process because of the necessary security architecture of backends, and [`validate_upload_access`](https://github.com/opensafely-core/job-server/blob/e0522251a37c2f4e41d41547854267f9bab4895b/jobserver/api/releases.py#L68-L92) is a well commented block of code which explains why each step for authorization is required.

PR #993, opened 2021-09-21
Require Auth for Applications

ghickman:
This adds a requirement that users are logged in, and either created the application or have the `application_manage` permission (the idea being that staff will use the latter).

### DIFF
--- a/applications/views.py
+++ b/applications/views.py
@@ -1,3 +1,4 @@
+from django.contrib.auth.decorators import login_required
 from django.shortcuts import get_object_or_404, redirect
 from django.template.response import TemplateResponse
 
@@ -6,6 +7,7 @@ from .models import Application
 from .wizard import Wizard
 
 
+@login_required
 def page(request, pk, page_num):
     application = get_object_or_404(Application, pk=pk)
     page = Wizard(application, form_specs).get_page(page_num)
@@ -45,6 +47,7 @@ def sign_in(request):
     return redirect("applications:terms")
 
 
+@login_required
 def terms(request):
     if request.method == "GET":
         return TemplateResponse(request, "applications/terms.html")
@@ -53,6 +56,7 @@ def terms(request):
     return redirect("applications:page", pk=application.pk, page_num=1)
 
 
+@login_required
 def confirmation(request, pk):
     application = get_object_or_404(Application, pk=pk)
     wizard = Wizard(application, form_specs)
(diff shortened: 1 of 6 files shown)


applications/views.py
```diff
@@ -1,15 +1,31 @@
...
+        return
+
+    raise Http404
+
+
 @login_required
 def page(request, pk, page_num):
     application = get_object_or_404(Application, pk=pk)
+
+    validate_application_access(request.user, application)
```
inglesp, line 27: Should we add this to the confirmation view too?

applications/views.py
```diff
@@ -1,15 +1,31 @@
...
 
 
+def validate_application_access(user, application):
+    if has_permission(user, "application_manage"):
+        return
+
+    if application.created_by == user:
+        return
+
+    raise Http404
```
inglesp, line 20: This might be the right solution for now, but looking at `application.created_by` is a new way of managing per-object permissions, and is different to the existing `RolesField` stuff.

inglesp (approved):

applications/views.py
```diff
@@ -1,13 +1,31 @@
...
```
iaindillingham, line 20: Why not 401?

iaindillingham (approved):
Just a question about status codes 🙂 I guess the answer is something like "consistency" or "ease of implementation" but I'm curious.

applications/views.py
ghickman: There's precedent with how we decide who can cancel a Job too, it spawned the conversation in #795

applications/views.py
ghickman: 401 (or even 403) would be more technically correct here, but it's common practice these days to hide pages someone shouldn't be able to see, eg a private GitHub repo you don't have permissions to view.

applications/views.py
ghickman: Yes, oversight on my part 👍

applications/views.py
iaindillingham: Ah, okay. That's a pity but I appreciate that GitHub, for example, 404 when really they mean unauthorized/forbidden.

sebbacon: The justification for Github doing a 404 is not to leak potentially private information; in this case, all applications are expected to be public in any case (at least eventually) so I'd have argued for a 401... but  `¯\_(ツ)_/¯`

ghickman: I wasn't aware applications would be public when I added this!  But yes 401 will likely be more appropriate here once we define the logic for who can view applications and when 👍

PR #998, opened 2021-09-22
Remove base form etc

inglesp:
This paves the way for allowing further per-field customisation driven by the form specs.


applications/views.py
```diff
@@ -30,11 +30,10 @@ def page(request, pk, page_num):
...
     if request.method == "GET":
-        ctx = {
-            "application": application,
-            "form": page.get_unbound_form(),
-            "page": page,
-        }
+        form = page.get_unbound_form()
+        ctx = page.form_spec.template_context(form)
+        ctx["application"] = application
+        ctx["page"] = page
```
ghickman: Now that we can splat dictionaries, what do you think about this?
```
ctx = {
    **page.form_spec.template_context(form),
    "application": application,
    "page": page,
}
```

tests/unit/jobserver/templatetags/test_snippet.py
```diff
@@ -1,14 +1,11 @@
-import os
+from pathlib import Path
 
-from django.conf import settings
+from django.test import override_settings
```
ghickman, line 3: TIR!

ghickman (approved):

applications/views.py
inglesp: I'm using `|` later on, but otherwise, yes!

PR #1006, opened 2021-09-24
Add Researchers Page

ghickman:
This adds the researchers page to our onboarding wizard with add&edit pages for each researcher a applicant wishes to add, and a delete button to remove them from an application.

Researchers are listed on the wizard page using a template override for the fieldset:


tomodwyer (approved):
Frontend does what I expect 👍

applications/views.py
```diff
@@ -20,6 +24,77 @@ def validate_application_access(user, application):
...
+        validate_application_access(request.user, self.application)
+
+        return super().dispatch(request, *args, **kwargs)
+
+    def form_valid(self, form):
+        researcher = form.save(commit=False)
+        researcher.application = self.application
+        researcher.save()
+
+        # TODO: how should we configure this so it's not hardcoded?
```
inglesp: Change `page_num` to `key`, and then use descriptive strings for the keys in the form spec?  A small amount of work would be required (probably in `wizard.py`) to get the next page.

tests/unit/applications/test_models.py
```diff
@@ -1,7 +1,31 @@
...
+
+def test_researcherregistration_get_edit_url():
+    researcher = ResearcherRegistrationFactory()
+
+    url = researcher.get_edit_url()
+
+    assert url == reverse(
+        "applications:researcher-edit",
+        kwargs={"pk": researcher.application.pk, "researcher_pk": researcher.pk},
+    )
```
inglesp, line 25: I'm not sure what the point of this kind of test is -- the test is exactly duplicating the code.  If we remove the tests and the coverage drops, that indicates either that these methods are never used, or that the code that uses the methods is not tested.

tests/unit/applications/test_views.py
```diff
@@ -14,28 +18,161 @@
...
-    user = UserFactory(roles=[CoreDeveloper])
-    assert validate_application_access(user, application) is None
+def test_researcherdelete_unknown_researcher(rf):
+    request = rf.post("/")
 
+    with pytest.raises(Http404):
+        ResearcherDelete.as_view()(request, pk=0, researcher_pk=0)
 
-def test_validate_application_access_with_owner():
+    # ResearcherEdit,
```
inglesp, line 145: stray comment?

inglesp (commented):
A few post-merge observations, but overall, looks good!

PR #1007, opened 2021-09-24
Allow editing form from confirmation page

StevenMaude:
This PR implements the "Change" buttons on the confirmation page.

1. Click the button for the relevant form entry.
2. You are then sent to the appropriate form page.
3. You can make a change to that page (or simply make no changes).
4. You can click the button the page to return back to the confirmation
    page, with any changes saved.

There is a current limitation: some form pages result in diverging paths
through the form (e.g. "type of research"). If you respond differently to your
original choice, you are still sent back to the confirmation page regardless,
instead of proceeding through the relevant section of the form.


tests/unit/applications/test_views.py
```diff
@@ -54,6 +54,47 @@ def test_confirmation_success(rf):
...
+    request = rf.get("/")
+    request.user = user
+
+    response = confirmation(request, pk=application.pk)
+    assert response.status_code == 200
+
+    application.refresh_from_db()
+    assert application.has_reached_confirmation
+
+    request = confirmation(request, pk=application.pk)
```
ghickman, line 72: did you mean to leave this line in?

ghickman (approved):
One note about an unexpected line, otherwise looks great 👍

tests/unit/applications/test_views.py
StevenMaude: Good catch; not sure how that one got in there :keyboard::oil_drum:

StevenMaude: Fixed now :+1:

StevenMaude: Oh, I remember this: I need to do some kind of check where:
* the confirmation page is loaded and the confirmation page is already reached.
* because otherwise the coverage drops, as there's a code branch not reached.

I'll tidy this up as it was an incomplete thought, but does run all of the code.

StevenMaude: Actually fixed this time.

PR #1012, opened 2021-09-27
Wizard Template Fixes

ghickman:
For each page in both the form wizard, and the pages beforehand (apply, sign in, terms) this adds:
* a meta title relevant to the page
* breadcrumbs
* a jumbotron with the title and sub title of the page

### DIFF
--- a/applications/templates/applications/sign_in.html
+++ b/applications/templates/applications/sign_in.html
@@ -1,5 +1,7 @@
 {% extends "base.html" %}
 
+{% block metatitle %}Apply: Sign In | OpenSAFELY Jobs{% endblock metatitle %}
+
 {% block jumbotron %}
 <div class="jumbotron jumbotron-fluid pt-md-2">
   <div class="container">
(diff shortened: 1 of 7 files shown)


applications/templates/applications/sign_in.html
```diff
@@ -1,12 +1,50 @@
...
+    </div>
+  </div>
+</div>
+{% endblock jumbotron %}
 
 {% block content %}
 <div class="container">
   <div class="row">
     <div class="col">
-      {% snippet "sign_in" %}
```
StevenMaude, line 9: What's the motivation behind moving the text here?

StevenMaude (approved):

applications/templates/applications/sign_in.html
ghickman: There was so little text left in the snippet that it didn't seem worth the downsides of breaking locality of behaviour in favour of a sentence or two.  I have a feeling this page is unlikely to change too much as well so really just making a call!

PR #1019, opened 2021-09-28
Override Field Type for Nullable Booleans

ghickman:
This overrides the field class for our radio button form fields (nullable boolean db fields) to use our YesNoField class which is configured to work with choices which use booleans.  We have to use a base form class for the factory function since we can't use the `field_classes` to override a field class for a field with choices.


inglesp (approved):
:(

PR #1021, opened 2021-09-28
Gitignore Updates

ghickman:
A collection of gitignore house keeping spurred by the switch to enabling WAL mode in SQLite.


.gitignore
```diff
@@ -1,18 +1,12 @@
-*~
-.#*
-*#
-model.log
-*/input.csv
-__pycache__
-.python-version
-*sqlite3
+*sqlite3*
```
inglesp, line 1: This looks a little bit too broad.  (I've got a file a project called `sqlite_wrapper.py`, which would almost get caught by this.)  Could we put the database files in a directory, and add that directory to .gitignore?

.gitignore
```diff
@@ -1,18 +1,12 @@
...
 .env
+.envrc
 .idea/
-node_modules/
 /assets/dist
 /assets/stats.html
 /coverage
+htmlcov
+node_modules/
+releases
```
inglesp, line 11: This'll confuse us if we extract the releases into their own app.

inglesp (approved):
Probably fine, but a couple of paranoid comments.

.gitignore
ghickman: Future us problems 🙂

.gitignore
ghickman: Good point 👍

PR #1023, opened 2021-09-28
Configure Sentry Releases

### LINKED ISSUE #1022: Configure Release Tracking in Sentry
We originally tried to set up release tracking as part of the dokku deploy but the plugin seemed to be missing a variable so only seems to run sporadically.  Since we're now deploying via GitHub Actions and Sentry have an [action for generating releases](https://github.com/marketplace/actions/sentry-release), lets use that as a new step in the `deploy` job.

ghickman:
This sets up our CI to create releases on Sentry using the action they maintain.

Fixes #1022


evansd (approved):
Nice!

PR #1031, opened 2021-09-28
Apply login_required to ProjectCreate

ghickman:
ProjectCreate checks request.user is a member of the project and 404s if
not, but login_required will get a user to login and return them to the
view so this is a slightly nicer UX for users who should have access but
are just logged out but doesn't change anything for non-members.

### DIFF
--- a/jobserver/views/projects.py
+++ b/jobserver/views/projects.py
@@ -72,6 +72,7 @@ class ProjectCancelInvite(View):
         return redirect(invite.project.get_settings_url())
 
 
+@method_decorator(login_required, name="dispatch")
 class ProjectCreate(CreateView):
     fields = ["name"]
     model = Project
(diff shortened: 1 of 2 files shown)


inglesp (approved):
The system "fails open", by which I mean if we get it wrong then we leave a sensitive endpoint open to access by anybody.

How could we "fail closed", to ensure this kind of thing doesn't happen again?

ghickman:
This didn't fail open previously?  The user still had to be a member of an Org and an AnonymousUser instance couldn't have been one.  This is a UX fix where a logged out user now gets redirected to the login page instead of seeing a 404.

> How could we "fail closed"

I suspect we'd need to separate the views for researchers and staff from those used by the public so we can use middleware to apply login_required.

PR #1036, opened 2021-09-29
Add a robots.txt file

tomodwyer:
To stop crawlers from trying (and failing) to access certain areas of the site.

PR #1038, opened 2021-09-29
Add `rel="nofollow"` to login links

evansd:
This is sort of redundant following #1036 which disallows these URLs in
`robots.txt`. But this might have a more immediate effect than the
`robots.txt` changes and so seems worth trying.


ghickman (changes requested):
There's a second login URL in the applications pages in the `applications/templates/applications/sign_in.html` template, could you do that one too?

evansd:
@ghickman Good spot, have amended

ghickman (approved):

PR #1040, opened 2021-09-29
Make Search Results Distinct

ghickman:


inglesp (approved):
All good.  But why is `.distinct()` required?  Perhaps a test would help.

ghickman:
> But why is .distinct() required?

Because the results weren't distinct 😜

We can get duplicate User results because we're looking up users using some related object fields (`project__name` and `org__name`) so if any Projects or Orgs have names which match the search term then the User is included in the results, even if we've already included them based on another field.

PR #1051, opened 2021-09-29
Drop page numbers

inglesp:


ghickman (approved):
Not sure I agree the tests are _more_ brittle with the move to named keys, but they're certainly easier to understand now 👍

PR #1053, opened 2021-09-30
Split Application model up

inglesp:
This does not introduce any changes to behaviour, but paves the way for us to add a status field to each page.

Merging now will reduce risks of a rebase being required later on.


applications/wizard.py
```diff
@@ -48,16 +49,29 @@ def form_class(self):
...
+    def instance(self):
+        """
+        Return instance of model containing data for this page.
+
+        If this page has already been saved, there will be an instance in the
+        database and we return that.  Otherwise, we return a new unsaved instance.
+        """
+        try:
+            return self.model.objects.get(application=self.application)
+        except ObjectDoesNotExist:
```
ghickman, line 67: You could always do `self.model.DoesNotExist` too.

tests/unit/applications/test_views.py
```diff
@@ -348,7 +348,14 @@ def test_page_post_with_invalid_continue_state(rf):
 
 def test_page_post_with_invalid_prerequisite(rf):
     user = UserFactory()
-    application = ApplicationFactory(created_by=user, is_study_research=False)
+    # application = ApplicationFactory(created_by=user, is_study_research=False)
```
ghickman, line 351: Missed comment?

ghickman (approved):
Looks great, couple of comments.  Could you name the migration too so it's easy to see what it's doing at a glance, something like "split_up_application_into_per_page_models".

PR #1064, opened 2021-10-01
Make post requests from a new login page

### LINKED ISSUE #1037: Avoid generating database writes on login view
Currently a bot sees a link in the header to `/login/github/?next=/` and clicks that link which triggers the auth flow. 

This flow creates a session in the database, which is bad.

Change the button so that it uses a `POST`. I'm not sure if this will Just Work, or if we'll have to intercept the click and turn it into a `GET`.

If it's easy, for bonus points, we've agreed a nice UX would be to move the login button to its own page where the button says "Sign in with Github" (perhaps also a "Register with Github" which actually goes to the same place, just to make it less confusing)

iaindillingham:
Previously, get requests were made to GH from each page. Now, post requests are made to GH from a new login page. Post requests don't create new sessions in the database; the new login page is simply a nice-to-have.

Thanks for the discussion about this PR, both here and on Slack ([open redirects](https://ebmdatalab.slack.com/archives/C63UXGB8E/p1633341732201500); [the applications flow](https://ebmdatalab.slack.com/archives/C63UXGB8E/p1633342236207800)).

I'm going to squash and merge, if this PR is approved, so the commit messages are terse. However, I have:

* ensured post requests are made to GH by switching from links to buttons and by decorating `social_django_views.auth` with `csrf_exempt` and `require_POST`
* created a new `Login` view that checks whether the value of the `next` URL attribute is safe; this guards against an [open redirect](https://cwe.mitre.org/data/definitions/601.html)
* persisted the current URL, which is perhaps better described as *the last non-login URL*, through the `next` URL attribute
* updated the `LOGIN_URL` setting

I haven't:

* tidied up our nomenclature: we "Login with GitHub" from the normal flow, but "Sign in to GitHub" from the applications flow. I'm happy to tidy up, but feel we should clarify our nomenclature first.
* asked @tomodwyer to review my use of Bootstrap. If this PR is approved, then I will do so before squashing and merging.

---

I wasn't able to determine how to test the changes introduced by this PR. Ideally, we'd like to test that the `auth-login` path points to the twice-decorated `social_django_views.auth` view function: once decorated with `csrf_exempt` and once decorated with `require_POST`. Whilst it's straightforward to fetch the view function that corresponds to a path, it's less straightforward to determine whether this view function has been decorated:

```python
from django.urls import resolve

match = resolve("/login/github/")
match.func  # <function social_django.views.auth(request, backend)>
match.func.__wrapped__  # <function social_django.views.auth(request, backend)> Same as above
```

I briefly looked at the [`inspect`](https://docs.python.org/3.9/library/inspect.html#module-inspect) module for guidance, but this seemed hacky.

I also took a somewhat "quick and dirty" approach to applying classes to the HTML elements in *jobserver/templates/login.html*. Perhaps @tomodwyer would be able to 👀?

Closes #1037

### DIFF
--- a/jobserver/templates/login.html
+++ b/jobserver/templates/login.html
@@ -30,12 +30,11 @@
 {% block content %}
 <div class="container">
   <div class="row">
-    <div class="col-sm-12">
+    <div class="col col-sm-8 offset-sm-2 col-md-6 offset-md-3 col-lg-4 offset-lg-4">
       <form method="POST" action="{% url "auth-login" "github" %}?next={{ next_url }}">
-        <div class="w-25 mx-auto">
-          <button type="submit" class="btn btn-outline-primary btn-block">Login with GitHub</button>
-          <button type="submit" class="btn btn-outline-secondary btn-block">Register with GitHub</button>
-        </div>
+        <button type="submit" class="btn btn-primary btn-block">Sign in with GitHub</button>
+        <p class="text-center my-2">or</p>
+        <button type="submit" class="btn btn-outline-secondary btn-block">Create a GitHub account</button>
       </form>
     </div>
   </div>
--- a/jobserver/templates/login.html
+++ b/jobserver/templates/login.html
@@ -1,5 +1,32 @@
 {% extends "base.html" %}
 
+{% block metatitle %}Login | OpenSAFELY Jobs{% endblock metatitle %}
+
+{% block breadcrumbs %}
+<nav class="breadcrumb-container" aria-label="breadcrumb">
+  <div class="container">
+    <ol class="breadcrumb rounded-0 mb-0 px-0">
+      <li class="breadcrumb-item">
+        <a href="/">Home</a>
+      </li>
+      <li class="breadcrumb-item active" aria-current="page">
+        Login
+      </li>
+    </ol>
+  </div>
+</nav>
+{% endblock breadcrumbs %}
+
+{% block jumbotron %}
+<div class="jumbotron jumbotron-fluid pt-md-2">
+  <div class="container">
+    <h1 class="pt-4 pt-md-0">Login</h1>
+    <p class="lead">OpenSAFELY Jobs uses your GitHub Account to sign to confirm your identity.</p>
+    <p>If you do not have a GitHub account, you will need to create one before signing in.</p>
+  </div>
+</div>
+{% endblock jumbotron %}
+
 {% block content %}
 <div class="container">
   <div class="row">
--- a/jobserver/templates/login.html
+++ b/jobserver/templates/login.html
@@ -4,7 +4,7 @@
 <div class="container">
   <div class="row">
     <div class="col-sm-12">
-      <form method="POST" action="{% url "auth-login" "github" %}?next={{ request.path }}">
+      <form method="POST" action="{% url "auth-login" "github" %}?next={{ request.GET.next|default:"/" }}">
         <div class="w-25 mx-auto">
           <button type="submit" class="btn btn-outline-primary btn-block">Login with GitHub</button>
           <button type="submit" class="btn btn-outline-secondary btn-block">Register with GitHub</button>
--- a/jobserver/templates/login.html
+++ b/jobserver/templates/login.html
@@ -21,8 +21,8 @@
 <div class="jumbotron jumbotron-fluid pt-md-2">
   <div class="container">
     <h1 class="pt-4 pt-md-0">Login</h1>
-    <p class="lead">OpenSAFELY Jobs uses your GitHub Account to sign to confirm your identity.</p>
-    <p>If you do not have a GitHub account, you will need to create one before signing in.</p>
+    <p class="lead">OpenSAFELY Jobs uses your GitHub account to confirm your identity.</p>
+    <p>If you do not have a GitHub account, then you will need to create one before signing in.</p>
   </div>
 </div>
 {% endblock jumbotron %}
--- a/jobserver/templates/login.html
+++ b/jobserver/templates/login.html
@@ -1,0 +1,16 @@
+{% extends "base.html" %}
+
+{% block content %}
+<div class="container">
+  <div class="row">
+    <div class="col-sm-12">
+      <form method="POST" action="{% url "auth-login" "github" %}?next={{ request.path }}">
+        <div class="w-25 mx-auto">
+          <button type="submit" class="btn btn-outline-primary btn-block">Login with GitHub</button>
+          <button type="submit" class="btn btn-outline-secondary btn-block">Register with GitHub</button>
+        </div>
+      </form>
+    </div>
+  </div>
+</div>
+{% endblock content %}
--- a/jobserver/templates/login.html
+++ b/jobserver/templates/login.html
@@ -4,7 +4,7 @@
 <div class="container">
   <div class="row">
     <div class="col-sm-12">
-      <form method="POST" action="{% url "auth-login" "github" %}?next={{ request.GET.next|default:"/" }}">
+      <form method="POST" action="{% url "auth-login" "github" %}?next={{ next_url }}">
         <div class="w-25 mx-auto">
           <button type="submit" class="btn btn-outline-primary btn-block">Login with GitHub</button>
           <button type="submit" class="btn btn-outline-secondary btn-block">Register with GitHub</button>
--- a/jobserver/templates/partials/header.html
+++ b/jobserver/templates/partials/header.html
@@ -17,7 +17,7 @@
       </ul>
 
       {% if not user.is_authenticated %}
-        <a class="btn btn-outline-light mt-2 mb-3 my-lg-0" href="{% url "social:begin" "github" %}?next={{ request.path }}" rel="nofollow">Login</a>
+        <a class="btn btn-outline-light mt-2 mb-3 my-lg-0" href="{% url "login" %}" rel="nofollow">Login</a>
       {% else %}
         <div class="dropdown mt-2 mb-3 my-lg-0">
           <button class="btn btn-link text-white nav-link dropdown-toggle px-0 px-md-2" id="navbarUserDropdown" type="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
--- a/jobserver/templates/partials/header.html
+++ b/jobserver/templates/partials/header.html
@@ -17,7 +17,7 @@
       </ul>
 
       {% if not user.is_authenticated %}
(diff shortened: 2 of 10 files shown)


jobserver/templates/partials/header.html
```diff
@@ -17,7 +17,7 @@
       </ul>
 
       {% if not user.is_authenticated %}
-        <a class="btn btn-outline-light mt-2 mb-3 my-lg-0" href="{% url "social:begin" "github" %}?next={{ request.path }}" rel="nofollow">Login</a>
+        <a class="btn btn-outline-light mt-2 mb-3 my-lg-0" href="{% url "login" %}" rel="nofollow">Login</a>
```
evansd, line 20: I think we still want `?next={{ request.path }}` here to persist the current URL.

jobserver/templates/login.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+
+{% block content %}
+<div class="container">
+  <div class="row">
+    <div class="col-sm-12">
+      <form method="POST" action="{% url "auth-login" "github" %}?next={{ request.path }}">
```
evansd, line 7: This `next` parameter will just bring the user back to the login page which I don't think is what we want. I think instead we need to save the original page where the login link was clicked (see below) and then use that parameter as the `next` value here.

jobserver/templates/partials/header.html
```diff
@@ -17,7 +17,7 @@
       </ul>
 
       {% if not user.is_authenticated %}
-        <a class="btn btn-outline-light mt-2 mb-3 my-lg-0" href="{% url "social:begin" "github" %}?next={{ request.path }}" rel="nofollow">Login</a>
```
evansd, line 20: There's another login link somewhere else. I can't remember exactly where but if you search `social:begin` you'll find it. I know this because I missed it out when I added the `rel="nofollow"` attribute and George had to point it out to me :smile:

evansd (commented):
Ah, you've gone for the gold-plated version with the whole new login page - excellent!

Couple of comments below. I've held off officially approving as this is just draft, but apart from the comments I've made it all looks good to me.

jobserver/templates/partials/header.html
iaindillingham: Yup, agreed! Thanks for spotting this.

ghickman:
@iaindillingham – it's just occurred to me that you'll need to switch the login URL [referenced in settings](https://github.com/opensafely-core/job-server/blob/98a340ece9eacd34f5c41ef71567670322e216b4/jobserver/settings.py#L168) to the new one too.

jobserver/templates/login.html
iaindillingham: I've used the value of `request.GET.next` for `next`, here. Consequently, a user will be returned to the page where they clicked the "Login" link in the header, rather than to the login page. I've added a template filter, for the case where a user arrives at the login page directly. In this case, they will be returned to the homepage.

jobserver/templates/partials/header.html
iaindillingham: `applications/templates/applications/sign_in.html` 🙂

jobserver/templates/login.html
evansd: I think this is probably fine, although technically this creates an "open redirect". I'm going to seek further opinions on the extent to which we need to care about this.

tomodwyer:
Give me a shout when you're happy with the functionality and I will 👀 the templates

jobserver/templates/login.html
```diff
@@ -4,7 +4,7 @@
 <div class="container">
   <div class="row">
     <div class="col-sm-12">
-      <form method="POST" action="{% url "auth-login" "github" %}?next={{ request.path }}">
+      <form method="POST" action="{% url "auth-login" "github" %}?next={{ request.GET.next|default:"/" }}">
```
evansd, line 7: This is a bit of a nitpick, but I'd be inclined to use a context variable here (`next_url` or something) which gets set in the view code rather than reference `request.GET.next` directly in the template. As it happens, the value of `GET.next` has already been validated, but it's not immediately obvious from looking at the template that that's the case; and in general referencing values directly from the request like this is a bit of a smell. It would also means you could do the default handling in the view as well, which would be a bit cleaner.

It should just be a case of doing something like this in the view:
```py
    def get(self, request, *args, **kwargs):
        next_url = request.GET.get("next", "/")
        if is_safe_path(next_url):
            return render(request, self.template_name, {"next_url": next_url})
        else:
            return bad_request(request, SuspiciousOperation)
```
```

File: jobserver/views/users.py [hunk 5]
```diff
@@ -1,9 +1,23 @@
...
 from django.utils.decorators import method_decorator
-from django.views.generic import UpdateView
+from django.views.defaults import bad_request
+from django.views.generic import TemplateView, UpdateView
 
 from ..forms import SettingsForm
+from ..utils import is_safe_path
+
+
+class Login(TemplateView):
```
evansd, line 13: Just a comment for info, rather than anything that needs changing, but you're not really using the TemplateView class here at all so this could just be a plain view function like this:
```py
def login_view(request):
    if is_safe_path(request.GET.get("next", "")):
        return render(request, "login.html")
    else:
        return bad_request(request, SuspiciousOperation)
```
Then in `urls.py` and elsewhere you'd just reference it as `login_view` rather than `LoginView.as_view()`.

Class-based views offer some very powerful features, but there's a conceptual simplicity to function-based views which might help you on your Django journey -- they're just functions which accept a `request` object and return a `response`.

evansd (approved):
Happy to approve. My nitpick is stylistic and doesn't affect the functionality.

jobserver/views/users.py
ghickman: Just make sure you return a `TemplateResponse` (it's the same API) rather than a bare `render`! (it adds template and context info to the response, which makes it easier to test)

evansd: Oh good shout, thanks George.

jobserver/templates/login.html
iaindillingham: Ah! Great, thank you. I was going to ask about how to handle the default, but forgot. I've modified `Login.get` and *jobserver/templates/login.html*.

jobserver/views/users.py
iaindillingham: Okay, so `render` returns a `HttpResponse` but this is less convenient than a `TemplateResponse`, because the latter is easier to test. From [Django's documentation](https://docs.djangoproject.com/en/3.2/ref/template-response/), this is because...

> Unlike basic `HttpResponse` objects, `TemplateResponse` objects retain the details of the template and context that was provided by the view to compute the response. The final output of the response is not computed until it is needed, later in the response process.

Unless you feel very strongly about view functions vs class-based views, I'm going to retain the class-based view. Thank you for the comment, though 👍🏻

tomodwyer (approved):
The man from front-end, he says yes ✅

tests/unit/jobserver/test_utils.py
```diff
@@ -9,6 +9,14 @@ def test_dotted_path():
     assert dotted_path(OutputChecker) == "jobserver.authorization.roles.OutputChecker"
 
 
+def test_is_safe_path_with_safe_path():
+    assert is_safe_path("/")
```
ghickman, line 13: Could you also check a path with a component/segment/some text in it alongside this check? (experience with path handling and single slashes has caused me issues in the past!)

jobserver/views/users.py
ghickman: To add to Dave's point, you're wiping out what TemplateView does for you by overriding all of `get()` ([full code here](http://ccbv.co.uk/projects/Django/3.1/django.views.generic.base/TemplateView/#get)).  However, since you need to add `next_url` to the context it all gets a bit bogged down with boilerplate if you try to keep the bits TemplateView is providing for you (context generation, template rendering, and response creation).

If you want to stick with a CBV then View is your best bet:

```
class Login(View):
    def get(self, request, *args, **kwargs):
        next_url = request.GET.get("next", "/")
        if is_safe_path(next_url):
            return TemplateResponse(request, self.template_name, {"next_url": next_url})
        else:
            return bad_request(request, SuspiciousOperation)
```

ghickman (changes requested):
Nearly there!

tests/unit/jobserver/test_utils.py
iaindillingham: I've added `assert is_safe_path("/status/")` to the same test.

jobserver/views/users.py
iaindillingham: I've converted the CBV to a view function. I'll also read up on Django views. Thanks for the comments @evansd and @ghickman.

ghickman (approved):

evansd:
@iaindillingham Not that this particularly matters but in case it's helpful for your own mental model of what's happening here, this description isn't really accurate:
> Previously, get requests were made to GH from each page. Now, post requests are made to GH from a new login page. Post requests don't create new sessions in the database

The background here is that there's a login process we do with Github and starting that process involves creating a session for the user. We used to start that process whenever the user made a GET request to `/login/github/`. Now we require a POST request to `/login/github/` (via your swanky new form).

A convention of HTTP is that GETs should always be "safe" in the sense that they don't modify any state. This means that web-crawlers will happily make GETs against a site in the belief that they shouldn't be able to do any damage this way, but they should never make POSTs.

Previously, crawlers would see the login link, follow it by issuing a GET and thus inadvertently start the login process and create a new session. Now, following the login link just takes them to a page with a form on it and in order to start the login process and create a new session they would need to submit that form with a POST. But crawlers don't do that. So we don't end up creating unnecessary sessions.

Hope that makes some sense and is helpful!

iaindillingham:
It does matter and it is helpful 🙂 

Just to check my understanding: The login process begins when [`social_django.views.auth`][1] is called. This function calls [`social_core.actions.do_auth`][2], which creates a session. As you say, this function used to be called when a GET request was made; it's now called when a POST request is made. Crawlers will make GETs but won't make POSTs. Consequently, crawlers used to create sessions, but now they don't.

It's not the case that POSTs don't create sessions; POSTs, *do* create sessions! It's simply the case that by switching from GET to POST, we've stopped crawlers creating sessions.

[1]: https://github.com/python-social-auth/social-app-django/blob/0fde29f12da5dcea1b94bdc3b5b14abacb3e503a/social_django/views.py#L20-L23
[2]: https://github.com/python-social-auth/social-core/blob/45a5abc6bbb2c890d85c811b8d7a82e6fb8f1aa0/social_core/actions.py#L7-L29

PR #1067, opened 2021-10-04
Add Applications to the Staff Area

ghickman:
This adds list and detail pages for Applications to the Staff area and makes applications searchable by their creator on the homepage.

The fields listed on the detail pages have all been manually added for now since we're not quite sure how it's going to work so I just did some vim macro magic with the intention that we should definitely do something better there once we have an idea how the application-dependent models are going to work.


tomodwyer (approved):
✅ templates match the staff area styles. Needs a review from a Pythonista for the `.py` files.

PR #1074, opened 2021-10-05
Avoid setting blank shortNames for single file outputs responses

### LINKED ISSUE #1073: A Single File Breaks the Outputs Viewer
When a page only has one file to display the outputs viewer shows no file list.  The API call returns a 200 and looks to be in the correct format, and there are no errors in the console.  Multiple files work as expected.

Example response with one file, based on a real single-file response:
```
{
    "files": [
        {
            "name": "output/cusum/picture.png",
            "id": "an_id",
            "url": "/api/v2/releases/file/an_id",
            "user": "user",
            "date": "2021-10-04T15:22:52.297821+00:00",
            "sha256": "abc123",
            "size": 12345,
            "is_deleted": false,
            "backend": "backend"
        }
    ]
}
```

ghickman:
We pull a common prefix from the filenames served to the outputs viewer
app.  We use this prefix to create a `shortName` value on the files
array for display in the file list UI.  However with one file the
longest common prefix is the entire filename, leading to an empty
shortName, and showing an empty file list.

Fixes #1073

### DIFF
--- a/assets/src/scripts/outputs-viewer/hooks/use-file-list.js
+++ b/assets/src/scripts/outputs-viewer/hooks/use-file-list.js
@@ -4,6 +4,11 @@ import useStore from "../stores/use-store";
 import { toastError } from "../utils/toast";
 
 export function longestStartingSubstr(array) {
+  if (array.length < 2) {
+    // don't match the entire string for single item arrays
+    return "";
+  }
+
   const A = array.concat().sort();
   const a1 = A[0];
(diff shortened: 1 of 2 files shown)


evansd (approved):
This is _such_ a great bug!

bloodearnest:
Excellent elision decisions

tomodwyer (approved):
I'm going to have a think about a nicer way to do this, but let's get it merged for the moment 👍

PR #1093, opened 2021-10-07
Show admin status of each page

inglesp:


applications/wizard.py
```diff
@@ -5,6 +5,12 @@
 from .forms import PageFormBase
 
 
+UNSTARTED = "unstarted"
+NEEDS_REVIEW = "needs review"
+NOT_APPROVED = "not approved"
+APPROVED = "approved"
```
ghickman, line 11: Would an Enum/TextChoices help with how these are used?

inglesp: They're not used in the database, so no need for TextChoices.

I'm not sure what benefit an enum would give?

PR #1095, opened 2021-10-07
Add Private Repo Report

### LINKED ISSUE #1091: Create a list of opensafely projects, including created-at date
When people sign up to use OpenSAFELY, we ask them to commit to publishing their repo code after 12 months, come what may.

We need to start monitoring this. A page listing all repos, their created date, and their associated project, sorted by created date (desc) would be ideal; in public even better.

Less ideal, but sufficient, would be a script we can run that does the same thing, run periodically, when we want to (or by cron job).

We need some kind of (albeit hacky) solution for a meeting on 14th October, if possible

ghickman:
This adds a list of private repos to the staff area, with links to the repo, Workspace, and Project.

Fixes #1091

### DIFF
--- a/jobserver/context_processors.py
+++ b/jobserver/context_processors.py
@@ -71,6 +71,11 @@ def staff_nav(request):
             "url": reverse("staff:project-list"),
         },
         {
+            "name": "Repos",
+            "is_active": _active(reverse("staff:repo-list")),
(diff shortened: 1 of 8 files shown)


sebbacon:
I've just realised we kind of need to see if the repo is private or not... sorry! Hope it's easy

PR #1097, opened 2021-10-07
Add progress bar to application pages

inglesp:
image

### DIFF
--- a/applications/templates/applications/page.html
+++ b/applications/templates/applications/page.html
@@ -40,6 +40,19 @@
 <article class="container" id="applicationForm">
   <div class="row">
     <div class="col-lg-8">
+      <div class="progress mb-4">
+        <div
+          class="progress-bar"
+          role="progressbar"
+          style="width: {{ progress_percent}}%"
+          aria-valuenow="{{ progress_percent}}"
+          aria-valuemin="0"
+          aria-valuemax="100"
+        >
+          {{ progress_percent }}%
+        </div>
+      </div>
+
       {% if rubric %}
         <div class="d-flex flex-column mb-3">
           {{ rubric }}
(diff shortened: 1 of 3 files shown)


applications/wizard.py
```diff
@@ -36,6 +36,17 @@ def get_next_page_key(self, key):
             if page.status == UNSTARTED:
                 return page.key
 
+    def progress_pc(self):
```
ghickman, line 39: pc == percent[age]?

applications/wizard.py
```diff
@@ -36,6 +36,17 @@ def get_next_page_key(self, key):
...
+        """Return user's progress through wizard as a percentage."""
+
+        num_pages = num_pages_done = 0
+        for page in self.get_pages():
+            num_pages += 1
+            if page.status != UNSTARTED:
+                num_pages_done += 1
```
ghickman, line 46: This took me a couple of read throughs to get my head around, I think it's because it's building the two counts at the same time.  Do you think this is more readable?

```
num_pages = len([p for p in self.get_pages()])
num_pages_done = len([p for p in self.get_pages() if p.status != UNSTARTED])
```

I'm not sure I like there being two loops, but it's unlikely to be a performance issue at this scale!

applications/templates/applications/page.html
```diff
@@ -40,6 +40,10 @@ <h1 class="display-4">{{ title }}</h1>
 <article class="container" id="applicationForm">
   <div class="row">
     <div class="col-lg-8">
+      <div class="progress mb-4">
+        <div class="progress-bar" role="progressbar" style="width: {{ progress_pc}}%" aria-valuenow="{{ progress_pc}}" aria-valuemin="0" aria-valuemax="100">{{ progress_pc }}%</div>
```
ghickman, line 44: Can you break the attributes down into multiple lines?  Makes it easier to scan down them!

PR #1098, opened 2021-10-07
Change form pages to unapproved on applicant edit

StevenMaude:
If a form page is approved, and subsequently edited by the applicant, rescind the approved status.

### DIFF
--- a/applications/views.py
+++ b/applications/views.py
@@ -132,6 +132,10 @@ def page(request, pk, key):
     else:
         form = page.get_bound_data_form(request.POST)
 
+        if form.has_changed():
+            page.instance.is_approved = False
+            page.instance.save()
+
         form.save()
         if not page.form_spec.can_continue(application):
             form.add_error(None, page.form_spec.cant_continue_message)
--- a/tests/unit/applications/test_views.py
+++ b/tests/unit/applications/test_views.py
@@ -19,6 +19,7 @@ from applications.views import (
     terms,
     validate_application_access,
 )
+from applications.wizard import Wizard
 from jobserver.authorization import CoreDeveloper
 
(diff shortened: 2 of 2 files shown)


applications/wizard.py
```diff
@@ -98,11 +98,11 @@ def get_unbound_data_form(self):
     def get_unbound_approval_form(self):
         return self.approval_form_class(instance=self.instance)
 
-    def get_bound_data_form(self, data, initial=None):
+    def get_bound_data_form(self, data, **kwargs):
```
StevenMaude, line 101: Not entirely sure what's better here, possibly `**kwargs`?

applications/views.py
```diff
@@ -127,11 +127,15 @@ def page(request, pk, key):
     except ValueError:
         raise Http404
 
+    unbound_form = page.get_unbound_data_form()
+
     if request.method == "GET":
-        form = page.get_unbound_data_form()
+        form = unbound_form
```
ghickman, line 133: Not sure we need this optimisation any more!

applications/views.py
```diff
@@ -127,11 +127,15 @@ def page(request, pk, key):
...
     else:
         form = page.get_bound_data_form(request.POST)
-
+        if form.has_changed():
+            page.instance.is_approved = False
+            page.instance.save()
```
ghickman, line 138: nit: could we keep a blank both before and after this block?  It can help to denote a separate block for some.

ghickman (changes requested):
Couple of tiny nits, nearly there!

applications/views.py
StevenMaude: :eyes: I got blinded to what I'd changed from what was originally present; thanks! Just edited the commit.

PR #1106, opened 2021-10-08
Validate Entire Application on Confirmation Page

ghickman:
*Rebuilt after switching to mirroring the staff template and using Field to define value and validity*

This adds final validation to the Wizard so we know if the confirmation can be submitted or not.  Validation is bubbled up through the form_spec helpers since Field already knows if it's valid or not so we build up from there to the Wizard.

### DIFF
--- a/applications/wizard.py
+++ b/applications/wizard.py
@@ -36,6 +36,9 @@ class Wizard:
             if page.status == UNSTARTED:
                 return page.key
 
+    def is_valid(self):
+        return all(p.form_spec.is_valid(p.page_instance) for p in self.get_pages())
+
     def progress_percent(self):
         """Return user's progress through wizard as a percentage."""
 
--- a/applications/form_spec_helpers.py
+++ b/applications/form_spec_helpers.py
@@ -33,6 +33,9 @@ class Form:
             "template_name": self.template_name,
         }
 
+    def is_valid(self, page_instance):
+        return all(fieldset.is_valid(page_instance) for fieldset in self.fieldsets)
+
     def review_context(self, page_instance):
         return {
             "title": self.title,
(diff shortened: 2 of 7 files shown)


applications/wizard.py
```diff
@@ -36,6 +36,55 @@ def get_next_page_key(self, key):
...
+        Build a structure for display on the confirmation page
+        """
+        for page in self.get_pages():
+            fields = [
+                {
+                    "label": field.label,
+                    "value": field.value(),
+                    "is_valid": self.is_field_valid(field, page.form_spec),
+                }
+                for field in page.get_unbound_data_form()
```
inglesp, line 95: I don't think we should be using `page.get_unbound_data_form()` here.  It's what I originally used for the confirmation page, but we've shown in the staff review page that we can render all the data for an application without using forms.

A suggestion:

1. add an `"is_valid"` key to `Field.review_context()`
2. re-write part of `confirmation.html` to use the same strategy for rendering pages as we do in [`application_detail.html`](https://github.com/opensafely-core/job-server/blob/main/staff/templates/staff/application_detail.html#L66-L82)

If you do this, I think you'll also be able to drop the `Wizard.is_field_optional` and `Wizard.is_field_valid` methods.

tests/factories/applications.py
```diff
@@ -31,123 +32,165 @@ class Meta:
     project = factory.SubFactory("tests.factories.ProjectFactory")
 
 
-class ContactDetailsPageFactory(factory.django.DjangoModelFactory):
+class AbstractPageFactory(factory.django.DjangoModelFactory):
```
inglesp, line 35: Ugh, thanks for doing this.

inglesp (changes requested):
Looks good, but I think the implementation could be simplified before we merge.

inglesp (commented):
Looks good.  How do we stop users updating an application once it has been submitted?

inglesp (approved):

PR #1119, opened 2021-10-12
Whitenoise skip compression for specific file types

tomodwyer:
So that `collectstatic` doesn't perform unnecessary work.


jobserver/settings.py
```diff
@@ -161,6 +161,28 @@
...
+    "xz",
+    "br",
+    "swf",
+    "flv",
+    "woff",
+    "woff2",
+    "svg",
+    "ico",
+    "webmanifest",
+    "json",
```
ghickman, line 184: Can you order these alphabetically please?  It'll be easier to see what we're using when we need to add another one later.

tomodwyer: Good point, done ✅

tomodwyer:
> I'm sure you've got good reasons for it, but out of interest what's the rationale for skipping `svg`, `json` and `webmanifest`? Being text-based they'd benefit from compression. Is it just that there are lots of them and its slowing the build down? If that's the case, and if you'd rather optimise build times than load times (which is perfectly reasonable in the context of this site) then have you considered just turning off compression altogether?
> 
> (I should add `ico` to the defaults list though.)

This is mostly because they are already being optimised by Vite during the build process, so whitenoise is just doubling the effort to compress them.

The plan is to turn off compression once we can put the Jobs site behind Cloudflare, and allow that to serve Brotli/Gzipped and minified assets for us.

ghickman (approved):

PR #1122, opened 2021-10-12
Add Django Upgrade

ghickman:
Similar to pyupgrade this looks for outdated syntax, but in our Django-specific code.


StevenMaude (approved):
:+1:

PR #1123, opened 2021-10-12
Identify applications by reference number

inglesp:
Each application is given a reference number.  This is a four-digit hex
code, which is a reversable hash of the primary key.

This is currently only used in URLs (changing eg /applications/1/ to
/applications/1edd/), but should be made visible in the UI so that users
and reviewers can user the reference number in conversation.

### DIFF
--- a/tests/unit/jobserver/test_hash_utils.py
+++ b/tests/unit/jobserver/test_hash_utils.py
@@ -1,0 +1,47 @@
+import pytest
+from django.http import Http404
+
+from jobserver import hash_utils
+
+
+def test_roundtrip_1():
+    # Verify that for every integer between 0 and 65535, hashing the integer and
+    # unhashing the result returns the same integer.
+    for m in range(2 ** 16):
+        assert hash_utils.unhash(hash_utils.hash(m)) == m
+
+
+def test_roundtrip_2():
+    # Verify that for every hash string between "0000" and "ffff", unhashing the has and
+    # hashing the result returns the same string.
+    for m_hash in range(2 ** 16):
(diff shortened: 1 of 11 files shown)


tests/unit/jobserver/test_hash_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+def test_roundtrip_1():
+    for m in range(2 ** 16):
+        assert hash_utils.unhash(hash_utils.hash(m)) == m
+
+
+def test_roundtrip_2():
+    for m_hash in range(2 ** 16):
+        h = hex(m_hash)[2:].rjust(4, "0")
+        assert hash_utils.hash(hash_utils.unhash(h)) == h
```
ghickman, line 15: Could you explain what this one is doing?  Particularly the definition of `h` is rather confusing to me!

inglesp: Have tried to explain notation in module docstring -- can you let me know whether it's clear enough?

ghickman: Not at all I'm afraid.

Things I would like explained in this test in particular:
* why are we running `hex()` on the int in `m_hash`?
* Why are we slicing the resulting string?
* Why are we padding that string out to 4 digits with zeros?
* And finally, what does the round trip through hash and unhash show us?

inglesp: The first three questions are answered by the output in the module docstring.  I tried to explain using words, but couldn't find a way to do so that wasn't more convoluted than showing an example.

Have added comments to the roundtrip tests which I hope clarifies things.

PR #1124, opened 2021-10-13
Store and display review details

StevenMaude:
Actually store `reviewed_by` and `last_reviewed_at` for pages in the form.

These fields were in the page model, but storing them wasn't implemented.

### DIFF
--- a/tests/unit/staff/views/test_applications.py
+++ b/tests/unit/staff/views/test_applications.py
@@ -68,7 +68,10 @@ def test_applicationdetail_without_core_dev_role(rf):
 
 
 def test_applicationdetail_post_with_complete_application(
-    rf, core_developer, complete_application
+    rf,
+    core_developer,
+    complete_application,
+    freezer,
 ):
     application = complete_application
     data = {
@@ -78,6 +81,11 @@ def test_applicationdetail_post_with_complete_application(
         "study-information-is_approved": "True",
     }
 
+    assert application.contactdetailspage.last_reviewed_at is None
+    assert application.contactdetailspage.reviewed_by is None
+    assert application.studyinformationpage.last_reviewed_at is None
+    assert application.studyinformationpage.reviewed_by is None
+
     request = rf.post("/", data)
     request.user = core_developer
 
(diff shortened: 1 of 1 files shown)


staff/views/applications.py
```diff
@@ -31,10 +32,14 @@ def post(self, request, *args, **kwargs):
         wizard = Wizard(application, form_specs)
         pages = list(wizard.get_pages())
 
+        review_time = timezone.now()
+
         for page in pages:
             form = page.get_bound_approval_form(request.POST)
             if form.instance.pk:
                 form.save()
+                page.instance.last_reviewed_at = review_time
```
StevenMaude, line 41: At the moment, each page has `last_reviewed_at`.

But, because the review is currently for the entire form instead of per page, not sure there's a semantic way to distinguish between pages reviewed and not reviewed :thinking:

StevenMaude: Actually, "approved_by" can be null.

tests/unit/staff/views/test_applications.py
```diff
@@ -137,18 +137,20 @@ def test_applicationlist_search(rf, core_developer):
     assert set_from_qs(response.context_data["object_list"]) == {app1.pk, app2.pk}
 
 
-def test_applicationdetail_submitted_review_updates_last_reviewed(
+def test_applicationdetail_updates_review_details_following_review(
```
inglesp, line 140: Rather than adding a new test here, I'd be tempted to include these tests in one or both of `test_applicationdetail_post_with_complete_application` / `test_applicationdetail_post_with_incomplete_application`.

staff/views/applications.py
```diff
@@ -31,10 +32,15 @@ def post(self, request, *args, **kwargs):
...
+        review_time = timezone.now()
+
         for page in pages:
             form = page.get_bound_approval_form(request.POST)
-            if form.instance.pk:
+            if form.instance.pk and form.has_changed():
                 form.save()
+                page.instance.last_reviewed_at = review_time
+                page.instance.reviewed_by = request.user
+                page.instance.save()
```
inglesp, line 43: I'd prefer it if we had `form.instance` here instead of `page.instance` -- although they're the same object, that isn't obvious from reading this code.

PR #1150, opened 2021-10-14
Change asserts to set values instead

StevenMaude:
When the tests were written, the application factory did not set these
values. It now does following 9f8b042bb51ac5cc325a1caa6b0cc2df4ac5738a
so the initial correct assertions now fail.


StevenMaude:
Merging #1124 caused [test failures](https://github.com/opensafely-core/job-server/runs/3894637377) post-merge; this is a fix for that.

PR #1173, opened 2021-10-19
Squash Migrations

ghickman:
A mix of the changes from #1099 and the migration squash from #1167.  It also turns off migrations on container boot so we can deploy the changes, at which point I'll:
1. put the site into maintenance mode
2. clear out the jobserver migrations from the db
3. fake the jobserver migrations
4. reset the applications migrations with `python manage.py migrate applications zero && python manage.py migrate applications ` (since we don't care about that data)

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 ./manage.py check --deploy
-./manage.py migrate
+# ./manage.py migrate
 ./manage.py ensure_admins
 ./manage.py collectstatic --no-input
 


jobserver/migrations/0002_add_backends.py
```diff
@@ -6,12 +6,21 @@
...
+        name="Test",
+        slug="test",
+        level_4_url="https://test.opensafely.org",
+    )
+    Backend.objects.create(
+        name="TPP",
+        slug="tpp",
+        level_4_url="http://localhost:8001",
         parent_directory="/d/Level4Files/workspaces",
+        is_active=True,
```
inglesp, line 23: Won't the backends already be in the database?  Or are you planning to fake this migration?

jobserver/models/core.py
```diff
@@ -142,6 +142,11 @@ def runtime(self):
         return Runtime(int(hours), int(minutes), int(seconds), int(total_seconds))
 
 
+class JobRequestManager(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset().prefetch_related("jobs")
+
+
```
inglesp, line 149: Should these changes be here?

inglesp (approved):
Looks good (`+1,180 −3,694`, how could I refuse?) but there might be some code that belongs to a different PR.

jobserver/models/core.py
ghickman: No 😖

PR #1184, opened 2021-10-20
Prepare for Switch to Postgres [again]

ghickman:
Reraising #1167 but with the installation of pyscopg2 removed, which was handled in #1180


inglesp (approved):
:rubber stamp:

Issue #1033, opened 2021-09-29
Migrate to postgres


sebbacon:
* We all have experience of postgres
* At some point we expect we will repeatedly start hitting the practical concurrency limits of sqlite
* Most of the dev work will be to work around `n+1` issues with `select_related` and friends. 
  * @ghickman suggested "I genuinely think just getting all the URLs from python manage show_urls and having python open then in your browser with DJDT configured to pre-open would take 50% of the work away"
* We think `JobRequestList` views are likely to be the pathological case to test against

ghickman:
pgloader appears to successfully migrate all the data into postgres with this work file:```load database
     from sqlite://db.sqlite3
     into postgresql://localhost/jobserver

 with include drop, create tables, create indexes, reset sequences

  set work_mem to '16MB', maintenance_work_mem to '512 MB';
```

The event log page is showing 83 queries in 93ms locally, most of those queries are the expected duplicates from the JobRequest properties not making use of prefetch_selected jobs.

The home page is showing a similar issue for the 5 JobRequests listed there.

inglesp:
You don't want pgloader to create the tables for you!  It won't create all the constraints the Django expects, and then you'll have trouble down the line.  Better is to run migrations against an empty postgres database, and then to use `with data only`.

</hard-won-experience>

ghickman:
Great to know 👍

I suspect we want to keep reset sequences so our PKs don't overlap?

inglesp:
> I suspect we want to keep reset sequences so our PKs don't overlap?

Good point!

StevenMaude:
It seems like we've switched to PostgreSQL now. So can this issue be closed?

One small clean-up task I previously noted: we still have `litecli` in dependencies, which could be removed if we've completed the migration definitively. Removing that would also shrinks our list of dependencies slightly as there are a few in there only used by `litecli`.

One clean-up task I previously didn't: the `sqlite.load` file in the root of the repository could be removed, again if we've completed migration.

ghickman:
@StevenMaude – still waiting on your answer in #1186 🙂

As you said there's a couple of cleanup tasks too, so I've kept this open as a central place.

ghickman:
cleanup tasks all done and we've been running this in production for a while 🎉

PR #1186, opened 2021-10-20
Document Running Postgres Locally

### LINKED ISSUE #1189: Update developer documentation following PostgreSQL migration
Following the sqlite3 to PostgreSQL change, there are now extra steps to get job-server running locally outside of Docker, over merely having sqlite3 installed.

I think you need to:

1. Have a PostgreSQL instance running, either on your own machine outside of Docker or via `docker-compose up db`.
2. Have `DATABASE_URL=postgres://user:pass@localhost:5432/jobserver` set when running Django-specific commands.
    * If you have a `.env` loading tool that loads on entering directory, and add this variable to `.env`, then this variable will be set.
    * Otherwise, if you're only using `just` to load environment variables, `just run` will work fine with the `DATABASE_URL` added to `.env`. However, you would need to add `DATABASE_URL` to the environment for any non-`just` commands such as `python manage.py migrate` (e.g. by prefixing the command with the variable, `DATABASE_URL=… python manage.py migrate`).

None of this is currently documented: anyone running job-server themselves will have to figure this out.

ghickman:
This adds documentation for running Postgres locally when you're not using docker-compose.

Ref #1033 
Fixes #1189

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -3,6 +3,8 @@
 - [Local development](#local-development)
   - [Native](#native)
     - [Prerequisites](#prerequisites)
+    - [`just` commands](#just-commands)
+    - [Postgres](#postgres)
     - [Steps](#steps)
   - [Docker Compose](#docker-compose)
   - [Frontend development (CSS/JS)](#frontend-development-cssjs)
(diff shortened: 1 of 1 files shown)


DEVELOPERS.md
```diff
@@ -37,12 +39,54 @@ _Note:_ you will need the [Bitwarden CLI tool](https://bitwarden.com/help/articl
...
+[This guide](http://meshy.co.uk/posts/postgresql-without-passwords) explains how to set up ident (password-less) auth, and set some options for faster, but more dangerous, performance.
+
+If you need to upgrade an installation then the Arch Wiki has the [best instructions](https://wiki.archlinux.org/title/PostgreSQL#Upgrading_PostgreSQL) I've found.
+
+
+##### Creating a database
+
+You'll need a database in Postgres to work with, run:
+
+  psql -c "create database jobserver"
```
madwort, line 78: alternatively, you can use `pg_restore --create` when restoring

DEVELOPERS.md
```diff
@@ -37,12 +39,54 @@ _Note:_ you will need the [Bitwarden CLI tool](https://bitwarden.com/help/articl
...
+
+
+##### Restoring Backups
+
+Backups of production can be restored to a local database using this command:
+
+  pg_restore --no-acl --no-owner -d jobserver jobserver.dump
```
madwort, line 85: you might also want `--clean` which will wipe a pre-existing database (if it exists)?

DEVELOPERS.md
```diff
@@ -37,12 +39,54 @@ _Note:_ you will need the [Bitwarden CLI tool](https://bitwarden.com/help/articl
...
+
+[Postgres.app](https://postgresapp.com/) is the easiest way to run Postgres on macOS, you can install it from homebrew (casks) with:
+
+  brew install --cask postgres-unofficial
+
+You will need to add its bin directory to your path for the CLI tools to work.
+
+Postico is a popular GUI for Postgres:
+
+  brew install --cask postico
```
madwort, line 62: oooh, interesting, at the last count I was using pg_admin for a macOS GUI (& the new versions  **really** suck since they moved to the webserver arch)

madwort (commented):
"more of a comment than a question"

DEVELOPERS.md
ghickman: This worked better than expected, you don't even have to quit all your connections 🤩

DEVELOPERS.md
```diff
@@ -37,12 +39,54 @@ _Note:_ you will need the [Bitwarden CLI tool](https://bitwarden.com/help/articl
...
 If you want to maintain your own virtualenv make sure you have activated it before running a `just` command and it will be used instead.
 
+
+#### Postgres
+
+##### Installing on macOS
+
+[Postgres.app](https://postgresapp.com/) is the easiest way to run Postgres on macOS, you can install it from homebrew (casks) with:
+
+  brew install --cask postgres-unofficial
```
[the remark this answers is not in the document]
ghickman: Hardly a nit 😁  Not sure why the indent isn't working, but I've matched the other command styles and that's working 👍

DEVELOPERS.md
```diff
@@ -37,12 +39,80 @@ _Note:_ you will need the [Bitwarden CLI tool](https://bitwarden.com/help/articl
...
+
+```sh
+psql -c "create database jobserver"
+```
+
+
+##### Restoring Backups
+
+Backups of production can be restored to a local database using a dump pulled from production.
+Instructions and process for this are TBD.
```
StevenMaude, line 98: question: What's to be decided here? The instructions that are included below, or restoring to production?

DEVELOPERS.md
```diff
@@ -37,12 +39,80 @@ _Note:_ you will need the [Bitwarden CLI tool](https://bitwarden.com/help/articl
...
+```
+
+
+##### Installing on Linux
+
+Install postgresql with your package manager of choice.
+
+[This guide](http://meshy.co.uk/posts/postgresql-without-passwords) explains how to set up ident (password-less) auth, and set some options for faster, but more dangerous, performance.
+
+If you need to upgrade an installation then the Arch Wiki has the [best instructions](https://wiki.archlinux.org/title/PostgreSQL#Upgrading_PostgreSQL) I've found.
```
StevenMaude, line 83: nitpick (non-blocking): You could drop the first-person pronoun and make this slightly more concise:

> If you need to upgrade an installation, the [ArchWiki](https://wiki.archlinux.org/title/PostgreSQL#Upgrading_PostgreSQL) is a good reference.

StevenMaude:
Looks broadly fine to me now, @ghickman.

* I've not tested the installation routes here, though I know that docker-compose works to run PostgreSQL.
* The other routes look plausible as they are. I don't have any macOS experience to throw in there.

I have some small optional suggestions, but that's all :+1:

DEVELOPERS.md
ghickman: Good point… this has changed a bit since I wrote this, will update, thank you!

PR #1187, opened 2021-10-20
Reinstate ensuring admins now we've migrated

ghickman:
This was disabled as part of #1184 so the deploy of that branch wouldn't fail due to an empty database.


StevenMaude (approved):
:partying_face:

PR #1210, opened 2021-10-25
Use Factories for all DB Objects in Tests

### LINKED ISSUE #1104: Migrations slow down tests
It takes ~4s to run a few simple unit tests:

```
$ time python -m pytest tests/unit/jobserver/views/test_index.py 
======================= test session starts ========================
platform linux -- Python 3.9.3, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
django: settings: jobserver.settings (from ini)
rootdir: /home/inglesp/work/ebmdatalab/opensafely/job-server, configfile: pyproject.toml
plugins: network-0.0.1, subtests-0.5.0, django-4.4.0, env-0.6.2, cov-3.0.0, Faker-8.10.1, freezegun-0.4.2, mock-3.6.1
collected 5 items                                                  

tests/unit/jobserver/views/test_index.py .....                  [100%]

========================== 5 passed in 3.05s ==========================

real	0m4.072s
user	0m3.948s
sys	0m0.114s

```

About two thirds of the time is in `django_db_setup`, most of which is running migrations.

The attached flame graph was produced with `time py-spy record -o tests.svg -- python -m pytest tests/unit/jobserver/views/test_index.py`.

Not a disaster, but the problem will get worse with more migrations.  In particular, certain migrations (see #1099) add multiple seconds to startup time.

ghickman:
This switches every use of Backend and Org in our tests to use factories.  Since we set up both in migrations the tests had leaned on those.  However as #1104 points out the migrations add a few seconds to test runs, regardless of number of tests being run.  This PR also sets up tests to run with `--no-migrations` by default so everyone can realise that saving locally, but adds `--migrate` to CI to ensure they're checked there.

Since we're no longer adding Backends via the migrations but they're semi-static data I've set them up as a list in the code with a function (and calling command) to make sure they exist.  I've used `update_or_create()` so that the structure in the code can be our source of truth.

Fixes #1104


ghickman:
@inglesp – success!  Added the config style you suggested, could you take a look? (and ideally check it works locally?)

justfile
```diff
@@ -122,6 +122,7 @@ test *ARGS: devenv
         --cov=tests \
         --cov-report=html \
         --cov-report=term-missing:skip-covered \
+        --no-migrations \
```
inglesp, line 125: Would you consider adding this to the `[tool.pytest.ini_options]` section of pyproject.toml instead?  That way, it'll be picked up even if the user doesn't use `just`.

jobserver/backends.py
```diff
@@ -7,6 +7,51 @@
...
+        "level_4_url": "https://test.opensafely.org",
+    },
+    {
+        "id": 3,
+        "name": "TPP",
+        "slug": "tpp",
+        "is_active": True,
+        "level_4_url": "http://localhost:8001",
+    },
+]
```
inglesp, line 53: I can see why these are ordered alphabetically -- but I wonder whether ordering them by ID would make it easier to work out what ID to assign to the next backend when we add more?  Maybe not worth doing now.

inglesp (approved):
Looks good!

jobserver/backends.py
ghickman: Yep, that makes much more sense 👍

justfile
ghickman: Oh yes, that's a much better idea actually 👍

PR #1232, opened 2021-10-29
Track Co-pilot Support Window

### LINKED ISSUE #1207: Track Copilot End Date on Projects
Co-pilots are give a fixed amount of time to projects, it would be useful to know when that is coming to an end.

When an application is accepted we should set a field with a default length of time (I think this is 4 weeks currently).  Staff should be able to change this in the Staff Area as things sometimes don't start on time.

We can list this value on the Project detail page in the Staff Area, showing a humanized value of time remaining.

ghickman:
This adds the Project.copilot_support_ends_at field to track when the support window for a Project ends and makes it editable on the ProjectEdit page.

Fixes #1207


iaindillingham (approved):
🚀

PR #1234, opened 2021-10-29
Notify Slack of Projects with Impending Lack of Co-pilot

### LINKED ISSUE #1208: Notifying of Impending Co-Pilot Windows Ending
Once #1207 is done we can add a scheduled task, running each morning, to notify a Slack channel (probably `#co-pilot-support`?) of the impending co-pilot windows which are going to end.

Let's start with windows ending in 7 days, and list one application per line in a single message each morning at 8:30 to avoid bothering the channel too much.

On a day with no messages to send we should still send a message to be explicit.

ghickman:
This adds a function, wrapped in a command, to notify slack of Projects with a co-pilot support window which ends in <5 days.

The expectation is this will be run daily or weekly to avoid spamming the channel.

Fixes #1208

### DIFF
--- a/jobserver/copiloting.py
+++ b/jobserver/copiloting.py
@@ -1,0 +1,61 @@
+from datetime import timedelta
+
+import structlog
+from django.conf import settings
+from django.contrib.humanize.templatetags.humanize import naturalday
+from django.utils import timezone
+from furl import furl
+
+from jobserver.models import Project
+from services.slack import client as slack_client
+
+
+logger = structlog.get_logger(__name__)
+
+
+def notify_impending_copilot_windows_closing(days=5):
+    """
+    Send a message to Slack about an upcoming copilot support windows closing
+
+    This sends a single message to the co-piloting Slack channel to notify of
+    Projects whose co-pilot support window ends in the given number of days. If
+    there aren't any, it doesn't send a message.
+
+    Each Project found is linked to its Staff Area page.
+
+    Note: URL syntax is using Slack's app surface linking syntax, which is
+    different from the markdown syntax:
+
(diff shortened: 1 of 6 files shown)


jobserver/copiloting.py
```diff
@@ -0,0 +1,52 @@
...
+from jobserver.models import Project
+from services.slack import client as slack_client
+
+
+logger = structlog.get_logger(__name__)
+
+
+def notify_impending_copilot_windows_closing(days=5):
+    """
+    Send a message to slack about an Application instance
```
iaindillingham, line 18: Nitpick: This could be more specific. What about:

> Sends a message to Slack that reports the number of closing copilot windows in the next `days` days. If there aren't any, then doesn't send a message.

iaindillingham (approved):

jobserver/copiloting.py
ghickman: @iaindillingham – updated partly with your suggestion but also with some notes on the link style, does that explain enough now?  Happy to add more if you think it's necessary!

ghickman:
@bloodearnest – added the cron bits we discussed from [django-extensions](https://django-extensions.readthedocs.io/en/latest/jobs_scheduling.html) and [dokku](https://dokku.com/docs/processes/scheduled-cron-tasks/).  I've enabled cron for each of the various times on the basis that it saves someone trying to debug this later, are you happy with that?

jobserver/copiloting.py
iaindillingham: It does, thank you 🙂

bloodearnest (approved):

PR #1241, opened 2021-11-02
Add Application States

### LINKED ISSUE #1136: Allow staff users to set a status on application forms
As an admin user, I want to set form submissions to various "states" in the admin interface, so that I can easily see what part of the process submitted forms are in.

## Status

- Approved fully
- Approved subject to
- Rejected
- Completed
- Ongoing

## To do

- [ ] Admin can set a submission to a single state
- [ ] Admin can add a comment on the current state

ghickman:
This adds both a `status` (defaulting to Ongoing) and `status_comment` field to Applications, an edit page for Applications in the Staff Area to edit those fields, displays their value on the detail page for an Application and with each Application on the list page, and allows filtering Applications by status on the list page.

I've not set statuses automatically as the mapping from them to our current process isn't clear, so I've chosen to let staff set the status manually until we can lock them down.

Fixes #1136


StevenMaude:
A few observations and possible bug reports from running this branch locally. Some of these might be intentional design decisions and not bugs, and that's fine.

On the whole, the PR is fine, it just might warrant:

* a quick check that the design choices are as intended;
* better handling of bad application states, if they are reproducible.

## Possible intentional design decisions that may want checking

1. I was able to set the state to "Approved Fully" while individual application pages were not approved, or when the application was not even complete. This may be an intentional decision.
2. The statuses aren't strictly mutually exclusive — I could have an "ongoing" application that's "approved subject to", for instance, but it seems I can only set one. This may be in line with how @amirmehrkar uses states currently.
3. The "Approve" button is enabled even if an application isn't actually "Approved Fully".

## Possible bugs

1. As an applicant, I created an application and didn't enter anything at all. Then I went to the Staff review page and pressed approve  "Approve". That caused an error: `RelatedObjectDoesNotExist at /staff/applications/7b74/approve/ Application has no studyinformationpage.` from `return {"project_name": self.application.studyinformationpage.study_name}` in `get_initial()`.
2. If you have a duplicate project name, you get a `IntegrityError at /staff/applications/7b74/approve/ duplicate key value violates unique constraint "jobserver_project_name_key" DETAIL:  Key (name)=(Test) already exists.`
3. The UI for actually approving an application, i.e. creating a project, is a little counterintuitive to me. Pressing Approve actually creates a project rather than setting the application. I don't know if another word other than "Approve" might be appropriate. But a staff member involved in this process who understands the difference may not be as confused as me by the different meanings of the same word.
4. As an applicant, I still seem to be able to modify an application and have those changes show, even if it is "Approved Fully" and the project has been created with the Approve button.

StevenMaude:
@ghickman: thanks for clarifying! I think if the use makes sense to Amir and anyone else, then it's fine.

Will open issues for the errors.

StevenMaude (approved):

sebbacon:
>> Pressing Approve actually creates a project rather than setting the application
> This was an intentional design decision, the artefact of a successful application (ie an approved application) is a Project

Is there an argument for making this clear in the UI? Principle of least surprise
"Approve and create corresponding project" for example?

ghickman:
I think that's something to discuss with Amir once we reach a point where it's useful to discuss the UX of that button (he has yet to try it), until then I think we can park the discussion!

PR #1246, opened 2021-11-02
More Query Optimisations

ghickman:
I somehow missed this round of optimisations on the same page as  #1236 

This is a tricky one, since both the Workspace and Project ReleaseList views are rendering N Releases _and_ N ReleaseFiles.  I couldn't find a way to tell Django to select_related the approrpiate objects for building URLs with the `get_FOO_url` methods of the ReleaseFiles.  This lead to every ReleaseFile on the page executing 2-3 extra queries.

I tried rebuilding WorkspaceReleaseList by getting all the ReleaseFiles for the Workspace, then grouping by Release using groupby, but hit the same issue because I wasn't passing QuerySets around again.

I suspect we might be able to break the view down but I have a feeling it will hurt readability and have already spent several hours on it.

### DIFF
--- a/jobserver/templates/_release_list.html
+++ b/jobserver/templates/_release_list.html
@@ -39,7 +39,11 @@
             <td class="d-flex justify-content-between align-items-center">
 
               <div>
-                <a href="{{ file.get_absolute_url }}" {% if file.is_deleted %}class="btn-link disabled"{% endif %}>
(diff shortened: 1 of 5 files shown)


iaindillingham (approved):
Rather a rubber stamp from me; I think I can see what you're trying to do and it makes sense. I've not dug into Django to be able to suggest an alternative; I certainly don't have that kind of knowledge in my head 🙁.

PR #1250, opened 2021-11-03
Replace Zustand with React Context

tomodwyer:
- Replace dependency Zustand with built-in React Context for storing global state
- Move `listVisible` out of global state
- Return image URL rather than storing the data, as the server displaying the image contains `cache-control` headers


jest.config.js
```diff
@@ -6,6 +6,7 @@ module.exports = async () => ({
     "!**/vendor/**",
     "!<rootDir>/assets/src/scripts/outputs-viewer/tests/*",
     "!<rootDir>/assets/src/scripts/outputs-viewer/index.jsx",
+    "!<rootDir>/assets/src/scripts/outputs-viewer/context/FilesProvider.jsx",
```
ghickman, line 9: Is this because we're not testing the code in FilesProvider directly?

assets/src/scripts/outputs-viewer/hooks/use-file.js
```diff
@@ -44,7 +33,7 @@ function useFile(file) {
             responseType: "blob",
           })
           .then((response) => response.data)
-          .then((blob) => convertBlobToBase64(blob))
+          .then(() => file.url)
```
ghickman, line 36: Why are we switching to the image URL from blobs?

ghickman (approved):
Could of questions more out of my own interest though!

jest.config.js
tomodwyer: Yes, this file just sets up the context provider in the way recommended by React. It would be largely pointless to test when the code is unlikely to change, and the tests don't rely on it.

assets/src/scripts/outputs-viewer/hooks/use-file.js
tomodwyer: I've updated the description to cover this. Previously React was storing the base64 blob, but the server is returning the correct caching headers, so the image isn't reloaded by the SPA for the same URL. That means there's no point React storing the data, when the URL will work just fine.

This now confirms that the image is available before we show the image on the client-side.

PR #1256, opened 2021-11-04
Add Dump Database Command

ghickman:
This adds a command to run pg_dump on our production database, putting it into the configured storage location for job-server in production.

This means developers can more easily copy down the production db for local work.

### DIFF
--- a/jobserver/jobs/hourly/dump_db.py
+++ b/jobserver/jobs/hourly/dump_db.py
@@ -1,0 +1,39 @@
+import pathlib
+import subprocess
+import sys
+
+from django_extensions.management.jobs import HourlyJob
+from environs import Env
+
+
+env = Env()
+
+
+class Job(HourlyJob):
+    help = "Dump the database to storage for copying to local dev environments"  # noqa: A003
+
+    def execute(self):
+        database_url = env.str("DATABASE_URL")
+        output = pathlib.Path("/storage/jobserver.dump")
+
+        if not output.exists():
+            print(f"Unknown output path: {output}", file=sys.stderr)
+            sys.exit(1)
+
+        try:
+            with output.open() as f:
+                # We use the "custom" output format for pg_dump to get smaller
+                # files, the docs have more detail on why it can be useful:
+                # https://www.postgresql.org/docs/14/app-pgdump.html
+                subprocess.check_call(
+                    [
+                        "pg_dump",
+                        "--format=c",
+                        "--no-acl",
+                        "--no-owner",
+                        database_url,
+                    ],
+                    stdout=f,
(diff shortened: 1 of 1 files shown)


jobserver/jobs/hourly/dump_db.py
```diff
@@ -0,0 +1,37 @@
...
+        if not output.exists():
+            print(f"Unknown output path: {output}", file=sys.stderr)
+            sys.exit(1)
+
+        try:
+            with output.open() as f:
+                subprocess.check_call(
+                    [
+                        "pg_dump",
+                        "-Fc",
```
bloodearnest, line 28: Huh, this is not clear to me.

Is it "--format=c"? Or "-F" (default) and "-c" (clean)?

If we are using a custom format, maybe a) use `--format=c` and b) a comment as to why? I'm guessing because of the no acl/owner stuff?

jobserver/jobs/hourly/dump_db.py
```diff
@@ -0,0 +1,37 @@
...
+                        "--no-acl",
+                        "--no-owner",
+                        database_url,
+                    ],
+                    stdout=f,
+                )
+        except Exception as e:
+            print(str(e), file=sys.stderr)
+            sys.exit(1)
```
bloodearnest, line 37: FWIW, this could be combined to just: 

`sys.exit(exc)`. 

https://docs.python.org/3/library/sys.html#sys.exit, 2nd paragraph.

bloodearnest (changes requested):

jobserver/jobs/hourly/dump_db.py
ghickman: It's --format=c, will update with long form!

The custom format typically creates much smaller files.  The docs also note that it provides more flexibility when restoring.

jobserver/jobs/hourly/dump_db.py
ghickman: Oh wow, TIL, useful!

jobserver/jobs/hourly/dump_db.py
ghickman: Do you think a comment would be useful here? I think I've only ever seen the custom format suggested when using pg_dump, so it's just the default style to me.

bloodearnest: So, I've generally just used text and gzipped it. So think a comment or maybe a link to doc explaining benefits would be good for future travellers

ghickman:
@bloodearnest – updated with a comment, does that work for you?

bloodearnest (approved):

PR #1257, opened 2021-11-04
Update ProjectCollaborator and ProjectDeveloper role descriptions

ghickman:
This updates ProjectDeveloper description and adds a description for ProjectDeveloper to explain to a user what they do.

### DIFF
--- a/jobserver/authorization/roles.py
+++ b/jobserver/authorization/roles.py
@@ -136,7 +136,7 @@ class ProjectCollaborator:
     """
 
     display_name = "Project Collaborator"
-    description = ""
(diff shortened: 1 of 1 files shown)


jobserver/authorization/roles.py
```diff
@@ -136,7 +136,7 @@ class ProjectCollaborator:
     """
 
     display_name = "Project Collaborator"
-    description = ""
+    description = "View unpublished outputs released to the Jobs site."
```
bloodearnest, line 139: "... released from Level 4 to the ..." ?

bloodearnest (approved):
+1 with minor wording suggestion

PR #1260, opened 2021-11-04
Remove backends env var

### LINKED ISSUE #1247: Remove BACKENDS and Related Filtering
We currently set BACKENDS and filter the base Backend QuerySet (that is Backend.objects.all()) based on its value.  Previously we added Backends via migrations and this allowed us to define which ones were available in an environment without further migrations.

However, this brought with it sharp edges, often for local dev setup, and most recently in production with the deployment of #1210 This tried to run `update_or_create` on all Backends, but since Backend.objects.all() returned all Backends _except_ Expectations it raised an IntegrityError as Expectations (with pk=2) already existed in the database.

Since we're already tracking some state on a Backend (is_active to decide if we care about it not checking in) we should move to putting all state on the model, rather than keeping some in the env.

So lets remove:
* `BACKENDS` as a configuration value
* `available_backends` and `get_configured_backends` from `jobserver/backends.py`
* the custom `BackendManager`

The Staff Area will need to grow the ability to add new Backends and edit any state we set on them (is_active is probably fine for now, but it will need help text to explain what that defines).

The docs will need updating to remove the BACKENDS sections.

ghickman:
This moves all Backend config into the application, dropping the BACKENDs env var, and the filtering which was based on this.

The Staff Area has grown the ability to create and edit Backends.

Fixes #1247


iaindillingham (approved):
The associated issue (#1247) mentions help text for `is_active`. This would be useful. What makes a backend inactive? What are the implications for job-server?

PR #1297, opened 2021-11-11
Remove Custom GitHub Auth Backend

### LINKED ISSUE #1133: GitHub login for users outside OpenSAFELY organisation
Currently users can only log in to the site if they are members of the @opensafely organisation. We want to open this up so all GitHub users can log in, enabling anyone to fill out an "Apply for research access to OpenSAFELY" form.

We need to check every existing page isn't affected by the removal of the requirement to be in the OpenSAFELY org, and there will be no impact to the security of the application.

ghickman:
This removes our custom auth backend, and the need for users to be a part of the `opensafely` GitHub organisation, so that we can move to having applications happen in job-server.

Fix #1133

### DIFF
--- a/jobserver/github.py
+++ b/jobserver/github.py
@@ -4,8 +4,6 @@ from datetime import datetime, timezone
 import requests
 from environs import Env
 from furl import furl
-from social_core.backends.github import GithubOAuth2
-from social_core.exceptions import AuthFailed
 
 
 env = Env()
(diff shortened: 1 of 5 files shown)


iaindillingham:
I've...
* Written all the URLs to a spreadsheet with `show_urls`
* Set up an "evil me" account. Because of the recent question about "evil me" accounts, I've used the more prosaic `testiaindillingham`
* Worked through each URL when logged in as `testiaindillingham` to check the response

I'm not sure how/whether to check...
* the social URLs (e.g. `social:begin`)
* job-server's API URLs (e.g. `api:user-detail`)
* invitation URLs (e.g. `project-accept-invite`)
* files, outputs, and releases URLs (e.g. `workspace-backend-files`)

For the most part, the responses are as I'd expect. I'm unsure whether the distinction between `404` and `404` is meaningful. For example, `project-membership-edit` returns `404` but `project-membership-remove` returns `405`.

ghickman:
> the social URLs (e.g. social:begin)

These should all be fine to leave untested.

> job-server's API URLs (e.g. api:user-detail)

These are all either publicly accessible or require backend auth so should be fine to leave.

> invitation URLs (e.g. project-accept-invite)
> files, outputs, and releases URLs (e.g. workspace-backend-files)

These definitely need testing as they require auth and either owernship and/or permissions to access.  Locally emails will be displayed in your devserver console so you can click links in there etc.  Files/releases/outputs require the ProjectCollaborator role to see on a Project, and can be generated with

    python manage.py release --workspace <workspace slug> --backend <backend slug> --user <username> outputs

> between 404 and 404

Did you mean 403 and 404 there?  Generally we've tried to use 404s to hide the existance of private items.  Can you check that those codes match the behaviour of the site when the opensafely-github-org-check is still in place?

> returns 405

This sounds like a bug, 405 is "Method not allowed" so there's likely an issue with the button calling that.

iaindillingham:
I've left the social URLs and the API URLs, as you suggest. I've tested invitation URLs and files, outputs, and releases URLs. Thanks for the note about the console and the management command 🙂 

Sorry, yes, `403` and `404` 🤦🏻‍♂️ This is my misunderstanding; I don't think there's a difference between the with-check state and the without-check state.

`405` is also my misunderstanding; I was naively copying from one window, where I was logged in as me, to another incognito window, where I was logged in as evil me. I neglected to issue `POST`s in the second window. The responses are as expected, now.

I'm afraid I don't understand the difference between _files_, _outputs_, and _releases_. I also don't understand how I find each in the UI. For example, when logged in as me:

* I've generously given myself the `ProjectCollaborator` role
* I've run `python manage.py release --workspace emis-qa-main --backend test --user iaindillingham outputs` to release some test outputs
* I see two URLs in the console:
  * http://localhost:8000/datalab/opensafely-testing/emis-qa-main/releases/01FMQHDCTYZ2FJM66DVCWWHJAZ/
  * http://localhost:8000/datalab/opensafely-testing/emis-qa-main/
* I don't see any links in the UI to take me to the first URL from the second URL. What am I missing?

A couple of issues I'm going to follow up:

* The `workspace-archive-toggle` button redirects to the homepage. Should it redirect to the workspace?
* `pg_restore` won't overwrite an existing database. Instead:

```sh
dropdb jobserver
createdb jobserver
pg_restore --clean --if-exists --no-acl --no-owner -d jobserver jobserver.dump
```

ghickman:
> I'm afraid I don't understand the difference between files, outputs, and releases. I also don't understand how I find each in the UI

I suspect this is because the `opensafely-testing` Project doesn't have new style releases enabled, all the links are hidden when it's disabled, which has bitten us before!  In the Staff Area you can edit a project to flip that on then both Project and Workspace pages will show links to released files, and Outputs once you publish some files in the UI.

> The workspace-archive-toggle button redirects to the homepage.

I picked the homepage when the site was a lot simpler because archiving is a soft-delete in this context so users were typically "done" with a workspace at that point.  So, yes, it's intentional, but it possibly needs revisiting at some point.

> pg_restore won't overwrite an existing database

This could be a tooling version issue let's discuss on Slack instead of cluttering the issue with it 🙂

iaindillingham:
I've enabled "new-style" releases on `opensafely-testing` and double-checked the outputs and releases URLs. Thanks for commenting on the other issues; they're tangential to the PR, as you say, so I'll move over to Slack.

iaindillingham (approved):

PR #1347, opened 2021-11-24
Update terms.md

amirmehrkar:
adding links and tweaks


iaindillingham (approved):
I've pushed a commit that fixes the formatting of the bulleted list; fixes a couple of links; and removes trailing whitespace.

PR #1353, opened 2021-11-24
Update service-evaluation-audit-rubric.md

amirmehrkar:


iaindillingham (approved):
I've pushed a commit that uses a numbered list; this means we don't have to keep track of the 1s, 2s, and 3s. The result looks like this:

Screenshot 2021-11-24 at 15-33-02 Apply Service evaluation or audit OpenSAFELY Jobs

PR #1386, opened 2021-11-26
Fix dump_db Command

ghickman:
This attempts to fix th dump_db command which was erroring with `pg_dump: error: could not write to output file: Bad file descriptor`.

Running `pg_dump` with `--file` directly in the container works so I'm hoping this works.

The removal of the try/catch should mean sentry flags up failures next time.


iaindillingham (approved):
🤞🏻

PR #1399, opened 2021-11-29
Tweak JS Dep Updates

ghickman:
We currently have Dependabot raising JS dependency bumps throughout the week and @tomodwyer bunching them into a single PR on Monday mornings.

This tries to reduce some of the noise in the build up to Tom's typical routine, and hopefully lowers the number of dependencies he is required to deal with since we don't need regular updates to dev tooling.


iaindillingham (approved):
Very sensible 🙂

PR #1428, opened 2021-12-08
Stop sending internal error events to Sentry

ghickman:
This is being done for every job included in an update, rather than any
job which transitions into this state, blowing through our Sentry quota
in a matter of days.


inglesp (approved):
:stamp:

PR #1430, opened 2021-12-10
Add more NHSD backends

bloodearnest:
Also remove the obfuscated databricks naming, and call it what it is.


justfile
```diff
@@ -1,3 +1,5 @@
+set dotenv-load := true
```
ghickman, line 1: Is this telling just to load a `.env` file?

bloodearnest: Yeah, it doesn't change anything, but just is deprecating loading .env by default, so recent just warns loudly about it, which was annoying

PR #1464, opened 2021-12-20
Add Browser Reloading

ghickman:
Enable automatic reloading of the browser with [django-browser-reload](https://github.com/adamchainz/django-browser-reload) 🎉

### DIFF
--- a/jobserver/urls.py
+++ b/jobserver/urls.py
@@ -272,6 +272,7 @@ urlpatterns = [
     path("status/", Status.as_view(), name="status"),
     path("workspaces/", RedirectView.as_view(url="/")),
     path("__debug__/", include(debug_toolbar.urls)),
+    path("__reload__/", include("django_browser_reload.urls")),
     path(
         "<str:org_slug>/",
         include(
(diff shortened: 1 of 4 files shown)


iaindillingham (approved):

jobserver/urls.py
```diff
@@ -272,6 +272,7 @@
     path("status/", Status.as_view(), name="status"),
     path("workspaces/", RedirectView.as_view(url="/")),
     path("__debug__/", include(debug_toolbar.urls)),
+    path("__reload__/", include("django_browser_reload.urls")),
```
inglesp, line 275: :seagull:

Given that the new library is only listed in the dev requirements, how does this work in production?

ghickman: [It doesn't](https://github.com/opensafely-core/job-server/runs/4584301749?check_suite_focus=true) 😬

I'll stick it in an if block first thing tomorrow.

inglesp: While we're here -- what about DDT?

ghickman: I was just thinking about that this morning, I'll pop browser reload into prod requirements, it saves possible settings.py problems down the line.  They're both turned off without DEBUG on.

PR #1465, opened 2021-12-21
Shift browser-reload into prod requirements

ghickman:
This avoids conditionals in urls.py and settings.py

### DIFF
--- a/requirements.dev.in
+++ b/requirements.dev.in
@@ -5,7 +5,6 @@
 # pip-compile --generate-hashes --output-file=requirements.dev.txt requirements.dev.in
 
 black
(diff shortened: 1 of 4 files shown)


iaindillingham (approved):
Why the shift?

ghickman:
@iaindillingham – the URL config and app name in settings.py's `INSTALLED_APPS` require the dependency to exist in the virutalenv, so in production this [broke the build/deploy](https://github.com/opensafely-core/job-server/actions/runs/1602752879).  I generally avoid conditionals in both urls.py and settings.py where I can because I've had poor experiences tracking down differences between envs with those structures.

PR #1466, opened 2021-12-22
Update CI Caching

ghickman:
This moves to using the setup-* action's caching (which is just the cache action under the hood but with expected defaults), and turns Pip caching back on but in the expected manner.  I've been trialling this on some personal projects without issues yet.

I'd like to trial the Pip caching here for ~1 month before rolling out to other projects.


inglesp (approved):
Looks sensible as far as I can tell!

PR #1474, opened 2021-12-23
Allow Applications to be Soft Deleted

### LINKED ISSUE #1306: Allow Applicants to Delete Applications
We should allow applicants the ability to delete their applications _unsubmitted_ applications.  Of course we don't want to lose any data by accident so lets add an `is_deleted` flag and only soft delete them.  We can hide soft deleted applications from the user view, but probably want to display them in the Staff Area with a strikethrough to show they're deleted.

Spun out from #1305

### LINKED ISSUE #1413: Multiple applications
It looks like there is a possibility of making several applications unintentionally. This certainly happened to me!

An applicant has made 3 three but only this one is valid https://jobs.opensafely.org/staff/applications/c058/

The other two can be deleted: https://jobs.opensafely.org/staff/applications/a17b/ and https://jobs.opensafely.org/staff/applications/df35/
and another one from different applicant https://jobs.opensafely.org/staff/applications/fe12/

@ghickman @iaindillingham should we consider if there application form needs a tweak to prevent people creating multiple forms. When I did it, I think a form was created everytime I clicked the first continue page? 

What's the trigger that actually creates the form?
I think a problem might be that people create forms and peruse them because they want to know what the full form entails, understandably. How can we give them that option without creating multiple forms?

ghickman:
This sets up applications to be deleteable by their authors on the authors list of applications, and in the Staff Area.  Applications are marked as deleted and hidden from the non-staff UI.

Fixes #1306
Fixes #1413


StevenMaude:
I tried this branch out and noticed a few UI features. These may be intentional, but flagging in case they're not:

1. Neither staff nor users appear to be able to undelete a soft-deleted application form.
2. Users can soft-delete approved application forms, which in combination with 1 might be an issue. (For staff, it means the application form has a cursed soft-deletion status. For users, it means they can't easily get back to the application, except by knowing the URL as in 3.)
3. Users can continue to edit their soft-deleted applications, if the URL for pages is known.
4. Staff can approve a soft-deleted application.

staff/views/applications.py
```diff
@@ -145,3 +145,22 @@ def get_queryset(self):
             qs = qs.filter(status=status)
 
         return qs
+
+
+@method_decorator(require_role(CoreDeveloper), name="dispatch")
+class ApplicationRemove(View):
+    def post(self, request, *args, **kwargs):
```
StevenMaude, line 152: bug(possible): Deleting from the Staff Area gives me a 405 (Method Not Allowed) error. Presumably this is because the [template has the Delete button](https://github.com/opensafely-core/job-server/pull/1474/files#diff-3ea9c38b376f0d215a1c2820c05570465bb3dede15279456963b3b1a5c3949a5R78) as a simple `href`.

applications/models.py
```diff
@@ -56,7 +63,14 @@ class Meta:
                     | ~Q(approved_at=None) & ~Q(approved_by=None)
                 ),
                 name="%(app_label)s_%(class)s_both_approved_at_and_approved_by_set",
-            )
+            ),
+            models.CheckConstraint(
+                check=(
+                    Q(deleted_at=None) & Q(deleted_by=None)
+                    | ~Q(deleted_at=None) & ~Q(deleted_by=None)
+                ),
```
StevenMaude, line 71: suggestion: (via @inglesp) it might help readers be clear on the order of evaluation by using parentheses around the Boolean expressions.

Although looking again, maybe the separation on two lines is fine.

StevenMaude (changes requested):
Overall, this looks fine :+1:  It was quite a large diff, but fairly easy to follow for a Django novice (with some guidance from @inglesp).

There's a possible bug to look at, and some UI considerations that could be worth checking the design against.

ghickman:
Thanks for flagging those @StevenMaude, I hadn't considered existing views at all!  I've updated the ones you mentioned and caught another couple too.

All the User and Staff views should now take the viewer back to the appropriate list page if they try to perform an action against a deleted Application, including just viewing it.  Both list pages have grown restore buttons so it's quick and easy to restore an application.

I've opted not to included a confirmation page since restoring is so easy now.

StevenMaude (approved):
:+1: this is definitely improved now, I can delete and restore.

It's slightly quirky that deleted applications are lying around; currently, it's more that they are marked as "not for submission". It might be later desirable to have a full deletion option:

* either automatically, after some days,
* or manually

(Though that has the risk of someone deleting a submission permanently themselves, by accident.)

So this is an improvement over the current behaviour as is, and approving.

PR #1479, opened 2021-12-23
Replace Missing ReleaseFiles

### LINKED ISSUE #1202: Inflate Release Files
When copying the database from production down to a local developer machine we don't take release files with it.

However having all Releases and Snapshot related views be broken by this isn't very useful.

Let's add a script to "inflate" all those paths with dummy content.  We know the filetype from the file path in the database so we can pick from various content templates, potentially adding some random data.

ghickman:
This builds dummy files for ReleaseFiles missing their on-disk files and is expected to be run for local dev.  I've only tackled the easy subset of file types so we didn't have to pull in pillow to dev requirements for this to work.  It still covers a majority of ReleaseFiles.

Fixes #1202

### DIFF
--- a/jobserver/management/commands/inflate_release_files.py
+++ b/jobserver/management/commands/inflate_release_files.py
@@ -1,0 +1,142 @@
+"""
+Create files on disk for ReleaseFiles missing them
+
+When we pull database dumps from production they won't bring down the files
+related to ReleaseFiles.  However this makes working with the file viewer
+harder since all links are effectively broken.  This command will generate
+files for a variety of extensions with semi-useful junk data, eg a CSV file
+with rows and columns.
+
+Each make_* function is given a ReleaseFile instance and is free to
+implement the creation of files as it sees fit.
+"""
+
+import csv
+import json
+
+from django.core.management.base import BaseCommand
+
+from ...models import ReleaseFile
+
+
+def make_csv(release_file):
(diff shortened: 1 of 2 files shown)


jobserver/management/commands/inflate_release_files.py
```diff
@@ -0,0 +1,129 @@
...
+from ...models import ReleaseFile
+
+
+def make_csv(release_file):
+    with release_file.absolute_path().open("w") as f:
+        columns = ["id", "created_at", "workspace_name", "release_id"]
+        writer = csv.DictWriter(f, columns)
+
+        writer.writeheader()
+        for _ in range(100):
```
StevenMaude, line 15: suggestion: The `100` that appears a few times feels a little magic. I assume it's somewhat arbitrary, but might be helpful to explain it.

jobserver/management/commands/inflate_release_files.py
```diff
@@ -0,0 +1,129 @@
+import csv
+import json
+
+from django.core.management.base import BaseCommand
+
+from ...models import ReleaseFile
+
+
```
StevenMaude, line 8: suggestion: Maybe a module-level docstring is worth adding to explain what this does? Or some explanation in the developer docs somewhere?

StevenMaude (approved):

PR #1481, opened 2021-12-24
Add Create Project to Staff Area

### LINKED ISSUE #1085: Add Project in Staff Area
When onboarding new researchers we don't always have a Project to add them to, so we should be able to add a Project from:

* the OrgDetail page (for that Org)
* the ProjectList page
* the UserSetProjects page

The latter two will need some UI to pick the Org the Project is part of.

ghickman:
This adds a create Project form, linked from the list page of Projects.  I've opted to not support a UserSetProject page since we don't seem to need that route currently.

Fixes #1085

### DIFF
--- a/staff/templates/staff/project_list.html
+++ b/staff/templates/staff/project_list.html
@@ -24,6 +24,12 @@
 <div class="jumbotron jumbotron-fluid jumbotron--danger pt-md-2">
   <div class="container">
     <h1 class="display-4">Projects</h1>
+
+    <div class="d-flex">
+      <a class="btn btn-primary" href="{% url 'staff:project-create' %}">
+        Create a Project
+      </a>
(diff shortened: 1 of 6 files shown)


staff/templates/staff/project_list.html
```diff
@@ -24,6 +24,12 @@
 <div class="jumbotron jumbotron-fluid jumbotron--danger pt-md-2">
   <div class="container">
     <h1 class="display-4">Projects</h1>
+
+    <div class="d-flex">
+      <a class="btn btn-primary" href="{% url 'staff:project-create' %}">
+        Create an Project
```
StevenMaude, line 30: nit: "an" should be "a"

StevenMaude (approved):

PR #1484, opened 2021-12-24
Upgrade to Django 4.0

ghickman:
This upgrades us to Django 4, fixing the URL tests to be more correct along the way.

### DIFF
--- a/jobserver/settings.py
+++ b/jobserver/settings.py
@@ -146,7 +146,6 @@ SESSION_SERIALIZER = "django.contrib.sessions.serializers.JSONSerializer"
 LANGUAGE_CODE = "en-us"
 TIME_ZONE = "UTC"
 USE_I18N = True
-USE_L10N = True
 USE_TZ = True
 
 
--- a/requirements.prod.txt
+++ b/requirements.prod.txt
@@ -144,9 +144,9 @@ dj-email-url==1.0.2 \
     --hash=sha256:15148141c6ef123636e4ca3663e95231ed94ca5ed267e91977e5a4397be8b34c \
     --hash=sha256:838fd4ded9deba53ae757debef431e25fa7fca31d3948b3c4808ccdc84fab2b7
     # via environs
-django==3.2.10 \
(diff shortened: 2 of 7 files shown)


jobserver/settings.py
```diff
@@ -146,7 +146,6 @@
 LANGUAGE_CODE = "en-us"
 TIME_ZONE = "UTC"
 USE_I18N = True
-USE_L10N = True
```
StevenMaude, line 149: note: This is now a default, so no change in behaviour.

tests/unit/jobserver/test_urls.py
```diff
@@ -145,9 +145,17 @@ def test_url_redirects(client, url, redirect):
 )
 def test_url_resolution(url, view):
     """Test each URL resolves to the expected view function or class"""
-    resolved_view = resolve(url).func
+    match = resolve(url)
 
+    # get the actual class for a CBV.  View.as_view() constructs a new wrapper
+    # function to be the callable which url resolution calls when a view is
+    # accessed.  We don't want that callalbe at this point in the test because
```
StevenMaude, line 152: nit: `callalbe` typo

requirements.prod.txt
```diff
@@ -402,6 +402,10 @@ python3-openid==3.2.0 \
     --hash=sha256:33fbf6928f401e0b790151ed2b5290b02545e8775f982485205a066f874aaeaf \
     --hash=sha256:6626f771e0417486701e0b4daff762e7212e820ca5b29fcc0d05f6f8736dfa6b
     # via social-auth-core
+pytz==2021.3 \
```
StevenMaude, line 405: issue: aeb47d24203cff9bd5f9911bb392b02ffc2429dc states:

> Upgrade DRF 3.12.4 -> 3.13.1 to drop pytz dependency

But `pytz` is still here, so either the commit message is not accurate, or the requirements need rebuilding.

StevenMaude (approved):
summary: :+1: there's a couple of small comments to look at, but overall this is good, so approving.

requirements.prod.txt
ghickman: Arg, they pulled it into their requirements rather than removing the dep, thanks 👍

PR #1520, opened 2022-01-11
Combine Dev Tooling Checks

ghickman:
This matches our general tooling setup across the group and removes some unnecessary time installing dependencies in CI.


CarolineMorton (approved):
Nice one

PR #1529, opened 2022-01-17
Switch to opensafely-pipeline for Pipeline Config Parsing

ghickman:
This switches us from pyyaml to [opensafely-pipeline](https://github.com/opensafely-core/pipeline/) for parsing pipeline config files (typically `project.yaml`).  This is pretty much switching `yaml.safe_load()` to `load_pipeline`, but there's scope later on to move some of the extra parsing into that library if we think it's useful elsewhere.

~Note: Keeping this in draft until there's a tag in the pipeline repo we can point at to install from.~


jobserver/pipeline_config.py
```diff
@@ -1,28 +1,24 @@
...
         # get latest status for this action from the lookup table
         status = status_lut.get(action, "-")
 
-        yield {"name": action, "needs": sorted(needs), "status": status}
+        yield {"name": action, "needs": needs, "status": status}
+
+    # ensure there's always a run_all action
+    if "run_all" not in config.actions:
+        all_actions = list(config.actions.keys())
+        yield {"name": "run_all", "needs": all_actions, "status": "-"}
```
iaindillingham, line 21: I'm sure this has been discussed elsewhere, but why must there always be a `run_all`?

iaindillingham (approved):
Just one question, about `run_all`.

jobserver/pipeline_config.py
ghickman: This is a funny quirk of OpenSAFELY history!

Originally we asked researchers to manually add a `run_all` action to their `project.yaml`s, there was nothing special about this, they simply had to list all other actions in the `needs` section and it ran them all.  Beautifully hacky and it worked quite well!

Of course this fell apart with the larger `project.yaml`s when there were so many dependencies that human error started to catch people out and it was a real chore to create that action.  Around that same time job-server was being set up and we added the create job page (`JobRequestCreate`) which renders the actions in a `project.yaml`.  This function drives the form construction and validation which deals with "what actions can a user run" as defined in the `project.yaml`, and we added a "fake" `run_all` action for everyone to avoid the manual hack.

job-runner now treats `run_all` as a special case since it might not exist in the `project.yaml` either.

Maybe one day we'll derive this from the create job form page with UI controls and ditch the magic string, but backwards compatibility of studies means we're tied to at least support it in job-runner forever more.

PR #1530, opened 2022-01-17
Order backends on Status page using ORM query instead of dictsort

ghickman:
dictsort's resolution logic was restricted in Django 4.0.1, dropping
support for indexing on dictionaries, to avoid a potentional information
disclosure bug.

https://docs.djangoproject.com/en/4.0/releases/4.0.1/#cve-2021-45116-potential-information-disclosure-in-dictsort-template-filter

### DIFF
--- a/jobserver/views/status.py
+++ b/jobserver/views/status.py
@@ -1,4 +1,5 @@
 from django.db.models import Count
+from django.db.models.functions import Lower
 from django.template.response import TemplateResponse
 from django.views.generic import View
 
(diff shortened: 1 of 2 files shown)


CarolineMorton (approved):
Thanks for linking out to that Django documentation. Very helpful.

jobserver/views/status.py
```diff
@@ -43,6 +44,6 @@ def get_stats(backend):
                 "show_warning": show_warning(last_seen),
             }
 
-        backends = Backend.objects.all()
+        backends = Backend.objects.order_by(Lower("name"))
```
CarolineMorton, line 47: Makes sense. We are applying specific logic via model functions here I think rather than generic ptyhon

ghickman: We'd have to sort the QuerySet in Python without this too, something like `sorted(backends, key=lambda b: b.name.lower())`.  That would convert it to a Python iterable so we'd lose any QuerySet functionality we might want to use.  We don't _need_ that here but always useful to be able to do something simple like this without ejecting from a QuerySet.

PR #1533, opened 2022-01-17
fix: set EMIS level_4_url to be localhost (same as TPP)

bloodearnest:
Also, drive by fix for just file not erroring properly.

Also disabled the Download button on level 4 output viewer pages. It doesn't work, but it should be obvious that it doesn't work. I've opted to disable rather than remove, as I like UIs that don't disappear or change layout under you, but I'm not sure there's any styling to support `a.btn[disabled]`, at least I couldn't find any.

### DIFF
--- a/jobserver/templates/workspace_backend_files.html
+++ b/jobserver/templates/workspace_backend_files.html
@@ -54,7 +54,7 @@
       You can see these outputs because you have access to the Level 4 environment on the
       {{ backend.name }} backend and have been assigned the ProjectDeveloper role.
     </p>
-    <a class="btn btn btn-success" href="{{ release.get_download_url }}">
+    <a class="btn disabled" aria-disabled="true" title="You cannot download files from the level 4, you must go through the release process first">
       Download
     </a>
   </div>
--- a/jobserver/backends.py
+++ b/jobserver/backends.py
@@ -9,7 +9,7 @@ backends = [
         "name": "EMIS",
         "slug": "emis",
         "is_active": True,
-        "level_4_url": "",
+        "level_4_url": "https://localhost:8001",
     },
     {
         "id": 2,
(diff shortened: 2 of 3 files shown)


jobserver/templates/workspace_backend_files.html
```diff
@@ -54,7 +54,7 @@ <h1>Level 4 outputs from {{ workspace.name }}</h1>
       You can see these outputs because you have access to the Level 4 environment on the
       {{ backend.name }} backend and have been assigned the ProjectDeveloper role.
     </p>
-    <a class="btn btn btn-success" href="{{ release.get_download_url }}">
+    <a class="btn" disabled="true" aria-disabled="true">
```
ghickman, line 57: This template is only used on the L4 outputs page, so this button never be enabled, maybe we should remove and put some text in its place to explain how to release outputs?

Disabling `<a>`s is a pain, you need to set the `disabled` class instead: https://getbootstrap.com/docs/4.0/components/buttons/#disabled-state

ghickman (changes requested):

jobserver/templates/workspace_backend_files.html
bloodearnest: So, in general, in the simple cases I prefer disabled UI components rather than hidden ones, as it provides more information, and keeps the layout stable. A disabled button says to me "we've disabled the download on purpose", whereas the absence doesn't say anything either way.

But I don't want to introduce new UI paradigms here

ghickman: Oh, you want to keep it matching the other SPA pages, gotcha!  Yep, lets do that.  Maybe add a `<p>` above/below just to explain that you can't download from a L4 page, outputs must be released and link to the releasing docs?

bloodearnest: Have added a title which should show up as a tooltip?

ghickman (approved):

PR #1542, opened 2022-01-19
Update Actions with Dependabot

ghickman:
Docs: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,3 +18,8 @@ updates:
     reviewers:
       - "tomodwyer"
     open-pull-requests-limit: 20
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"


.github/dependabot.yml
```diff
@@ -18,3 +18,8 @@ updates:
     reviewers:
       - "tomodwyer"
     open-pull-requests-limit: 20
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
```
madwort, line 25: why not `daily`, if there are potential security implications? (nb. I'm assuming this is going to get copy-pastaed into every other repo we have)

ghickman: This started as "I'm fed up of dealing with dependabot updates most days of the week", but I don't expect to see many updates to actions from dependabot _in this repo_. The core actions are all use rolling tags, eg `setup-python` gained support for [caching dependencies in `v2.3.0`](https://github.com/actions/setup-python/releases/tag/v2.3.0) but we're using `v2` which is a pointer to that.

madwort: just reviewed the current versions, and yeah, I see what you mean - ok then!

madwort (approved):

PR #1562, opened 2022-01-27
Run Repo Requests in Parallel on ProjectDetail Pages

ghickman:
On ProjectDetail pages with several repos the lookup for private/public
status was noticably slowing the page down.  This switches to
ThreadPoolExecutor to run through them in parallel (mostly).

### DIFF
--- a/jobserver/views/projects.py
+++ b/jobserver/views/projects.py
@@ -1,3 +1,6 @@
+import concurrent
+import operator
+
 import requests
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
@@ -17,6 +20,12 @@ from ..emails import send_project_invite_email
 from ..forms import ProjectInvitationForm, ProjectMembershipForm
 from ..github import get_repo_is_private
 from ..models import Org, Project, ProjectInvitation, ProjectMembership, Snapshot, User
+
+
+# Create a global threadpool for getting repos.  This lets us have a single
+# pool across all requests and saves the overhead from setting up threads on
+# each request.
+repo_thread_pool = concurrent.futures.ThreadPoolExecutor(thread_name_prefix="get_repo")
 
 
 @method_decorator(login_required, name="dispatch")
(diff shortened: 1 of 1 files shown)


jobserver/views/projects.py
```diff
@@ -161,21 +166,26 @@ def get_outputs(self, workspaces):
...
                 is_private = None
 
-            yield {
+            return {
                 "name": f.path.segments[1],
                 "url": url,
                 "is_private": is_private,
             }
 
+        with concurrent.futures.ThreadPoolExecutor() as executor:
```
bloodearnest, line 184: I'd think this might be better as a global instance. Otherwise it will have to create the all the threads every request.

I would also recommend setting `thread_name_prefix="get_repo"` or similar, as when you have threading issues in the future, you'll wish you had.

It might be a good idea to set `max_workers` as well, perhaps, but it defaults to `min(32, cpu + 4)`, as long as its not unbounded I think thats fine.

bloodearnest: It also applies the max_workers across requests. Currently, it could be 32 threads per concurrent request, in theory!

ghickman: Fixed and rebased in.

ghickman:
@bloodearnest – is this good to go now?

jobserver/views/projects.py
```diff
@@ -161,21 +172,25 @@ def get_outputs(self, workspaces):
...
 
-            yield {
+            return {
                 "name": f.path.segments[1],
                 "url": url,
                 "is_private": is_private,
             }
 
+        futures = [repo_thread_pool.submit(get_repo, url=url) for url in repo_urls]
+        for future in concurrent.futures.as_completed(futures):
```
bloodearnest, line 191: So, I think we can slightly simplify this using [Executor.map()](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map). We can set an explicit timeout for *all* requests too, which is usually a good idea when doing concurrency.

So, maybe replace the last 3 lines with:

```python
# use the threadpool to parallelise the repo requests
yield from repo_thread_pool.map(get_repo, repo_urls, timeout=30)
```

bloodearnest (approved):
+1 with a suggestion

jobserver/views/projects.py
ghickman: Much nicer, and TIL!

PR #1566, opened 2022-01-28
Rework our use of Org.github_orgs

### LINKED ISSUE #1554: Remove Use of Org.github_orgs in JobRequestCreate
We should be able to get the repo owner from the Workspace (available the `workspace.repo_owner` property) rather than relying on a populated Org.github_orgs.

### LINKED ISSUE #1555: Hard Fail on Missing Org.github_orgs in WorkspaceCreate
We use Org.github_orgs to pull the list of repos (and their branches) from GitHub on page load WorkspaceCreate.  If a User's org has an empty `github_orgs` we should just return a blank page with a short message as to why it's happened.

ghickman:
This is a mixture of changes revolving around making `Org.github_orgs` easier to work with and more useful to the project.

It removes the use of it entirely from JobRequestCreate (#1554) since we can get the same information from the Workspace.

WorkspaceCreate now shows a different template when we can't find any GitHub orgs for the user (#1555).

Org.github_orgs has been switched from a JSONField to an ArrayField to better map that it's a list strings.

And finally the Staff Area has grown the ability to add and remove GitHub orgs on the OrgDetail page.  This was also a test of using HTMX for basic enhancements, it went quite well!

Fixes #1554 
Fixes #1555


staff/templates/staff/org_detail.html
```diff
@@ -126,6 +154,7 @@ <h2 class="h3">Projects</h2>
 {% endblock staff_content %}
 
 {% block extra_js %}
+<script type="text/javascript" src="{% static 'vendor/htmx.js' %}"></script>
```
tomodwyer, line 157: Could we use the minified version here instead?

```suggestion
<script type="text/javascript" src="{% static 'vendor/htmx.min.js' %}"></script>
```

vite.config.js
```diff
@@ -40,6 +40,10 @@ const config = {
           src: "./node_modules/bootstrap/dist/js/bootstrap.bundle.min.*",
           dest: "./assets/dist/vendor",
         },
+        {
+          src: "./node_modules/htmx.org/dist/htmx.js",
```
tomodwyer, line 44: Could we use the minified version here instead?

```suggestion
          src: "./node_modules/htmx.org/dist/htmx.min.js",
```

staff/templates/staff/org_detail.html
```diff
@@ -60,6 +60,34 @@ <h1 class="display-4">{{ org.name }}</h1>
...
+            </form>
+          </div>
+        </div>
+        {% endfor %}
+        <div class="list-group-item list-group-item-success">
+          <span
+            hx-get="{% url 'staff:org-add-github-org' slug=org.slug %}"
+            hx-swap="outerHTML">
+            Add a GitHub organisation
+          </span>
```
tomodwyer, line 87:
```suggestion
          <button
            class="btn ml-n2"
            hx-get="{% url 'staff:org-add-github-org' slug=org.slug %}"
            hx-swap="outerHTML"
            type="button">
            Add a GitHub organisation
          </button>
```

Just on tweak for semantic HTML purposes, otherwise I'm happy ✅

tomodwyer (approved):

PR #1580, opened 2022-02-04
Event log styles

tomodwyer:
WIP 🚧

### DIFF
--- a/jobserver/templates/_job_requests.html
+++ b/jobserver/templates/_job_requests.html
@@ -1,92 +1,119 @@
 {% load humanize %}
 {% load runtime %}
 
-<div class="container-fluid">
-  {% include "_job_request_pagination.html" %}
-
-  <div class="job-requests">
-    <div class="d-flex">
-      <div class="status"></div>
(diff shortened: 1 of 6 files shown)


jobserver/templates/_job_requests.html
```diff
@@ -11,21 +11,26 @@
...
         <th>View jobs</th>
       </tr>
     </thead>
     <tbody>
       {% for group in page_obj %}
       <tr>
-        <td>
-          <div class="status status-icon {{ group.status }}"></div>
-          <span class="sr-only">{{ group.status }}</span>
+        <td style="width: 0">
```
ghickman, line 22: Does this hide the child `<div>`?

tomodwyer: Because tables are weird, it makes the `tbody > td` as small as possible given the content, which means the "Status" heading sets the width of the column.

ghickman: ah, "tables are weird" I can get behind!

PR #1585, opened 2022-02-08
Link to workspace branch instead of just repo

### LINKED ISSUE #1583: Workspace page should link to branch, not just repo.
Frequently, when looking at a workspace, I want to go to its repo.

The current links take you to the repo, but not the branch the workspace is bound to. This has caused confusion multiple times.

It would be nice if the links took you the branch on github.

ghickman:
Fixes #1583


bloodearnest (approved):
Can we update the body text here too?

https://github.com/opensafely-core/job-server/blob/main/jobserver/templates/workspace_detail.html#L180

PR #1590, opened 2022-02-09
feat: notify slack for new-style releases

bloodearnest:
Slight refactor of the slack stuff, and an attempt to provide useful
links in the message, which we can do with new style releases, as we
have things to link to!


tests/conftest.py
```diff
@@ -77,3 +80,23 @@ def incomplete_application():
...
+def slack_messages(monkeypatch, enable_network):
+    messages = []
+
+    actual_post = services.slack.post
+
+    def post(text, channel):
+        messages.append((text, channel))
+
+        if slack_token and slack_test_channel:  # pragma: no cover
+            actual_post(text, slack_test_channel)
```
ghickman, line 99: Did you see if pytest-network would let us enable the network for just this block? (assumption: this is what `enable_network` was added for?)

bloodearnest: yep, although the pytest-network pypi README is wrong:
https://pypi.org/project/pytest-network/ vs
https://github.com/best-doctor/pytest_network

PR #1592, opened 2022-02-10
Try splitting up dev and prod dependency update frequency

ghickman:


.github/dependabot.yml
```diff
@@ -5,9 +5,15 @@ updates:
...
+      - dependency-type: "production"
+
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    allow:
+      - dependency-type: "development"
 
   - package-ecosystem: "npm"
```
bloodearnest, line 18: Can we deprioritise dev for npm too? I feel like that would be a big win

benbc (approved):

PR #1612, opened 2022-02-15
Approve and auto-merge Dependabot's Python PRs

ghickman:
Following [these](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/automating-dependabot-with-github-actions#approve-a-pull-request) [two](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/automating-dependabot-with-github-actions#enable-auto-merge-on-a-pull-request) sections in the docs the `gh` commands should now approve and toggle auto-merge on PRs which pass the `tests` check (a requirement for merging in this repo).

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -1,6 +1,6 @@
-name: Explore dependabot metadata
+name: Auto-merge Dependabot Python updates
 
-on: pull_request_target
+on: pull_request
 
 permissions:
   pull-requests: write
@@ -11,16 +11,12 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.actor == 'dependabot[bot]' && contains(github.event.pull_request.labels.*.name, 'python') }}
     steps:
-      - name: Dependabot metadata
-        id: dependabot-metadata
-        uses: dependabot/fetch-metadata@v1.1.1
-        with:
-          github-token: "${{ secrets.GITHUB_TOKEN }}"
-
-      - name: Get metadata for Dependabot PRs
+      - name: Enable auto-merge for Dependabot PRs
         run: |
-          echo ${{ toJSON(steps.dependabot-metadata) }}
-          echo ${{ toJSON(github.event) }}
+          set -eu
+
+          gh pr review --approve "$PR_URL"
+          gh pr merge --auto --merge "$PR_URL"
         env:
           PR_URL: ${{ github.event.pull_request.html_url }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


benbc (commented):
Do we still need the `dependabot-metadata` step? Do we still need the `contents` permissions?

This looks to me as though it's going to automatically merge all the Dependabot PRs. I don't remember that we agreed that as an option in the team catch-up last week. (It could be the right thing to do, but I think we should agree it as a team.)

If we _are_ auto-merging, can we revert the earlier change and check the dev dependencies more frequently again?

ghickman:
> Do we still need the dependabot-metadata step?

Good catch, we're not using the conditional on the subsequent job so I've removed that (and rebased).

> Do we still need the contents permissions?

I believe so.  The `permissions` key here is actually reducing what the workflow is able to do by default with `secrets.GITHUB_TOKEN`.  From my understanding of the permissions ([sort of useful docs](https://docs.github.com/en/rest/reference/permissions-required-for-github-apps#permission-on-contents)) we need `contents: write` to talk to the repo's PR merge endpoint, which is likely how the `gh pr merge` command is doing things.

> This looks to me as though it's going to automatically merge all the Dependabot PRs. I don't remember that we agreed that as an option in the team catch-up last week. (It could be the right thing to do, but I think we should agree it as a team.)

We haven't agreed it as a team, but I want to see how it works in practice before I give options and suggestions to the team, so my plan is to run this on job-server for a few days since it has a comprehensive test suite (for the Python code) and then feed back to the team once I've seen it in practice.

StevenMaude:
A couple of thoughts:

1. As far as I remember, Dependabot doesn't bump the versions in the `.pre-commit-config.yaml` — we've been upgrading those manually. What process would we use to keep these in sync?
2. If we do decide to auto-merge, I'd advocate having the PR sit open for some kind of delay before merging

    From recent memory, this feels more like a common worry with npm, more than PyPI. But the risk of supply chain compromise is still there regardless of software platform:

    * a package owner's GitHub — where some owners auto-publish builds via Actions — can be compromised
    * [a package owner's package publication account can be compromised](https://portswigger.net/daily-swig/popular-npm-package-ua-parser-js-poisoned-with-cryptomining-password-stealing-malware)
    * [package owners can go rogue](https://research.swtch.com/npm-colors)
    * [a rogue maintainer can take over a package](https://blog.npmjs.org/post/180565383195/details-about-the-event-stream-incident.html)
     
    Dependabot [removed auto-merge](https://github.com/dependabot/dependabot-core/issues/2243) for this reason, and hasn't reintroduced it yet.

    If there's a critical security vulnerability in a commonly used package, we can always manually merge early.

.github/workflows/auto-merge.yml
```diff
@@ -1,4 +1,4 @@
-name: Explore dependabot metadata
+name: Auto-merge Dependabot Python updates
 
 on: pull_request_target
```
StevenMaude, line 3: question: What's the need for `pull_request_target` over `pull_request`? The [example in GitHub's docs](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/automating-dependabot-with-github-actions#enable-auto-merge-on-a-pull-request) uses `pull_request`.

ghickman:
@StevenMaude

> As far as I remember, Dependabot doesn't bump the versions in the .pre-commit-config.yaml — we've been upgrading those manually. What process would we use to keep these in sync?

We'll have to come up with a process.  As I said to Ben above, this is to test how auto-merging will work in practice, it's not the addition of auto-merging as a team decision.  I intend to trial it and report back to the team with options.

> I'd advocate having the PR sit open for some kind of delay before merging

That would be ideal but I don't see a way to delay actions happening.  They're event based so I think the action would have to either sit in a pending state or pause itself (eating up worker time).  I'm happy to keep an eye out for ways to do this but it's out of scope for this PR which is exploratory.

.github/workflows/auto-merge.yml
ghickman: Just a mistake left in from testing, fixed and rebased.

benbc (approved):

PR #1634, opened 2022-03-01
Expose Published Files to Unauthenticated User

ghickman:
ReleaseFiles are currently served to the SPA (and directly when opening in a new tab) by the `ReleaseFileAPI` view.  This checks a file can be viewed by checking details about the requesting user and the workspace.  Unfortunately this doesn't let us check if a file has been published, or handle the user being anonymous.

To work around this situation this PR:
* removes the default BasicAuthentication auth class from DRF so unauthenticated users can get through to views, since we handle our own permissions inside each view.
* Adds a new `PublishedSnapshotFile` view under `/<org>/<project>/<workspace>/published/<file_id>/` from which we can serve a file.

This lets us serve publicly viewable files from a completely separate URL, meaning we could, if necessary, apply caching to those URLs should we be slashdotted.

There's also a drive-by optimisation since my local copy of the SRO measures Ouputs view was doing ~2100 queries and taking several seconds to load.

### DIFF
--- a/jobserver/models/outputs.py
+++ b/jobserver/models/outputs.py
@@ -158,6 +158,21 @@ class ReleaseFile(models.Model):
 
     def get_api_url(self):
         """The API url that will serve up this file."""
+        if getattr(self, "is_published", False):
+            # This is a hack so the view for viewing a published ReleaseFile
+            # can pass down the knowledge that a ReleaseFile has been
+            # published, avoid O(N) queries on views which aren't expecting to
+            # have to add `.select_related("snapshot")` to a QuerySet.
+            return reverse(
+                "published-file",
+                kwargs={
+                    "org_slug": self.workspace.project.org.slug,
+                    "project_slug": self.workspace.project.slug,
+                    "workspace_slug": self.workspace.name,
+                    "file_id": self.id,
+                },
+            )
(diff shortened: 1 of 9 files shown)


jobserver/models/outputs.py
```diff
@@ -158,6 +158,21 @@ def get_absolute_url(self):
...
+            return reverse(
+                "published-file",
+                kwargs={
+                    "org_slug": self.workspace.project.org.slug,
+                    "project_slug": self.workspace.project.slug,
+                    "workspace_slug": self.workspace.name,
+                    "file_id": self.id,
+                },
+            )
+
```
bloodearnest, line 175: Hmm, am not sure about this.

This means, if you view the *release* for this file, it could use the public url for some files (that are part of a published snapshot) and the private url for others.

That seems a bit odd, but I guess we don't really care, as it's all hidden away

bloodearnest (approved):

jobserver/models/outputs.py
ghickman: Yeah, I didn't really love how this panned out but because we just pass ReleaseFile instances down to `generate_index` there wasn't an obvious path.

PR #1635, opened 2022-03-02
fix: do not ensure_backends on deployment

bloodearnest:
This was rewriting the db config in production.

ensure_backends command still works locally if needed in dev.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -5,7 +5,6 @@ set -euo pipefail
 ./manage.py check --deploy
 ./manage.py migrate
 ./manage.py ensure_admins
-./manage.py ensure_backends
 ./manage.py collectstatic --no-input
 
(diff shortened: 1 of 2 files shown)


ghickman (changes requested):
I'm not sure we should be keeping `ensure_backend` around as a dev tool, it's essentially a bespoke version of `loaddata`.  Could you replace it with a JSON fixture file and document that one can load that with `loaddata`?

ghickman (approved):

bloodearnest:
> I'm not sure we should be keeping `ensure_backend` around as a dev tool, it's essentially a bespoke version of `loaddata`. Could you replace it with a JSON fixture file and document that one can load that with `loaddata`?

Makes sense, I will do in follow up PR

PR #1637, opened 2022-03-02
chore: replace ensure_backends command with fixture

bloodearnest:
Added `just load-dev-data` as a wrapper


justfile
```diff
@@ -155,3 +155,6 @@ rebuild-static:
     npm ci
     npm run build
     $BIN/python manage.py collectstatic --no-input
+
+load-dev-data:
```
ghickman, line 159: This will need devenv as a dependency?

ghickman (changes requested):

ghickman (approved):

PR #1644, opened 2022-03-09
Add our backend access policy to the staff area

ghickman:
Describe to Staff Area users how they should pick which backends to give to a user they are setting up.

### DIFF
--- a/staff/templates/staff/user_detail.html
+++ b/staff/templates/staff/user_detail.html
@@ -63,6 +63,14 @@
             A list of backends {{ user.username }} has access to.
           </p>
 
+          <p>
+            Once you've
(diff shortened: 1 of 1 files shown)


amirmehrkar:
1. could the link be https://docs.google.com/spreadsheets/d/1rmFdaUz2BFedLL9cgVk-PDWINeLAeupR0TYoS8Btlh0/edit#gid=0&range=Q1 (so it goes to the right column?)

2. confirmed the user has signed a DAA **and their application has been approved** grant them access to only the backends their study requires. You may need to check with Amir about their application if it's status is not "Approved Fully".

@ghickman

ghickman:
@amirmehrkar – done for both, could you approve the PR so I can merge it please? 🙂

amirmehrkar:
youve not mentioned contacting me - but that's not essential.

amirmehrkar (approved):

PR #1646, opened 2022-03-09
Cancelling jobs from a job request

### LINKED ISSUE #1629: Explain what cancelling a request/job does
I submit a request, containing actions A, B, C, and C depends on B depends on A. 

Job for A completes successfully. B is still running, and we decide we want to cancel B and change the script. But we want to keep the completed outputs from A. 

If we cancel B (big red button on the job page at `.../<project>/<workspace>/<request>/<job-id>`), then B is cancelled, C is cancelled (because it depends on B), but the status of A remains "success".

If we cancel the entire request (big red button on the request page at `.../<project>/<workspace>/<request>`), then A, B, and C are all cancelled. 

Can we:

* check that the above is true
* explain it more concisely and add the explanation somewhere near those big red buttons on each page

rebkwok:
Cancelling a job request previously added all the jobs to the request's cancelled_actions.  This resulted in jobs that had already completed (both failed and succeeded jobs) being marked as cancelled too.  This PR filters the request's jobs to cancel only those that have not yet completed.

Also adds some explanatory text to the individual job Cancel button.

Fixes #1629

### DIFF
--- a/jobserver/templates/job_detail.html
+++ b/jobserver/templates/job_detail.html
@@ -84,7 +84,7 @@
         </button>
       </form>
       <p class="text-muted">Cancelling this job will cancel both the current job and any subsequent jobs that depend on it.</p>
-      {% endif %}
+    {% endif %}
   </div>
 </div>
 {% endblock jumbotron %}
--- a/jobserver/templates/job_detail.html
+++ b/jobserver/templates/job_detail.html
@@ -83,7 +83,8 @@
           Cancel
         </button>
       </form>
-    {% endif %}
+      <p class="text-muted">Cancelling this job will cancel both the current job and any subsequent jobs that depend on it.</p>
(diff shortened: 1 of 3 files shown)


jobserver/templates/job_detail.html
```diff
@@ -83,7 +83,8 @@ <h1><code class="text-body">{{ job.action }}</code></h1>
           Cancel
         </button>
       </form>
-    {% endif %}
+      <p class="text-muted">Cancelling this job will cancel both the current job and any subsequent jobs that depend on it.</p>
+      {% endif %}
```
ghickman, line 87: Could you dedent this 2 spaces to match the opening block please?

ghickman (approved):
Just one small tweak, thanks for fixing this!

PR #1661, opened 2022-03-15
Check Code Changes to get Language Type

ghickman:
The current version of this workflow hasn't been triggering, from which I can assume that labels are applied after the PR is raised, so they're not present in the event payload we're receiving.

This looks at the PR branch to look for Python-specific requirements changes done in a Dependabot PR.  Mirroring how we look for tag updates in Data Builder it sets an output which we check in the subsequent step so non-Python updates (from Dependabot) should be ignored _without_ a failing Job.

---

This extra set of steps probably isn't required for most of our projects, which can rely on tests passing alone.  However, for job-server don't want to automatically merge Javascript changes without testing them out manually.  We're confident-enoiugh that our tests will catch problems with the Python dependencies.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -1,4 +1,4 @@
-name: Auto-merge Dependabot Python updates
+name: Approve and enable auto-merge for Dependabot updates
 
 on: pull_request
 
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-merge.yml
```diff
@@ -9,8 +9,27 @@ permissions:
 jobs:
   dependabot:
     runs-on: ubuntu-latest
-    if: ${{ github.actor == 'dependabot[bot]' && contains(github.event.pull_request.labels.*.name, 'python') }}
+    if: ${{ github.actor == 'dependabot[bot]' }}
     steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1.1.1
```
bloodearnest, line 16: Hmm, not sure if this counts as an external action or not, so should be pinned via hash or not.

dependabot is a github thing, so I think that counts as internal?

.github/workflows/auto-merge.yml
```diff
@@ -9,8 +9,27 @@ permissions:
...
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Checkout branch
+        uses: actions/checkout@v2
+
+      - name: Check we're dealing with Python updates
+        run: |
+          if [[ $(git diff --name-only HEAD~1) == *"requirements"* ]]; then
```
bloodearnest, line 25: This seems a bit brittle to me.

I wonder if we can use `gh pr view --json labels -q .labels` ? gh cli is preinstalled, authed, and has the current PR context.

bloodearnest: Because we're fetching them via the api, the fact they are not in the event data shouldn't matter?

ghickman: Definitely up for giving it a go.  My assumption is dependabot applies them after the PR is created because it can't set them at PR creation time which is why we don't get them in the initial event payload, so there's some potential for a race condition.  At least this is checking the diff for expected changes!

Relatedly, I only expect us to use this sort of check for repos where we don't want all dependabot PRs merged, so eg the pipeline or databuilder repos wouldn't need this kind of check.

bloodearnest: Yeah, given the wait time for an action to run, I would be very surprised if the labels hadn't already been applied to the PR by the time this command runs

bloodearnest (approved):

PR #1671, opened 2022-03-18
Link Project to Application UI

ghickman:
This adds a page to the Staff Area to help us (likely, only me though) link Projects to their Application.  We have a number of Projects which were created via the Staff Area since we weren't using the Approve Application UI (but now are) and so aren't linked to their relevant Application.  We need to manually reconcile that and this is intended to help with that process.

### DIFF
--- a/staff/views/projects.py
+++ b/staff/views/projects.py
@@ -13,6 +13,7 @@ from django.views.generic import (
     View,
 )
 
+from applications.models import Application
 from jobserver.authorization import CoreDeveloper
 from jobserver.authorization.decorators import require_role
 from jobserver.authorization.utils import roles_for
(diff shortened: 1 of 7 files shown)


staff/views/projects.py
```diff
@@ -149,6 +151,39 @@ def post(self, request, *args, **kwargs):
...
+        applications = (
+            Application.objects.filter(project=None)
+            .exclude(status__in=ignored_statuses)
+            .select_related("created_by")
+            .order_by("-created_at")
+        )
+
+        return super().get_context_data(**kwargs) | {
+            "applications": applications,
+        }
```
rebkwok, line 184: first time I've seen this syntax in the wild :slightly_smiling_face:

rebkwok (approved):

staff/views/projects.py
ghickman: It's really cut down the boiler plate when doing simple extensions of the GCBV bits!

PR #1676, opened 2022-03-21
Improve release listings ADR

lucyb:


doc/adr/0001-avoid-listing-file-counts.md
```diff
@@ -0,0 +1,23 @@
+# 1. Avoid listing file counts unless user is authenticated
+
+Date: 2022-03-21
+
+## Status
+
+Accepted
+
```
CarolineMorton, line 8: This is very nice

CarolineMorton (approved):
This looks good to me. I think the ADR is a good approach as well and it will be nice to get into the habit of doing this especially as the team grows and iterates.

PR #1678, opened 2022-03-22
Add an Org via the Approve Application Page

ghickman:
When approving an Application we need the Staff member to select an Org and decide on a title for the Project they're about to create.  When the Application is for a new Org the user has to go to the Org section of the Staff Area and create it there, which for a single field form feels like a bit much.  This uses HTMX to call a view which returns a Bootstrap modal with the Org create form in it.  On success it redirects the user back to the approval page with the new Org's slug in the query args so the Org dropdown can pre-select that Org for the user.

I attempted to break down the OrgCreate page's template so we could re-use the form but it ended up being a bit tricky.  The Form instance is generated under the hood by OrgCreate (which is a configured CreateView) so there's no definition to reuse in the HTMX view, and the HTML itself needed to handle being inside a modal (with the submit button being in the modal footer) so it was more readble to not combine them.

### DIFF
--- a/staff/forms.py
+++ b/staff/forms.py
@@ -41,14 +41,8 @@ class PickUsersMixin:
 
 
 class ApplicationApproveForm(forms.Form):
+    org = forms.ModelChoiceField(queryset=Org.objects.order_by("name"))
     project_name = forms.CharField(help_text="Update the study name if necessary")
-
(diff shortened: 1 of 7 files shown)


tomodwyer (approved):
HTML and JS looks good to me 👍

PR #1709, opened 2022-03-31
Revert "fix: move otel django instrumenter to gunicorn.conf"

madwort:
Reverts opensafely-core/job-server#1708


madwort:
deploy failed

PR #1714, opened 2022-03-31
feat: add some more job metadata to opentelemetry

madwort:


jobserver/api/jobs.py
```diff
@@ -127,8 +129,12 @@ def post(self, request, *args, **kwargs):
                             job.status not in completed
                             and job_data["status"] in completed
                         )
+
                         span.set_attribute("previous_status", job.status)
                         span.set_attribute("status", job_data["status"])
+                        span.set_attribute("started_at", job.started_at)
```
madwort, line 135: this is wrong!!!

jobserver/api/jobs.py
```diff
@@ -118,6 +118,8 @@ def post(self, request, *args, **kwargs):
                     )
 
                     span.set_attribute("job_id", str(job.id))
```
bloodearnest, line 120: If we're going to keep it, I think we should call this `django_id` or `db_id`. It's internal to job-server db and should never be used externallly.

jobserver/api/jobs.py
```diff
@@ -118,6 +118,8 @@ def post(self, request, *args, **kwargs):
...
+                    span.set_attribute("identifier", job.identifier)
```
bloodearnest, line 121: I would call this "job_id"

bloodearnest (approved):

jobserver/api/jobs.py
madwort: willing to go to `job_identifier`? lots of potential confusion between id / identifier in this class...

Issue #1706, opened 2022-03-31
Handle Missing GitHub Org in WorkspaceCreate


ghickman:
Currently we return a 500 when a User with no OrgMemberships tries to create a Workspace, despite having permissions to do so with the relevant Project.

The User can't fix this situation but we also can't reasonably auto-assign Users to Orgs.  However, we can at least return a more specific error page so that co-pilots know which relationship the User is missing.

Issue #1705, opened 2022-03-30
Tag Users with Missing Options


ghickman:
Further to #1704 we can mark users in the user list with pill badges who are missing memberships to relevant objects.  Red badges for Backend and Org, yellow for Project.

Ideally we would also add a filter to the sidebar to quickly find users missing relationships to each of the 3 objects.

Issue #1704, opened 2022-03-30
Warn when Users are Missing Required/Expected Options


ghickman:
When a user first signs up to the site they have no BackendMemberships, ProjectMemberships, or OrgMemberships.  To make proper use of the site they need at least one of each, although ProjectMembership isn't technically _required_.  We don't want to ask for their Org on registration because we control assignment to Orgs, and BackendMembership is explicitly controlled by us, with DAAs defining sign-off for access to backends.

However, that means it's on Bennett staff to make sure a user is in the correct state.  We have a reasonable process in place for co-pilots but we still end up with users missing the required pieces because they don't enter the pipeline at the correct point.

We can't do much about this, however we should be able to make it obvious when users are in a non-ideal state.

The UserDetail staff page can show a red (danger) info box at the top of the page for required things (1+ BackendMemberships & 1+ OrgMemberships) and a yellow (warning) info box for ideal things (1+ ProjectMemberships).  I imagine some frontmatter text and then a list of the items which need resolving (linked to the sections further down the page) would be ideal here.

PR #1715, opened 2022-03-31
Smooth Setting Up Users

ghickman:
This is a collection of changes aimed at making the setup of new Users easier for either co-pilots/staff or users themselves.

Specifically it flags up when a User is missing memberships to any one of Backends, Orgs, or Projects on their detail page in the Staff Area (fixing #1704), on the User list page using badges (fixing #1705), and tells a user to contact their copilot/support when they aren't linked to an Org (fixing #1706).

### DIFF
--- a/staff/views/users.py
+++ b/staff/views/users.py
@@ -1,5 +1,7 @@
 import inspect
 
+import structlog
+from django.core.exceptions import FieldError
 from django.db import transaction
 from django.db.models import Exists, OuterRef, Q
 from django.db.models.functions import Lower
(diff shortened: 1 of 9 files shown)


staff/views/users.py
```diff
@@ -103,6 +103,21 @@ def get_context_data(self, **kwargs):
     def get_queryset(self):
         qs = super().get_queryset()
 
+        # lazily build up some queries for annotation below (Exists uses a
+        # subquery, hence the use of OuterRef)
+        backends = Org.objects.filter(members=OuterRef("pk"))
+        orgs = Org.objects.filter(members=OuterRef("pk"))
+        projects = Project.objects.filter(members=OuterRef("pk"))
```
inglesp, line 116: TIL!

inglesp (approved):
Code changes look fine, but if we can detect something's wrong, why do we expect users to tell us?

PR #1716, opened 2022-03-31
feat: even more metadata for opentelemetry

madwort:


bloodearnest (approved):

jobserver/api/jobs.py
```diff
@@ -119,6 +119,9 @@ def post(self, request, *args, **kwargs):
 
                     span.set_attribute("db_id", str(job.id))
                     span.set_attribute("job_identifier", job.identifier)
+                    span.set_attribute("backend", job_request.backend)
```
ghickman, line 122: This will be the related Backend object, is that expected?

jobserver/api/jobs.py
```diff
@@ -119,6 +119,9 @@ def post(self, request, *args, **kwargs):
...
+                    span.set_attribute("created_by", job_request.created_by)
+                    span.set_attribute("workspace", job_request.workspace)
```
ghickman, line 124: This will be the related Workspace object, is that expected?

jobserver/api/jobs.py
madwort: yeah, maybe I want `job_request.backend.slug`?

madwort: (I guess I was hoping something along the line would convert this to a string for me)

madwort: https://github.com/opensafely-core/job-server/pull/1717 perhaps?

PR #1717, opened 2022-04-01
fix: opentelemetry metadata explicit conversion to string

madwort:
* let's see whether this helps


ghickman (commented):
`str(model_object)` will return whatever is defined in the Model's `__str__` (there's a default too), is that what we want here? Or do we want specific string values? Maybe primary key or slug?

ghickman (approved):

PR #1746, opened 2022-04-20
feat: simple instrumentation for event-log

madwort:


jobserver/views/job_requests.py
```diff
@@ -216,55 +217,63 @@ class JobRequestList(FormMixin, ListView):
     template_name = "job_request_list.html"
 
     def get_context_data(self, **kwargs):
-        # only get Users created via GitHub OAuth
-        users = User.objects.exclude(social_auth=None)
-        users = sorted(users, key=lambda u: u.name.lower())
+        tracer = trace.get_tracer(__name__)
```
ghickman, line 220: Given `__name__` is for the module, can we define `tracer` outside of class/methods, similar to how we get a `logger` in various modules?

jobserver/views/job_requests.py
```diff
@@ -216,55 +217,63 @@ class JobRequestList(FormMixin, ListView):
...
-        if username:
-            qs = qs.filter(created_by__username=username)
-
-        workspace = self.request.GET.get("workspace")
-        if workspace:
-            raise_if_not_int(workspace)
-            qs = qs.filter(workspace_id=workspace)
-
-        return qs
+        tracer = trace.get_tracer(__name__)
```
ghickman, line 244: Same here.

jobserver/models/core.py
```diff
@@ -24,6 +25,7 @@
 
 env = Env()
 logger = structlog.get_logger(__name__)
+tracer = trace.get_tracer(__name__)
```
madwort, line 28: you mean here, @ghickman ? I presume it'll work?

jobserver/views/job_requests.py
ghickman: Ah, no, I meant the top of this file.  `__file__` is a magic variable which holds the name of the _current_ module.  So for loggers we are essentially naming them with the same name as the module they are declared in, then all logging calls retain where they are called from.  I think the tracer is supposed to work in the same way, so L21 of this module should do it.

madwort: oh right, yeah

ghickman (approved):

jobserver/views/job_requests.py
madwort: for the tracer, I thought that because it's using `__name__` it's named for the thing that spawned the request (in the prod case gunicorn?). So, I think the auto instrument initialisation in `gunicorn.conf.py` is creating the tracer, and we're picking up the same one here (hence why we're able to augment the existing traces). So it's not per-module in that sense, and if we start using it more widely it might make more sense to put it in one central place (rather than every module), but I don't really mind atm.

PR #1749, opened 2022-04-21
fix: pre-commit is installed by the devenv

madwort:
* therefore the justfile install should depend on devenv

### DIFF
--- a/justfile
+++ b/justfile
@@ -77,7 +77,7 @@ devenv: prodenv requirements-dev && install-precommit
 
 
 # ensure precommit is installed
-install-precommit:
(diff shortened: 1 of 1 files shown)


iaindillingham (approved):

ghickman:
`devenv` has `install-precommit` set up as a post-dependency so this would make it a circular dependency.  That requires just>=0.9.9 though, are you running a version later than that?

madwort:
I'm running just-0.10.5. I hit this because I manually trashed my venv in order to attempt to replicate a CI failure - which then went away by re-running CI, sooooo

PR #1761, opened 2022-04-28
Require Usernames for Token-based Auth

ghickman:
This switches our token-based auth to the format `username:token`, so that we can look up the relevant User with the username.

### DIFF
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -6,7 +6,12 @@ from django.db.models import Value
 from django.shortcuts import get_object_or_404
 from django.utils import timezone
 from rest_framework import serializers
-from rest_framework.exceptions import NotAuthenticated, ParseError, ValidationError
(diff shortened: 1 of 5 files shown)


jobserver/api/releases.py
```diff
@@ -77,8 +82,16 @@ def validate_release_access(request, workspace):
 
     This validation uses a User PAT OR Django's regular User auth.
     """
-    auth_header = request.headers.get("Authorization")
-    if User.is_valid_pat(auth_header):
+    if auth_header := request.headers.get("Authorization"):
```
bloodearnest, line 85: I AM THE WALRUS

bloodearnest (approved):
This makes sense to me.

I've realsised that the current design limits users to 1 PAT only.

No necessarily a problem, but it makes rotation brittle, and more awkward if we wanted to use the same user in different settings. 

But not relevant to this PR

PR #1785, opened 2022-05-09
Fix downloading published outputs

### LINKED ISSUE #1784: Stop Redactions Breaking Publishing
When we redact released files they are still included when publishing outputs from a workspace.

This causes two problems:
* The published outputs page includeds deleted files by default
* The download button is broken

While we do need to be able to delete published files I don't think we need to include them when creating a new Snapshot.

The download button should be able to skip deleted files.

ghickman:
This fixes two pieces of downloading Snapshots (published outputs).

The download URL was being created on the wrong object and we now only include existing ReleaseFiles in a Snapshot.  ReleaseFiles can be deleted after a Snapshot is created still but we won't add them by default.

Fixes #1784

### DIFF
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -301,16 +301,20 @@ class SnapshotCreateAPI(APIView):
         file_ids = set(data["file_ids"])
         files = ReleaseFile.objects.filter(pk__in=file_ids)
 
-        rfile_ids = set_from_qs(files)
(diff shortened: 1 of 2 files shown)


iaindillingham (approved):
I think I understand the problem and the solution, so ✅. I'm not familiar with job-server's terminology, though. When we move into teams, then hopefully you'll get better quality reviews 🙂

PR #1800, opened 2022-05-11
Attempt to remove configured reviewers from dependabot

ghickman:
🤞


iaindillingham (approved):
Not quite sure why we're adding empty properties here, but I assume it's because we tried unsetting something previously, and it didn't work.

PR #1809, opened 2022-05-20
Add some additional instructions for running a local database

rebkwok:
Add instructions on creating the postgress user with necessary permissions for the local db.

Also changed the default user from "user" to "jobsuser" because psql doesn't like a user with the name "user".

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -26,12 +26,10 @@
 ### Development credentials
 
 _Note:_ you will need the [Bitwarden CLI tool](https://bitwarden.com/help/article/cli/) installed in order to access passwords, but it is not a requirement.
-- There is an existing `dotenv-sample` template that you can use to base
-  your own `.env` file on.
+
+- Create a `.env` file; there is an existing `dotenv-sample` template that you can use to base your own `.env` file on.
 - Use `bw` to login to the Bitwarden account.
-- When logged in to Bitwarden, run `scripts/dev-env.sh
-  <env_file_target>` to retrieve and write the credentials to the target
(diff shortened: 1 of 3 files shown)


DEVELOPERS.md
```diff
@@ -93,7 +91,11 @@ If you need to upgrade an installation the [ArchWiki](https://wiki.archlinux.org
 You'll need a database in Postgres to work with, run:
 
 ```sh
-psql -c "create database jobserver"
+psql -c "CREATE DATABASE jobserver"
+psql -c "
+CREATE ROLE jobsuser PASSWORD 'pass' NOSUPERUSER CREATEDB;
+GRANT ALL PRIVILEGES on database jobserver to jobsuser;
+"
```
ghickman, line 98: We should probably find a better way to standardise this, I know OSI had a run at putting postgres inside docker-compose by default.  This works but isn't necessary on macOS when using Postgres.app, but I think this is fine to document until we have a better story in general!

ghickman (approved):

DEVELOPERS.md
rebkwok: Hmm. I guess that's why it wasn't documented.  Do you also not need to create the user with Postgres.app?  I got a bunch of unexpected django postgres errors trying to run tests.  I could add that the create role/grant bit is only required on linux?

ghickman: Yeah Postgres.app sets up ident auth so your OS user gets the equivalent of these steps.  You could add these are only for linux for now and then we can tackle this probably later for all the sites :)

PR #1830, opened 2022-05-30
Generate UUID to force a fresh download on reload

tomodwyer:
Currently the outputs viewer is caching `fetch` requests too aggressively. We want to refetch files from the server on page reload.

This PR generates a cryptographically random UUID each time the React app is mounted, and appends this as a query string to the fetched URL. This forces a file download on first attempt, but allows for cached queries when navigating between files.

### DIFF
--- a/assets/src/scripts/outputs-viewer/tests/jest-setup.js
+++ b/assets/src/scripts/outputs-viewer/tests/jest-setup.js
@@ -42,3 +42,11 @@ function noOp() {
 if (typeof window.URL.createObjectURL === "undefined") {
   Object.defineProperty(window.URL, "createObjectURL", { value: noOp });
 }
+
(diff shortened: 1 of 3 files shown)


assets/src/scripts/outputs-viewer/tests/jest-setup.js
```diff
@@ -42,3 +42,11 @@ function noOp() {
 if (typeof window.URL.createObjectURL === "undefined") {
   Object.defineProperty(window.URL, "createObjectURL", { value: noOp });
 }
+
+if (typeof window.crypto === "undefined") {
```
bloodearnest, line 46: Also oblig xkcd ref https://xkcd.com/221/

bloodearnest (approved):

PR #1835, opened 2022-06-02
feat: add links to two simple visualisations

madwort:
* matching documentation here: https://github.com/opensafely/documentation/pull/753 
* very open to suggestions on better layout/styling
* looks like this: 
Screenshot 2022-06-02 at 11 41 55
* nb. these links will be visible to everyone, but only users with honeycomb logins will be able to view the visualisations. There may be an argument for only showing the links to e.g. admin users.


tomodwyer:
CleanShot 2022-06-09 at 11 49 50@2x

Updated styles

ghickman (approved):

PR #1836, opened 2022-06-02
fix:  change test_jobrequest_completed_at_success behaviour

madwort:
* The `test_jobrequest_completed_at_success` looks at a glance as though it tests that the calculated completed_at for a job_request matches the value for a job. However, the jobs do not have a value for completed_at, so the test is actually testing `None == None`. 
* Demonstrate that the assert from `test_jobrequest_completed_at_while_incomplete` also passes.


bloodearnest (approved):

ghickman (approved):
👌

PR #1841, opened 2022-06-02
feat: add basic visualisation links for jobs

madwort:


jobserver/views/jobs.py
```diff
@@ -49,11 +52,22 @@ def get(self, request, *args, **kwargs):
             request.user, "job_cancel", project=job.job_request.workspace.project
         )
 
+        honeycomb_can_view_links = has_role(self.request.user, CoreDeveloper)
+
+        honeycomb_context_starttime = job.created_at - timedelta(minutes=1)
+
+        honeycomb_context_endtime = timezone.now()
+        if job.completed_at is not None:
+            honeycomb_context_endtime = job.completed_at + timedelta(minutes=3)
```
ghickman, line 61: Why completed_at + 3 minutes here?

ghickman (approved):
Looks good.  Couple of comments on the timedeltas, might be easier to put the explanations in as comments for the next reader too?

jobserver/views/jobs.py
madwort: Good question, I don't remember now why this is longer than the starttime - would it be more aesthetic to make them match?

madwort: (that they're different makes it look like they're different for a reason, if they're not, then, I should change it!)

jobserver/views/jobs.py
```diff
@@ -49,11 +52,22 @@ def get(self, request, *args, **kwargs):
             request.user, "job_cancel", project=job.job_request.workspace.project
         )
 
+        honeycomb_can_view_links = has_role(self.request.user, CoreDeveloper)
+
+        honeycomb_context_starttime = job.created_at - timedelta(minutes=1)
```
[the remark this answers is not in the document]
ghickman: This is great 👍

PR #1859, opened 2022-06-14
update slack channel for release notifications

LFISHER7:
`opensafely-outputs` was getting quite noisy. This moves notifications for releases from `opensafely-outputs` to the newly created `opensafely-releases`.

@ghickman is this all I need to do are is there more to it?


bloodearnest (approved):

ghickman:
@LFISHER7 – yep! You'll need to update the tests as well though 🙂

PR #1864, opened 2022-06-16
Add JobRequest POST API

### LINKED ISSUE #1822: Implement  a JobRequest API
This would allow the creation of jobs from OSI and openssfely-cli in future.

We can probably use the job-runner API JobRequest JSON definition?

ghickman:
This adds a POST handler to `api/v2/job-requests/` so OSI can create JobRequests.  It didn't yield any nice abstractions from the UI view unfortunately, but it did push me to make all the view authentication explicit and opt-in which is a nice win.

Couple of things to note:
* The `JobRequestAPIListCreate` view is a simple dispatcher to the relevant view based on HTTP verb
*  The project_definition field is being validated by our pipeline library just to make sure it's valid

Fixes #1822


jobserver/api/jobs.py
```diff
@@ -234,6 +236,73 @@ class Meta:
...
+            ]
+            model = JobRequest
+
+        def validate_project_definition(self, value):
+            load_pipeline(value)
+
+            return value
+
+    def post(self, request, *args, **kwargs):
+        auth_header = request.headers.get("Authorization")
```
lucyb, line 283: How come this is checking the token in a different way to the APIViews below? Only asking because I'm trying to learn how the authentication works and the views below use an initial method..

jobserver/urls.py
```diff
@@ -73,7 +73,7 @@
 
 
 api_urls = [
-    path("job-requests/", JobRequestAPIList.as_view()),
+    path("job-requests/", JobRequestAPIListCreate.as_view()),
```
lucyb, line 76: Why are you replacing the list view? What's going to happen to that?

ghickman: I'm not!  Have a look at the JobRequsetAPIListCreate view (it's tiny), hopefully the docstring explains, if not I'd consider that a bug 🙂

jobserver/api/jobs.py
ghickman: We have 4 ways of authenticating with our API:
* no authentication (for the public list of JobRequests)
* backend token (mostly for a job-runner instance to sync with job-server)
* session auth (for the output viewer)
* bot user token (for OSI)

Typically in an APIView auth[n|z] is checked in the `initial` method (which most of our views do), because it fires before the relevant verb handler method.  But in this case we should never get a non-POST request because of how we invoke the view and I thought it was simpler to reduce the view to just the POST handler method.

lucyb: Ah ha, thank you. That makes sense. I think I was also trying to understand the difference between the backend token and the bot user token, which I found confusing on the first read.

jobserver/urls.py
lucyb: Oh yes. Don't mind me, I see now.

PR #1872, opened 2022-06-22
fix: update honeycomb org name

madwort:
* new organisation also has environments, so add that too


ghickman (approved):

madwort:
thanks @ghickman you approved before I had time to tag you in slack...!

PR #1876, opened 2022-06-23
Add per-backend status pages

ghickman:
This adds a duration field to the Backend model to track how long before we consider a Backend missing and a page which returns 200 or 503 based on if we've exceeded that duration or not.

I've defaulted the duration to 5 minutes since that's our current threshold but this will allow us to have a different duration for each Backend.

The duration field has been added to the Staff Area but I'll add editing in a subsequent PR.

Ref opensafely-core/sysadmin#36


jobserver/migrations/0019_add_backend_missing_threshold.py
```diff
@@ -0,0 +1,20 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("jobserver", "0018_move_project_numbers_to_number_field"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="backend",
+            name="missing_threshold",
```
bloodearnest, line 17: nit: I think "alert_timeout" might be a better name? I think it better describes what it does, "missing" is a bit vague?

jobserver/views/status.py
```diff
@@ -1,12 +1,49 @@
...
+from django.shortcuts import get_object_or_404
 from django.template.response import TemplateResponse
+from django.utils import timezone
 from django.views.generic import View
 
 from ..backends import show_warning
 from ..models import Backend
 
 
+class PerBackendStatus(View):
```
bloodearnest, line 12: My instinct would for these to be API calls, rather than regular views

Given they're will be hit a lot, doesn't seem worth rendering a template?

Typical form for status checks just plain text response. They should be lightweight and fast.

Having a human readable page is maybe also worthwhile, but shouldn't be the same url I think?

ghickman: That's a point, we could just leave the UI for backend statuses to the existing `/status/` page.  I'll switch these to empty responses (freshping only cares about status code) and put the debugging information on the cards on the main status page.

ghickman:
@bloodearnest – updates all round:

* updated field name
* per-backend pages changed to be empty plain responses with the relevant status code
* added the new field to the main status page and tweaked their "missing" display based on that
* converted the show_warning function to use `alert_timeout` as its timedelta so it can make the calculation based on the timeout set on a backend

jobserver/views/status.py
```diff
@@ -1,20 +1,40 @@
...
+        # how long ago did we last see this backend?
+        time_since_last_seen = timezone.now() - last_seen
+
+        # do we consider the backend missing?
+        is_missing = time_since_last_seen > backend.alert_timeout
 
-            return last_seen.strftime("%Y-%m-%d %H:%M:%S")
+        # set a status code of 503 if the backend was last seen longer than the
+        # configured value ago
+        status_code = 503 if is_missing else 200
```
bloodearnest, line 31: Suggestion: return a text/plain response with `last_seen` as isodate

Freshping only needs the status code, but it's useful to have something there if we hit it with curl to test stuff, and the actual last seen value might be useful

bloodearnest (approved):

jobserver/views/status.py
ghickman: ooh yes 👍

PR #1887, opened 2022-06-28
Disable Creating Jobs

ghickman:
This adds a feature flag to disable creating jobs.

Previously we relied on the user's BackendMemberships QS being the full set of backends a user could access.  With this change I've hoisted that lookup into dispatch so we can selectively exclude some backends if the flag is enabled.  Then the access check and form kwargs builder can both use that maybe-modified QS.

### DIFF
--- a/jobserver/views/job_requests.py
+++ b/jobserver/views/job_requests.py
@@ -1,6 +1,7 @@
 import functools
 from datetime import timedelta
 
+from django.conf import settings
 from django.contrib import messages
 from django.core.exceptions import MultipleObjectsReturned
 from django.db import transaction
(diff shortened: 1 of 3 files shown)


jobserver/views/job_requests.py
```diff
@@ -160,16 +170,10 @@ def get_context_data(self, **kwargs):
...
-        kwargs["actions"] = [a["name"] for a in self.actions]
-
-        # get backends from the current user
-        backends = Backend.objects.filter(
-            memberships__in=self.request.user.backend_memberships.all()
-        )
-        kwargs["backends"] = backends_to_choices(backends)
-
-        return kwargs
+        return super().get_form_kwargs() | {
```
lucyb, line 173: This is much nicer. The pipe operator is a new one for me and I like the usage of it here

jobserver/views/job_requests.py
```diff
@@ -103,9 +104,18 @@ def dispatch(self, request, *args, **kwargs):
...
             return redirect(self.workspace)
 
+        # some backends might need to be disabled.  This view only uses
+        # backends the user can see so we look them up here, removing the
+        # relevant ones from the QS before we check if there are any below.
+        # The form, in get_form_kwargs, will also use the backends constructed
+        # here to be consistent.
+        self.backends = request.user.backends.all()
+        if settings.DISABLE_CREATING_JOBS:
+            self.backends = self.backends.exclude(Q(slug="emis") | Q(slug="tpp"))
```
lucyb, line 114: I understand why it's been done this way, but I dislike that we need to know that DISABLE_CREATING_JOBS affects only TPP and EMIS. I don't have a better solution though and hopefully it will only be here for a short time!

jobserver/views/job_requests.py
```diff
@@ -103,9 +104,18 @@ def dispatch(self, request, *args, **kwargs):
...
+
         # jobs need to be run on a backend so the user needs to have access to
         # at least one
-        if not request.user.backends.exists():
+        if not self.backends.exists():
```
lucyb, line 118: The JobRequestPickRef uses the same logic. Does that need to be changed too?

lucyb (approved):

PR #1890, opened 2022-06-29
link to workspaces on event log

ghickman:


CarolineMorton (approved):
Lol - short but sweet

PR #1915, opened 2022-07-08
Test GitHub API with Fake object

### LINKED ISSUE #485: Replace GitHub Patches with FakeAPI
Many of the tests, particularly those exercising the views make use of `mock.patch`, often many of them at once.  Many of those patches are for calls to functions in, or wrapping those in, `github.py`.  Patches can be tricky to get right, as #424 shows.

Peter and Simon have suggested a way to avoid this would be to create a Fake API which is used in tests.

Ideally this would be activated for a test using a fixture.

ghickman:
This enables the use of a verified fake for testing our code which talks to GitHub's API.

It does this by:
* wrapping the functions in github.py in a class
* passing the wrapper to the views which talk to GitHub as a kwarg
* testing all uses of the API code with a Fake
* adding verification tests which hit the new `opensafely-testing` org to confirm the wrapper code works
* comparing the output of the real API to that of the fake, in terms of keys and value types
* only runs coverage on the code which talks to the real API in CI [by default]

Fixes #485


TESTING.md
```diff
@@ -73,6 +73,15 @@ We also provide DRF's testing tools as fixtures to mirror pytest-django.
 [pytest-django](https://pytest-django.readthedocs.io/en/latest/) is the bridge between pytest and Django, exposing various Django testing tools (Client, RequestFactory, etc) as fixtures.
 
 
+## Verified Fakes
+We use a fake object, `FakeGitHubAPI`, to test our uses of GitHub's API.
+However, we want to [verify that fake](https://pythonspeed.com/articles/verified-fakes/), so we have a set of verification tests in `tests/verification/`.
+
+`just test` will run those tests, but `just-dev` will not, this way CI verifies our faster tests are correct.
```
bloodearnest, line 80:
```suggestion
`just test` will run those tests, but `just test-dev` will not, this way CI verifies our faster tests are correct.
```

justfile
```diff
@@ -96,19 +96,29 @@ run-telemetry: devenv
...
+        --module pytest \
+        {{ args }}
+
+
+test-dev *args:
+    {{ just_executable() }} test-base \
+        '-m "not verification"' \
+        {{ args }}
+
+    # run with || so they both run regardless of failures
```
bloodearnest, line 113: why do we want them to both run regardless? Does report exit with 1 if not 100%

bloodearnest (commented):
I am -0 on this.

It adds a lot of complexity.

I wonder if marking tests with a `githubapi` marker would allow them to be skipped in local testing if needed for much less complexity? Or would that involve so many tests so as to be useless?


I think the value proposition is increased if we added this to our osgithub api wrapper. This would mean maintaining the fake and verification there, and not adding that complexity to job-server, as well as solidifying our github api story in multiple places.

justfile
ghickman: Yes 😞

bloodearnest (approved):

PR #1929, opened 2022-07-12
Require User.fullname is set

### LINKED ISSUE #1900: Ask that users add their names into job server

### LINKED ISSUE #1903: SPIKE Mandate entering name during registration
How easy would it be to add middleware to send people to the settings page to enter their details, with a redirect back to the existing registration flow.

ghickman:
This redirects all users with an empty `fullname` value to their settings page where they will see an alert saying they need to fill in their fullname.

I'm always wary of adding more middleware but since we have so many (~50%) users who need to fill it in this is the easiest way to force them to the page where they can do so.  It also shouldn't result in more db requests since `request.user` is set on the request by a preceeding middleware (`AuthenticationMiddleware`).

I considered using Social Auth's partial pipeline machinery to kick a user out of the pipeline if they didn't have a name set on GitHub, so they could fill it in on a job-server page before returning to the pipeline.  However that seemed like more work and wouldn't cover our existing user base.

Fixes #1900 
Fixes #1903

### DIFF
--- a/jobserver/middleware.py
+++ b/jobserver/middleware.py
@@ -1,0 +1,53 @@
+from django.shortcuts import redirect
+
+
+class RequireNameMiddleware:
+    """
+    Redirect users without `fullname` set to the settings page
+
+    We get a User's username from GitHub when they register with the site.  If
+    their GitHub profile has a name set we also get that, however we have many
+    (~50% at time of writing) users who have not set that.  We want to require
+    all users have their name set.  This middleware redirects any user with an
+    empty `fullname` value to their settings page.
+
+    We exempt various URL prefixes which don't make sense to do this on, eg if
+    we're not expecting traditional users there or it doesn't make sense for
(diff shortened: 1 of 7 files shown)


jobserver/middleware.py
```diff
@@ -0,0 +1,52 @@
...
+            "/apply",
+            "/complete",
+            "/disconnect",
+            "/favicon.ico",
+            "/login",
+            "/logout",
+            "/robots.txt",
+            "/settings",
+            "/staff",
+            "/status",
```
bloodearnest, line 37: Hmm, this might bite us if we change urls for any of these. But hopefully this is a short term thing!

Minor suggestion - order this most-popular to least, which will make the `startswith()` clause below more efficient overall. OTTOMH I think `/api` and `/staff` and `/robots.txt` are likely to be very common.

bloodearnest (approved):

PR #1946, opened 2022-07-18
Add various security headers

ghickman:
Ref opensafely-core/interactive.opensafely.org#179
Ref #2004


jobserver/settings.py
```diff
@@ -259,6 +262,16 @@
 X_FRAME_OPTIONS = "SAMEORIGIN"
 
 
+# CSP
+# https://django-csp.readthedocs.io/en/latest/configuration.html
+CSP_DEFAULT_SRC = ["'self'"]
+CSP_IMG_SRC = ["*", "data:"]
```
lucyb, line 268: Why '*'? Is that something that will be changed in future?

jobserver/settings.py
```diff
@@ -259,6 +262,16 @@
...
+CSP_SCRIPT_SRC = ["'self'"]
```
lucyb, line 269: We're also using plausible on job-server

jobserver/settings.py
```diff
@@ -259,6 +262,16 @@
 X_FRAME_OPTIONS = "SAMEORIGIN"
 
 
+# CSP
+# https://django-csp.readthedocs.io/en/latest/configuration.html
+CSP_DEFAULT_SRC = ["'self'"]
```
lucyb, line 267: From a bit of random clicking around the site, I think localhost will need to be explicitly allowed somewhere for viewing the L4 outputs

jobserver/settings.py
lucyb: Also, the avatars coming from github seem to be of type "text/html / Redirect". Not sure if they'll be picked up under the img_src section?

ghickman:
@tomodwyer – looking at the outputs viewer (testing or interactive workspaces are good examples) CSP is complaining about inline styles.  Any idea how we deal with that in a React app?

jobserver/settings.py
lucyb: Hmm, that might be tricky. Env vars would be easier to deal with. I'm not sure I have a nice solution. My ideas are 
1) hard code the release-hatch values in this file and live with the duplication (how often do we add new backends?)
2) use '*' for the relevant section; is it script-src? Then, tidy it up once we have something that works. I don't like the idea of doing something that's insecure, just in case we don't come back to it. However, it's probably better than not having CSPs at all

tomodwyer:
> @tomodwyer – looking at the outputs viewer (testing or interactive workspaces are good examples) CSP is complaining about inline styles. Any idea how we deal with that in a React app?

Oh yeah, I did that specifically to bypass the Bootstrap styles. I can separate it out in to a new stylesheet. I'll crack on with that now

jobserver/settings.py
```diff
@@ -259,6 +262,41 @@
...
+]
+CSP_MANIFEST_SRC = ["'self'"]
+
+# Duplicate the *_ELEM settings for Firefox
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1529338
+CSP_SCRIPT_SRC = CSP_SCRIPT_SRC_ELEM = ["'self'", "https://plausible.io"]
+CSP_STYLE_SRC = CSP_STYLE_SRC_ELEM = ["'self'"]
+
+# which directives to set a nonce for
+CSP_INCLUDE_NONCE_IN = ["script-src-elem"]
```
lucyb, line 286: Does "script-src" also need to be in this list?

tomodwyer: It does indeed, fixed ✅

ghickman:
@tomodwyer – I've updated the local declaration for fonts (we needed `data:` instead of the vite devserver URL.

We're still getting an error for `style-src-elem` wanting `unsafe-inline`.  I can't trigger it in dev mode and I can't work out what's triggering it in prod mode.  Could you take a look?

I _think_ that's the last itme to check off for this to be merged 🤞

lucyb:
> unsafe-inline

Searching for `style=` shows a number of places where inline styles are still used.

One option might be to roll out this change with unsafe-inline and create another ticket to fix it later.

tomodwyer (approved):

lucyb (approved):

Issue #1871, opened 2022-06-22
Add Project Number to Applications List


ghickman:
When looking at the list of applications it would be useful to show Project.number for applications which have an approved project with a number.

We should also be able to search on that page (and the home page) for project numbers.

ghickman:
Done in #1953

PR #1953, opened 2022-07-20
Link to project on application list page

ghickman:
This adds a link to a Project if it's linked on the Application list page, using the number instead of the title:

Ref #1871

### DIFF
--- a/staff/templates/staff/application_list.html
+++ b/staff/templates/staff/application_list.html
@@ -133,6 +133,10 @@
 
           </div>
 
+          {% if application.project %}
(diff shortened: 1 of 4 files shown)


lucyb (approved):
Looks good.

Since this will close the original ticket, which mentions "We should also be able to search on that page (and the home page) for project numbers.".. Is this something that should be done later or not worth it?

ghickman:
@lucyb – I missed that, thanks!  More searchable fields is always good and they're trivial to add so popped them into this one.

Issue #1961, opened 2022-07-22
Track size of ReleaseFile


ghickman:
We don't currently track the size of ReleaseFiles in the database.  The actual files are on disk so can be looked up but we have realised it would be good to have those in the database too.  ReleaseFiles never change (unless they are deleted) so they should never get out of sync.

We already have a `size` property which returns the on-disk size in megabytes which we should keep (maybe with a keyword to change the size format).

We can add a `filesize` field (to match `filehash`) which stores the size as an integer of bytes.

We [create ReleaseFiles in one place](https://github.com/opensafely-core/job-server/blob/4d4ecd8311ac6644e3c109765043603048b6ca45/jobserver/releases.py#L141-L148) so can set the filesize there.

Backfilling this should be a simple migration which enumerates all the existing ReleaseFiles and gets their size in bytes, we can then make the field non-nullable.

ghickman:
Fixed in #2031

PR #1963, opened 2022-07-25
Fixup Migrations&Constraints

ghickman:
This is a bit of an odd collection of changes.  They all come from running a quick spike on #1961.

We have a missing migration for the state of Org.logo_file's upload path.  There's no database change to be made by this but the model state tracks everything so a migration was still missing, so I've fixed that and also added a check into `just check` to help us catch this in CI going forward.  This led to uncovering the next part…

The code has a Check constraint on both ReleaseFile's deleted_at and deleted_by either being set or not.  The constraint exists in both the code and migrations, but _not_ in the production db.  I suspect this is due to a bad migration squash where the constraint got added but wasn't deployed before the squashed migrations which would have been migrated with `--fake`.  I've removed the constraint from the initial migration and re-added it as a separate migration so we can get it into production.

This then led me to looking at all our check constraints to make sure they all existed in the db and why I've also added/normalised how we test all of those.

Finally I realised that we had sevaral forms of writing these constraints:
* `(Q(field1=None) & Q(field2=None)) | (~Q(field1=None) & ~Q(field2=None))`
* `Q(field1=None, field2=None) | ~Q(field1=None, field2=None)`
* `Q(field1__isnull=True, field2__isnull=True) | Q(field1__isnull=False, field2__isnull=False)`

The SQL produced by these is all subtlely different.  I couldn't find a way to test that they had any different outcome so I don't think we have any bugs there, however I have normalised all the forms to the final one as I think it's the clearest to what's going on.


.github/workflows/main.yml
```diff
@@ -86,11 +86,13 @@ jobs:
       - name: Run tests
         env:
           DATABASE_URL: postgres://user:password@localhost/jobserver
+          GITHUB_TOKEN: empty
           GITHUB_TOKEN_TESTING: ${{ secrets.OPENSAFELY_GITHUB_TESTING_ORG_PAT }}
           SECRET_KEY: 12345
           SOCIAL_AUTH_GITHUB_KEY: test
           SOCIAL_AUTH_GITHUB_SECRET: test
         run: |
+          just check-migrations
```
lucyb, line 95: If I'm understanding this correctly, it will just print out the migrations that are required, it won't fail, which I think is the intention?

lucyb: Something like this might work?

`if $BIN/python manage.py makemigrations --dry-run; then exit 1; fi`

I was able to test it on OSI, because apparently that is missing a migration right now :grimacing:

lucyb: Maybe saved by stack overflow https://stackoverflow.com/a/48442671

ghickman: Arg, no, thanks!

lucyb (changes requested):
Looks good, just the --check flag to add for the migrations. Seems like a good solution that's worth copying over to other apps in future (like OSI).

Must have been quite tricky digging through all this, but worth it and the overall solution looks nice.

ghickman:
@lucyb – added the `--check` flag, is this good to go now?

lucyb (approved):

PR #1970, opened 2022-07-27
Filter out non research repos

ghickman:
This flips the way in which we filter repos to present to the User when creating a Workspace.

We currently get repos from the `researchers` team  relying on our permissions script to ensure repos are in that.  There's a lag time for that script running (every 15 minutes) and it causes some amount of support overhead to ask Users to wait for that to update.

This change removes the team check in favour of relying on our recent use of tagging non-research repos with the `non-research` topic.

### DIFF
--- a/jobserver/github.py
+++ b/jobserver/github.py
@@ -267,6 +267,13 @@ class GitHubAPI:
                     name
                   }
                 }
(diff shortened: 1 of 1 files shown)


bloodearnest:
Fixes #1971

PR #1974, opened 2022-07-29
Expose db maintenance mode

ghickman:
Store the flags job-runner sends up in a JSON column and then expose the current mode via the `/status/<backend>/db-availability/` URL.

We only have flags to send up currently but we have wanted a generic place for a while hence no reference to flags on the Backend model (could it just be called state?), and also a JSONB column because we really want job-runner to define the structure here.

I [tested this](https://github.com/opensafely-core/job-server/pull/1974/files?show-viewed-files=true&file-filters%5B%5D=#diff-9bee776c8c3e90248db2ad41ce2e665c0fb25a3cef562281afce957b4ab9883dR504-R507) by recreating what I think a flags header looks like based on the data in job-runner-on-TPP's live database and the [flag building code in job-runner](https://github.com/opensafely-core/job-runner/blob/4041d0d7442c8a385c23c9a6d9f0a37e1ceebc0c/jobrunner/sync.py#L76-L79).

I've only exposed this for TPP since that's the only database we actually support it on.

I've set up a freshping check and freshstatus integration so that when this endpoint returns a 503 we'll put the TPP service on status page into "Under Maintenance".

Fixes opensafely-core/sysadmin#38


jobserver/views/status.py
```diff
@@ -10,6 +10,19 @@
...
+    def get(self, request, *args, **kwargs):
+        backend = get_object_or_404(Backend, slug=self.kwargs["backend"])
+
+        if backend.slug != "tpp":
+            return HttpResponse()
+
+        if backend.jobrunner_state.get("mode", {}).get("v", "") == "db-maintenance":
+            return HttpResponse(status=503)
+
+        return HttpResponse()
```
bloodearnest, line 23: Can we return a plain text reponse here for human debuggers?

e.g.
```
"DB maintenance mode not implemented in {backend.name}"
"DB maintenance mode off (since {mode['ts']})"
"DB maintenance mode on (since {mode['ts']})"
```

jobserver/urls.py
```diff
@@ -189,6 +189,16 @@
     path("<str:pk>/<path:path>", ReleaseDetail.as_view(), name="release-detail"),
 ]
 
+status_urls = [
+    path("", Status.as_view(), name="status"),
+    path("<slug:backend>/", PerBackendStatus.as_view(), name="status-backend"),
+    path(
+        "<slug:backend>/db-availability/",
```
bloodearnest, line 196: I'd be inclined to use the term `maintenance` rather than `availability`. The former has a more specific documented meaning, and is "business as usual", where as the db being "unavailable" is arguably a separate check and a more serious problem

tests/unit/jobserver/api/test_jobs.py
```diff
@@ -476,6 +478,53 @@ def test_jobapiupdate_post_with_errors(api_rf, mocker, error_message):
...
+            "updated_at": now,
+            "completed_at": None,
+        }
+    ]
+
+    flags = json.dumps(
+        {
+            "mode": {"v": "", "ts": "1659092411.5025"},
+            "paused": {"v": " ", "ts": "1657099752.16788"},
+            "pause": {"v": "", "ts": "1657099560.72109"},
```
bloodearnest, line 506: lol, this was a type on my part, it's "paused"

bloodearnest (approved):
Some comments

jobserver/urls.py
ghickman: From the slack discussion we've chosen to keep this as is to avoid `GET 200 /maintenance` meaning that db maintenance is off and causing confusion, but we all agree we don't like it.

PR #1975, opened 2022-07-29
Save flag states as a dict instead of the json string

ghickman:


jobserver/api/jobs.py
```diff
@@ -190,7 +191,7 @@ def post(self, request, *args, **kwargs):
         # store backend state sent up from job-runner.  We might rename the
         # header this is passed in at some point but now this is good enough.
         if flags := request.headers.get("Flags", {}):
```
lucyb, line 193: The fallback here shouldn't be an empty dict, as I think that will cause an error for json.loads?

lucyb (changes requested):

jobserver/api/jobs.py
ghickman: Good catch! Fixed and rebased.

lucyb (approved):

Issue #1968, opened 2022-07-26
Spike on GH badges


lucyb:
Spike on using GH badges

### Background

- We only expect to spend a max of N hours on this exploratory work
- It would be very useful to have a way to link repo READMEs back to useful things (e.g. job server workspace page)
- Asking users to add links and remember to update things is prone to error
- We could use badges to add links dynamically and colour-coding to show state
- Could badges be used to incentivise other actions by researchers? (e.g. add link to published papers, etc)
- Could we also show published outputs, released outputs and number of jobs run? This would show how active a repo is and would help provide Amir with useful information.

### Acceptance criteria

- Notes have been made in google doc explaining what badges could be used for and how much effort would be involved in implementing this (possibly with different options, and a proposed best solution)
- There is a link to this doc from the “Make Amir’s Life Easier” doc https://docs.google.com/document/d/1N5XC0faHmfC_XxJiH3wSGsftfY402G3hc4ezDrO6nZ8/edit?usp=sharing
- The findings have been communicated to the rest of the pipeline team and Catherine

ghickman:
There are 3 obvious and easy badges that I think have relevance for this:
* number of jobs for a workspace
* number of released files for a workspace
* number of published outputs for a workspace

These are all workspace specific but a given branch on a repo can only be tied to one workspace (we have a uniqueness constraint to enforce this) so I think keeping these workspace specific makes sense here.

We can't support "does this repo have a paper published" yet since we don't have that information tracked in the platform but it would be easier to add later on.

ghickman:
Doc updated with some of the bits above and some thoughts on driving users to do take actions.

PR #1987, opened 2022-08-02
Add badges for number of jobs and released files for a given workspace

ghickman:
This adds a couple of extra badges and reorganises the badge views into their own module.

Refs #1968

### DIFF
--- a/jobserver/urls.py
+++ b/jobserver/urls.py
@@ -25,6 +25,7 @@ from jobserver.api.releases import (
     WorkspaceStatusAPI,
 )
 
+from .views.badges import WorkspaceNumJobs, WorkspaceNumPublished, WorkspaceNumReleased
 from .views.index import Index
 from .views.job_requests import (
     JobRequestCancel,
@@ -68,7 +69,6 @@ from .views.workspaces import (
     WorkspaceLog,
     WorkspaceNotificationsToggle,
     WorkspaceOutputList,
-    WorkspaceOutputsBadge,
 )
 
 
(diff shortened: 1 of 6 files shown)


bloodearnest (commented):
I'm not sure the url requirements work here.

Currently, this would require the urls in the README to have explicit org, project, and workspace in order to work.

At the point of generating the initial README, we don't know all that. Indeed, job-server doesn't even know that yet, as the repo has to exist before such things can be determined. That means for these badges to be useful, a user would have to manually edit and commit the readme to supply the missing information.


When we generate, the initial README, we do know the name of the repo, we could include that in the badge url, e.g.

`https://jobs.opensafely.org/badges/opensafely/foo/$BADGETYPE`

Now we can render a badge based on the workspaces linked to the repo.

We don't know the branch however, which affects workspace currently (another reason to bad multiple branches!), and the branch of course can change.

However, we could maybe hack around this by using `referer`, which in the case of viewing a readme in a branch would look like:

`https://github.com/openafely/foo/tree/branch`

Then the badge url could use this info to pick the write workspace to use in the case of multiple workspaces.

ghickman:
It seems like we're not [the first to think of the referrer](https://github.com/github/markup/issues/1072), unfortunately GitHub has yet to implement anything that would help us there.

I don't see a way forward for this change now.  If we merge as is then the readme can't include badges to start with and if we include them for only workspaces pointing at `main` then any workspaces for a branch are showing incorrect data.

I'm going to close this PR because we need to make a call on how we update readmes before we can add badges.

bloodearnest:
I checked manually, and my browser does send a referer header that would be enough

ghickman:
I tested locally with ngrok and didn't get a referer header for the badge URL.  Did you check the request made for an image link in a markdown file?

bloodearnest:
Looks like I tested the wrong thing, sorry.

So supporting auto updating badges by default is tricker when there multiple workspaces (another reason to ban them)

We could show total jobs/released/published for all workspaces for this repo. That will still be live.
 
And if there is >1 workspace for a repo, could we link to a (new?) page that let you choose the workspace?

PR #1992, opened 2022-08-03
Wrap up asset creation in skipable just targets

ghickman:
This copies our python stack in flow, where each target compares some
file modification times, and only runs the target if the check file is
older than the source.

### DIFF
--- a/justfile
+++ b/justfile
@@ -150,8 +150,56 @@ run: devenv
     $BIN/python manage.py runserver
 
 
-# update npm deps, build payload, and collect for Django
-rebuild-static:
+# Remove built assets and collected static files
+assets-clean:
+    rm -rf assets/dist
+    rm -rf staticfiles
+
+
+# Install the Node.js dependencies
+assets-install:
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -150,8 +150,55 @@ run: devenv
...
+    #!/usr/bin/env bash
+    set -eu
+
+    # get most recently modified src file
+    newest=$(find assets/src -type f | xargs ls -tr | tail -n 1)
+
+    # exit if most recently modified file in assets/src has not changed since
+    # we built the assets. -nt == "newer than", but we negate with || to avoid
+    # error exit code
+    test $newest -nt assets/dist/.written || exit 0
```
bloodearnest, line 183: I think this might be simplified with single `find`:

`find assets/src -type f -newer assets/dist/.written | grep -q . || exit 0`

This will exit with a 1 if we find no files that are newer than .written, and with 0 if we do.

bloodearnest (approved):
Looks good, one suggestion

PR #1997, opened 2022-08-05
Researcher agreement

### LINKED ISSUE #1988: Publishing Repos: Researcher Agreement
### User story
As the owner of a project with a private repo who has chosen to make my repo public, I have a way to register that choice in the system, so that I can make my repo public with minimal effort

### Further info
Tom O already has a clear idea about how this should look and can wireframe it for us

### Acceptance Criteria
* A link takes the user to a page where they can agree to make their team’s repo public
* The page must clearly display the names of all the workspaces associated with the repo, with a link from each workspace to the branch on GitHub, and the names of their owners (currently creators)
* There must be a way to get the same key information that is communicated via email at 11 months (implementation: the email may direct here?) (I think - ask Tom O for advice on this from a UX perspective!)
* The user must also actively agree that the READMEs on each branch of the repo related to their workspaces are correct (follows OpenSAFELY guidelines)
* Once all users have clicked “agree” an automatic email confirmation must be sent to all workspace owners (the same as those who would receive the email at 11 months - though that email will not have been sent if this is done prior to 11 months)
* Once all users have clicked “agree” Amir must be automatically alerted somehow (whether this should be an email or another route is still to be determined - decision should be made based on things like technical feasibility, what is easiest for Amir, and how it fits into the process for Amir giving approval) - SUGGESTION: change this to auto-alert to co-pilots slack channel when there are no outputs to GitHub
* The page, user journey, and content must have been reviewed and approved by Amir

ghickman:
This adds the researcher side of signing off a repo, in preparation for it to be converted to a public repo.

It adds 3 core features:
* a feature flag
* the repo sign off page, linked to all the related workspaces
* a banner on the workspace detail page pointing the sign off page

The UX of the sign off page and workspace banner need to be reviewed by Amir and we'd like to get them in front of some internal researchers so we're going to merge them as is (modulo any glaring mistakes!).

The feature flag is so we can roll this out on a couple of repos used by internal researchers to check our assumptions around usage.

There is some code tidy up to do from this, predominantly leaning on the Repo model for everything except the RepoList view.  That will come in a subsequent PR.

Fixes #1988


bloodearnest (approved):
This is big, and messy due the inherent mess with repos/workspaces, but it seems straightforward solution to that problem to me

PR #2002, opened 2022-08-09
Remove ProjectCoordinator role

### LINKED ISSUE #1982: Remove the ProjectCoordinator role
This role has been effectively deprecated since we introduced it.  We realised that allowing external users the ability to invite more users to the platform would bypass our IG checks while they are not part of job-server so we made sure to not give this role to anyone other than Amir.  We have since set up members on Projects via the Staff Area so the invitation machinery has stayed unused.

We should:
* Remove the ProjectCoordinator role
* Remove the ability to invite users to Projects

ghickman:
This removes the ProjectCoordinator role, a role designed for a user to administer projects.  We found with use of the system that handing this over to external users was an issue for our IG so we ended up not giving it out at all and instead using the Staff Area to manage projects.

Along with the role I've pulled out the Project invitiation system, which meant the non-staff-area-settings page has also gone (it makes no sense for a standard user to be able to edit a Project name).

---

While pulling out the templates for pages mentioned above I found some unused templates which I've also removed.

Fixes #1982


lucyb (commented):
Louis has been updating the jobs site docs. I don't think your changes affect what he's done, but I'm sharing for awareness: https://github.com/opensafely/documentation/pull/897/files

lucyb (approved):
Thank you for simplifying a whole bunch of code :)

PR #2016, opened 2022-08-11
Add more honeycomb template links

madwort:
* add to job_detail, job_request_detail and workspace
* some simple tests
* Replaces https://github.com/opensafely-core/job-server/pull/1898


ghickman (approved):
Great to see more honeycomb links going in!

@bloodearnest mentioned we will probably pull in visualisations instead of these at some point.  But if we add any more I think it's worth adding a query&link builder. Debugging one of them recently did not spark joy 😆

PR #2018, opened 2022-08-15
extend base exclude with various project and tool directories

ghickman:
This switches us to `extend-exclude` so we get the base flake8 excludes for free and adds various project and tooling-related directories to the ignore.  Locally this dropped flake8 runtime from ~6.5 to ~4.5s for me.


lucyb (approved):

bloodearnest:
Given the size of the exclude list, is there a way to have an allowlist of includes? That feels more maintainable to me.

lucyb:
> Given the size of the exclude list, is there a way to have an allowlist of includes? That feels more maintainable to me.

There doesn't seem to be. It's also the sort of thing we're likely to forget to update if we add new directories.

PR #2031, opened 2022-08-16
Store size of ReleaseFile on upload

ghickman:
This adds size of a ReleaseFile as a column to the model, replacing the file-on-disk-lookup-wrapped-in-a-property.  File size is set on API upload and there's now a custom `__format__` which takes size units to get the correct output.


jobserver/models/outputs.py
```diff
@@ -142,6 +143,19 @@ class Meta:
             )
         ]
 
+    def __format__(self, format_spec):
+        match format_spec:
+            case "b":
+                return f"{self.size}b"
```
bloodearnest, line 149:
```suggestion
                return f"{self.size:,}b"
```

Large bytes are hard to read, so let's use comma separators?

jobserver/models/outputs.py
```diff
@@ -142,6 +143,19 @@ class Meta:
...
+            case "Kb":
+                value = round(self.size / 1024, 2)
+                return f"{value}Kb"
```
bloodearnest, line 152:
```suggestion
                return f"{value:,}Kb"
```

jobserver/models/outputs.py
```diff
@@ -142,6 +143,19 @@ class Meta:
...
+            case "Mb":
+                value = round(self.size / (1024 * 1024), 2)
+                return f"{value}Mb"
```
bloodearnest, line 155:
```suggestion
                return f"{value:,}Mb"
```

tests/unit/jobserver/models/test_outputs.py
```diff
@@ -188,12 +188,17 @@ def test_releasefile_ulid():
...
+    release = ReleaseFactory(ReleaseUploadsFactory({"file1.txt": b"a"}))
+
     rfile = release.files.first()
-    assert rfile.size == pytest.approx(0.001, 0.1)
-    rfile.absolute_path().unlink()
-    assert rfile.size == 0
+    rfile.size = 3.2 * 1024 * 1024  # 3.2Mb
+    rfile.save()
+
+    assert f"{rfile:b}" == "3355443.2b"
```
bloodearnest, line 198:
```suggestion
    assert f"{rfile:b}" == "3,355,443.2b"
```

tests/unit/jobserver/models/test_outputs.py
```diff
@@ -188,12 +188,17 @@ def test_releasefile_ulid():
...
+    assert f"{rfile:Kb}" == "3276.8Kb"
```
bloodearnest, line 199:
```suggestion
    assert f"{rfile:Kb}" == "3,276.8Kb"
```

bloodearnest (approved):
Looks good

Added suggestions to use comma separator for thousands

PR #2036, opened 2022-08-19
Handle new-style releases and their uploads

### LINKED ISSUE #2026: Releases UI: modify current Release creation
- modify current release to branch on json shape and create ReleaseFile before upload
   - new json shape should be still be filename keys, but should have other attributes in a 
     dict value 
   - name, hash, size are db columns, metadata is json column, as may change
   - add uploaded_at column (in migration, copy created_at -> uploaded_at)
 - modify current ReleaseFile upload api to branch on ReleaseFile existance
   - Mostly just changing `releases.handle_file_upload`
   - if exists, save file and set `uploaded_at`
   - if it does, use current code path, and assume old json shape

ghickman:
This handles "new style" releases both when creating the release and when uploading the files to that release.

"new style" here refers how we're going to create a Release and N ReleaseFile objects when the release is first requested, allowing for reviews of the files to be created (handled in a subsequent change), and then upload the actual file content at a later stage, linking that to the relevant ReleaseFile objects via the path (as we do now).

Fixes #2026


jobserver/releases.py
```diff
@@ -101,9 +118,17 @@ def check_not_already_uploaded(filename, filehash, backend):
 
 @transaction.atomic
 def create_release(workspace, backend, created_by, requested_files, **kwargs):
-
-    for filename, filehash in requested_files.items():
-        check_not_already_uploaded(filename, filehash, backend)
+    # we infer a new style payload from it being a dict or list.  Once
+    # release-hatch and SPA are both migrated to the new style of doing things
```
bloodearnest, line 122:
```suggestion
    # the new style release UI is completed and in use,
```

jobserver/releases.py
```diff
@@ -112,6 +137,20 @@ def create_release(workspace, backend, created_by, requested_files, **kwargs):
...
     )
+
+    if new_style:
+        for f in requested_files:
+            ReleaseFile.objects.create(
+                release=release,
+                workspace=release.workspace,
+                created_by=created_by,
+                name=f["name"],
+                path=f["path"],
```
bloodearnest, line 148: There is no path field in the json, just name, which is effectively the path.

https://github.com/opensafely-core/release-hatch/blob/main/hatch/schema.py#L27

bloodearnest: The path for a release file seems to be path on the server, so should be NULL here?

jobserver/releases.py
```diff
@@ -125,38 +164,60 @@ def handle_file_upload(release, backend, user, upload, filename, **kwargs):
...
+            raise
+
+    # New flow
+
+    # _is_ on disk
+    if existing.absolute_path().exists():
+        # existence of a ReleaseFile isn't an error case now but a file-on-disk
+        # having already been uploaded is
+        raise ReleaseFileAlreadyExists(
+            f"This version of '{filename}' has already been uploaded from backend '{backend.slug}'"
```
bloodearnest, line 207: I wonder if this should be a 4xx rather than a 5xx?

The client has done something wrong here

jobserver/releases.py
```diff
@@ -125,38 +164,60 @@ def handle_file_upload(release, backend, user, upload, filename, **kwargs):
...
 
-    return rfile
+    # We have a ReleaseFile but no file-on-disk, but we still need to confirm
+    # the uploaded files hash matches what was sent to us when the ReleaseFile
+    # and Release were created.
+    if existing.filehash != calculated_hash:
+        raise Exception
+
+    absolute_path.write_bytes(data)
+    existing.uploaded_at = now()
```
bloodearnest, line 217: Need to also add this I think:
```
existing.path = str(relative_path)
```

jobserver/releases.py
```diff
@@ -125,38 +164,60 @@ def handle_file_upload(release, backend, user, upload, filename, **kwargs):
...
+    existing.save(update_fields=["uploaded_at"])
```
bloodearnest, line 218:
```suggestion
    existing.save(update_fields=["uploaded_at", "path"])
```

jobserver/releases.py
```diff
@@ -125,38 +164,60 @@ def handle_file_upload(release, backend, user, upload, filename, **kwargs):
...
-            workspace=release.workspace,
-            created_by=user,
-            name=filename,
-            path=str(relative_path),
-            filehash=calculated_hash,
-            mtime=mtime,
-            size=absolute_path.stat().st_size,
-            **kwargs,
+    # Check if this filename for this release has been uploaded before.
+    existing = ReleaseFile.objects.filter(
```
bloodearnest, line 170: `existing` is the wrong name here I think. It only makes sense for the following if clause, it reads a bit oddly for the rest of the code.

I'd prefer `rfile` or similar?

bloodearnest (changes requested):
Looks good. Couple of nit picks, and one issue that needs addressing

There's no "path" field in the json, just "name", which is in fact a path. Sorry.

Either way, ReleaseFile.path should be set to `str(relative_path)` same as old flow. I've suggested in comments to set it on upload. So there's no .path if it's never uploaded. But you may prefer to set ahead of time anyway.




2) I think either a) ReleaseFile.path needs setting on upload, not on creation and thus be nullable, or b) ReleaseFile.path needs sett

jobserver/releases.py
ghickman: I think that's what it's doing, we're catching that exception in the API and raising a ValidationError from it which I would expect to give a 4xx?

bloodearnest: ok, great

bloodearnest (approved):

PR #2040, opened 2022-08-22
Add ReleaseFile review API

### LINKED ISSUE #2028: Releases UI: Add approve/reject API and models
New ReleaseFileReview model
  - columns: status (unknown/approved/reject), comments (text), created_at/by
  - fk to ReleaseFile
  - must share same `created_at` across all instances created in a review POST request, rather than the ORM default.

New API
  - url: /releases/release/<released_id>/reviews (as release-hatch only knows the release id, not the file ids)
  - accepts json dict with filename keys and dict values, similar to other APIs
  - created multiple ReleaseFileReview instances as needed.
  - does not require *all* files to be included in the json (to support reviewing files in batches)

ghickman:
This adds the ReleaseFileReview model, extra review-related fields to the Release and ReleaseFile, and the API handler for submitting reviews.

Fixes #2028


jobserver/api/releases.py
```diff
@@ -259,6 +303,48 @@ def get(self, request, file_id):
...
+                validate_files(db_files, payload_files)
+            except UnknownFiles as e:
+                sentry_sdk.capture_exception(e)
+                raise ValidationError(e)
+
+            # create review objects for each file in the files list
+            for file in payload_files:
+                rfile = db_files.get(name=file["name"], filehash=file["sha256"])
+
+                created_at = timezone.now()
```
bloodearnest, line 337: Move this timezone.now() to be outside the loop, so it's the same for all ReleaseFileReview's created by this review?

jobserver/api/releases.py
```diff
@@ -259,6 +303,48 @@ def get(self, request, file_id):
...
+    def post(self, request, *args, **kwargs):
+        release = get_object_or_404(Release, id=self.kwargs["release_id"])
+        db_files = release.files.all()
+
+        serializer = self.ReleaseSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        payload_files = serializer.validated_data["files"]
+
+        with transaction.atomic():
+            release.metadata = serializer.validated_data["metadata"]
```
bloodearnest, line 323: Metadata is read only at this point - it should only be set when the release is created, as the end user sets it.

bloodearnest (commented):
One minor thing needs changing (don't set Release.metadata), and few suggestions

jobserver/api/releases.py
ghickman: Yep 🤦‍♂️

bloodearnest (approved):

PR #2046, opened 2022-08-23
Create GitHub issue on release

### LINKED ISSUE #2027: Releases UI: Create Github issue when Release created
This should use the current [issue templates](https://github.com/ebmdatalab/opensafely-output-review/issues/new/choose) as a base, and possibly enhance it since we have info about the files.

Talk to @LFISHER7 about the specifics (e.g. internal vs external)

Maybe include a direct link to the appropriate SPA page (which will only work on level 4)

We should store the issue url in the db.

ghickman:
This creates an issue on GitHub for the release when it's created via the API.  It uses the templates from the output reviewing repo.

There's one open question around how we decide if a user is considered internal but the rest can be reviewed now.

Fixes #2027

### DIFF
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -33,6 +33,8 @@ from jobserver.models import (
 from jobserver.releases import serve_file
 from jobserver.utils import set_from_qs
 
+from ..github import _get_github_api
+
 
 logger = structlog.get_logger(__name__)
 
@@ -212,6 +214,7 @@ class ReleaseWorkspaceAPI(APIView):
     """Listing current files and creating new Releases for a workspace."""
 
     authentication_classes = [SessionAuthentication]
+    get_github_api = staticmethod(_get_github_api)
     parsers = [JSONParser]
 
     def post(self, request, workspace_name):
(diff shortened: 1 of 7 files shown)


bloodearnest (commented):
Some thoughts:

I think the release id should be included in the issue title and body.

Ideally, we'd also have direct links to view the release.

Obviously, you'll only be able to view the SPA view of the release on level 4.

Is our current non-SPA release view on job-server directly linkable? If so would be good to add that.


We should generate the list of files for the internal issue, if that's a requirement.

ghickman:
@LFISHER7 – are you happy with all of Simon's suggestions?

LFISHER7:
Those suggestions sound good to me. For internal reviews, I don't think the list of files is needed; we have only included these because internal reviews have not attached the word form with the list of files.

ghickman:
We know the list of files in the system… should we put that in both templates?

ghickman:
@bloodearnest – updated to include most of the points above.  I've also added titles to all the links to improve readability.

bloodearnest (approved):

PR #2047, opened 2022-08-23
Use the new API for first leg of releasing

ghickman:
This moves the creation of release to our "standardised" payload API for releasing files (where the data shape is the same for several actions).

I've left the forking logic in `create_release` because all the current factories make use of it and I want to unblock E22 testing.


jobserver/migrations/0045_use_django_json_encode_for_release_requested_files.py
```diff
@@ -0,0 +1,21 @@
...
+    dependencies = [
+        ("jobserver", "0044_add_review_fields_to_release_and_releasefile"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="release",
+            name="requested_files",
+            field=models.JSONField(
+                encoder=django.core.serializers.json.DjangoJSONEncoder
```
bloodearnest, line 18: Huh, presumably this is a migration because it needs to load/dump all the current json stored in the db?

bloodearnest (approved):

jobserver/migrations/0045_use_django_json_encode_for_release_requested_files.py
ghickman: This is because the model definition changed.  Django migrations track all changes to models, a subset of which change the db state, so that they can auto-generate subsequent migrations.

PR #2054, opened 2022-08-25
Build a testing onion for Release[File]s

### LINKED ISSUE #2038: Convert Releases factories to FactoryBoy factories
When we added the outputs code originally we required that ReleaseFiles were created with files on disk, relying on the lack of a file to denote "this has been redacted".  Build factories to support this flow with FactoryBoy was difficult and produced confusing code.  We opted to build custom objects that looked a lot like FactoryBoy factories but used our business logic under the hood to create the appropriate model instances and on-disk file(s).

With the shift to our new flow where we create a Release and the relevant ReleaseFiles for files a researcher would like to have reviewed (ie with no file-on-disk) we need a way to test our code with a few more options.  The current factories have some downsides, they require you create files, and they don't always let you set all the fields (because they're trying to be helpful and set them for you).

However, the new flow means we should be able to move FactoryBoy factories with some pytest fixtures to make setting up all the pieces in one go easier.

We should end up with these factories:
* `ReleaseFactory`
* `ReleaseFileFactory`
* `ReleaseFileReviewFactory`

and these fixtures:
* `create_release` -> takes a list strings (filenames), and kwargs for ReleaseFactory.  Creates the release, and relevant ReleaseFiles, with on-disk files, roughly mirroring our current set up.
* `create_release_file` -> creates a release file (and file on-disk) with the given release, path, and name.  This would be called by `create_release` for each of the requested files.

We might also need an easy way to create a Release and ReleaseFiles with existing files.

I've spiked this new layout as part of the Releases UI work and put the existing work on the `new-style-release-factories` branch.

ghickman:
This switches all our tests which touch Release and/or ReleaseFile objects to use some combination of FactoryBoy factories and fixtures which wrap up the logic necessary for combining them.

The goal here was to support our recent full use of the two-legged upload process whereby we can have ReleaseFiles without files on disk.  This means many of our tests can have greatly reduced setup (as little as a bare factory invocation) or have the creation of the appropriate objects abstracted to a fixture.

The fixtures are made up of a few layers of functions with low level ones like `file_content` generating random byte strings and high level ones like `release` which generates all the pieces required to have a Release, a related ReleaseFile, and a file on-disk.

I've laid out how this all works in TESTING.md, which I hope is clear enough to fully explain it!

Reviewing this is going to be a bit of a pig I'm afraid.  I've pulled everything I can into separate commits but there's still one very meaty commit where I move all the tests over.

Fixes #2038

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,9 +1,14 @@
+import hashlib
 import os
+import random
+import string
 import textwrap
+from pathlib import Path
 
 import pytest
 import structlog
 from django.conf import settings
+from django.utils import timezone
 from structlog.testing import LogCapture
 
 import services.slack
@@ -12,6 +17,7 @@ from jobserver.authorization.roles import CoreDeveloper
 
 from .factories import OrgFactory, OrgMembershipFactory, UserFactory
 from .factories import applications as application_factories
+from .factories.releases import ReleaseFactory, ReleaseFileFactory
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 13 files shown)


tests/conftest.py
```diff
@@ -83,6 +94,97 @@ def incomplete_application():
...
+        # write the file to disk
+        rfile.absolute_path().write_bytes(file_content)
+
+        return rfile
+
+    return func
+
+
+@pytest.fixture
+def build_release_path(tmp_path):
```
lucyb, line 157: Still reading through (had a few meetings, have one more), but I was wondering if `build_release_path` and `build_release_file` needed to be fixtures? They're only used in conftest, so they feel like utility functions?

bloodearnest (approved):
Nice clean up

TBH, I'm -0 on the pytest fixtures as they currently are. I think I'd prefer adding methods on the factory object or similar to create the right models/files, rather than *having* to use pyteset fixture. I'm not really sure what that gives us. 

The only things it really gets from the fixture is the tmp_path - file contents doesn't really need to be a fixture IMO

That said, this is an improvement on the current fixtures post-two-leg-refactor, so I'm happy to discuss more later on.

tests/conftest.py
ghickman: They don't need to be fixtures but I exposed them in case they were of use in the future.  `content_file` for example ended up being really useful for example.

PR #2055, opened 2022-08-25
Build a prototype releases UI

tomodwyer:
🚧


bloodearnest (approved):
Very light review from me, as there's various react techniques I'm not familiar with, but the high level structure makes some sense to me

PR #2057, opened 2022-08-25
feat: serve helpful content to SPA for missing files

bloodearnest:
ReleaseFiles might not be able to be served: if they have been redacted,
if they have not yet been uploaded, and if the file is simply not there.

Currently, all the above cases, the API returns a 404, for which the SPA
displays a generic page and pop up error. This doesn't give the user
much clue as to why the file cannot be served.

This change serves place holder content, specific to each of the above
cases, which informs the user why they cannot view the file contents.

 - it returns these as 200s, rather than the more correct 404, as I am
   not sure if the SPA can handle 404s with content.
 - it returns a plain text message. This might be prettier as a html
   template with links, perhaps.
 - it assumes the SPA displays the content-type in the response, rather
   than based on filename.

### DIFF
--- a/jobserver/releases.py
+++ b/jobserver/releases.py
@@ -4,12 +4,12 @@ import zipfile
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
 
+import sentry_sdk
 from django.db import transaction
 from django.http import FileResponse
 from django.utils.http import http_date
 from django.utils.timezone import now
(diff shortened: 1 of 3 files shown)


ghickman (approved):
Love it 👍

bloodearnest:
It doesn't work with the SPA at all sadly

PR #2058, opened 2022-08-25
Display review data

ghickman:
Add per-Release and per-ReleaseFile metadata

**Release**
We're just saving raw data to the metadata for Release currently so this isn't the most useful but we can at least show that we've captured parts of it?

**ReleaseFile**

### DIFF
--- a/assets/src/scripts/outputs-viewer/App.jsx
+++ b/assets/src/scripts/outputs-viewer/App.jsx
@@ -141,6 +141,7 @@ function App({
                   fileName={selectedFile.name}
                   fileSize={selectedFile.size}
                   fileUrl={selectedFile.url}
+                  metadata={selectedFile.metadata}
                 />
                 <Card.Body>
                   <Viewer
--- a/assets/src/scripts/outputs-viewer/components/Metadata/Metadata.jsx
+++ b/assets/src/scripts/outputs-viewer/components/Metadata/Metadata.jsx
@@ -2,7 +2,7 @@ import React from "react";
 import prettyFileSize from "../../utils/pretty-file-size";
 import { selectedFileProps } from "../../utils/props";
 
-function Metadata({ fileDate, fileName, fileSize, fileUrl }) {
+function Metadata({ fileDate, fileName, fileSize, fileUrl, metadata }) {
   const fileDateValue = () => {
     if (Date.parse(fileDate)) {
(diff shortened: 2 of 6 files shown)


assets/src/scripts/outputs-viewer/components/Metadata/Metadata.jsx
```diff
@@ -46,6 +53,19 @@ function Metadata({ fileDate, fileName, fileSize, fileUrl }) {
         )}
         <li className="list-inline-item spacer">{prettyFileSize(fileSize)}</li>
       </ul>
+      <div>
+        <div className="mb-3">
+          <strong>What does this file contain?</strong>
```
bloodearnest, line 58: This should be the key value of the dict

assets/src/scripts/outputs-viewer/components/Metadata/Metadata.jsx
```diff
@@ -46,6 +53,19 @@ function Metadata({ fileDate, fileName, fileSize, fileUrl }) {
...
+          <div>{context}</div>
+        </div>
+
+        <div>
+          <strong>
+            How has the author ensured that this file is non-disclosive?
```
bloodearnest, line 64: Same as above.

assets/src/scripts/outputs-viewer/App.jsx
```diff
@@ -137,6 +137,8 @@ function App({
               {isReviewEdit ? <FileReview selectedFile={selectedFile} /> : null}
               <Card>
                 <Metadata
+                  context={selectedFile.metadata.context}
```
bloodearnest, line 140: This should be just `metadata=metadata`

The jsx component should iterate through the key/value pairs.

To match the current document, we'd expect a dict like:

````
{
    "Context": "user input context",
    "Controls": "user input controls",
}
```

bloodearnest (changes requested):

ghickman:
Updated to show what it looks like.  I'm not sold on using the keys for the labels as that ties the display to the labels of the form, but I've updated it to that so we can see what it looks like for the demo.

bloodearnest:
The keys have no semantic meaning - the can be the whole text of the form label if they want.

In fact, a dict of labels: values is a much better terminology

bloodearnest (approved):

PR #2059, opened 2022-08-26
Use deleted_at to drive is_deleted

ghickman:
Now that ReleaseFiles are created without a file on-disk we want ReleaseFile.is_deleted to use one of the two deleted_* fields we explicity set when deleting.  Since those fields are required to both be filled _or_ empty by the db check constraint I've picked `deleted_at` since it's a cheaper field to look up.

### DIFF
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -198,7 +198,7 @@ def generate_index(files):
                 date=rfile.created_at.isoformat(),
                 sha256=rfile.filehash,
                 size=get_size(rfile),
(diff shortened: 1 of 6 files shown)


bloodearnest (approved):
This is an improvement on the current situation.

I guess we should use is_deleted in other places (e.g. serve_file) as the official way to check if a file is supposed to be deleted.

ghickman:
Ah yes, I'll add that to this PR too since the idea was to improve correctness!

PR #2060, opened 2022-08-26
Validate file list apis

### LINKED ISSUE #2049: Validate generate_index() with Serializers
We're consolidating all communication with release-hatch, job-server and the SPA around the structure validated by the 3 serializers: `ReleaseSerializer`, `FileSerializer`, `ReviewSerializer`.  release-hatch validates these with pydantic models and we'll investigate consolidating these at a later date.

Currently the listing of files in job-server is handled by the [generate_index](https://github.com/opensafely-core/job-server/blob/95eca99cb60c2e9b8e81187bcc67ad2c36711eec/jobserver/api/releases.py#L182-L206) function.  We should validate the output of that function with the serializers.  This saves us having to wire up and translate all the various model fields for the serializer fields.

ghickman:
Use FileSerializer to validate the output of generate_index to ensure we match the incoming payloads for various APIs.

I've avoided using the serializer to pull the field data since this feels a lot more readable.

Small drive-by fix on getting file size from the db as I was there too.

Fixes #2049

### DIFF
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -182,12 +182,6 @@ def validate_snapshot_access(request, snapshot):
 def generate_index(files):
     """Generate a JSON list of files as expected by the SPA."""
 
-    def get_size(rfile):
-        try:
-            return rfile.absolute_path().stat().st_size
-        except FileNotFoundError:  # pragma: no cover
-            return None
-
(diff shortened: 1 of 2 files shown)


bloodearnest (approved):
Nice

PR #2091, opened 2022-09-05
Rename maintenance mode url

### LINKED ISSUE #2081: Improve TPP maintenance mode alert
Thanks to some feedback from Dave, 

Rename the freshping alert, so that it appears as:

> TPP Maintenance Mode: DB service is down
> TPP Maintenance Mode: DB service is up

Change the current job server url from:

> https://jobs.opensafely.org/status/tpp/db-availability/ 

to:
> https://jobs.opensafely.org/status/tpp/maintenance-mode/

lucyb:
This endpoint shows up in the alert/message when maintenance mode is switched
on. We hope to make that message clearer by renaming the endpoint.

Note: fresh ping will need to be updated when this is merged.

Fixes #2081

### DIFF
--- a/jobserver/urls.py
+++ b/jobserver/urls.py
@@ -48,7 +48,7 @@ from .views.releases import (
     SnapshotDownload,
     WorkspaceReleaseList,
 )
-from .views.status import DBAvailability, PerBackendStatus, Status
+from .views.status import MaintenanceMode, PerBackendStatus, Status
 from .views.users import Settings, login_view
 from .views.workspaces import (
     WorkspaceArchiveToggle,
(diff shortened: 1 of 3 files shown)


ghickman (approved):
Code looks fine 👍

I'm not sure about the naming though.  We had exactly this discussion when first adding this feature!  When maintenance mode is on, we'll now get a 503 from `/maintenance-mode`.  At the time we decided a failure code when the mode was on would be confusing.  I don't see any discussion in the ticket around that, was that discussed elsewhere?

lucyb:
> Code looks fine +1
> 
> I'm not sure about the naming though. We had exactly this discussion when first adding this feature! When maintenance mode is on, we'll now get a 503 from `/maintenance-mode`. At the time we decided a failure code when the mode was on would be confusing. I don't see any discussion in the ticket around that, was that discussed elsewhere?

Hmm, yes you're right. I think we're going round in circles a bit. 

Sorry for not linking to the thread, it's [here](https://bennettoxford.slack.com/archives/C0270Q313H7/p1661761793916429). Basically, Dave had the same questions that we had to start with. Tom O has now changed the alert name to "TPP Maintenance Mode: DB service", which may resolve the issue. 

Happy to bin this change and call it done for now. Don't want to make things more confusing again.

PR #2103, opened 2022-09-08
Add orgs to the JobRequest API endpoint

ghickman:
This adds the `orgs` field to the JobRequest API endpoint so we can pass it through to job-runner to enhance our metrics with.

We're using a `SerializerMethodField` here because a JobRequest only has one Org currently (via it's Workspace then Project) but in the future we expect that to change so we're adding a little future proofing.


bloodearnest (commented):
Can you add project too like https://github.com/opensafely-core/job-server/pull/2095/files?

I think you'll need the select related changes and also the test too?

Or shall we just merge this and I can rebase mine?

bloodearnest:
actually, I could just pull this change into my PR?

ghickman:
@bloodearnest – yeah just pull this change into yours and I'll close this one 👍

PR #2137, opened 2022-09-26
Display number of running and queued jobs

### LINKED ISSUE #2090: Display the number of running and queued jobs on the job-server status page
Display the number of running and queued jobs on the job-server [status page](https://jobs.opensafely.org/status/) for each backend. So we can set researchers' expectations around how long their jobs might be queued for and so anyone on tech support can easily identify if there are potential problems with a particular backend.

lucyb:
Display the number of running and queued jobs on the job-server [status page](https://jobs.opensafely.org/status/) for each backend. This should allow us to set researchers' expectations around how long their jobs might be queued for and allow anyone on tech support to more easily identify if there are potential problems with a particular backend.

Fixes #2090

Screenshot from 2022-09-26 17-17-13

### DIFF
--- a/jobserver/views/status.py
+++ b/jobserver/views/status.py
@@ -68,12 +68,20 @@ class Status(View):
                 .filter(num_jobs=0)
                 .count()
             )
-            running = Job.objects.filter(
-                job_request__backend=backend, status="running"
-            ).count()
-            pending = Job.objects.filter(
-                job_request__backend=backend, status="pending"
-            ).count()
+            counts_by_status = (
+                Job.objects.filter(
+                    job_request__backend=backend,
+                    status__in=["running", "pending"],
+                )
+                .values("status")
+                .annotate(count=Count("status"))
+            )
+            pending = running = 0
+            for result in counts_by_status:
+                if result["status"] == "pending":
+                    pending = result["count"]
+                if result["status"] == "running":
+                    running = result["count"]
 
             try:
                 last_seen = (
--- a/jobserver/views/status.py
+++ b/jobserver/views/status.py
@@ -7,7 +7,7 @@ from django.utils import timezone
 from django.views.generic import View
 
 from ..backends import show_warning
-from ..models import Backend
+from ..models import Backend, Job
 
 
 class DBAvailability(View):
@@ -68,6 +68,12 @@ class Status(View):
                 .filter(num_jobs=0)
                 .count()
             )
+            running = Job.objects.filter(
+                job_request__backend=backend, status="running"
+            ).count()
+            pending = Job.objects.filter(
+                job_request__backend=backend, status="pending"
+            ).count()
 
             try:
                 last_seen = (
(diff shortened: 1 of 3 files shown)


jobserver/views/status.py
```diff
@@ -68,6 +68,11 @@ def get_stats(backend):
                 .filter(num_jobs=0)
                 .count()
             )
+            running = 0
+            pending = 0
+            for jr in backend.job_requests.all():
+                running += jr.jobs.filter(status="running").count()
+                pending += jr.jobs.filter(status="pending").count()
```
ghickman, line 75: This will execute two queries per JobRequest.  You can traverse the Job to JobRequest relationship in the filter with a `__` (double underscore) to have the database get all data you want, like this:

```
running = Job.objects.filter(job_request__backend=backend, status="running").count()
pending = Job.objects.filter(job_request__backend=backend, status="pending").count()
```

ghickman (changes requested):

jobserver/views/status.py
lucyb: Thank you!!

evansd: While we're query golfing, you should be able to do it one using something like (untested):
```py
from django.db.models import Count

Job.objects.filter(
    job_request__backend=backend,
    status__in=["running", "pending"],
).values("status").annotate(count=Count("status"))
```
Which will result in a single `COUNT()` query with a `GROUP BY` on `status`.

ghickman: Very nice!  I so rarely use `.values()` that I'd not even considered you could use `.annotate()` with it 👌

evansd: You'd struggle to call it a transparently obvious syntax, wouldn't you? I just remember needing to do this before.

lucyb: Thank you

I did experiment with this syntax when I was looking at JobRequest rather than Job. It's really useful, if hard to read. I'm not sure it's worth the additional complexity, given that performance isn't particularly important on this page?

ghickman: I'd say keep this one and add a comment to say it's getting a dictionary of status and count.  I'd certainly like an easy place to reference in the future when I need this again.  While this page isn't heavily used by external users, it is something we use for debugging so having it be performant under heavy load is always useful.

bloodearnest (commented):
I think that adding to the status page is useful, but I don't think users really look at this page.

I think that to manage user expectations, this info should ideally also be shown on the submit jobs page, when its most relevant.

E.g. in the backend selection list, we could maybe alter the radio button text from `TPP` to `TPP (4 running, 2 in queue)` or similar?

Might be a separate PR, but I think worth doing to actually get the user benefit

jobserver/views/status.py
lucyb: Right, I've pushed the change now..

@ghickman Is this clear enough without needing a code comment? Is there a clearer name I can use for `counts_by_status`?

PR #2138, opened 2022-09-27
Use our User.name property for string representations of a User

ghickman:
We use the string repr in the Staff Area's search so it's useful to be consistent with how Users are displayed in various other places.

### DIFF
--- a/jobserver/models/core.py
+++ b/jobserver/models/core.py
@@ -682,6 +682,9 @@ class User(AbstractBaseUser):
             ),
         ]
 
+    def __str__(self):
+        return self.name
+
     def clean(self):
         super().clean()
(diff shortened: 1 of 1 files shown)


jobserver/models/core.py
```diff
@@ -682,6 +682,9 @@ class Meta:
             ),
         ]
 
+    def __str__(self):
```
lucyb, line 685: I don't quite know where this will be used, but would it also be useful to also show the email or username in brackets after their display name?

ghickman: In this instance it's used to show any users matching a search query on the Staff Area index page.  We typically use `self.name` for a user around the site, both in the Staff Area and for general use.  Where we're displaying this, it's the text for links to user detail pages so probably not useful there.

PR #2139, opened 2022-09-28
Handle longer backend names on log pages

ghickman:
This expands the backend column to account for the graphnet-test backend:

**before:**

**after:**


tomodwyer (approved):
Approving this quick fix, but added a ticket to fix the underlying issue (of not using a `<table>` for this tabular data)

PR #2141, opened 2022-09-28
Repo detail fixes for the Staff Area

ghickman:


lucyb (approved):

staff/templates/staff/repo_detail.html
```diff
@@ -36,13 +36,13 @@ <h1 class="display-4">{{ repo.name }}</h1>
       <li>
         <strong>Status:</strong>
         <span>
-          {% if repo.is_private %}Private{% else %}Public{% endif %}
+          {% if repo.private %}Private{% else %}Public{% endif %}
```
lucyb, line 39: `repo.is_private` is used in other places, like the project_detail template. Does that need to be changed too?

ghickman: No, those are all looking up `is_private` from an API query in the view.  This view is Extra Special™ and `repo` here is the full repo payload from GitHub so we can look the key up directly.  It does make me think we should put this under some kind of dashboards menu in the Staff Area though…

lucyb: Definitely. I'll continue the conversation on slack, since it's separate to this PR now.

PR #2150, opened 2022-10-03
Build production images in CI

### LINKED ISSUE #2102: Deploy image built on GHA
Our other websites have all moved over to deploying via dokkus `from-image` deployment method.  We'd like all the sites to be deployed in the same way.

We should look at abstracting our deployment workflow step out to the `.github` repo as part of this too.

We think this might also fix the ram issue detailed in #2101

ghickman:
Fixes #2102


docker/justfile
```diff
@@ -0,0 +1,47 @@
+export DOCKER_USERID := `id -u`
+export DOCKER_GROUPID := `id -g`
```
bloodearnest, line 2: Don't need these - they were for job-runner only!

docker/entrypoints/dev.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -euo pipefail
+
+./manage.py migrate
+./manage.py ensure_groups
+./manage.py collectstatic --no-input --clear | grep -v '^Deleting '
```
bloodearnest, line 7: Can we do `just collectstatic` to avoid this if it's not needed?

bloodearnest (commented):
Looks good

I think you need to move some dependencies into build-dependencies.

Also, in dev and test, rather than forcing collectstatic, can we install just into the dev image, and use its clever collectstatic logic? Make for a nicer dev/test experience with the docker stack

docker/entrypoints/dev.sh
ghickman: I gave this a solid shot and couldn't find a way to pair the just and docker stacks yet so I've left it as it is a the pragmatic way forward, and duplicated the pytest & coverage commands into the docker-compose.yaml.

docker/Dockerfile
```diff
@@ -0,0 +1,161 @@
...
+
+# Adjust this metadata to fit project. Note that the base-docker image does set
+# some basic metadata.
+LABEL org.opencontainers.image.title="job-server" \
+      org.opencontainers.image.description="Job Server" \
+      org.opencontainers.image.source="https://github.com/opensafely-core/job-server"
+
+# Expose the port to Dokku
+# https://dokku.com/docs/networking/port-management/#applications-using-expose
+EXPOSE 5000
```
bloodearnest, line 124: This is wrong? It runs on 8000 by default?

.github/workflows/main.yml
```diff
@@ -121,14 +148,31 @@ jobs:
...
-        uses: opensafely-core/dokku-deploy-github-action@v1
-        with:
-          app-name: "job-server"
-          dokku-host: ${{ secrets.DOKKU_HOST }}
-          ssh-private-key: ${{ secrets.DOKKU2_DEPLOY_SSH_KEY }}
-          remote-branch: "main"
+      - name: Build docker image
+        run: |
+          # docker-test step will build the dev image by default, so build the prod image
+          just docker-build prod
```
bloodearnest, line 156: We need to test the prod image in some way here. It's currently broken, but the tests don't catch that.

We should probably have a just docker-smoketest, which builds and runs prod service, and makes sure we can hit it and get 200 response?

docker/Dockerfile
```diff
@@ -0,0 +1,161 @@
...
+
+# copy application code
+COPY . /app
+
+# collect static files
+RUN SECRET_KEY=dummy-key python /app/manage.py collectstatic --no-input
+
+# We set command rather than entrypoint, to make it easier to run different
+# things from the cli
+CMD ["/app/docker/entrypoints/prod.sh"]
```
bloodearnest, line 134:
```suggestion
ENTRYPOINT ["/app/docker/entrypoints/prod.sh"]

# We set command rather than entrypoint, to make it easier to run different
# things from the cli
CMD ["gunicorn", "-c", "gunicorn.conf.py", "jobserver.wsgi"]
```

bloodearnest (changes requested):
The prod version of the image is broken, see comments.

Also, the `./manage.py collectstatic --no-input` needs removing from entrypoints/prod.sh

docker/Dockerfile
ghickman: We're using 5000 in production but I can't see how gunicorn is being told to use that.  There's nothing in the gunicorn config or Procfile.  Any ideas?

bloodearnest: Ahh, I missed the procfile. That's why the prod image works, out-of-band magic!

With docker image deploys procfiles are a bad idea, as dokku then ignores the configured default CMD, and passes the procfile command. My suggested commit removes the need for it, so I say delete it.

bloodearnest: I suspect the 5000 thing is *because* of the procfile somehow, as thats dokkus default port.

bloodearnest: using 5000 in prod is fine, but I've slight preference for 8000, as that's what it runs on in dev

ghickman: Updated to 8000, and the Procfile so this all becomes explicit 👍

.github/workflows/main.yml
```diff
@@ -108,11 +114,70 @@ jobs:
...
+        run: |
+          cat >> .env <<EOL
+          DATABASE_URL=postgres://user:password@172.17.0.1/jobserver
+          DJANGO_VITE_DEV_MODE=False
+          GITHUB_TOKEN=
+          SECRET_KEY=12345
+          SOCIAL_AUTH_GITHUB_KEY=test
+          SOCIAL_AUTH_GITHUB_SECRET=test
+          EOL
+          docker-compose -f docker/docker-compose.yaml up --detach prod
```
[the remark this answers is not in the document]
bloodearnest: This would work when running locally, but it would breaking in production

The issue is that gunicorn is misconfigured - it's listening on the default of 127.0.0.1, rather than 0.0.0.0

This is likely a consequence of moving away from dokku's magic Procfile config, which whilst looked similar, and used our gunicorn config, it somehow still worked.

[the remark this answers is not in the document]
bloodearnest: The fix is to add `bind = "0.0.0.0"` to gunicorn.conf.py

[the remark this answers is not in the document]
bloodearnest: Then our prod image works the same whether run locally or in dokku

lucyb:
Really small change to the docs given that the docker-compose file has moved (sorry)

```
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -55,6 +55,7 @@ If you want to maintain your own virtualenv make sure you have activated it befo
 If you're comfortable with running a database in docker-compose you can run
 
 ```sh
+cd docker
 docker-compose up db
 ```

docker/entrypoints/dev.sh
bloodearnest: I've fixed this, as part of solving the staticfiles ownership conflict

PR #2157, opened 2022-10-04
Add a normal ReposList page

### LINKED ISSUE #2156: All repos view in staff area
* [ ] Create a dashboards menu item
* [ ] Move Amir's dashboard under dashboards
* [ ] Create a new repos list page in the staff area showing a list of all repos from the database

ghickman:
This pulls the existing repos page into a new dashboards menu and adds a "normal" repos page, which only lists repos that job-server knows about.

I've added orgs as a filter in leiu of anything more obvious.

Fixes #2156

### DIFF
--- a/staff/views/repos.py
+++ b/staff/views/repos.py
@@ -12,14 +12,14 @@ from django.shortcuts import get_object_or_404, redirect
 from django.template.response import TemplateResponse
 from django.utils import timezone
 from django.utils.decorators import method_decorator
-from django.views.generic import UpdateView, View
+from django.views.generic import ListView, UpdateView, View
 from first import first
 from furl import furl
 
 from jobserver.authorization import CoreDeveloper
 from jobserver.authorization.decorators import require_role
 from jobserver.github import _get_github_api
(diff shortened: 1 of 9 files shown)


staff/views/repos.py
```diff
@@ -147,113 +259,27 @@ def get_form_kwargs(self):
 
 
 @method_decorator(require_role(CoreDeveloper), name="dispatch")
-class RepoList(View):
-    get_github_api = staticmethod(_get_github_api)
-
-    @csp_exempt
```
lucyb, line 153: Do we need to make the new view `csp_exempt`?

lucyb (approved):
I think this is much better, thank you.

I think the repo detail page could do with some input from Tom O, at the moment it is very focused on getting people to 'sign off' the repo. However, that might not be relevant to someone accessing it from the repo list page. Not a big priority though!

staff/views/repos.py
ghickman: Nope, that's so the sort&filter JS on the dashboard can be loaded.

PR #2169, opened 2022-10-10
chore: upgade otel packages and add requests and psycopg2

bloodearnest:
These should be enabled by automatically in production.


ghickman (approved):
Looks great.  Is the `just run-prod` command something we can meaningfully test in CI?  If not, lets not hold this PR up looking for a solution to that.

bloodearnest:
I'm not too fussed about the just command working, but we should probably add a test docker-test to ensure it starts up and responds to requests ok?

PR #2178, opened 2022-10-11
feat: add new honeycomb direct links

bloodearnest:
This change adds honeycomb direct links to the trace for a job.

To do this, job-runner needs to send us the trace context, which we're
not doing yet, but will once this change is deployed.

I've removed all the current honeycomb links, as they are based on the
old tracing data. I've also refactored the code for rendering those
links, to make it easier to add new onces without changing the template.

We will probabaly add more template links again, once we have a better
idea of what would be useful.

This change also explicit sets up tracing in tests, same as how we've
done it in job-runner. We don't actually use this for any tests yet, as
we do not emit any manual traces to test at the minute. But it was
needed in order to be able to have the Job factory generate a suitable
mock trace_context.


jobserver/views/jobs.py
```diff
@@ -59,28 +60,44 @@ def get(self, request, *args, **kwargs):
...
+                        "trace_id": job.trace_id,
+                        "trace_start_ts": honeycomb_starttime_unix,
+                        "trace_end_ts": honeycomb_endtime_unix,
+                    }
+                )
+                # breakpoint()
+                links["Job Trace"] = trace_link.url
+
+            if links:
+                context["honeycomb_links"] = links
```
ghickman, line 99: Can we move the creation of `honeycomb_links` to before the creation of context so we don't need to mutate it please.  That way it's always in the context and we can rely on the falsey-ness of it being empty.

jobserver/views/jobs.py
```diff
@@ -59,28 +60,44 @@ def get(self, request, *args, **kwargs):
...
+            if job.trace_id:
+                trace_link = jobs_honeycomb_url / "trace"
+                trace_link.add(
+                    {
+                        "trace_id": job.trace_id,
+                        "trace_start_ts": honeycomb_starttime_unix,
+                        "trace_end_ts": honeycomb_endtime_unix,
+                    }
+                )
+                # breakpoint()
```
ghickman, line 95: Missed breakpoint

tests/conftest.py
```diff
@@ -25,11 +29,28 @@
 from .factories import applications as application_factories
 
 
+# set up tracing for tests
+provider = TracerProvider()
+trace.set_tracer_provider(provider)
+test_exporter = InMemorySpanExporter()
+provider.add_span_processor(SimpleSpanProcessor(test_exporter))
```
ghickman, line 36: Should these be in a session scoped fixture?  I'm not sure if that would have any meaningful difference…

tests/unit/jobserver/views/test_jobs.py
```diff
@@ -202,7 +198,7 @@ def test_jobdetail_with_core_developer_with_completed_at(rf):
     assert response.status_code == 200
     assert "Cancel" not in response.rendered_content
     assert "Honeycomb" in response.rendered_content
-    assert "%22end_time%22%3A1655298060%2C" in response.rendered_content
+    assert "trace_end_ts=1655298060" in response.rendered_content
```
ghickman, line 201: 👌

ghickman (approved):
One small tweak, but looking good!

jobserver/views/jobs.py
bloodearnest: done

jobserver/models/core.py
```diff
@@ -145,6 +149,18 @@ def runtime(self):
 
         return Runtime(int(hours), int(minutes), int(seconds), int(total_seconds))
 
+    @property
```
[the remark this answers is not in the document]
bloodearnest: In theory, but it's only accessed in one place currently.

jobserver/views/jobs.py
```diff
@@ -59,28 +60,40 @@ def get(self, request, *args, **kwargs):
...
-            "honeycomb_this_action_all_workspaces_link": f"https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/datasets/job-server?query=%7B%22time_range%22%3A2419200%2C%22granularity%22%3A0%2C%22breakdowns%22%3A%5B%22status%22%2C%22workspace_name%22%2C%22job_identifier%22%5D%2C%22calculations%22%3A%5B%7B%22op%22%3A%22MAX%22%2C%22column%22%3A%22current_runtime%22%7D%2C%7B%22op%22%3A%22COUNT%22%7D%5D%2C%22filters%22%3A%5B%7B%22column%22%3A%22name%22%2C%22op%22%3A%22%3D%22%2C%22value%22%3A%22update_job%22%7D%2C%7B%22column%22%3A%22action%22%2C%22op%22%3A%22%3D%22%2C%22value%22%3A%22{job.action}%22%7D%5D%2C%22filter_combination%22%3A%22AND%22%2C%22orders%22%3A%5B%7B%22column%22%3A%22current_runtime%22%2C%22op%22%3A%22MAX%22%2C%22order%22%3A%22descending%22%7D%5D%2C%22havings%22%3A%5B%5D%2C%22limit%22%3A100%7D",
+            "honeycomb_links": {},
         }
 
+        if honeycomb_can_view_links:
+            # TODO: make this configurable?
+            jobs_honeycomb_url = furl(
+                "https://ui.honeycomb.io/bennet/environments/production/datasets/jobrunner"
```
madwort, line 84: the organisation name here is incorrect, should be `bennett-institute-for-applied-data-science`

PR #2180, opened 2022-10-12
fix: add config instructions for honeycomb

madwort:


INSTALL.md
```diff
@@ -17,6 +17,10 @@ dokku config:set job-server DATABASE_URL='postgres://localhost/jobserver'
 dokku config:set job-server EMAIL_BACKEND='anymail.backends.mailgun.EmailBackend'
 dokku config:set job-server GITHUB_TOKEN='xxx'
 dokku config:set job-server MAILGUN_API_KEY='xxx'
+dokku config:set job-server OTEL_EXPORTER_OTLP_ENDPOINT='https://api.honeycomb.io'
+dokku config:set job-server OTEL_EXPORTER_OTLP_HEADERS='x-honeycomb-team=xxx,x-honeycomb-dataset=job-server'
+dokku config:set job-server OTEL_PYTHON_DISABLED_INSTRUMENTATIONS='psycopg2'
```
bloodearnest, line 22: I wouldn't encode this permanently - I hope we can reenable when we have pro account?

bloodearnest (approved):

INSTALL.md
bloodearnest: Scratch that - it's quite heavy event wise, so we should probably leave it off for now

madwort: added a comment that this is due to quota, so a future person could decide at the time whether it still applies

PR #2185, opened 2022-10-13
Restyle error pages

### LINKED ISSUE #2174: Restyle of error template
Update the error page templates on Job Server to match the new component style system.

## Acceptance criteria

- The following error page templates are updated:
  - 403.html
  - 404.html
  - 500.html
- Sign off from Seb or Ben G required

tomodwyer:
Closes #2174

Pages can be previewed at the URL corresponding to their error code.

## 400

localhost_8000_400

## 403

localhost_8000_403

## 404

localhost_8000_404

## 500

localhost_8000_500


jobserver/templates/error.html
```diff
@@ -0,0 +1,24 @@
+{% extends 'base-tw.html' %}
+
+{% block content %}
+<div class="min-h-full px-4 py-16 sm:px-6 sm:py-24 md:grid md:place-items-center lg:px-8">
+  <div class="sm:flex">
+    <p class="text-4xl font-extrabold text-oxford-600 sm:text-5xl">{{ error_code }}</p>
+    <div class="sm:ml-6">
+      <div class="sm:border-l sm:border-slate-200 sm:pl-6">
+        <h1 class="text-4xl font-extrabold text-slate-900 tracking-tight sm:text-5xl">{{ error_name }}</h1>
+        <p class="mt-2 text-base text-slate-500">{{ error_message }}</p>
```
lucyb, line 10: Technically, this text isn't quite lining up with the title and the buttons. I don't know if it's worth fixing though

jobserver/views/errors.py
```diff
@@ -0,0 +1,66 @@
+from django.template.response import TemplateResponse
+
+
+def bad_request(request, exception=None):
```
lucyb, line 4: The exception parameter isn't being used, so can be removed. We can always add it back in if needed later. Same with the other functions below.

jobserver/views/errors.py
```diff
@@ -0,0 +1,66 @@
...
+        status=400,
+        context={
+            "error_code": "400",
+            "error_name": "Bad request",
+            "error_message": "An error has occurred displaying this page.",
+        },
+    )
+
+
+def csrf_failure(request, reason=""):
```
lucyb, line 17: I couldn't see this used anywhere in the code, so went digging in Interactive to see if I could find out how it worked. That has this entry in the settings.py file

`CSRF_FAILURE_VIEW = "interactive.views.csrf_failure"`

Presumably, we need something similar here or in errors.py?

lucyb (changes requested):
This looks good. A couple of comments, mostly on the python side.

I've realised this will be the first public-facing change visible to all users. Seems like a good place to start!

jobserver/views/errors.py
ghickman: The error handling views all have to accept a request and exception 🙂  (ie [handler400](https://docs.djangoproject.com/en/4.1/ref/urls/#django.conf.urls.handler400))

sebbacon (approved):

lucyb (approved):

jobserver/templates/error.html
tomodwyer: Re-reading the code today, it was overly complex for what I was trying to achieve. Pushing an update shortly which should align everything a bit better.

tomodwyer: Pushed 🚀

PR #2188, opened 2022-10-13
switch set-output for $GITHUB_OUTPUT

ghickman:


lucyb (approved):
Thanks for doing this.

I'd forgotten why it needed to be done, so in case anyone else comes across it this, here's the announcement about deprecating set-output: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

PR #2192, opened 2022-10-14
Remove transition and scaling effects

tomodwyer:
As they are causing cross-browser and cross-device issues.


lucyb (approved):
:sparkles:

Issue #2195, opened 2022-10-17
Track Project status


ghickman:
We want to track the status of a Project using `status` and `status_description` fields.

The status fields should be a TextChoice subclass with the following fields:
* Retired
* Postponed
* Ongoing
* Completed

We can make this non-nullable after moving all existing projects to `ongoing` and defining that all projects created before <deploy date> with a status of `ongoing` need auditing.

The `status_description` field should be a `TextField`, defaulting to `""`, and not be nullable.  It will let researchers describe the reason for setting a status.

The staff area should be able to edit both of these new fields, and filter projects on the ProjectList view by their status.

However, the ideal here is that researchers change these fields themselves.  We can add a ProjectEdit view to the jobserver app which lets those with the ProjectDeveloper role use it (disabling the link otherwise).

Parent: opensafely/research-template#92

PR #2208, opened 2022-10-18
Project audit fields

ghickman:
This adds the updated_at and updated_by fields to a Project.

Note: We can't require updated_by because we don't have historic data for some created_by fields.


lucyb (approved):
These changes look good. Only thing I'm not sure about is why you're making the change (you might have told me, but I've forgotten. Adding it into a commit message is helpful).

Also, would it be worth writing any tests for the save actions?

ghickman:
@lucyb – that's fair!  I'm working on adding statuses for #2195 and realised we didn't have updated_* fields on a model we're about to let users edit.  I've broken them out into a separate PR to make reviewing easier.  Updated tests 👍

PR #2210, opened 2022-10-19
Add status fields to Project

### LINKED ISSUE #2195: Track Project status

ghickman:
This adds the two status fields to the Project model, and makes them editable both in the staff area in the UI by Project members.

Status field in use on a Project page:

Fixes #2195


lucyb (approved):

lucyb:
This looks good.

I notice the status isn't being shown anywhere at the moment. What's the plan for that? You could ask Tom O to take a look potentially?

ghickman:
@tomodwyer – what do you think of the status label I've added here?  If there's a description it'll render in a `<p>` below the label.  I didn't see a better location in the card stack for it so went with top position, happy to be told otherwise!

tomodwyer:
@ghickman Seems fine for the moment, it's clear enough and will get picked up in the restyle work 🚀

PR #2211, opened 2022-10-19
Build the nonzero_exit log path directly in the JobDetail view

ghickman:
We have configured Tailwind to look for CSS classes in .html files.  The status_tools template tag generated the output of nonzero_exit codes from TPP inside its python file, meaning tailwind didn't know to include the relevant CSS classes.

This moves the HTML into the template so Tailwind can pick it up, and the log path generation into the view.  If we handle further error codes from TPP or any other backend we can go back to looking those up dynamically using a method which Tailwind can support.


jobserver/templates/job_detail_tw.html
```diff
@@ -129,7 +128,14 @@ <h1 class="text-3xl break-all md:bread-words md:text-4xl font-bold text-slate-90
           {% if job.action in job.job_request.cancelled_actions %}
             <p>A User has requested this Job is cancelled.</p>
           {% endif %}
-          {% status_hint job %}
+          {% if log_path %}
```
lucyb, line 131: What if the job exited with an error on a non-TPP backend, wouldn't we still want to show an error message here, just without the log path?

ghickman: No, this error (according to the old template tag) is only specific to TPP.  I'm just copying the existing logic over to unstick Tom 🙂

lucyb (approved):

PR #2220, opened 2022-10-21
Show restyled Job Details view to all users

tomodwyer:
As currently it only displays to logged in users.


jobserver/views/jobs.py
```diff
@@ -93,10 +93,7 @@ def get(self, request, *args, **kwargs):
                 "Job Request concurrency"
             ] = format_jobrequest_concurrency_link(related_job_request)
 
-        if has_role(request.user, CoreDeveloper):
-            template_name = "job_detail_tw.html"
-        else:
-            template_name = "job_detail.html"
+        template_name = "job_detail.html"
```
ghickman, line 96: Now that we're not using `template_name` else where we can just use the name string directly in the `TemplateResponse()` call.

ghickman (approved):

PR #2243, opened 2022-10-26
Add Workspace.purpose

### LINKED ISSUE #2221: Add Workspace.purpose
We want to display a purpose for each workspace, so researchers can describe why they started a given workspace.

* [x] Do our internal researchers think this is something they could do across the vast majority of their own workspaces?
* [ ] Add the field.
* [ ] Display on the WorkspaceDetail page.

Parent: opensafely/research-template#92

ghickman:
This adds a purpose field to the Workspace model, making it editable for researchers.  The create workspace page has it as a required field so it'll be there going forward.  We'll need to ask researchers to backfill this over time.

Fixes #2221


bloodearnest (approved):
LGTM

I was slightly surprised how much code it took to add a new form to edit a model, given that's django's strength, but nothing is ever that simple is it!

ghickman:
@bloodearnest – this is because I wanted to set a help text but I generally dislike doing that on a model (it feels like it really breaks separation of concerns to set display values on the model) so we lost out on using a simple UpdateView with `fields = ["purpose"].  However, I'm generally keen to move away from ModelForms, they're _great_ for a handful of fields but they always hide some complexity and cause debugging down the line.

PR #2248, opened 2022-10-27
Display contacts for workspaces as email addresses

### LINKED ISSUE #2245: Add contacts emails to RepoDetail in the staff area
We want to make it easy to email the relevant people when a repo is both past its 12 months since first code run date AND hasn't been signed off.

For #1989 we're expecting to email people but until we have that in, for the repos which won't be captured by those (existing repos past the cut off date which need dealing with in the mean time), and generally to get points of contact we want to display email addresses for the creators of workspaces.

This should be a `<pre>` block containing a string with a distinct list of creators, each in the format `Full Name <email>`.

ghickman:
This displays the email addresses of workspace contacts on a RepoDetail page.  It uses the format `name <email>` for users with email addresses.

Fixes #2245

### DIFF
--- a/staff/views/repos.py
+++ b/staff/views/repos.py
@@ -204,7 +204,17 @@ class RepoDetail(View):
         )
         workspaces = [build_workspace(w, users) for w in workspaces]
 
+        def build_contact(user):
+            return (
+                f"{user.fullname} <{user.notifications_email}>"
+                if user.fullname
+                else user.notifications_email
+            )
+
+        contacts = ", ".join({build_contact(w["created_by"]) for w in workspaces})
+
         context = {
+            "contacts": contacts,
             "first_job_ran_at": first_job_ran_at,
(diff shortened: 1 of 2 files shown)


staff/views/repos.py
```diff
@@ -204,7 +204,13 @@ def build_workspace(workspace, all_users):
         )
         workspaces = [build_workspace(w, users) for w in workspaces]
 
+        def build_contact(user):
+            return f"{user.fullname} <{user.email}>" if user.fullname else user.email
```
lucyb, line 208: In other places we're using `user.notifications_email` to send emails to users, would it make sense to use that here too if the main purpose is to use it for contacting users?

lucyb (approved):
Looks good, just one question about which email address to use (given that we're using `user.notifications_email` to contact the workspace creators in other places)

PR #2254, opened 2022-10-27
Fix link to publishing repos guidance

### LINKED ISSUE #2238: Update the user's repo-sign-off page to remove references to update READMEs
Now that we've moved all project status data into job-server we need to remove references to updating readmes from the sign off page we present users with.

Users should instead be directed to relevant docs pages on keeping their project data up to date.

ghickman:
This doesn't change any text on the page since we still need to double check READMEs until the majority of them have been updated, so I've left the changes at just setting the docs link now that we know it.

Fixes #2238

### DIFF
--- a/jobserver/templates/sign_off_repo.html
+++ b/jobserver/templates/sign_off_repo.html
@@ -94,7 +94,10 @@
             </li>
             <li class="d-flex align-items-center mb-2">
               {% icon_check_outline class="text-success mr-2" %}
-              <span>Updated README's on each branch to match the <a href="#">OpenSAFELY guidance</a></span>
+              <span>
+                Updated READMEs on each branch to match the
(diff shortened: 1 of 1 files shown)


jobserver/templates/sign_off_repo.html
```diff
@@ -94,7 +94,10 @@ <h2>Read and agree</h2>
             </li>
             <li class="d-flex align-items-center mb-2">
               {% icon_check_outline class="text-success mr-2" %}
-              <span>Updated README's on each branch to match the <a href="#">OpenSAFELY guidance</a></span>
+              <span>
+                Updated README's on each branch to match the
```
StevenMaude, line 98: nit:

```suggestion
                Updated READMEs on each branch to match the
```

PR #2257, opened 2022-10-31
Timeline component updates

tomodwyer:
Various component updates for the Timeline, in preparation for #2250

- Change `job_time` prop to `time`
- Add a `card_footer` component
- Add `padding_bottom` prop to timeline-item
- Move card components in to component folder


jobserver/templates/job_detail.html
```diff
@@ -55,41 +55,41 @@ <h1 class="text-3xl break-words md:text-4xl font-bold text-slate-900">
 <div class="grid grid-cols-1 gap-6 lg:grid-flow-col-dense mt-2 md:mt-8 lg:grid-cols-3">
   {% #card title="Timeline" class="lg:col-start-1 lg:col-span-1" container=True %}
     <ul class="flow-root">
-      {% #timeline_item background="blue" title="Created by the backend:" time=job.created_at %}
+      {% #timeline_item background="blue" title="Created by the backend:" time=job.created_at padding_bottom=True %}
```
ghickman, line 58: It looks like you're doing this to every use of `#timeline_item`, will there be uses with no padding at the bottom?

ghickman (approved):

jobserver/templates/job_detail.html
tomodwyer: Yes, although now you've said that, I will invert it to `no_padding` – as it's more likely to need the padding, than not.

Issue #1962, opened 2022-07-22
Remove prompt for entering full name


lucyb:

ghickman:
There's another part to this ticket since we don't want to stop forcing users to have a full name.  The social auth pipeline needs to grow an action to request a User's fullname if one doesn't come from GitHub.  We should be able to do this with PSA's partial pipeline functionality where we kick them from the pipeline to a view and then back to it on successful post.

ghickman:
PSA overwrites the values we have in our User model each time a user logs in (we've seen this with `fullname`s being wiped by users with no fullname on GitHub), we'll need to have a look at the pipeline to see if we still want to use all the existing functions to handle that as part of this.

PR #2268, opened 2022-11-02
Only set fullname from GitHub if not already set

ghickman:
This fixes the issue where changing/setting your name on job-server got overwritten with the data from GitHub the next time you logged into job-server.  This was particularly annoying when a user had no name set on GitHub, set one on job-server, then was prompted to set it again the next time they logged into job-server.

Ref #1962


lucyb (approved):
Glad you've gotten to the bottom of this problem!

PR #2256, opened 2022-10-28
Copiloting dashboard

ghickman:
Add a simple copiloting dashboard.  This is part spike output/PoC and part a test to see how adding more dashboards go.

Currently the query which drives this page is quite inefficient, seemingly because of the Count() calls but I wanted to get this out to see how useful it was in practice.

Fixes: #2251

PR #2270, opened 2022-11-03
Look up file counts for copilot dashboard in a second query

ghickman:
In production this page is hitting the 30s timeout for gunicorn and the worker is being killed.

When annotating the number of files for a project the query projects query jumps from ~120ms to ~20s locally.  I can't see anything obviously wrong from the query planner and it seems to specifically be the files annotation causing this slow down.  So instead this does a separate query, grouping in memory by project and coverting to a dict of
project_pk: number_of_files, which is merged with the project data.

With this change I'm seeing page load times of ~160ms.

Refs #2256

### DIFF
--- a/staff/views/dashboards/copiloting.py
+++ b/staff/views/dashboards/copiloting.py
@@ -1,3 +1,6 @@
+import itertools
+
+from csp.decorators import csp_exempt
 from django.db.models import Count, Min
 from django.db.models.functions import Least
 from django.utils.decorators import method_decorator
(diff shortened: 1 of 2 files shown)


lucyb (approved):
I'm assuming that the memory usage isn't particularly high with this change. Don't know if that's worth checking?

Would be interesting to see the query plan after this change as well. How many files are in listed in the database?

ghickman:
**Before:**
[Query plan](https://app.pgmustard.com/#/explore/fd9b0fd2-a088-47f2-9b43-f839fd46cc25)

```
Memory [post qs]:    2.015625MB
```

**After:**
[Query plan](https://app.pgmustard.com/#/explore/f81ca830-2ac6-40e9-9f14-25589d531c79?highlighted=1)

Without `.iterator()` on the ReleaseFile queryset:
```
Memory [post qs]:    0.015625MB
Memory [post build]: 18.546875MB
Memory [total diff]: 18.5625MB
```

With `.iterator() on the ReleaseFile queryset:
```
Memory [post qs]:    0.03125MB
Memory [post build]: 12.015625MB
Memory [total diff]: 12.046875MB
```

Putting `.iterator` on the main queryset seems to maybe knock about 1/2 a mb off the post build memory usage.  I've left it on there because it won't hurt but I'm dubious it's actually doing anything meaningful.

PR #2282, opened 2022-11-04
Add a projects dashboard

ghickman:
This is ostensibly identical to the copilots dashboard but doesn't exclude the Datalab or LSHTM orgs.  We expect these dashboards to diverge over time, hence putting them on separate pages.

### DIFF
--- a/staff/views/dashboards/projects.py
+++ b/staff/views/dashboards/projects.py
@@ -1,0 +1,77 @@
+import itertools
+
+from csp.decorators import csp_exempt
+from django.db.models import Count, Min
+from django.db.models.functions import Least
+from django.utils.decorators import method_decorator
+from django.views.generic import TemplateView
+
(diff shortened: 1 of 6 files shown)


staff/views/dashboards/projects.py
```diff
@@ -0,0 +1,77 @@
...
+from django.utils.decorators import method_decorator
+from django.views.generic import TemplateView
+
+from jobserver.authorization import CoreDeveloper
+from jobserver.authorization.decorators import require_role
+from jobserver.models import Project, ReleaseFile
+
+
+@method_decorator(require_role(CoreDeveloper), name="dispatch")
+@method_decorator(csp_exempt, name="dispatch")
```
lucyb, line 15: Forgot to ask previously, but why are these dashboards marked as csp_exempt?

lucyb (approved):

PR #2287, opened 2022-11-07
Add outputs-on-GitHub filter to RepoList

ghickman:
This syncs down the topic `github-releases` to a `has_outputs` field on our Repo model with a daily scheduled job.  It lets us remove some API code (although doesn't make the API payload any smaller) for one view, but more importantly allows us to add it as a filter to the RepoList page of the staff area.

While working on some data tidy up and thinking about how staff might want to dig through repos I realised this would be a really useful boolean to have available.

Since we rarely change the repos with the `github-releases` topic the daily sync should be more than enough to keep our local data in the correct state.

I originally envisioned this field and sync process would go away at some point but I think it could be useful even when repos are public, likely for some of our internal dashboards and metrics.


staff/views/repos.py
```diff
@@ -100,7 +100,7 @@ def enhance(repo):
             return repo | {
                 "first_run": first_run,
                 "has_jobs": has_jobs,
-                "has_releases": "github-releases" in repo["topics"],
+                "has_outputs": "github-releases" in repo["topics"],
```
bloodearnest, line 103: I think `has_outputs` is not specific enough

I think it should explicitly mention github, so maybe `has_github_outputs`?

bloodearnest (approved):

PR #2304, opened 2022-11-08
Sign off repos with outputs role

### LINKED ISSUE #2293: Add SignOffRepoWithOutputs role
We want to stop staff accidentally signing off repos with outputs hosted on GitHub and a new role seems the ideal route to take for that.

We can add a tooltip to the `Sign off` button on the staff area's repo detail page which explains why the button is disabled:
* a researcher has not yet signed it off
* the user does not have the SignOffRepoWithOuputs role

ghickman:
This gates the use of the sign off button for repos in the staff area, adding tooltips to explain why the button has been disabled.

Fixes #2293


lucyb (approved):
:+1:

PR #2314, opened 2022-11-09
set to address for send_repo_signed_off_notification_to_researchers

ghickman:
Mailgun doesn't allow emails to be sent without a to address.  However we don't want to expose email addresses to recipients so we're using the notifications address here for lack of a better option.

Fixes: #2291


lucyb (approved):
Looks good, but you broke a test apparently :)

PR #2317, opened 2022-11-10
Specify python version in runtime.txt for dependabot

ghickman:
Now that python 3.11 is out and dependabot has helpfully upgraded to it without telling anyone we're seeing dependabot failures (eg [slippers](https://github.com/opensafely-core/job-server/pull/2305)] and [slack-sdk](https://github.com/opensafely-core/job-server/pull/2306)).

Configuring the python version with `.python-version` is has a [known conflict with pyenv](https://github.com/dependabot/dependabot-core/issues/4062), and _requires_ we use bugfix versions of python.

Dependabot-core will also look at the value of [a runtime.txt file](https://github.com/dependabot/dependabot-core/blob/00f0e52b630af6d55318951e8cb21a9a32eabbb7/python%2Flib%2Fdependabot%2Fpython%2Ffile_parser%2Fpython_requirement_parser.rb#L77-L86).  In lieu of a better option this is a test to see if that will work, and if it will work with a minor version.


StevenMaude (approved):
:crossed_fingers:

PR #2372, opened 2022-11-24
Add a dashboard to track repos with multiple projects

ghickman:


templates/staff/dashboards/repos_with_multiple_projects.html
```diff
@@ -0,0 +1,123 @@
...
+                {% else %}
+                  {% for workspace in repo.workspaces %}
+                    <a href="{{ workspace.get_staff_url }}">
+                      {{ workspace.name }}
+                    </a>
+                  {% endfor %}
+                {% endif %}
+              </td>
+
+              {% if repo.has_github_outputs %}
```
lucyb, line 100: Do we need to know whether a repo has github outputs here? Does it change anything about the data tidying process?

lucyb (approved):
Looks good. Will be really interesting to hear the feedback from researchers/co-pilots.

templates/staff/dashboards/repos_with_multiple_projects.html
ghickman: I don't _think_ so, but we store it in the database so I figured hey why not!

PR #2388, opened 2022-11-25
Improve the status page layout and content

tomodwyer:
So that it makes more sense to users and developers.

## Before

CleanShot 2022-11-25 at 09 59 45@2x

## After

CleanShot 2022-11-25 at 09 59 52@2x

### DIFF
--- a/templates/status.html
+++ b/templates/status.html
@@ -31,33 +31,38 @@
         </div>
       </div>
       <div>
-        {% #description_item title="Runner last seen" stacked=True %}
+        {% #description_item title="Last contact with "|add:backend.name|add:" backend" stacked=True %}
           <time
             class="
-              font-medium text-slate-900 group relative cursor-pointer
-              {% if backend.last_seen %}text-bn-ribbon-900{% endif %}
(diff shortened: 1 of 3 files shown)


ghickman (approved):

templates/status.html
```diff
@@ -31,33 +31,38 @@ <h2 id="workspaces-title" class="text-xl font-semibold tracking-tight text-slate
...
+        {% #description_item title="Pending jobs" stacked=True %}
           {{ backend.queue.pending }}
         {% /description_item %}
+        {% #description_item title="Running jobs" stacked=True %}
+          {{ backend.queue.running }}
+        {% /description_item %}
       </div>
+      {% #card_footer no_container class="border-t border-t-slate-200" %}
+        <p class="text-sm font-medium text-slate-600">
+          {{ backend.name }} backend is unavailable if OpenSAFELY Jobs does
```
madwort, line 62:
```suggestion
          {{ backend.name }} backend is missing if OpenSAFELY Jobs does
```

madwort (changes requested):

madwort (approved):

PR #2397, opened 2022-11-25
Run tests in parallel with pytest-xdist

ghickman:
This setups up the test suite for running in parallel, but doesn't run that way by default.  Users can opt into using it by suffixing test commands with `-n <number|auto>`.

Locally this takes a full test run from ~59 to ~26s.

I've not configured this for CI because we only have one vCPU there so we get no improvement.  There's also a bug where running inside of docker seems to lose 92% of our coverage.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ omit = [
   "jobserver/settings.py",
   "jobserver/wsgi.py",
 ]
+parallel = true
 source = [
   "applications",
   "interactive",
@@ -136,3 +137,4 @@ lines-after-imports = 2
 [tool.ruff.per-file-ignores]
 "gunicorn.conf.py" = ["INP001"]
 "manage.py" = ["INP001"]
+"sitecustomize.py" = ["INP001"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ skip_covered = true
 show_contexts = true
 
 [tool.pytest.ini_options]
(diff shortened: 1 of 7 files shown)


sitecustomize.py
```diff
@@ -0,0 +1,4 @@
+import coverage
```
bloodearnest, line 1: Isn't this what coverage-enable-subprocess does?

pyproject.toml
```diff
@@ -52,7 +46,7 @@ skip_glob = [".direnv", "venv", ".venv"]
 use_parentheses = true
 
 [tool.pytest.ini_options]
-addopts = "--disable-network --tb=native --no-migrations --ignore=./release-hatch"
+addopts = "--disable-network --tb=native --no-migrations --ignore=./release-hatch -n auto"
```
bloodearnest, line 49: I think we shouldn't switch this on by default. This so we can always run normally to debug parallel test issues.

We could have a `just test` which is then parallel by default, and maybe a `just test-serial` for those cases?

Also, locally I don't see any significant benefit from >4 workers, so maybe add `--maxprocesses=4` here instead?
So we can run parallel with `just test -n N`

bloodearnest (commented):
Works for me!

Although I've spotted a couple of tests with occasional failure, one todo with datetime, the other was a clashing db index.

One nice thing about parllel tests is it shakes out many test isloation issues!

bloodearnest (approved):

Issue #2308, opened 2022-11-08
Display friendly message for Jobs exiting with a syntax error (or similar user error)


lucyb:
Job exited with error code 1" is confusing because that reads like it's an internal error of some sort. However, it's almost certainly an issue with the user's code. This could be rewritten to be more friendly to users and give them more of an indication of what went wrong

[For example](https://jobs.opensafely.org/university-of-bristol/post-covid-events/post-covid-mentalhealth/12992/ekyylwulyz6a7knd/)
> Job exited with error code 1
> The job exited with an error. You can check the log output by logging into the TPP server and looking at:
   /d/Level4Files/workspaces/post-covid-mentalhealth/metadata/cox_ipw-cohort_prevax-sub_priorhistory_true-self_harm- 
   cov_bin_history_self_harm.log

This could just say 
> The job exited with an error. 
> 
> You can check the log output by logging into the TPP server and looking at:
   /d/Level4Files/workspaces/post-covid-mentalhealth/metadata/cox_ipw-cohort_prevax-sub_priorhistory_true-self_harm- 
   cov_bin_history_self_harm.log

🧵 https://bennettoxford.slack.com/archives/C03ELD5FEAU/p1667928052164089?thread_ts=1667921921.272879&cid=C03ELD5FEAU

💡 This should also link to the logs in the outputs-viewer (with a "vpn required" lozenge)

PR #2399, opened 2022-11-25
Update job error message

lucyb:
:question: I'm not sure if I'm generating the URL for the logs correctly. Is there a way I can find out locally?

Linked to: https://github.com/opensafely-core/job-runner/pull/524
Refs: https://github.com/opensafely-core/job-server/issues/2308


jobserver/views/jobs.py
```diff
@@ -64,17 +64,24 @@ def get(self, request, *args, **kwargs):
...
                 job.job_request.backend.parent_directory,
                 job.job_request.workspace.name,
                 "metadata",
                 f"{job.action}.log",
             )
+            log_path_url = os.path.join(
+                job.job_request.workspace.get_files_url(),
+                "metadata",
+                f"{job.action}.log",
+            )
```
ghickman, line 79: While `os.path.join` will work in most cases I'd rather use `furl` here since we're constructing a URL.

lucyb: That's a fair point. Is it the correct URL though?

bloodearnest:
I think we should drop the line about the path on disk.  Old school users will know about it anyway - we want to direct newer users to use the output viewer.  And it will be gone altogether at some point.

lucyb:
> I think we should drop the line about the path on disk. Old school users will know about it anyway - we want to direct newer users to use the output viewer. And it will be gone altogether at some point.

Will it still be useful for people on tech support? It's what I've been using. Should I be doing something different or do we just need to document that that's where logs live?

bloodearnest:
> > I think we should drop the line about the path on disk. Old school users will know about it anyway - we want to direct newer users to use the output viewer. And it will be gone altogether at some point.
> 
> Will it still be useful for people on tech support? It's what I've been using. Should I be doing something different or do we just need to document that that's where logs live?

It's the path on the level 4 VM, which is all regular users have access to

I just use the source of those files in the VM, `/srv/medium-privacy/`, to look at logs for debugging. I only go look at the actuall files on the level 4 VM  if there's a problem with the TPP sync script, or issues with osrelease/release-hatch running there.

lucyb:
> > > I think we should drop the line about the path on disk. Old school users will know about it anyway - we want to direct newer users to use the output viewer. And it will be gone altogether at some point.
> > 
> > 
> > Will it still be useful for people on tech support? It's what I've been using. Should I be doing something different or do we just need to document that that's where logs live?
> 
> It's the path on the level 4 VM, which is all regular users have access to
> 
> I just use the source of those files in the VM, `/srv/medium-privacy/`, to look at logs for debugging. I only go look at the actuall files on the level 4 VM if there's a problem with the TPP sync script, or issues with osrelease/release-hatch running there.

I've swapped the messages around for now, so the link to the outputs viewer is first, which should make it more likely people will try that method. Happy to remove it in a little while if no one complains about the change. I'm also hoping that Tom O can perform some CSS magic when he returns - maybe he can grey out the old way of doing it or mark it as deprecated.

jobserver/views/jobs.py
lucyb: Updated to use furl

jobserver/views/jobs.py
```diff
@@ -64,17 +65,25 @@ def get(self, request, *args, **kwargs):
...
                 job.job_request.backend.parent_directory,
                 job.job_request.workspace.name,
                 "metadata",
                 f"{job.action}.log",
             )
+            log_path_url = (
+                furl(job.job_request.workspace.get_files_url())
+                / "metadata"
+                / f"{job.action}.log"
+            )
```
ghickman, line 80: Let's use make `log_path_url` be the result of calling `.url` on this structure.  I was surprised to see that property being called in the test since the variable ends in `_url`, since I had assumed it was the string, not a furl instance we were passing around.

jobserver/views/jobs.py
```diff
@@ -64,17 +65,25 @@ def get(self, request, *args, **kwargs):
...
+            log_path_url.path.normalize()
 
         context = {
```
ghickman, line 83: Do we need this because `get_files_url()` returns a trailing slash?

ghickman (approved):

jobserver/views/jobs.py
lucyb: Yes, unfortunately.

Issue #2310, opened 2022-11-09
Move to Python 3.11


ghickman:
It's out now, we should dig into moving to it.

I've tried this locally and it raises some show stoppers currently:
* `ruamel-yaml-clib` is removed from requirements.prod.txt (maybe just a macOS thing?)
* pyyaml from PyPI produces a hash which is unknown to our requirements.prod.txt (removing it from `requirements.prod.txt` and re-running `requirements-prod` gets the hashes for the 3.11 artefacts).

This also impacts #2102 which will need to move to 3.11 as well.

PR #2450, opened 2022-12-19
3.11

ghickman:
Fixes: #2310


ghickman:
@bloodearnest – I've copied reports' method for getting python from deadsnakes.  I don't understand why I had to take the python dependencies out of `docker/dependencies.txt` and while it currently works I'm keen to get this right before anything else copies it!  Could you take a look and see if this is the right way to do it?

docker/Dockerfile
```diff
@@ -13,6 +13,14 @@ FROM ghcr.io/opensafely-core/base-docker:22.04 as base-python
 # docker clean up that deletes that cache on every apt install
 RUN rm -f /etc/apt/apt.conf.d/docker-clean
 
+# ensure fully working base python3.11 installation using deadsnakes ppa
+# see: https://gist.github.com/tiran/2dec9e03c6f901814f6d1e8dad09528e
+# use space efficient utility from base image
+RUN --mount=type=cache,target=/var/cache/apt \
+    echo "deb http://ppa.launchpad.net/deadsnakes/ppa/ubuntu jammy main" > /etc/apt/sources.list.d/deadsnakes-ppa.list &&\
+    /usr/lib/apt/apt-helper download-file 'https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xf23c5a6cf475977595c89f51ba6932366a755776' /etc/apt/trusted.gpg.d/deadsnakes.asc &&\
```
bloodearnest, line 21:
```suggestion
    /usr/lib/apt/apt-helper download-file 'https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xf23c5a6cf475977595c89f51ba6932366a755776' /etc/apt/trusted.gpg.d/deadsnakes.asc
```

docker/Dockerfile
```diff
@@ -13,6 +13,14 @@ FROM ghcr.io/opensafely-core/base-docker:22.04 as base-python
...
+    root/docker-apt-install.sh python3.11 python3.11-venv python3.11-distutils
```
bloodearnest, line 22:
```suggestion
```

bloodearnest (commented):
I've add some suggested changes, but I can't do it all, as I can't suggest changes to deleted lines.

Basically, add the ppa as one RUN command, don't install python there.

Then, put those 3.11 packages into dependencies.txt.  This puts all the actual installs into 1 RUN command and thus one layer, rather than spreading them out over too (which saves us an apt update in the build process), and I think is clearer from a documentation pov too.

bloodearnest (approved):

Issue #2373, opened 2022-11-24
Project 12


ghickman:
Application: https://jobs.opensafely.org/staff/applications/2b8a/

Title: Investigating events following SARS-CoV-2 infection

Projects:
* https://jobs.opensafely.org/university-of-bristol/post-covid-events/ (use this as the new one)
* https://jobs.opensafely.org/university-of-bristol/post-covid-bristol/
* https://jobs.opensafely.org/university-of-bristol/post-covid-neurodegenerative/
* https://jobs.opensafely.org/university-of-bristol/post-covid-vaccinated/
* https://jobs.opensafely.org/university-of-bristol/post-covid-unvaccinated/

To find:
* ~Project(s)~
* ~Workspace(s)~
* Member(s)
* ~Repo(s)~

ghickman:
@LiamHart-hub – all done.  Please tell Venexia to keep an eye out for anything odd in their workspaces, permissions, etc in case we've missed something.

LiamHart-hub:
Will do. 

Thanks George for actioning this.

PR #2451, opened 2022-12-20
Add a merge_projects command

ghickman:
For #2373 we need to move several objects around to achieve the desired state.  We've done this a couple of times before and it seems like we might have to do it again so it seems like the ideal to put it into code now.

### DIFF
--- a/jobserver/management/commands/merge_projects.py
+++ b/jobserver/management/commands/merge_projects.py
@@ -1,0 +1,79 @@
+import sys
+
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from jobserver.models import Project, User
+
+
+class MergeException(Exception):
+    pass
+
+
+@transaction.atomic()
+def merge_projects(username, primary_slug, other_slugs):
+    try:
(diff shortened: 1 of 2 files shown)


jobserver/management/commands/merge_projects.py
```diff
@@ -0,0 +1,79 @@
...
+
+    for project in others:
+        # set up redirect
+        primary.redirects.create(
+            created_by=operator,
+            old_url=project.get_absolute_url(),
+        )
+
+        # move workspaces
+        project.workspaces.update(project=primary)
```
lucyb, line 40: How will this handle workspaces with the same names?

lucyb (approved):
This looks really good and like it will be very useful.

I appreciate that it will only be a very small number of people using it, but is there anywhere sensible to document it?

jobserver/management/commands/merge_projects.py
ghickman: They're globally unique!  We use workspace names on the backends for the directories we run jobs inside of, and they predate projects as the "core" object.

PR #2468, opened 2022-12-29
Support adding Interactive users via the staff area

ghickman:
As part of providing the OSI v2 in job-server we're going to manually onboard users, setting up everything they need in one place.  This is also a small test for possibly doing the same for all users to see if handling onboarding in one step improves our general situation.

Applications will be done externally (is the current plan) so we're going to track them via a URL.  It's not the cleanest solution given we have applications inside job-server too but the form for this view has validation to stop them both being populated.  It seems that we can't enforce that in the database with a CheckConstraint because I can't see a way to target a reverse FK in that.

This includes some HTMX additions and improvments to make the OrgCreate view more versatile for this kind of usecase.  I've mirrored those changes into the ProjectCreate view, meaning a user can create both an Org and Project for the new user from the UserCreate form and have the selectors for those populated correctly.

**UI**

This changes our auth config slightly by setting the email field as the username.  This doesn't appear to break our OAuth flow, a new user registering will still have their username set to their GitHub username thanks to PSA mapping values from GitHub to our model's fields regardless of auth system config, which is the expected outcome.

I haven't added a login form for email-based logins since that still needs a discussion on where to expose it, and design work.

Since we have no password [re]set view I'll add the email with a link to that, and that view in a subsequent PR.


templates/staff/user_list.html
```diff
@@ -24,6 +24,12 @@
 <div class="jumbotron jumbotron-fluid jumbotron--danger pt-md-2">
   <div class="container">
     <h1 class="display-4">Users</h1>
+
+    <div class="d-flex">
+      <a class="btn btn-primary" href="{% url 'staff:user-create' %}">
+        Create a User
```
lucyb, line 30: Could this be called "Create Interactive User", please? It caused me some confusion at first.

templates/staff/user_create.html
```diff
@@ -0,0 +1,176 @@
...
+  <div class="container">
+    <ol class="breadcrumb rounded-0 mb-0 px-0">
+      <li class="breadcrumb-item">
+        <a href="{% url 'staff:index' %}">Staff area</a>
+      </li>
+      <li class="breadcrumb-item">
+        <a href="{% url 'staff:user-list' %}">Users</a>
+      </li>
+      <li class="breadcrumb-item active" aria-current="page">
+        Create a user
```
lucyb, line 23: Similarly, it might help reduce confusion to call this page "Create Interactive user"

lucyb (commented):
I noticed on the `create-user-improvements` branch that I couldn't see the InteractiveReporter role in the list on the staff/user_detail page. This made me think that the user I'd created didn't have a role at first.

Also, I get an error when viewing the staff/user_detail page that the user doesn't have any backends set and that this will be needed for using the site. This doesn't have to be fixed in this PR, but should be a ticket if not.

lucyb (commented):
I am 2/3 of the way through reviewing this, but it's late so I'll return to it tomorrow. Really sorry for not getting further, but I will finish it tomorrow!

jobserver/models/core.py
```diff
@@ -525,7 +525,7 @@ class Meta:
         ]
 
     def __str__(self):
-        return f"{self.org.name} | {self.name}"
+        return f"{self.org.name} | {self.title}"
```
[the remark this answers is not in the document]
ghickman: When looking through the list of projects to set up a user I realised we weren't displaying the project number, which is included when we use `title`.  I figured it would be a little easier to find the projects with that visible!

templates/staff/user_create.html
```diff
@@ -0,0 +1,176 @@
...
+            {% for error in form.non_field_errors %}
+              <li class="text-danger">{{ error }}</li>
+            {% endfor %}
+          </ul>
+        {% endif %}
+
+        <fieldset class="mb-3">
+          <legend class="h3 mb-3">Set the user's details</legend>
+
+          {% include "components/form_text.html" with field=form.application_url label="Application URL" name="application_url" %}
```
lucyb, line 64: Just had a thought about this... is there any need to have the application url on the create user form? It might get quite annoying when adding several users and there's more potential for error. Given that it should be set per project, should it actually just be a field on the create project modal instead?

ghickman: Are we intending to add multiple Interactive users to a single project?  I was under the impression that it would effectively be 1 user/application, but I might just have that wrong in my head!

lucyb: Ah good point. I'd assumed that there might be more than one now that we have a more complicated application process.

lucyb (approved):

PR #2472, opened 2022-12-30
Add basic interactive page

ghickman:
This adds the minimum required to get a page stood up at `/<org>/<project>/interactive` so we can start integrating the form work.

The views have been gated on the CoreDeveloper role while we work on this feature.  My expectation is for this to be downgraded to the interactive role.


interactive/dates.py
```diff
@@ -0,0 +1,19 @@
+from calendar import WEDNESDAY
+from datetime import date, timedelta
+
+
+def date_of_last_extract():
```
lucyb, line 5: I wrote a set of tests for this function to sanity check this logic. You might be able to reduce them or refactor, but it will be worth keeping some of them

```
def test_date_of_last_extract_sun_to_previous_wed(freezer):
    freezer.move_to("2022-05-15")
    assert date_of_last_extract() == date(2022, 5, 4)


def test_date_of_last_extract_mon_to_previous_wed(freezer):
    freezer.move_to("2022-05-16")
    assert date_of_last_extract() == date(2022, 5, 4)


def test_date_of_last_extract_tues_to_previous_wed(freezer):
    freezer.move_to("2022-05-17")
    assert date_of_last_extract() == date(2022, 5, 11)


def test_date_of_last_extract_wed_to_previous_wed(freezer):
    freezer.move_to("2022-05-18")
    assert date_of_last_extract() == date(2022, 5, 11)


def test_date_of_last_extract_thurs_to_previous_wed(freezer):
    freezer.move_to("2022-05-19")
    assert date_of_last_extract() == date(2022, 5, 11)
```

ghickman: Oops, I forgot to pull those over, fixed 👍

interactive/views.py
```diff
@@ -0,0 +1,72 @@
...
+        # * commit to repo
+        # * create a job request
+        # * create an issue
+        # * send confirmation email
+
+        # create some required objects so we can create skeleton views as we
+        # build up the functionality of interactive
+        job_request = self.project.interactive_workspace.job_requests.create(
+            backend=Backend.objects.get(slug="tpp"),
+            created_by=self.request.user,
```
lucyb, line 48: Very trivial point, but here you have:

`created_by=self.request.user`

Whereas with the analysis request you have:

`created_by=request.user,`

lucyb (approved):
Looks good.

I'm not sure about linking an AnalysisRequest directly to a project. It feels weird when we have workspaces. I can't think of a strong argument for changing it right now though

Issue #2383, opened 2022-11-25
Limit study purpose field to 300 words


ghickman:
Request from Amir. How difficult would this be to do? Are there alternatives? NOTE: Catherine needs to report back to Amir on decision

Add text box word count to the ‘Study Purpose’ project description to not go beyond the allocated 300 words that we allow for project descriptions.

lucyb:
A character limit, rather than a word limit would be easier. We'd need to create a custom validation rule for a word limit.

Otherwise, should be straight forward to add validation in the form and on the server. Will need Tom O's contribution for the styling though.

lucyb:
Catherine has confirmed with Amir that a character limit is fine. We can assume 5 characters per word (plus whitespace).

PR #2479, opened 2023-01-04
Add a character limit to study purpose

ghickman:
This sets a 1500 character limit on the study purpose field at the UI, form, and db levels.

I've used keyup on in the JS because the keydown event was firing before the textarea saw the value.

Fixes: #2383


assets/src/scripts/application-form.js
```diff
@@ -8,3 +8,46 @@ const links = content.getElementsByTagName("a");
...
+// Select the character counter element.
+const characterCounterElement = document.getElementById("character_counter");
+
+// Select the element that shows the number of characters typed in the textarea.
+const typedCharactersElement = document.getElementById("typed_characters");
+
+// Define the maximum number of characters allowed.
+const maximumCharacters = textAreaElement.getAttribute("max_length");
+
+// Add a "keydown" event listener on the textarea element.
```
madwort: `keydown` -> `keyup`

assets/src/scripts/application-form.js
```diff
@@ -8,3 +8,46 @@ const links = content.getElementsByTagName("a");
...
+  if (typedCharacters > maximumCharacters) {
+    return false;
+  }
+  //
+  // Display the number of characters typed.
+  typedCharactersElement.textContent = typedCharacters;
+
+  /**
+   * Change the character counter text colour to "orange" if the typed
+   * characters number is between 200 to 250. If more, then change the colour to "red".
```
madwort: `200 to 250`?

assets/src/scripts/application-form.js
```diff
@@ -8,3 +8,46 @@ const links = content.getElementsByTagName("a");
...
+textAreaElement.addEventListener("keyup", () => {
+  // Count the number of characters typed.
+  const typedCharacters = textAreaElement.value.length;
+
+  /**
+   * Check if the typed characters is more than allowed characters limit.
+   * If yes, then return false.
+   */
+  if (typedCharacters > maximumCharacters) {
+    return false;
```
madwort: I'm assuming you've tried this out, and if you did something like copy-pasted in a big block of text this would do the right thing (i.e. a keyup event per char in the input, so everything works as intended)?

madwort: ah, just saw the next commit!

assets/src/scripts/application-form.js
madwort: fixed in a later commit!

ghickman: I think this is from an old commit, it's fixed by the latest one.

madwort (approved):

assets/src/scripts/application-form.js
ghickman: Just tested this, and the keyup of paste triggers it 🚀

PR #2480, opened 2023-01-05
fix: honeycomb.status_link used job.id not job.identifier

bloodearnest:
Added missing test that would have caught this


tests/unit/jobserver/test_honeycomb.py
```diff
@@ -65,6 +65,30 @@ def test_trace_link():
...
+
+    assert parsed.stacked
+    assert parsed.query == {
+        "breakdowns": ["name"],
+        "calculations": [{"op": "CONCURRENCY"}],
+        "end_time": 1665594060,
+        "filter_combination": "AND",
+        "filters": [
+            {"column": "scope", "op": "=", "value": "ticks"},
+            {"column": "job", "op": "=", "value": job.identifier},
```
madwort, line 82: my preference for this kind of test is to hardcode the identifier in the test, as per https://github.com/opensafely-core/job-server/blob/main/tests/unit/jobserver/test_honeycomb.py#L70-L94 , but wdyt? 

```suggestion
            {"column": "job", "op": "=", "value": '01234567890123456'},
```

tests/unit/jobserver/test_honeycomb.py
```diff
@@ -65,6 +65,30 @@ def test_trace_link():
     assert "bennett-institute-for-applied-data-science" in url
 
 
+@pytest.mark.freeze_time("2022-10-12 17:00")
+def test_status_link():
+    job = JobFactory(completed_at=timezone.now())
```
madwort, line 70:
```suggestion
    job = JobFactory(completed_at=timezone.now(), identifier="01234567890123456")
```

madwort (approved):

Issue #2385, opened 2022-11-25
Add radio button to service evaluation audit page of application


ghickman:
Request from Amir. How difficult would this be to do? Are there alternatives? NOTE: Catherine needs to report back to Amir on decision

Request:
Insert tick button onto service-evaluation audit rubric
https://github.com/LiamHart-hub/job-server/blob/main/snippets/service-evaluation-audit-rubric.md

[Insert tick button]
Yes, I am a member of the Bennett Institute/LSHTM conducting a study conforming to the Service Restoration Observatory and Aftershocks workstream

ghickman:
It's unclear why we need this, could we work that out before we look at specific controls to add?

CLStables:
It's because the paperwork Amir needs is different if the project sits under one of those workstreams vs. other projects that don't. He wants a way to quickly tell if a project is one of these special cases or not. Does that help?

PR #2481, opened 2023-01-05
Add checkbox for service evals only being internal users

ghickman:
Fixes: #2385


applications/form_specs.py
```diff
@@ -345,6 +345,15 @@
         rubric=snippet("service-evaluation-audit-rubric"),
         footer=snippet("service-evaluation-audit-footer"),
         fieldsets=[
+            Fieldset(
+                label="",
+                fields=[
+                    Field(
+                        name="is_member_of_bennett_or_lshtm",
+                        label="I am a member of the Bennett Institute/LSHTM conducting a study conforming to the Service Restoration Observatory and Aftershocks workstream",
```
madwort, line 353: double-checking - `Service Restoration Observatory and Aftershocks workstream` - should this be `Service Restoration Observatory or Aftershocks workstream`?

madwort (approved):

applications/form_specs.py
ghickman: Looking at the page (body updated with screenshot for context!), I think `and` is correct here because of the previous paragraph.

PR #2503, opened 2023-01-11
Integration test the submission of an application

ghickman:
This adds an integration test of the full application process, successfully submitting an application.  It is intended to avoid obvious bugs like #2502


lucyb (approved):
Thanks for doing this, it's a nice solution to help reduce errors in future.

PR #2507, opened 2023-01-16
Redirect interactive users to form

ghickman:
This redirects an interactive user to the appropriate interactive page when they log in.

It assumes that someone using the SetPassword view with only InteractiveRole is an interactive user for now, so we don't annoy staff users or those who start with interactive and expand their usage of the site to the typical flow.

### DIFF
--- a/tests/unit/interactive/test_views.py
+++ b/tests/unit/interactive/test_views.py
@@ -1,14 +1,28 @@
+import pytest
+from django.core.exceptions import PermissionDenied
+
 from interactive.models import AnalysisRequest
 from interactive.views import AnalysisRequestCreate
+from jobserver.authorization import InteractiveReporter
 
-from ...factories import BackendFactory, ProjectFactory, WorkspaceFactory
+from ...factories import (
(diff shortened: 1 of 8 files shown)


tests/unit/interactive/test_views.py
```diff
@@ -1,14 +1,24 @@
...
+    BackendFactory,
+    ProjectFactory,
+    ProjectMembershipFactory,
+    UserFactory,
+    WorkspaceFactory,
+)
 
 
-def test_analysisrequestcreate_get_success(rf, core_developer):
+def test_analysisrequestcreate_get_success(rf):
```
lucyb, line 14: I think you need some tests here to check that PermissionDenied is raised when someone doesn't have the correct role?

lucyb (approved):
Looks good, just might need some extra tests.

tests/unit/interactive/test_views.py
ghickman: Oops, yep 👍

Issue #2496, opened 2023-01-10
Show report from AnalysisRequest


ghickman:
Once an AnalysisRequest's JobRequest has completed we should have a report to show the user.

The URL for the AnalysisRequestCreate page will be `/<org>/<project>/interactive/`.  After completing the form the user will be redirected to `/<org>/<project>/interactive/<pk>/` where we'll show them a success message.  Once we have a report we can display it on that page.

The reports service displays reports by having them converted to HTML using `nbconvert`, and then wrapping them in a block of HTML if there's no existing html tag to make sure we have a consisent document structure for styling.  This works well in reports and opens up the possibility that we could roll a reusable action to call nbconvert in the correct way at a later date.  We can make use of the way reports does things though by implementing its Report model here without all the GitHub baggage, and FKing to the appropriate ReleaseFile.

We'll need a way for a staff user to create the report for a given ReleaseFile.  Until we have an outputs-viewer which can select and perform actions on files we'll have to create something which an output reviewer can use, so likely a page in the staff area that takes a ReleaseFile ID and a Project and writes a Report object to the db.  This page can also email the requester that their report is ready to be looked at, and the URL to view it.

PR #2521, opened 2023-01-19
Add reports for interactive requests

ghickman:
This is a first pass at adding "reports" for interactive requests (AnalysisRequests under the hood).

I am currently defining a report as:

> An HTML output from a research pipeline which the OpenSAFELY platform will render with some visual improvements

My goal was to cover what OSIv1 does, what reports does, and what OSIv2 needs.  I think we're all generally settled on OSIv2 pipelines producing the necessary HTML inside the pipeline, which matches how the reports service consumes content.  I have taken all the bits of Reports' `Report` model that I think are necessary to support interactive's reports with an eye that we end up with one of those models support both uses cases with Reports' curation being handled outside of that model.

One immediately nicety to shake out of that change was getting the report content for an AnalysisRequest has been neatly wrapped up in the `report_content` property.

Mirroring OSIv1 I've kept the assumption that outputs will use the AnalysisRequest's PK for the filename, allowing us to auto-generate reports as they're uploaded to the UI.

The AnalysisRequestDetail page, where users are redirected after submitting a request, will display the report when one exists.  This means users can refresh the page to see the rendered report once released.  There are various bells and whistles (particularly around notifications) still to add to this flow but I'm keen to stand up a walking skeleton for OSIv2 which we can build on (and enable parallel development if necessary).

To aid in local dev testing I've added the `add_osi_report` management command (and documented it in DEVELOPERS.md).  It scripts the following bits:
1. create a JobRequest
1. create an AnalysisRequest for the JobRequest
1. create a copy the report template with the right file name
1. create a Release for the file
1. upload the file to the API
1. print a link to the AnalysisRequest page so you can see the report

Steps 1-2 mirror what happens in the AnalysisRequestCreate view.
Steps 3-5 mirror what osrelease from a backend would do
Step 7 just saves you some time.

Fixes: #2496


jobserver/api/releases.py
```diff
@@ -320,6 +324,35 @@ def post(self, request, release_id):
 
         slacks.notify_release_file_uploaded(rfile)
 
+        # if we're dealing with an HTML file and the name matches an
+        # AnalysisRequest PK create a report for it
+        name, ext = os.path.splitext(filename)
+        if ext == ".html":
+            # TODO: this try block exists to handle TimeflakePrimaryKeyBinary raising
```
lucyb, line 331: Is it worth moving the contents of this if-statement into a separate function? It might be a bit distracting if you're looking at this and don't care about Analysis Requests and Reports.

jobserver/management/commands/add_osi_report.py
```diff
@@ -0,0 +1,121 @@
...
+
+    return r.json()["release_id"]
+
+
+def submit_request(*, backend, user, workspace):
+    # create a JobRequest and AnalysisRequest to mirror a user making a request
+    job_request = workspace.job_requests.create(
+        backend=backend,
+        created_by=user,
+        sha="",
```
lucyb, line 61: Oh, I've just had an interesting thought.. sorry, wish I'd realised this earlier..

We need to know the Analysis Request ID in order to make the commit, because it will need to be added to the project.yaml.

This means we need to create and save the Analysis Request, before we make the Job Request.

So, AnalysisRequest.job_request needs to be nullable and the `AnalysisRequest.objects.create` should be done first.

Walking skeletons are useful, I guess?

jobserver/models/reports.py
```diff
@@ -0,0 +1,48 @@
...
+        "ReleaseFile",
+        on_delete=models.PROTECT,
+        related_name="reports",
+    )
+
+    title = models.TextField()
+    slug = AutoSlugField(max_length=100, populate_from=["title"], unique=True)
+    description = models.TextField()
+
+    authors = models.TextField(null=True)
```
lucyb, line 18: I think we should leave authors and doi out until we start using this for Reports Reports, as they will be easy to add later and we might end up implementing it differently potentially.

lucyb (commented):
Pausing my review for a meeting. Will continue reading through later.

templates/staff/report_detail.html
```diff
@@ -0,0 +1,93 @@
...
+        <a href="{{ report.created_by.get_staff_url }}">{{ report.created_by.name }}</a>
+      </li>
+
+      <li>
+        <strong>Created at:</strong> {{ report.created_at }}
+      </li>
+    </ul>
+
+    <div class="d-flex">
+      <a class="btn btn-primary mr-1" href="{{ report.release_file.get_absolute_url }}">View on ReleaseFile</a>
```
lucyb, line 42: > View on ReleaseFile

This sounds a bit clunky. Maybe "View released output"?

templates/staff/report_detail.html
```diff
@@ -0,0 +1,93 @@
...
+</div>
+{% endblock jumbotron %}
+
+{% block staff_content %}
+<div class="container">
+  <div class="row">
+    <div class="col col-lg-9 col-xl-8">
+     <section class="card mb-3">
+        <div class="card-header">
+          <h2 class="card-title">Details</h2>
```
lucyb, line 54: I think showing details of the AR is going to be really useful here, but not necessary for this PR.

templates/staff/report_list.html
```diff
@@ -0,0 +1,142 @@
...
+          "
+        >
+            {{ project.title }}
+        </a>
+        {% endfor %}
+      </div>
+      {% endif %}
+
+      {% if users %}
+      <h3 class="h4">Authors</h3>
```
lucyb, line 89: Would "creators" rather than "authors" be more accurate on the staff pages? I'm thinking that more accurately covers both Interactive and Reports reports.

outputs/report.html
```diff
@@ -0,0 +1,332 @@
+<!DOCTYPE html>
```
lucyb, line 1: It took me some digging to figure out what report.html was for. Could it be renamed to something like "sample_report.html" or something to make its purpose clearer, please?

jobserver/management/commands/add_osi_report.py
```diff
@@ -0,0 +1,121 @@
...
+        "Content-Disposition": f'attachment; filename="{path.name}"',
+        "Content-Type": "application/octet-stream",
+        "OS-User": user.username,
+    }
+    url = f"{BASE_URL}release/{release_id}"
+    r = requests.post(url, data=path.read_bytes(), headers=headers)
+    r.raise_for_status()
+
+
+class Command(BaseCommand):
```
lucyb, line 91: This works! It's a really nice way to test it.

lucyb (approved):
:notebook: :tada:

jobserver/models/reports.py
ghickman: Good point 👍

jobserver/management/commands/add_osi_report.py
ghickman: oh no 😱

I think we can avoid a null by wrapping this code in the view in a transaction, so we'll create an AnalysisRequest instance with `AnalysisRequest(…)`, create and save the JobRequest, update the AnalysisRequest instance and save it to the db.  I think we'll need to see how the implementation of `create_commit()` plays out before we make any changes with this or the view though!

templates/staff/report_detail.html
ghickman: Hands up who did `s/Site/ReleaseFile` 😂

outputs/report.html
ghickman: Possibly, everything in `outputs` is a sample though!

PR #2525, opened 2023-01-20
Drop timeflake for ULID as pk of AnalysisRequest

ghickman:
This drops timeflake as the field type of AnalysisRequest's pk for a ULID.  `timeflake` has been left in as a dependency so a subsequent PR can rebuild the interactive migrations, removing the need to have timeflake installed.

We found in use the timeflake PK made certain common operations behave in unexpected ways.  Querying in the `.filter(pk="blah")` and `.get(pk="blah")` forms, that is passing a string to the query raise a ValidationError.  This made common uses, like the `filter(pk="<some string>").first()` pattern unwieldy.  In postgres this field uses a `uuid`, so even passing the raw value through raised an exception (Django exposed this as a DataError).  While both technically the correct way to respond it makes the field less usable in practice.

Since we're already using ULIDs this way for ReleaseFile.pk we opted try that out here too.


bloodearnest (approved):
The ulid stuff looks fine.

The complexity of redirects worries me a bit, but it's a hard problem to solve.

PR #2531, opened 2023-01-23
Add missing constraints

ghickman:
This adds database-level check constraints to our models for the paired up fields we have in many places, ie those which end with `_at` and `_by`, where we want to ensure that they're either both set or both not set.

While I was doing this I realised that the test_code.py module had grown quite large and split it up into per-model modules.  It's done in one commit so stepping through the commits will be much easier to read here.  I left the models in the core.py module as I don't think splitting them up brings much benefit.

We'll need add constraints to the Org, OrgMembership, Project, and ProjectMembership models once we can work out how to backfill the necessary data.


lucyb (approved):
Looks good. I like the refactor of those unit tests.

Issue #2499, opened 2023-01-10
List previous analysis reports


ghickman:
Any given project can have N AnalysisRequests, each with 0 or 1 related reports.

We need to find a way to surface those for returning users.

One option is to have the homepage for interactive users be just their analysis requests and reports.

ghickman:
From Brian in #2724:

> As a user I'd like to be able to see previous analysis I have requested/run.

> At the moment if you enter via the login you cannot see your analyses - you must go back to your emails to find the correct url. Adding "analysis history" to a users account would be the obvious solution - maybe eventually it might be part of the home screen when you login?

> My experience is that over time the usefulness of having something like this increases.

> In week after I run an analysis I can normally find my analysis in email/slack. Longer it goes I find it difficult to find and tempted to request the analysis again!

lucyb:
Has this now been done?

ghickman:
Nope, just waiting on a review for #2537 :)

PR #2537, opened 2023-01-23
List analysis requests

ghickman:
This adds two pages to list analyses, one for the current user (if logged in), and one to live under a given project.

I've not linked to the projects one yet since this is a barebones implementation (particularly UI wise!) to play around with and discuss further.

Fixes: #2499


jobserver/context_processors.py
```diff
@@ -73,6 +73,11 @@ def nav(request):
     is_active = functools.partial(_is_active, request)
 
     items = [
+        NavItem(
+            name="Analyses",
```
lucyb, line 77: Maybe this should be called `Interactive`, instead? Then it's more consistent with where we've used Interactive in other places. I like having this page in this location though, it seems like it will be very useful to people.

jobserver/urls.py
```diff
@@ -246,6 +246,11 @@
 
 project_urls = [
     path("", ProjectDetail.as_view(), name="project-detail"),
+    path(
+        "analyses/",
```
lucyb, line 250: Having this at the project level feels like the right thing to do, but I don't know how will link from the project page, whereas if we have it at the workspace level, we could easily insert something between Workspace purpose and Releases for Interactive Reports/Analyses. What do you think?

Maybe we need to let Tom O have a think about how the user journey will work before making any changes.

jobserver/context_processors.py
ghickman: Actually I think this is going to get confusing if we switch to Interactive!

We currently have:

* `/analyses/` -> list of requested analyses for the current user
* `/org/project/interactive/` -> interactive form
* `/org/project/analyses/` -> list of requested analyses for the project
* `/org/project/workspace/analyses/` -> list of requested analyses for the workspace (I'm adding this one based on your comment below)

I think we use Interactive for the nav item we might cause confusion with the URL not matching the name?

lucyb: Interactive in this case is a noun rather than an adjective, just to be picky! Might be worth sharing some screenshots and getting other people's opinions though.

Are the `*/analyses/` pages going to just have lists of requested analyses or are they going to be reports (whether pending, released or published) or both?

I think we've talked with Catherine about having `/org/project/interactive/` show a list of previous requests (or reports) and have a button at the top to create a new analysis request, so maybe we don't need `/org/project/analyses/` as well?

jobserver/urls.py
ghickman: It'll probably start life as a card, at the bottom of which ever column seems the least daft.  It's definitely not ideal but I know Tom is planning to re-think this page with timeline-style cards which would be perfect for this.

jobserver/context_processors.py
ghickman: Dictionary says otherwise!  This is the literally only way I will ever know what type a word is.

The `*/analyses/` pages were going to show lists of existing AnalysisRequests, based on the context of the viewer.  There _might_ be a report on the page the user clicks through to, but there isn't a guarantee of one.

Vaguely, but I'll pivot that page to show those for this PR.

ghickman: Change of plan, we need some more planning around URLs to display a list of Analyses on the `*/interactive` page, because it changes around how the form works.  Lets discuss at the next tech meeting for OSI!

For now I'll drop the project-specific page and add a workspace-specific one.

lucyb: > Dictionary says otherwise! This is the literally only way I will ever know what type a word is.

This is totally not the slightest bit important, but while interactive is an adjective, **Interactive** is the name of a thing, so it's a proper noun. It's also totally possible to turn [adjectives into nouns](https://en.wikipedia.org/wiki/Nominalized_adjective) in English (yes, I did need to google that).

ghickman:
@lucyb – Tom and I have completely reworked this.  I've updated the PR body to reflect that, could you take another look at it?

templates/workspace_analysis_request_list.html
```diff
@@ -0,0 +1,46 @@
...
+    {% breadcrumb title="Analyses" active=True %}
+  {% /breadcrumbs %}
+{% endblock breadcrumbs %}
+
+{% block content %}
+<h1 class="mb-6 text-3xl break-words md:text-4xl font-bold text-slate-900">
+  {{ workspace.name }} analyses
+</h1>
+
+{% #card title=workspace.name subtitle="This page shows all the analyses for "|add:workspace.name %}
```
lucyb, line 22:
```suggestion
{% #card title=workspace.name subtitle="This page shows all the Interactive analyses for "|add:workspace.name %}
```

templates/workspace_detail.html
```diff
@@ -277,6 +276,20 @@ <h1 class="mb-2 text-3xl tracking-tight break-words font-bold text-slate-900 sm:
         {% /list_group %}
       {% /card %}
 
+      {% if analyses %}
+      {% #card title="Analyses" button=True button_text="View more" button_href=workspace.get_analyses_url %}
```
lucyb, line 280:
```suggestion
      {% #card title="Interactive analyses" button=True button_text="View more" button_href=workspace.get_analyses_url %}
```
I wonder if people are going to get confused about what "Analyses" are, especially if they're moving between traditional OpenSAFELY projects and Interactive ones, so I think it would be a good idea to mention Interactive in the title. We could also add a subtitle explaining these were created using OpenSAFELY Interactive.

templates/_components/card/card.html
```diff
@@ -1,7 +1,7 @@
 <section {% attrs class %} aria-labelledby="{{ title | slugify }}-title">
   <div class="bg-white shadow">
     {% if title or subtitle or button %}
-    <div class="flex flex-wrap gap-2 items-start justify-between px-4 py-2 sm:px-6 sm:py-4 sm:flex-nowrap">
+    <div class="flex flex-wrap gap-2 items-center justify-between px-4 py-2 sm:px-6 sm:py-4 sm:flex-nowrap">
```
lucyb, line 4: I don't see any practical difference between items-start and items-center here. What am I missing?

lucyb (approved):
Looks good. 

This is probably one for Tom O after it's merged, but it would be good for the "View more" button to match what's done on the home page. It would also be nice to see the status of analysis requests in the card on the workspace (as we do for published/unpublished reports), to save having to click into them.

lucyb:
> @lucyb – what status were you thinking for the analyses? Something like `finished`?

Ideally, yes, something like In Progress/Finished.

However it needs to match up with the workspace_analysis_request_list.html page, which is currently just showing the job request status. So, maybe that's good enough for now and we can improve it later?

PR #2543, opened 2023-01-24
fix: switch to MAX aggregation for honeycomb metrics

bloodearnest:
Also, in order to show job request metrics, switch the job request link
to use tick traces rather than job traces, and add metrics.

### DIFF
--- a/jobserver/honeycomb.py
+++ b/jobserver/honeycomb.py
@@ -112,8 +112,8 @@ def status_link(job):
         breakdowns=["name"],
         calculations=[
             {"op": "CONCURRENCY"},
-            {"op": "AVG", "column": "cpu_percentage"},
-            {"op": "AVG", "column": "memory_used"},
+            {"op": "MAX", "column": "cpu_percentage"},
+            {"op": "MAX", "column": "memory_used"},
         ],
         stacked=True,
         omit_missing=True,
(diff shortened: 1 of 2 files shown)


jobserver/honeycomb.py
```diff
@@ -130,16 +130,17 @@ def jobrequest_link(job_request):
...
-        calculations=[{"op": "CONCURRENCY"}],
+        breakdowns=["job", "name"],
+        calculations=[
+            {"op": "CONCURRENCY"},
+            {"op": "MAX", "column": "cpu_percentage"},
+            {"op": "MAX", "column": "memory_used"},
+        ],
         orders=[{"op": "CONCURRENCY", "order": "descending"}],
         stacked=True,
+        omit_missing=True,
```
sebbacon, line 141: bit: what's the significance of this option? Worth a comment maybe?

sebbacon (approved):

jobserver/honeycomb.py
bloodearnest: We're switching trace types. Ticks traces can have missing data, whereas job traces generally do not (as they are historical). So omit_missing values prevents ugly/misleading graphs when that happens.

It's the equivalent of the "Omit missing values" UI option
image

Issue #2542, opened 2023-01-24
Retrieve codelists from opencodelists for Analysis Form


lucyb:
The React app will need to display the codelists for selection. Job Server will retrieve these from opencodelists

* DM+D for medications
* SNOMED CT for events

Note: we made need other restrictions, such as only active and public codelists. Will need to retrieve the organisation, created date and description, and maybe others.

PR #2548, opened 2023-01-25
Get codelists for Interactive form

ghickman:
This adds an API wrapper for the OpenCodelists API, copying the method we used for GitHub, testing with a Fake and then verifying the fake separately.

The form and the way it's used in the view are a rough version that will almost certainly need to change when we integrate the react app form (eg we're likely going to POST JSON, not form data at the view), but these changes have been focussed on getting the codelists and validating their slugs.  We can address the method of submission when integrating.

Fixes: #2542

### DIFF
--- a/tests/verification/test_opencodelists.py
+++ b/tests/verification/test_opencodelists.py
@@ -1,0 +1,39 @@
+import pytest
+
+from interactive.opencodelists import OpenCodelistsAPI
+
+from ..fakes import FakeOpenCodelistsAPI
+from .utils import compare
+
+
+pytestmark = [
+    pytest.mark.verification,
+]
+
+
+@pytest.fixture
+def opencodelists_api():
+    """create a new API instance for these tests"""
+    return OpenCodelistsAPI()
+
+
+def test_get_codelists(enable_network, opencodelists_api):
+    args = ["snomedct"]
+
+    real = opencodelists_api.get_codelists(*args)
+    fake = FakeOpenCodelistsAPI().get_codelists(*args)
+
+    compare(fake, real)
+
+    assert real is not None
+
+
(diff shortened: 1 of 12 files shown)


jobserver/opencodelists.py
```diff
@@ -0,0 +1,78 @@
+import requests
```
lucyb, line 1: This should probably go in the top level services directory or into the interactive directory, rather than the jobserver one.

tests/verification/test_opencodelists.py
```diff
@@ -0,0 +1,39 @@
...
+    assert real is not None
+
+
+def test_get_codelists_with_unknown_coding_system(enable_network, opencodelists_api):
+    args = ["test"]
+
+    real = opencodelists_api.get_codelists(*args)
+    fake = FakeOpenCodelistsAPI().get_codelists(*args)
+
+    compare(fake, real)
```
lucyb, line 37: I'm not sure what the comparison with the fake api is adding here, given that we're checking below that the expected response is always an empty list?

lucyb (approved):

jobserver/opencodelists.py
ghickman: Services is for wrapping external services when there's config or primitives to be gained from them, rather than API wrappers.  (Slack is the odd-one-out there!)  Putting it in interactive is a great shout though 👍

tests/verification/test_opencodelists.py
ghickman: It compares the two results without me having to think too hard about what they are!  I could reformat the test to directly check the response from each api object is `[]`, but having done this for lots of methods in the GitHub tests I found that consistency was much easier to read.

Issue #2522, opened 2023-01-19
Enable Notification button broken


brianmackenna:
On this [page](https://jobs.opensafely.org/phc-university-of-oxford/scarlet-fever-and-invasive-group-a-strep-cases-during-the-covid-19-pandemic/strepa_scarlet/) the notification button gives me a 404

I have tried to check on other jobs but I don't appear to have permissions to get notifications in any running! 

![image](https://user-images.githubusercontent.com/40460354/213511705-5f282759-8354-49ba-9d6d-01d9aae69efa.png)

PR #2552, opened 2023-01-26
only show the notifications button for users who are allowed to use it

ghickman:
Toggling notifications on a workspace sets the default for notifications about the completion of a JobRequest.  Only a user with the workspace_toggle_notifications permission (via the ProjectDeveloper role) should be to see and use this button.

Fixes: #2522

### DIFF
--- a/tests/unit/jobserver/views/test_workspaces.py
+++ b/tests/unit/jobserver/views/test_workspaces.py
@@ -463,6 +463,26 @@ def test_workspacedetail_authorized_private_repo_show_change_visibility_banner(r
     assert response.context_data["show_publish_repo_warning"]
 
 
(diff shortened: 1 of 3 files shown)


tests/unit/jobserver/views/test_workspaces.py
```diff
@@ -463,6 +463,26 @@ def get_repo_is_private(self, owner, name):
     assert response.context_data["show_publish_repo_warning"]
 
 
+def test_workspacedetail_authorized_toggle_notifications(rf):
```
lucyb, line 466: Might also be good to check that someone who's not a ProjectDeveloper has "user_can_toggle_notifications" set to False?

lucyb (approved):

PR #2562, opened 2023-01-30
Interactive spa base url

ghickman:
To allow the interactive SPA to keep its pages in the URL we need to tell it what URL prefix it can ignore since that prefix is already handled by Django.

### DIFF
--- a/templates/interactive/analysis_request_create.html
+++ b/templates/interactive/analysis_request_create.html
@@ -24,6 +24,7 @@
 {{ medications|json_script:"codelist_medications" }}
 <div
   id="osi"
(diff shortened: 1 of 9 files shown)


templates/interactive/analysis_request_create.html
```diff
@@ -24,6 +24,7 @@
 {{ medications|json_script:"codelist_medications" }}
 <div
   id="osi"
+  data-base-url="base_url"
```
tomodwyer, line 27:
```suggestion
  data-base-path="{{ base_path }}"
```

tomodwyer (approved):

PR #2568, opened 2023-01-31
remove pgcli from prod, we do not need it and its many awkward dependencies

bloodearnest:


ghickman (approved):

bloodearnest:
I'm really glad we did this. pgcli was pulling in psychopg2, which meant we head both psychopy2 and pyschopg2-binary installed. That seems like a foot gun. Which were we actually importing?

Issue #2593, opened 2023-02-06
Form validation schema


ghickman:
What do we want to check in the form submission from the react app? and what format does the analysis code invocation function want the data in?

The form currently submits a JSON payload in this format:

```
{
  "frequency": "monthly",
  "codelistA": {
    "label": "Abdominal aortic aneurysm diagnosis codes",
    "organisation": "NHSD Primary Care Domain Refsets",
    "value": "nhsd-primary-care-domain-refsets/aaa_cod/20210127",
    "type": "codelist_event"
  },
  "codelistB": {
    "label": "Active and inactive ethnicity codes",
    "organisation": "NHSD Primary Care Domain Refsets",
    "value": "nhsd-primary-care-domain-refsets/ethnall_cod/20210127",
    "type": "codelist_medication"
  },
  "timeValue": "12",
  "timeScale": "months",
  "timeEvent": "before",
  "filterPopulation": "adults",
  "demographics": ["age"]
}
```

#### `frequency`
One of the following strings:
* `monthly`
* `quarterly`
* `yearly`

#### `codelistA`
`label`, `organisation`, and`value` are all valid values as returned from `get_codelists()`
`type` is one of the following strings:
* `codelist_event`
* `codelist_medication`

#### `codelistB`
`label`, `organisation`, and`value` are all valid values as returned from `get_codelists()`
`type` is one of the following strings:
* `codelist_event`
* `codelist_medication`

#### `timeValue`
A number.

Does this have upper and lower bounds?

#### `timeScale`
One of the following strings:
* `months`

#### `timeEvent`
One of the following strings:
* `before`
* `after`

#### `filterPopulation`
One of the following strings:
* `all`
* `adults`
* `children`

#### `demographics`
An optional list of strings:
* `age`
* `ethnicity`
* `imd`
* `region`
* `sex`

#### Analysis code schema
TBD

tomodwyer:
> #### `codelistA`
>
> `label`, `organisation`, and`value` are all valid values as returned from `get_codelists()` `type` is one of the following strings:
>
> - `codelist_event`
> - `codelist_medication`

Some codelists listed currently do not have an organisation. All codelists have a `label` and `value`

If you would prefer, I can change the type to: `event` and `medication`?

> #### `timeValue`
>
> A number.
>
> Does this have upper and lower bounds?

Yes, it has to be greater than zero, and less than 5 years.

```js
// The following finds validation failures for 5 years
[...]

PR #2599, opened 2023-02-06
Validate the submitted form payload

ghickman:
Refs: #2593

### DIFF
--- a/interactive/forms.py
+++ b/interactive/forms.py
@@ -1,36 +1,77 @@
+import functools
+
 from django import forms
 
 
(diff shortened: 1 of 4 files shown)


tomodwyer (approved):
Matches the validation rules from the React app

lucyb (approved):
I think this is fine for now, although I think the time_value/time_scale/time_event fields might need to become optional based on whether a second codelist has been selected or not, given that they describe the relationship between the two codelists (if I've understood correctly). Something to check!

Issue #2594, opened 2023-02-06
Duplicate job requests


iaindillingham:
Two very similar job requests were created, three seconds apart. [The earlier job request][2] failed, reporting `__error__`. [The later job request][3] succeeded. The stack trace isn't available through [Sentry][1] but it is available through Honeycomb's [Job Trace page][5]. It mentions `assert_new_jobs_created`, which @madwort may be investigating. See also opensafely-core/job-runner#517.

@lucyb and I discussed the issue in [this Slack thread][4].

[1]: https://ebm-datalab.sentry.io/issues/2629669572/?project=5443358&referrer=slack
[2]: https://jobs.opensafely.org/lshtm/long-term-kidney-outcomes-after-sars-cov-2-infection/long_term_kidney_outcomes_after_covid/14953/uda5yjgfxge33pmv/
[3]: https://jobs.opensafely.org/lshtm/long-term-kidney-outcomes-after-sars-cov-2-infection/long_term_kidney_outcomes_after_covid/14954/zvhf5jt7uyvkzruj/
[4]: https://bennettoxford.slack.com/archives/C0270Q313H7/p1675557576841149
[5]: https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/datasets/jobrunner/trace/mCjukpiBvQy?fields[]=c_name&fields[]=c_service.name&span=76e301c1ad366d5a

lucyb:
This issue occurred again [here](https://bennettoxford.slack.com/archives/C0270Q313H7/p1675698457875629).

Looking at the `Created by the backend` field in Job Server on the Job Request page, in both instances the duplicate jobs were created within one second.

Testing locally, I've found that it's possible to submit a job request twice by double-clicking the Submit button on the `job_request_create.html` page. 

I've tried swapping the base template to the new tw version and using the new button component, but that seems to have the same issue too.

As a first step, we should stop users from double-clicking that Submit button.

tomodwyer:
I've pushed a quick JavaScript fix for the UI, which stops users from double-clicking on submit buttons:
#2606

PR #2606, opened 2023-02-08
Don't allow users to double-click on submit

tomodwyer:
Use JS to disable the button for 1s after a user clicks it.

Linked to #2594


lucyb (approved):
Thank you for doing this!

Issue #2498, opened 2023-01-10
Integrate Louis' report code into form submission


ghickman:
Louis has been working on the report code for v2.  Once he's at a point where he feels we can start integrating we should wire it into the AnalysisRequestCreate page.

We'll store the code as a template and interpolate in the values submitted by the form.

TBD: can we use simple values (eg `codelist_1=<slug from opencodelists>`) or do we need to generate larger snippets to handle the logic contained by the windowing functionality we now provide?

TBD: if we roll the functionality in #2494 into a new service do we keep the template and interpolation code in job-server or do we encapsulate in the external service?

lucyb:
Note: this work is dependent on the outcome of https://github.com/opensafely-core/interactive.opensafely.org/issues/465. It may simplify the solution a lot if some/all of it was moved into Reusable Actions.

If not, another option may be to store the analysis/report code in a separate repo and clone it each time. That has some drawbacks though.

lucyb:
The data that will need to be passed from the AnalysisRequest into the code that creates a git commit with all the analysis/report code in can be found in the params being passed into the report code at the moment over here:

https://github.com/opensafely/medication-reviews/blob/1576fbd74d2fe651a03ce7dbefb6004626430631/project.yaml#L245-L253

The codelist slugs and names will likely be used to download the relevant CSVs and added directly to the repo rather than passed into cohortextractor as parameters.

lucyb:
If I understand correctly, this ticket should cover passing the necessary data from the form submission/analysis request creation into the module that generates a commit into the relevant repo (see https://github.com/opensafely-core/job-server/issues/2494)

PR #2609, opened 2023-02-08
Create an analysis commit

ghickman:
This takes the validated submitted form data and creates a commit in the relevant repo, using the code we've created from our templated code repo.

The templated code repo and the tool to smush our context into it to create content for our commit is currently in progress so calling that function is commented out, but everything leading up to it has been added.

This code is heavily based off what was in v1's submit.py, modified to support 1 or 2 codelists, and structured such that most of the new submit.py could be hoisted out to a new service.

I've used the Analysis object to pass around data since we needed one place to stash data as we moved through the different places, but the AnalysisRequest has pushed that data down into the template_data dict and I wanted a structure which let the submission code rely on the presence of variables.

Unfortunately the tests which access git are significantly slower than the rest of the test suite so we're adding 3-5s for a handful of tests.

Note: I've used `GIT_CONFIG_GLOBAL` ([docs](https://git-scm.com/docs/git#Documentation/git.txt-codeGITCONFIGGLOBALcode)) to disable reading a local user's git config so our local dev and testing start with a stable base.

Refs: #2498


interactive/submit.py
```diff
@@ -0,0 +1,240 @@
...
+        title=analysis.title,
+        template_data=asdict(analysis),
+    )
+
+    # update the Analysis structure so we can pass around a single object
+    # if/when we pull the create_commit function out into another service
+    # this structure would be the JSON we send over
+    analysis.identifier = analysis_request.pk
+
+    sha, project_yaml = create_commit(
```
bloodearnest, line 54: So, we should probably think about the case where this commit/push succeeds, but the subsequent job request creation fails, and thus this *entire* transaction will not commit, and we've got a random commit unrelated to any existing AnalysisRequest or JobRequest in the git repo.

I don't think it's a big problem - the commit is never run - but we should probably be aware it's a possible failure mode.

Possibly, rather than wrapping the entire function, we could AnalysisRequest creation in its own transaction, and same for the JobRequest creation. That might leave us in a better state when something breaks.

But it might be better to leave as is, perhaps. Distributed systems are hard.

interactive/submit.py
```diff
@@ -0,0 +1,240 @@
...
+        # check this commit does not already exist
+        raise_if_commit_exists(analysis.repo, analysis.identifier)
+
+    # 1. create tempdir with AR.pk suffix
+    suffix = f"template-{analysis.identifier}"
+    with tempfile.TemporaryDirectory(suffix=suffix) as working_dir:
+        working_dir = Path(working_dir)
+
+        # 2. clone the given analysis code template repo to tempdir
+        git("clone", template_repo, working_dir)
```
bloodearnest, line 158: worth doing a shallow clone here, for speed in tests if nothing else?

bloodearnest (approved):
Looks good to me.

There's a bunch of minor details around codelist downloading and such that we'll need to change when I add the current template smush :tm: functionality in, but I can make those changes then.

Not a blocker, as it preexists this PR, but seeing it in action, I'm so far not enthused by the explicit dependency injection for codelists api. I think I prefer out existing pattern of pytest fixtures and monkeypatching. To my eye, the cost of the additional clutter to the production code is not worth it.

interactive/submit.py
ghickman: Yeah, splitting this up is really tricky.  We don't actually save the AnalysisRequest until the very end so that we can rely on there being a related JobRequest, so we're just passing around an instance here, without any db row until the final `save()` call on it.

interactive/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+    slug: str
+    type: str  # noqa: A003
+
+    # TODO: what are these again?
+    path: str | None = None
+    description: str | None = None
+
+
+@define
+class Analysis:
```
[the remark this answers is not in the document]
ghickman: It is, the Django Form will be too and I just don't have a nice way to handle all of that yet.  I have a feeling we will need to package these up together somehow and I wouldn't be opposed to reaching for something like pydantic since it did both and we're only using the Django form for validation (and haven't even tackled sending failures back yet!)

Are you happy to leave this as-is until we have a better understanding of what's coming?  I can put a comment in to the effect of what I've said above.

interactive/submit.py
ghickman: Done for both clones as I don't think we care about history for either repo.

ghickman:
> Not a blocker, as it preexists this PR, but seeing it in action, I'm so far not enthused by the explicit dependency injection for codelists api. I think I prefer out existing pattern of pytest fixtures and monkeypatching. To my eye, the cost of the additional clutter to the production code is not worth it.

Definitely in agreement for the current code.  My plan is for most of submit.py to be lifted-n-shifted to the external service, which removes nearly all of the dependency injection.  We can then decide if that's a useful path for the external service.

I held my nose about it for this for consistency and it's a nice way to test, but we should definitely do something nicer if the "external" parts of submit.py stick around.

interactive/__init__.py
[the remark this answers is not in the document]
bloodearnest: Yeah, comment would be good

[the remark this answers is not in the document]
bloodearnest: And pydantic does feel like a good fit to me

PR #2613, opened 2023-02-09
feat: ADR for revised database backup policy

madwort:


docs/adr/0007-backup-policy.md
```diff
@@ -0,0 +1,23 @@
...
+
+Date: 2023-02-08
+
+## Status
+
+Accepted
+
+## Context
+
+The database for job-server is currently stored on a managed cluster which is able to backups for seven days. A recent regression in the codebase caused some database fields to be truncated, and was not spotted until about six days after deployment. In this case, we were able to repair the truncated data using an old backup, but if it had gone a little bit longer without being spotted we would have experienced minor data loss.
```
lucyb, line 11:
```suggestion
The database for job-server is currently stored on a managed cluster which is able to keep daily backups for seven days. A recent regression in the codebase caused some database fields to be truncated, and was not spotted until about six days after deployment. In this case, we were able to repair the truncated data using an old backup, but if it had gone a little bit longer without being spotted we would have experienced minor data loss.
```

docs/adr/0007-backup-policy.md
```diff
@@ -0,0 +1,23 @@
...
+
+Our current opinion is that although seven days is sufficiently long enough to spot recover in the event of catastrophic dataloss, we should keep backups for longer to help recover from more subtle dataloss/degradation.
+
+The data stored in the database is not considered sensitive, and because of our desire & policy for openness a significant proportion of the contents are (or will become) public information.
```
lucyb, line 15: I'm not sure about this sentence. We will be keeping usernames and passwords in that database, which is sensitive information.

docs/adr/0007-backup-policy.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Decision
+
+In addition to the seven-day backups provided by our database cluster, we will use an external service to retain backups of the jobserver database for a longer period. The backup will be generated daily, and we will keep the last 90 days of backups.
```
lucyb, line 19: I would be specific about the external service that we're using here.

Why the last 90 days of backups, out of interest? Why not 12 months?

docs/adr/0007-backup-policy.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Consequences
+
+We will have a much bigger window to spot & recover from any data degradation issues. The daily backup file is currently ~39MB, which implies a total backup history of around 3.33GB. This will easily fit within current cloud storage allowance, and would only cost around $0.06/mo at current prices if we exceed our allowance.
```
lucyb, line 23: There are some specific consequences of the service we're choosing, which is worth putting here I think. For example, the data will be held in Frankfurt and it will not be encrypted at rest. Not sure if there's anything else?

lucyb: Just a thought, but who will have access to manage the backups? Is it anyone with a DO account?

docs/adr/0007-backup-policy.md
ghickman: We also have user data from the applications form which would stay private, and OAuth tokens&data from GitHub.  However, I do see what Tom is getting at here, and we have generally wanted to make everything apart from user-specific data publich over time.  Myabe this isn't the best place to record that though?

madwort: sorry, yes, this is condensed from two ideas & doesn't quite make sense on its own. Because of the question-mark about storing the backup in Frankfurt, we had two different ideas about sensitivity - one is "user identifying information to which data protection rules may apply" (e.g. private data about an individual) and one is "internal secrets that we need to protect" (e.g. keys/tokens). I'm talking about the former here, and attempting clumsily to note that although we store usernames, we also publically publish the name of the author & link to their github commits with their public github username.

docs/adr/0007-backup-policy.md
madwort: It is anyone with a DO account (which is currently the ad hoc "admin team") and therefore fewer people than are able to access the production server & database (which is everyone on the tech team).

docs/adr/0007-backup-policy.md
madwort: Yes, I'm not sure how to justify any specific length of time, as the thing we're trying to address is to give ourselves & end-users longer to notice data degradation due to accident. In the recent instance it took us 6 days, but it could just as easily take a year plus one day. 90 days is a significantly bigger window than we have now. The backup file is small, though, so we could make it longer? I don't really want to commit to keeping these backups in perpetuity.

lucyb: 90 days seems reasonable. There's a risk to keeping data for longer, as we'd need to ensure we'd be able to delete user data if needed. It was Seb who originally mentioned 12 months, so if he's happy with 90 days that's fine.

docs/adr/0007-backup-policy.md
lucyb: Sounds good

lucyb (approved):

PR #2625, opened 2023-02-10
Update terms.md

LiamHart-hub:
Please can a PDF copy of the application form be hyperlinked from the word 'here' 
[OS Application form.pdf](https://github.com/opensafely-core/job-server/files/10720335/OS.Application.form.pdf)
Do let me know if a separate PDF copy needs to be supplied. 
Thanks


ghickman:
Fixed in #2683

PR #2626, opened 2023-02-10
fix: Merge code used to actually render the current templates

bloodearnest:
This should work with the main branch of interactive-reports, when the
current PR lands with it.

This method assumes the templates for a specific interactive report are
in a directory call `reports/NAME`, where name is hardcoded to 'v2' for
now. The files in that directory are recursively copied/rendered as
needed into the target repo. This simplifies the `create_commit()`
function quite a bit.

Expanded the default template_repo fixture to include some more files,
to properly exercise the templating.


interactive/submit.py
```diff
@@ -218,7 +288,8 @@ def git(*args, check=True, text=True, **kwargs):
     cmd = ["git"] + [str(arg) for arg in args]
 
     cwd = kwargs.get("cwd", os.getcwd())
-    cleaned = [arg.replace(settings.GITHUB_WRITEABLE_TOKEN, "*****") for arg in cmd]
+    token = settings.GITHUB_WRITEABLE_TOKEN or "*** no token ***"  # avoid a token of ""
+    cleaned = [arg.replace(token, "*****") for arg in cmd]
```
ghickman, line 292: Thank you 😂

ghickman (approved):
This all looks great, and I'm quite pleased with how your templating idea has worked out!

My one worry is using the term `report` to talk about the template when we already have reports as a concept both in this code base and on the platform.  We even have report templates in `templates/` and the output of an AnalysisRequest… is a report! I think this could cause some confusion over time when we have to explain to someone which report we're talking about.

I know it's a terrible word to spell, but how do you feel about `s/report/analysis/`?

bloodearnest:
> This all looks great, and I'm quite pleased with how your templating idea has worked out!
> 
> My one worry is using the term `report` to talk about the template when we already have reports as a concept both in this code base and on the platform. We even have report templates in `templates/` and the output of an AnalysisRequest… is a report! I think this could cause some confusion over time when we have to explain to someone which report we're talking about.
>
> I know it's a terrible word to spell, but how do you feel about `s/report/analysis/`?

I agree `report` is not ideal. It comes from the repo being called `interactive-reports`.

I don't think `analysis` is really any better though, as it too is very generic. And the pluralisation is bound to trip us up, spelling wise.

I think probably need to qualify any term "interactive", so my suggestion is `InteractiveAnalysis` for class type things, and `interactive-analysis-templates` for the the repo name?

As for the subdirectory in the template repo, perhaps we could dodge the question and just call it `templates/`?

interactive/submit.py
bloodearnest: This particular spelling is a hack to avoid coverage shouting at me for not covering it when set!

ghickman:
> I don't think analysis is really any better though, as it too is very generic. And the pluralisation is bound to trip us up, spelling wise.

It really is the worst word.

> I think probably need to qualify any term "interactive", so my suggestion is InteractiveAnalysis for class type things, and interactive-analysis-templates for the the repo name?

Both sound good 👍

> As for the subdirectory in the template repo, perhaps we could dodge the question and just call it templates/?

Given that's going to be inside the other repo, I think that's idea really 👍

Issue #2617, opened 2023-02-09
Automate the creation of a GitHub repo when creating a new Interactive user


lucyb:
We know what the name of the repo should be when we create/select a Project via the Interactive user creation form, `<project.slug>-interactive`, so we should be able to create/select a repo on GitHub with that information.

Interactive repos have some extra admin associated with them:
* part of the `interactive` team
* tagged with the `interactive` label (topic?)
* created from the `research-template` template repo

Issue #827, opened 2021-08-17
Investigate Implementing CQRS


ghickman:
With the addition of releasing outputs features, the first real set of changes to add both write functionality and data owned by job-server we saw some useful patterns emerge from both the HTML and API views, many of the views have ended up in the form:

1. check requested object exists
2. check permissions to access object
3. do business logic
4. maybe render a template
5. return http response

We've pulled that business logic out in a few places and it's starting to look a lot like the CQRS (Command-Query Responsibility Segregation) pattern where read and write action [definitions] are separate both from each other and the views which use them.

We have some prior art in the form of opencodelists which uses `actions.py` modules in various places to test logic.

I'd like to investigate the various forms of this pattern (eg the [Cosmic Python book](https://www.cosmicpython.com/book/chapter_12_cqrs.html) has a whole chapter on it) and see if they or a facsimile would suit job-server.

Some, non-exhaustive, goals to achieve:
* pull business logic out of views
* better document logic

PR #2634, opened 2023-02-14
Create interactive repo when creating a user

ghickman:
This updates the Create interactive user page to also create a repo on GitHub if it doesn't already exist.

I was struggling to find nice patterns to create methods on Project/User which match those in `interactive/commands.py`.  I felt like the wrong layer to have a Model know about an API, so in the end I tried pulling all the functionality into "commands", as per [CQRS terminology](https://martinfowler.com/bliki/CommandQuerySeparation.html).  I'm not sold on following CQRS strictly but I've been [curious to see how useful it would be in job-server for a while](https://github.com/opensafely-core/job-server/issues/827) and this feels like a reasonable place to give it a go, being fairly separate from the rest of the system, and tasks which require more than one Model or API call.

Fixes: #2617

### DIFF
--- a/interactive/commands.py
+++ b/interactive/commands.py
@@ -1,0 +1,71 @@
+from jobserver.authorization import InteractiveReporter
+from jobserver.github import _get_github_api
+from jobserver.models import OrgMembership, ProjectMembership, Repo, User
+
+
+def create_repo(*, name, get_github_api=_get_github_api):
+    """
+    Create an interactive GitHub repo
+
+    The repo might already exist so we try to get it first, and then we ensure
+    it has the correct topics set.
+    """
+    api = get_github_api()
+
+    repo = api.get_repo("opensafely", name)
+
+    if repo is None:
+        repo = api.create_repo("opensafely", name)
+
+    # add interactive topic to the repo if it's not already there
+    topics = set(repo["topics"])
+    if "interactive" not in topics:
+        topics.add("interactive")
+        api.set_repo_topics("opensafely", name, list(topics))
+
+    api.add_repo_to_team("interactive", "opensafely", name)
+
+    return repo["html_url"]
+
+
+def create_user(*, creator, email, name, org, project):
+    """Create an interactive user"""
+    user = User.objects.create(
+        fullname=name,
+        email=email,
+        username=email,
+        created_by=creator,
+        is_active=True,
+    )
+
+    OrgMembership.objects.create(
+        created_by=creator,
+        org=org,
+        user=user,
+    )
+
+    ProjectMembership.objects.create(
+        created_by=creator,
+        project=project,
+        user=user,
+        roles=[InteractiveReporter],
+    )
+
+    return user
+
+
+def create_workspace(*, creator, project, repo_url):
+    """Create an interactive workspace"""
+    repo, _ = Repo.objects.get_or_create(url=repo_url)
+
+    workspace, _ = project.workspaces.get_or_create(
+        name=project.interactive_slug,
+        defaults={
+            "repo": repo,
+            "branch": "main",
+            "created_by": creator,
+            "purpose": "??",
+        },
+    )
+
+    return workspace
--- a/jobserver/github.py
+++ b/jobserver/github.py
@@ -18,6 +18,16 @@ session.headers = {
 }
 
 
+class GitHubError(Exception):
+    """Base exception to target all other exceptions we define here"""
+
+    pass
+
+
+class RepoAlreadyExists(GitHubError):
+    pass
+
+
 class GitHubAPI:
     """
     A thin wrapper around requests, furl, and the GitHub API.
@@ -43,11 +53,17 @@ class GitHubAPI:
         self.session = _session
         self.token = token
 
+    def _delete(self, *args, **kwargs):
+        return self._request("delete", *args, **kwargs)
+
     def _get(self, *args, **kwargs):
         return self._request("get", *args, **kwargs)
 
     def _post(self, *args, **kwargs):
         return self._request("post", *args, **kwargs)
+
+    def _put(self, *args, **kwargs):
+        return self._request("put", *args, **kwargs)
 
     def _request(self, method, *args, **kwargs):
         """
@@ -144,6 +160,31 @@ class GitHubAPI:
 
         return f.url
 
+    def add_repo_to_team(self, team, org, repo):
+        path_segments = [
+            "orgs",
+            org,
+            "teams",
+            team,
+            "repos",
+            org,
+            repo,
+        ]
+        url = self._url(path_segments)
+
+        payload = {
+            "permission": "maintain",
+        }
+
+        headers = {
+            "Accept": "application/vnd.github.v3+json",
+        }
+        r = self._put(url, headers=headers, json=payload)
+
+        r.raise_for_status()
+
+        return
+
     def create_issue(self, org, repo, title, body, labels):
         path_segments = [
             "repos",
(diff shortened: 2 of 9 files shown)


jobserver/github.py
```diff
@@ -168,6 +184,57 @@ def create_issue(self, org, repo, title, body, labels):
 
         return r.json()
 
+    def create_repo(self, org, repo):
+        path_segments = [
+            "repos",
+            "opensafely",
+            "research-template",
+            "generate",
+        ]
```
bloodearnest, line 193: Hmm, so I'm not sure about using the research-template for interactive, as it's setup with base files we don't want, and a README that's misleading in this case. Also, it runs an initial action to render the repo template, which given we're automating things could be a bit messy.

I think it might be better to create a blank repo, and include a standard boiler plate README.md in every commit that spells out that this was a repo for an Interactive Project, with links to the project/workspace on jobs.opensafely.org.

interactive/commands.py
```diff
@@ -0,0 +1,65 @@
+from jobserver.authorization import InteractiveReporter
+from jobserver.github import _get_github_api
+from jobserver.models import OrgMembership, ProjectMembership, Repo, User
+
+
+def create_repo(*, name, get_github_api=_get_github_api):
```
lucyb, line 6: I think this is a nice way of moving logic out of the view and model, and has resulted in some cleaner code.

My only, minor, comment is that I'm always a bit wary of naming something after the particular pattern that it implements. There's usually something that expresses the intent better within that specific domain. I'm not sure what this is right now though... we're creating jobserver things, but for Interactive. Come to think of it, all of the imports are from jobserver and this code is called from jobserver too. Is it in the right place?

Things to think about, but no reason to not merge it.

lucyb (approved):

jobserver/github.py
```diff
@@ -389,6 +456,28 @@ def get_repos_with_dates(self, org):
...
+        url = self._url(path_segments)
+
+        payload = {
+            "names": topics,
+        }
+
+        headers = {
+            "Accept": "application/vnd.github.v3+json",
+        }
+        r = self._put(url, headers=headers, json=payload)
```
bloodearnest, line 475: I think this implementation may wipe any existing topics that have been manually added? 

Whilst likely not a problem in the interactive case, I see a foot gun waiting to happen. e.g. we could accidentally remove our manually created `github-releases` topics via this API call.

interactive/commands.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    api = get_github_api()
+
+    repo = api.get_repo("opensafely", name)
+
+    if repo is None:
+        repo = api.create_repo("opensafely", name)
+
+    api.set_repo_topics("opensafely", name, ["interactive"])
+
```
bloodearnest, line 21: I've created an [Interactive](https://github.com/orgs/opensafely/teams/interactive) team as a way to group interactive repos with `write` permission?. Do you think we could add this newly created repo to that team?

I think topics are informative and useful for searching. But adding them to a team means we can control privileges better.

e.g. we could give the interactive bot account write access to just the repos on this team.

jobserver/github.py
bloodearnest: Alternatively, maybe an alternate interactive specific template might be better? And you can pass the template in as an arg to create_repo?

bloodearnest (commented):
Looks sensible to me

If I'm right about `set_repo_topics()`, I think we need to fix that.

I think the repo template question is worth figuring out too.

tests/verification/test_github.py
```diff
@@ -197,5 +250,24 @@ def test_graphql_error_handling(enable_network, github_api):
...
+    args = [
+        "opensafely-testing",
+        "github-api-testing-topics",
+        ["testing"],
+    ]
+
+    real = github_api.set_repo_topics(*args)
+    fake = FakeGitHubAPI().set_repo_topics(*args)
+
+    # does the fake work as expected?
```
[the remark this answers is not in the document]
ghickman: Verification tests exist to verify that a Fake responds like the real thing, which is why we only run these tests in CI by default!

jobserver/github.py
```diff
@@ -168,6 +184,57 @@ def create_issue(self, org, repo, title, body, labels):
...
         return r.json()
 
+    def create_repo(self, org, repo):
+        path_segments = [
+            "repos",
+            "opensafely",
+            "research-template",
+            "generate",
+        ]
+        url = self._url(path_segments)
```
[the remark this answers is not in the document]
ghickman: Oh this is a good point 🤔

So the API module methods are supposed to represent what the API does, not what we do with them, following along with the idea that these modules are thin wrappers of endpoints.  With that in mind, and your point that I've hidden that this is creating a repo from a template… that should be expressed in the name! (It's even hitting a different endpoint to the normal repo one)

interactive/commands.py
ghickman: I see your point about the code all coming from job-server, and I did think about putting this module into the `jobserver` app, but they're so specific to interactive that they would all need renaming to denote that.

My long term hope/goal is that we can move all onboarding to a flow similar to this, at which point I'd move them into the main app because they're not special-cased, but until then they're specifically for an interactive flow.

jobserver/github.py
ghickman: Good point.  I'll get the current topics and merge in anything we pass to avoid the footgun 👍  (You can't do add/remove via the API, you have to do GET/PUT)

interactive/commands.py
ghickman: Absolutely lets add it to the team 👍

jobserver/github.py
ghickman: The current research-template's README is already a boilerplate!  I'm not sure we need anything in there about interactive?

A separate template might well work but I'm leaning towards not doing that. We're going to wipe the working directory each time we generate code anyway.  Also it saves maintaining another repo!

I've added the ability to pass the template org/repo and renamed the method slightly to make it clearer what we're doing under the hood.

bloodearnest: Yes, but it's boilerplate for a very different thing - a manually authored study.

I think the interactive repos should have a README something like:

```
# OpenSAFELY Interactive Repository for Project Foo

This repo contains the code for the [OpenSAFELY Interactive] study for [project Foo].  Unlike regular studies, this code is not hand written, and is instead automatically generated from [a template], 

Any published results from this study will be available here: ...
```

This could possibly be rendered on every commit we do, and is not integral to walking skeleton, so we can defer decision.

But my other concern I think make cause issues in tests.

When you clone a repo from the research-template, it runs an initial action that updates the README template and then deletes itself. This takes a minute or two to run.

It's entire possible, especially in automated testing, that we a) create repo b) clone c) initial setup job finishes, making our clone out of date b) we commit/push, and thus error.

jobserver/github.py
bloodearnest: yep, familiar with this pain :(

ghickman: topics are part of the get_repo response so I can check them there 🎉  Updated the calls in the create_repo command to make sure we use the existing ones and only update if it's not already there.

jobserver/github.py
ghickman: Adding it to the analysis template repo for now sounds fine but I'm keen to only have one readme template if possible, that was a big point of the work when changing it for repo sign-offs!  I have a feeling we can handle the differences on job-server rather than in repo READMEs _but_ we should discuss this with the OSI team too (I'll drop a note in channel).

ghickman: Wouldn't making it part of the template repo mean it was the same for all actions…?  My goal was to have it all the same so now I'm a bit confused why having it in the template directory with nothing changed would not achieve that…!

bloodearnest: Not as currently implemented no, as each templated analysis is separate.

Additionally, we'd need add additional data in the template context.

I think this belongs in InteractiveAnalysisTemplate - its a step analogous to downloading a codelist, the same process for rendering all templated analysis.

ghickman: Switched to creating a repo without the research template as discussed 👍

PR #2646, opened 2023-02-17
Submission walking skeleton fixes

ghickman:
A handful of tweaks to get a successful submission of the interactive form, through to pushing the commit to the remote.


bloodearnest (approved):
Thanks makes sense

Sorry, completely missed the id/identifier thing. But id is nicer for templating as we use it a lot!

PR #2653, opened 2023-02-20
Ignore rate limit errors when testing without auth

ghickman:
This has been intermittently failing CI for a few months and I finally got annoyed enough at it to roll this.

### DIFF
--- a/tests/verification/test_github.py
+++ b/tests/verification/test_github.py
@@ -1,5 +1,6 @@
 import pytest
 from environs import Env
+from requests.exceptions import HTTPError
 
 from jobserver.github import GitHubAPI, RepoAlreadyExists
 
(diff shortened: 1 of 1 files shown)


tests/verification/test_github.py
```diff
@@ -281,4 +282,14 @@ def test_set_repo_topics(enable_network, github_api, clear_topics):
 
 
 def test_unauthenticated_request(enable_network):
-    assert GitHubAPI().get_repo("opensafely-testing", "github-api-testing")
+    try:
```
lucyb, line 285: Is it worth keeping this test if it's causing us so many problems? Do we learn anything from it that we don't get from any of the other tests? I'm a bit worried that we'll miss problems as a result of this change, but I also appreciate that rate limiting is an inherent part of the GH API.

ghickman: I don't think so, if anything the test failing is incorrect, we've correctly talked to the API without authentication!  The rate limiting is an implementation detail of the API, rather than an error with the way we're talking to it.

lucyb (approved):

PR #2664, opened 2023-02-20
Various usability fixes for OSIv2

tomodwyer:


lucyb (approved):
Thank you very much for making those changes. It has improved things, I think.

PR #2668, opened 2023-02-21
Script running and releasing of an OSI study for local dev

ghickman:


jobserver/management/commands/osi_release.py
```diff
@@ -52,31 +53,6 @@ def create_release(*, backend, path, user):
     return r.json()["release_id"]
```
lucyb, line 53: `url = f"{BASE_URL}workspace/opensafely-internal-interactive"`

Is this being hard-coded going to cause any problems, given that I'm using a different workspace?

lucyb (approved):

jobserver/management/commands/osi_release.py
ghickman: Yes, yes it is, fixed 👍

PR #2676, opened 2023-02-22
interactive data fixes

bloodearnest:
- fix: add start/end dates to Analysis template data
- s/snomedct/snomed, as that is the value expected by cohortextractor

### DIFF
--- a/interactive/__init__.py
+++ b/interactive/__init__.py
@@ -1,5 +1,7 @@
 from attrs import define
 
+from interactive import dates
+
 
 @define
 class Codelist:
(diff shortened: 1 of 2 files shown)


interactive/__init__.py
```diff
@@ -34,3 +36,5 @@ class Analysis:
     time_value: str
     title: str
     id: str | None = None  # noqa: A003
+    start_date: str = dates.START_DATE
```
lucyb, line 39: I think this is fine for the moment, but might need to be passed in from the form once that's updated rather than be in two different places.

lucyb (approved):
I used `osi_run` to check this locally and was able to successfully run the code that was generated from an analysis request. It works! :tada:

PR #2678, opened 2023-02-22
Make AnalysisRequest.slug unique

ghickman:
We generate AnalysisRequest titles from the form the user submits, which means there are only so many titles we can ever generate.  Slugs also suffer from this, being generated from titles.  Suffixing the slug with the pk retains the read-ability of the slug but keeps it unique.


ghickman:
@lucyb – updated to avoid the titles-being-generated issue.

tomodwyer (approved):

Issue #2669, opened 2023-02-21
ADR for allowing the download of released Interactive Reports


lucyb:
Decision in this thread https://bennettoxford.slack.com/archives/C04NHSU5DRB/p1675943852606299

The ADR itself has been mostly written, it just needs some tweaks: https://docs.google.com/document/d/1sZNT0LhrWpxVtFxSLrqzhxKRFMVRzGIa56g2kCQcmW4/edit?usp=sharing

Required for https://github.com/opensafely-core/job-server/issues/2670

PR #2684, opened 2023-02-22
ADR Allow download of released Interactive reports

lucyb:
Fixes: #2669


docs/adr/0009-download-released-reports.md
```diff
@@ -0,0 +1,49 @@
+# 1. Allow download of released reports
```
ghickman, line 1: Should this be `9.`?

ghickman (approved):
Looks great 👍

Can we use semantic line breaks at the end of sentences?

docs/adr/0009-download-released-reports.md
lucyb: Yes. *sigh*

lucyb:
> Can we use semantic line breaks at the end of sentences?

This isn't something I've used previously. I can guess what they are from your ADR8. Why are they used and is it something we've used elsewhere? I find it harder to read the markdown when they there.

PR #2714, opened 2023-02-27
ignore delete_repo not failing in coverage

ghickman:
This path only exists so the method doesn't raise an exception when the repo doesn't exist.  This covers the verification tests trying to tear down a repo that has gone away or never existed for some reason.  We don't want it to fail in that situation because this is used in teardown code, not test cases themselves.


lucyb (approved):
:mage:

PR #2738, opened 2023-03-01
chore: refactor interactive templating to be importable

bloodearnest:
This makes it easier for interactive-templates to import the code to use
in tests

### DIFF
--- a/interactive/submit.py
+++ b/interactive/submit.py
@@ -1,16 +1,15 @@
-import copy
 import os
 import shutil
 import subprocess
 import sys
 import tempfile
-import typing
 from pathlib import Path
 
-from attrs import asdict, define
+from attrs import asdict
 from django.conf import settings
 from django.db import transaction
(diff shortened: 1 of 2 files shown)


interactive/submit.py
```diff
@@ -1,16 +1,15 @@
...
-import typing
 from pathlib import Path
 
-from attrs import asdict, define
+from attrs import asdict
 from django.conf import settings
 from django.db import transaction
-from jinja2 import Environment, FileSystemLoader, StrictUndefined
+
+from interactive import AnalysisTemplate
```
ghickman, line 12: Can we import from `.` to match all the other imports here?

ghickman (approved):

interactive/submit.py
bloodearnest: For consistency, sure.

I still maintain importing from . is Just Not Right, but that ship has sailed in job-server :)

PR #2740, opened 2023-03-01
Stop github token from being persisted in db

lucyb:
Move the github url parsing out of the view and insert the secure token at the point of use instead.

### DIFF
--- a/interactive/submit.py
+++ b/interactive/submit.py
@@ -36,7 +36,6 @@ def submit_analysis(
     This will stay in job-server while create_commit() is intended to move to
     an external service in the future.
     """
-    # TODO: wrap form kwargs up in a dataclass?
 
     # create an AnalysisRequest instance so we have a PK to use in various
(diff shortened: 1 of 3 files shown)


bloodearnest (commented):
I much prefer the first commit's solution to the second.

Magically applying it to all arguments passed to `git()` feels footgunish for something with a secret.

From the 2nd commit message.

> This method is only relevant when calling `git`, so I've moved it closer to that function. 

I think the opposite here - it *only* applies on ls-remote and the clone, and not other git operations, so we should just use it there.

We're fixing an issue with being too liberal with secrets, feels like this is still being too liberal.

lucyb:
No worries. I've removed that second commit now. I think there's probably a slightly nicer way of doing it, but that's a bigger piece of refactoring that I don't want to worry about right now.

bloodearnest (approved):

Issue #2670, opened 2023-02-21
Download released Interactive Reports


lucyb:
See 🧵 https://bennettoxford.slack.com/archives/C04NHSU5DRB/p1675943852606299
See also the [ADR](https://github.com/opensafely-core/job-server/issues/2669)

Allow Interactive reports that have been released (not published) to be downloaded by providing a "download" button that triggers a print to PDF. 

The html report should contain some text to ensure that users are aware of the rules/conditions of sharing and provide them with a link to [our policy](https://www.opensafely.org/policies-for-researchers/#details-all-datasets).

The generated PDF should have a watermark to make it clear that it's "confidential".

tomodwyer:
PDFs can be downloaded, styling will need work once the report on page styles have been signed off.

tomodwyer:
Closed by https://github.com/opensafely-core/job-server/pull/3103

PR #2749, opened 2023-03-03
Allow users to download Interactive reports

tomodwyer:
Starter files for downloading reports. This PR adds the dependencies and scaffolding code to enable downloads – but as we don't have formatted and outputted reports yet, it will only download the "empty" state.

- Move debug information to the Staff Area (and syntax highlight the JSON)
- Add [html2pdf](url) dependency
- Load dependency on "Download" button click
- Add a watermark to the download
- Disable buttons until the report is generated
- Add an empty/loading state whilst the report is generating
- **Fix:** A fix for links that require a character straight after a link (`append_after`)

Linked to https://github.com/opensafely-core/job-server/issues/2670

### DIFF
--- a/assets/src/scripts/analysis-request-detail.js
+++ b/assets/src/scripts/analysis-request-detail.js
@@ -1,13 +1,1 @@
-import hljs from "highlight.js/lib/core";
-import json from "highlight.js/lib/languages/json";
-import "highlight.js/styles/github.css";
-
-hljs.registerLanguage("json", json);
-
-document.getElementById("analysisRequest").textContent = JSON.stringify(
-  JSON.parse(document.getElementById("analysisRequestData").textContent),
-  null,
-  2
-);
-
-hljs.highlightAll();
+// Silence
--- a/assets/src/scripts/analysis-request-detail.js
+++ b/assets/src/scripts/analysis-request-detail.js
@@ -1,0 +1,13 @@
+import hljs from "highlight.js/lib/core";
+import json from "highlight.js/lib/languages/json";
+import "highlight.js/styles/github.css";
+
+hljs.registerLanguage("json", json);
+
+document.getElementById("analysisRequest").textContent = JSON.stringify(
+  JSON.parse(document.getElementById("analysisRequestData").textContent),
+  null,
+  2
+);
+
+hljs.highlightAll();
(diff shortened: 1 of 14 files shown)


templates/interactive/analysis_request_detail.html
```diff
@@ -46,7 +51,7 @@ <h1 class="text-3xl tracking-tight break-words sm:text-4xl font-bold text-slate-
     <dl class="border-t border-slate-200 sm:divide-y sm:divide-slate-200">
       {% #description_item title="Created at" %}
         {{ analysis_request.created_at }}
-        <time datetime="{{ analysis_request.created_at|date:"Y-m-d H:i:sO" }}"></time>
+        <time datetime='{{ analysis_request.created_at|date:"Y-m-d H:i:sO" }}'></time>
```
ghickman: Should we be using single quotes here normally?

assets/src/scripts/analysis-request-detail.js
```diff
@@ -1,13 +1 @@
...
-hljs.registerLanguage("json", json);
-
-document.getElementById("analysisRequest").textContent = JSON.stringify(
-  JSON.parse(document.getElementById("analysisRequestData").textContent),
-  null,
-  2
-);
-
-hljs.highlightAll();
+// Silence
```
ghickman: What does this do?

templates/_components/link.html
```diff
@@ -8,4 +8,4 @@
   {% if new_tab %}
     target="_blank" rel="noopener noreferrer"
   {% endif %}
->{{ text }}</a>
+>{{ text }}</a>{{ append_after }}
```
ghickman, line 11: Why do we need to handle this in the component rather than in the containing element?

ghickman (approved):

templates/interactive/analysis_request_detail.html
tomodwyer: Oh I think was just my editor not formatting the file correctly (thank you incorrect VSCode plugin). I will switch this back.

templates/_components/link.html
tomodwyer: Annoyingly, if you have EditorConfig set to add a new line at the end of your file, Slippers + HTML will add an extra space after the anchor element, meaning you can't line up a full stop (for example) after a link. This was the easiest way I could find to fix this, rather than wrapping things in `spaceless` tags all over the templates.

Open to other ideas though!

ghickman: Oof, that's going to catch out various editor configs, this seems like a solid solution 👍

Issue #2747, opened 2023-03-02
Add Django messages support to base tailwind template


ghickman:
Our outgoing base template supports these [in the alerts partial](https://github.com/opensafely-core/job-server/blob/main/templates/partials/alerts.html#L19-L32).

Any view can add to this list of messages, which is cleared once they've been rendered for the user, so they typically go somewhere prominent.

[Django docs on messages](https://docs.djangoproject.com/en/4.1/ref/contrib/messages/#displaying-messages).

tomodwyer:
Closed by https://github.com/opensafely-core/job-server/pull/2762

PR #2750, opened 2023-03-03
Magic login link

ghickman:
⚠️ Blocked on #2747 ⚠️

This removes our use of passwords from email-only users (they were already disabled for social-auth-users) and sets up login via an expiring (15 minute) magic link.

Some things to consider when viewing:
* ~should we bake expiry onto the user (as I'm doing in this PR) or define it at runtime as a kwarg-with-default to `from_login_token`?~
* ~Is 1 hour too much/little for an expiry?~
* ~Should we explicitly disallow social-auth users from using this method of login?~


jobserver/models/core.py
```diff
@@ -939,11 +967,14 @@ def get_full_name(self):
         """Support Django's User contract"""
         return self.fullname
 
-    def get_password_reset_url(self):
-        uid = urlsafe_base64_encode(force_bytes(self.pk))
-        token = PasswordResetTokenGenerator().make_token(self)
+    def get_login_url(self, max_age=timedelta(hours=1)):
+        token = new_ulid_str()
```
bloodearnest, line 971: I don't think ulid is the best choice for a secret token here, as it's partially predictable by design. 

`secrets.token_urlsafe(64)` is a better option, I think?

jobserver/models/core.py
```diff
@@ -939,11 +967,14 @@ def get_full_name(self):
         """Support Django's User contract"""
         return self.fullname
 
-    def get_password_reset_url(self):
-        uid = urlsafe_base64_encode(force_bytes(self.pk))
-        token = PasswordResetTokenGenerator().make_token(self)
+    def get_login_url(self, max_age=timedelta(hours=1)):
```
bloodearnest, line 970: 15mins seems more appropriate for a login action to me.

lucyb (commented):
I definitely like having single use links that expire after one hour. We definitely don't want to have magic links having around indefinitely.

I think that we should not allow social auth users to log on via this method, as it bypasses the controls we have and will cause a lot of confusion.

I also think, as a belt and braces measure, that we should only allow users to log in if they have the InteractiveReporter role and nothing else. This stops someone from having their access increased without us knowing and without them changing their auth method. I appreciate this will probably have to be tweaked in future, but it seems like a sensible starting point while we figure out the rest of the story.

jobserver/models/core.py
```diff
@@ -877,6 +892,19 @@ def clean(self):
         super().clean()
         self.email = self.__class__.objects.normalize_email(self.email)
 
+    @classmethod
+    @transaction.atomic()
+    def from_login_token(cls, token):
+        user = User.objects.get(
```
bloodearnest, line 898: So, I think we should add the username/id into the url also, and look up based on both username *and* token.  This prevents link from ever logging you in as the wrong user, if for some reason the two tokens where the same.  Unlikely, I know, but easily preventable so we probably should do it.

Also, we may want to distinguish between token expired and invalid token, possibly. So it might be worth separating the expiry check out from the get query.

I think we'd want to log whether it is one or other, even if what we say to the user is the same in both cases, to help us debug auth failures more easily.

Also, if we do the expiry check separately, and it has expired, we may want to reset it in the db right now, just to be clean and safe.

If we do respond to the user differently depending on reason for failure, then its more helpful to the user, but arguably leaks information about the token's existence or not.

jobserver/views/users.py
```diff
@@ -1,82 +1,64 @@
...
+        return next_url
+
+
+class LoginWithURL(View):
+    def get(self, request, *args, **kwargs):
+        try:
+            user = User.from_login_token(self.kwargs["token"])
+        except (User.DoesNotExist) as e:
+            print(e)
+            messages.error(request, "An error occurred, please try again")
```
bloodearnest, line 55:
```suggestion
            messages.error(request, "Invalid token, please try again")
```

bloodearnest (changes requested):
Looks good, couple of changes suggested to tighten security

First two have more detailed PR comments. They are argubly not essential, but are easy to do and safer.

1) use `secrets.token_urlsave` not `new_ulid_str`
2) add explicit username to email and use that in lookup to cross-user login errors

And the 3rd suggestion is below. I think we definitely should do this one, as much for avoiding the bad UX as the security:


I think we need to combine the user clicking the login request with setting a cookie with the token in too, and requiring that cookie to also be present and valid when handling the link before logging them in.

This improves security by ensuring that the user has to evaluate the login from the same browser that they requested the link - an attacker can't just use a link it finds in an email.

It also prevents bad UX when an email client or spam filter "helpfully" pre-fetches/crawls the GET in the email link, thus expiring the token so it doesn't work for the user.

bloodearnest:
> * Should we explicitly disallow social-auth users from using this method of login?

IMO, yes. This would allow an attacker to potentially bypass github 2fa login to run jobs. And I don't think it provides a usabilty win for those users, as they have to have a GH account anyway.

jobserver/models/core.py
lucyb: 15 mins might be too short for NHSmail though :disappointed: 

Maybe 30 mins would be a resonable compromise?

jobserver/models/core.py
ghickman: I always forget about the secrets module 👌

jobserver/views/users.py
ghickman: Should we expose that there was a problem with the token specifically?  I'm not sure it adds anything meaningful for the user to know this, it's not like they can do anything about it!

bloodearnest: I think it does.

"Error" suggests something went wrong on our end, possibly nothing to do with their login attempt.

But this is either a token mismatch, or an expired token.  The former is also what happens when they click on the same url again.

"An error" is just too open to be useful to guide them what went wrong

jobserver/views/users.py
```diff
@@ -1,22 +1,64 @@
...
+
+        return next_url
+
+
+class LoginWithURL(View):
+    def get(self, request, *args, **kwargs):
+        try:
+            user = User.from_login_token(self.kwargs["token"])
+        except (User.DoesNotExist) as e:
+            print(e)
```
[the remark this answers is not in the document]
ghickman: Yep 🤦‍♂️

Logging incoming 👍

jobserver/models/core.py
```diff
@@ -837,6 +836,9 @@ class User(AbstractBaseUser):
     pat_token = models.TextField(null=True, unique=True)
     pat_expires_at = models.DateTimeField(null=True)
 
+    # this token is generated for users when they want to log in
+    login_token = models.TextField(null=True, unique=True)
```
bloodearnest, line 840: I don't think we need to store the token in db any more - we can just add it to the session.

That way there's no other way except via the session.

templates/emails/login.txt
```diff
@@ -0,0 +1,5 @@
+We have received a request to log into OpenSAFELY for your account. If you think this was made in error please ignore this email.
+
+Otherwise, please the link below to log in, it is valid for {{ timeout_minutes }} minutes:
```
bloodearnest, line 3:
```suggestion
Otherwise, please click the link below to log in, it is valid for {{ timeout_minutes }} minutes.
```

jobserver/models/core.py
```diff
@@ -877,6 +879,22 @@ def clean(self):
...
 
+    @classmethod
+    @transaction.atomic()
+    def from_login_token(
+        cls,
+        email,
+        signed_token,
+        max_age=timedelta(minutes=settings.LOGIN_URL_TIMEOUT_MINUTES),
+    ):
+        token = TimestampSigner(salt="login").unsign(signed_token, max_age=max_age)
```
bloodearnest, line 890: Worth a comment explaining that we're verifying the token in the email is valid and hasn't expired?

jobserver/views/users.py
```diff
@@ -1,82 +1,101 @@
...
+            email = request.session.pop(INTERNAL_LOGIN_SESSION_TOKEN)
+            user = User.from_login_token(email, self.kwargs["token"])
+        except (BadSignature, KeyError, SignatureExpired, User.DoesNotExist) as e:
+            logger.exception("Login failed")
+
+            # also log to our metrics host
+            span = trace.get_current_span()
+            span.record_exception(e)
+
+            messages.error(request, "Invalid token, please try again")
```
bloodearnest, line 85: Worth adding something here like:
```The link will only work in the same browser you requested the login email from.```

I do wonder if we should handle the missing session data a bit differently, perhaps.

File: jobserver/views/users.py [hunk 10]
```diff
@@ -1,82 +1,101 @@
...
+
+        return next_url
+
+
+class LoginWithURL(View):
+    def get(self, request, *args, **kwargs):
+        try:
+            email = request.session.pop(INTERNAL_LOGIN_SESSION_TOKEN)
+            user = User.from_login_token(email, self.kwargs["token"])
+        except (BadSignature, KeyError, SignatureExpired, User.DoesNotExist) as e:
```
bloodearnest, line 78: Not sure about lumping all these failure cases together, tbh. But I don't have a concrete reason not to.

jobserver/views/users.py
```diff
@@ -1,82 +1,101 @@
...
-                # that they don't have a password to reset
+                # we don't want to expose users email address to a bad actor
+                # via the login page form so we're emailing them with a link
+                # to the login page to click the right button.
+                # We can't email them with a link to the social auth entrypoint
+                # because it only accepts POSTs.
                 send_github_login_email(user)
             else:
-                send_reset_password_email(user)
+                self.request.session[INTERNAL_LOGIN_SESSION_TOKEN] = user.email
```
bloodearnest, line 49: I think we should refuse to send an email if they're not interactive users? Same validation as like the POST handler, basically

bloodearnest (commented):
Looking much nicer.

Left a few comments, tweaks mostly.

jobserver/views/users.py
ghickman: Yeah it feels weird!  But I think specifying the exact use cases we care about then logging those particular instances means it _should_ be ok.

jobserver/views/users.py
```diff
@@ -1,82 +1,119 @@
...
-        # * unknown user
-        # * GitHub user
-        # * email-only user
-        # we want to show the same outcome regardless of state so a bad actor
-        # can't work out work out who uses the service by email
         if user:
+            if not has_role(user, InteractiveReporter):
+                messages.error(
+                    self.request,
+                    "Only users who have signed up to Interactive can log in via email",
```
bloodearnest, line 47:
```suggestion
                    "Only users who have signed up to OpenSAFELY Interactive can log in via email",
```

jobserver/views/users.py
```diff
@@ -1,82 +1,119 @@
...
-            return "/"
-
-        # if the user is logging in via this view, so with an email address and
-        # password, _and_ has the InteractiveReporter role then they should
-        # definitely be an interactive user.
-
-        # TODO: add a page for users who have 2+ projects
-        project = self.request.user.projects.first()
-        return project.get_interactive_url()
+                self.request.session[INTERNAL_LOGIN_SESSION_TOKEN] = user.email
```
bloodearnest, line 59: We need to store the token in the session to compare.

This means restucturing all the auth bits to be unrelated to the mode (which they are not really anymore)

e.g.

```
token = secrets.token_urlsafe(64)
self.request.session[INTERNAL_LOGIN_SESSION_TOKEN] = (user.email, token)
signed_token = TimestampSigner(salt="login").sign(token)
url = reverse("login-with-url", kwargs={"token": signed_token})
send_login_email(
    user, 
    url=url,
    timeout_minutes=settings.LOGIN_URL_TIMEOUT_MINUTES,
)
```

jobserver/views/users.py
```diff
@@ -1,82 +1,119 @@
...
+        if not is_safe_path(next_url):
+            next_url = ""
+
+        return next_url
+
+
+class LoginWithURL(View):
+    def get(self, request, *args, **kwargs):
+        try:
+            email = request.session.pop(INTERNAL_LOGIN_SESSION_TOKEN)
```
bloodearnest, line 86: Similar to the above, we need to compare the unsigned url token to the stored session token.

If we don't do this, then an attacker could potentially use a signed token for any login request.

Also, we should probably only delete the session data on *successful* login, or else copy/pasting to the right browser won't work.

Altogether, I think this looks something like this:


```

try:
    email, token = request.session.get(INTERNAL_LOGIN_SESSION_TOKEN)
    url_token = TimestampSigner(salt="login").unsign(
        self.kwargs["token"],
        max_age=timedelta(minutes=settings.LOGIN_URL_TIMEOUT_MINUTES),
    )
    if url_token != token:
        raise BadToken()
    user = User.objects.get(email=email)
    request.session.pop(INTERNAL_LOGIN_SESSION_TOKEN)
except (BadSignature, BadToken, KeyError, SignatureExpired, User.DoesNotExist) as e:
    ...
    
```

jobserver/views/users.py
bloodearnest: I think we need to do something different in the case of the session KeyError - I think that represents the user using the wrong browser, and we should response with a helpful page about copy/pasting.

But I think that can wait for a later PR

bloodearnest (commented):
I think we've crossed wires somewhere - we still need to compare tokens, but we can store it in the session, not the db.

I've added some code snippets which are more explicit about what I mean.


I've also had some thoughts about the signing, I suspect we should probably sign the *entire* url, and somehow include the user in that too, but I think that can wait for a later PR.

Issue #2751, opened 2023-03-03
Dependabot PR checks aren't running in the merge queue


ghickman:
We've turned on GitHub's [merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions) but Dependabot PRs aren't triggering any checks to run when they get added to the merge queue.  Eventually they are removed from the queue after ~3h of non-activity.

It looks like we need to trigger checks on the [`merge_group` target](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions) in our `main.yml`.

madwort:
next assumption - permissions?

PR #2769, opened 2023-03-08
switch to pull_request target so we don't run CI on every push

ghickman:
Ref: #2751

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,9 @@ env:
 
 on:
   merge_group:
+  pull_request:
   push:
+    branches: [main]
 
 jobs:
   check:


.github/workflows/main.yml
```diff
@@ -9,7 +9,7 @@ env:
 
 on:
   merge_group:
-  push:
+  pull_request:
```
bloodearnest, line 12: I think we'll still want push, but only on main? So that we run checks post-merge?

Or does the merge queue make that unessesary now?

so
```
 push:
   branches: [main]
```

ghickman: Hah, yep, that would be useful wouldn't it!

bloodearnest (approved):

Issue #2622, opened 2023-02-10
ADR: OSIv2 submit method


ghickman:
In v1 we created commits in a particular way in a particular place, with v2 this has changed and we should explain what's changed and why.

PR #2773, opened 2023-03-08
Describe the changes to analysis submission in OSIv2

ghickman:
Fix: #2622

### DIFF
--- a/docs/adr/0010-creating-interactive-queries-v2.md
+++ b/docs/adr/0010-creating-interactive-queries-v2.md
@@ -1,0 +1,22 @@
+# 10. Creating Interactive queries v2
+Date: 2022-04-14
+
+## Status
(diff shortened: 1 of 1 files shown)


bloodearnest (approved):

lucyb:
This is a very well written and clear ADR. Unfortunately, it currently repeats a lot of ADR 0008. I would suggest adding some detail about how we're now creating one repo per project rather than just having a single repo for all of OSI and how we're pulling in the details from a template repo.

ghickman:
@lucyb – I'd completely forgotten about ADR 0008 🤦‍♂️  I've updated the text to link to that rather than copy it, and called out that we'll have one repo per project for interactive bits.  How does that read?

lucyb (approved):

Issue #2774, opened 2023-03-09
Add a jobs endpoint


ghickman:
The URL we use for our JobDetail page is too long for Sentry to display (or keep?), which makes debugging a particular job much harder.

@evansd has suggested we have a top-level URL which takes a job identifier and redirects to the appropriate detail page.  We can then use that URL for the `job` tag on our sentry events allowing us to click through to jobs from Sentry.

The URL should, of course, be `/jobsworth/<job identifier>/`.

PR #2775, opened 2023-03-09
Use the JobDetail redirector URL in Sentry

ghickman:
Fix: #2774


evansd (approved):
:tada: Well done past George!

PR #2777, opened 2023-03-09
Check all roles for magic link login access

ghickman:
`has_role` looks at the "global" roles of a user, optionally searching any extra context we give it.  Rather than look up all the possible org and project memberships of the user in the view this adds a property to the User model which grabs them as a set so we can have simple membership checks.

### DIFF
--- a/jobserver/views/users.py
+++ b/jobserver/views/users.py
@@ -14,7 +14,7 @@ from django.views.generic import FormView, UpdateView, View
 from furl import furl
 from opentelemetry import trace
 
-from jobserver.authorization import InteractiveReporter, has_role
+from jobserver.authorization import InteractiveReporter
 
 from ..emails import send_github_login_email, send_login_email
 from ..forms import EmailLoginForm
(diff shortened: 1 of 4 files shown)


jobserver/views/users.py
```diff
@@ -48,7 +48,7 @@ def form_valid(self, form):
         user = User.objects.filter(email=form.cleaned_data["email"]).first()
 
         if user:
-            if not has_role(user, InteractiveReporter):
+            if InteractiveReporter not in user.all_roles:
```
lucyb, line 51: I know this is the existing behaviour, but are we potentially revealing the existence of a user account by returning back this message? Would it be better just to return back the message on line 80 regardless, but only email them the login link if they have the Interactive role?

jobserver/views/users.py
```diff
@@ -48,13 +48,6 @@ def form_valid(self, form):
...
 
         if user:
-            if not has_role(user, InteractiveReporter):
-                messages.error(
-                    self.request,
-                    "Only users who have signed up to OpenSAFELY Interactive can log in via email",
-                )
-                return redirect("login")
-
             if user.social_auth.exists():
```
lucyb, line 51: You might have spotted it by the time you see this comment, but the else on line 65 needs to become:

`elif InteractiveReporter not in user.all_roles:`

lucyb (approved):

Issue #2767, opened 2023-03-07
Email requester when their Interactive report is ready to view


lucyb:
Once an output checker has released the outputs from an analysis request, the requester should be notified by email that their report is ready to view. The email should contain a direct link to the report and a reminder about not sharing the results beyond their immediate team (Catherine to confirm wording).

This was the message from v1 for comparison
```
Hi {{ name }},

Your analysis titled "{{ title }}" using OpenSAFELY interactive has now finished running. You can view the results at:

{{ url }}

These results are not publicly accessible and are only viewable when logged into your OpenSAFELY Interactive account. Please do not share these results with anyone. If you would like to share the results, please send an email to us at team@opensafely.org

```

PR #2789, opened 2023-03-10
Email the requester when a report for an AnalysisRequest is uploaded

ghickman:
Fix: #2767

### DIFF
--- a/interactive/emails.py
+++ b/interactive/emails.py
@@ -1,0 +1,21 @@
+from django.conf import settings
+from furl import furl
+from incuna_mail import send
+
+
+def send_report_uploaded_notification(analysis_request):
+    analysis_request_url = furl(settings.BASE_URL) / analysis_request.get_absolute_url()
+
+    context = {
+        "name": analysis_request.created_by.name,
+        "title": analysis_request.title,
+        "url": analysis_request_url,
+    }
+
+    send(
+        to=analysis_request.created_by.email,
+        sender="notifications@jobs.opensafely.org",
+        subject="Your OpenSAFELY Interactive report is ready to view",
+        template_name="emails/notify_report_uploaded.txt",
+        context=context,
+    )
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -20,6 +20,7 @@ from rest_framework.parsers import FileUploadParser, JSONParser
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
+from interactive.emails import send_report_uploaded_notification
 from interactive.models import AnalysisRequest
 from jobserver import releases, slacks
 from jobserver.api.authentication import get_backend_from_token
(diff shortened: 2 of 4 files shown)


interactive/emails.py
```diff
@@ -0,0 +1,21 @@
...
+    context = {
+        "name": analysis_request.created_by.name,
+        "title": analysis_request.title,
+        "url": analysis_request_url,
+    }
+
+    send(
+        to=analysis_request.created_by.email,
+        sender="notifications@jobs.opensafely.org",
+        subject=f"A report has been released for your interactive study: {analysis_request.title}",
```
lucyb, line 18: The word "released" is quite Job Server specific, I'm not sure Interactive users will know what that means. Also, being extremely picky, I'm not sure Interactive users will be doing "studies" either, although that's a conversation for another time.

How about this for the subject: "Your OpenSAFELY Interactive report is ready to view"

jobserver/api/releases.py
```diff
@@ -67,19 +68,22 @@ def maybe_create_report(rfile, filename, user):
...
+        title=analysis_request.title,
+        description="TODO fill from AR",
+        created_by=user,
+        updated_by=user,
+    )
+    analysis_request.report = report
+    analysis_request.save(update_fields=["report"])
+
+    send_report_uploaded_notification(analysis_request)
+    # TODO: notify some staff…?
```
lucyb, line 86: Super tiny nitpick: I'm not sure if we'll need to notify anyone else, so this TODO is probably redundant.

templates/emails/notify_report_uploaded.txt
```diff
@@ -0,0 +1,7 @@
+Hi {{ name }},
+
+Your analysis titled "{{ title }}" using OpenSAFELY interactive has now finished running. You can view the results at:
+
+{{ url }}
+
+These results are not publicly accessible and are only viewable when logged into your OpenSAFELY Interactive account. Please do not share these results with anyone. If you would like to share the results, please send an email to us at team@opensafely.org
```
lucyb, line 7: If we're using report in the title, then we should probably be consistent within the body of the email (i.e s/results/report/).

This second part has changed! People can share the results within their team now. I have some suggested wording here, but can you confirm it with Catherine please because I can't find the correct IG document right now?

> This report is not publicly accessible and is only viewable when logged into your OpenSAFELY Interactive account. Please only share these results with your immediate team. If you would like to share the report more widely, please send an email to us at team@opensafely.org

ghickman:
@lucyb – all changes done 👍

templates/emails/notify_report_uploaded.txt
```diff
@@ -0,0 +1,7 @@
+Hi {{ name }},
+
+Your analysis titled "{{ title }}" using OpenSAFELY interactive has now finished running. You can view the results at:
```
lucyb, line 3:
```suggestion
Your analysis titled "{{ title }}" using OpenSAFELY Interactive has now finished running. You can view the results at:
```

lucyb (approved):

PR #2792, opened 2023-03-13
Document why and how the magic login links work

ghickman:


lucyb (approved):
This looks good. It nicely explains why we've made this decision.

Do we have any record of slack (or other) conversations with Amir where we can point to his approval? If we don't have it yet, then it might be useful to add it later on just to make life easier for the future traveller.

bloodearnest (approved):

PR #2793, opened 2023-03-13
feat: add interactive_templates as a package

bloodearnest:
Now we just import the schema from the `interactive_templates` package,
and call its render method.  The removes all references and use of
a template repo, and tests that convered the actualy rendering logic.

As the interactive_templates packaging branch cannot land until this
change is deployed, we pin to specific branch commit of it in this
change. After this is deployed, we'll switch the dependency to point to
a version tag on main.

### DIFF
--- a/interactive/submit.py
+++ b/interactive/submit.py
@@ -156,19 +156,19 @@ def create_commit(
     with tempfile.TemporaryDirectory(suffix=suffix) as repo_dir:
         repo_dir = Path(repo_dir)
 
-        # 4. clone the given interactive repo
+        # 2. clone the given interactive repo
         git("clone", "--depth", "1", repo_url, repo_dir)
 
-        # 5. clear working directory because each analysis is fresh set of files
+        # 3. clear working directory because each analysis is fresh set of files
         clean_working_tree(repo_dir)
 
-        # 6. render the files into the interactive repo
+        # 4. render the files into the interactive repo
         render_analysis(analysis, repo_dir)
 
-        # 7. write a commit to the given interactive repo
+        # 5. write a commit to the given interactive repo
         sha = commit_and_push(repo_dir, analysis)
 
-        # 8. return contents of project.yaml (from disk) and sha
+        # 6. return contents of project.yaml (from disk) and sha
         project_yaml = (repo_dir / "project.yaml").read_text()
 
     return sha, project_yaml


interactive/submit.py
```diff
@@ -148,60 +143,33 @@ def get_repo_with_token(repo):
...
-        suffix = f"repo-{analysis.id}"
-        with tempfile.TemporaryDirectory(suffix=suffix) as repo_dir:
-            repo_dir = Path(repo_dir)
+    suffix = f"repo-{analysis.id}"
+    with tempfile.TemporaryDirectory(suffix=suffix) as repo_dir:
+        repo_dir = Path(repo_dir)
 
-            # 4. clone the given interactive repo
-            git("clone", "--depth", "1", repo_url, repo_dir)
+        # 4. clone the given interactive repo
```
ghickman, line 159:
```suggestion
        # 2. clone the given interactive repo
```

interactive/submit.py
```diff
@@ -148,60 +143,33 @@ def get_repo_with_token(repo):
...
+        git("clone", "--depth", "1", repo_url, repo_dir)
 
-            # 5. clear working directory because each analysis is fresh set of files
-            clean_working_tree(repo_dir)
+        # 5. clear working directory because each analysis is fresh set of files
```
ghickman, line 162:
```suggestion
        # 3. clear working directory because each analysis is fresh set of files
```

interactive/submit.py
```diff
@@ -148,60 +143,33 @@ def get_repo_with_token(repo):
...
+        clean_working_tree(repo_dir)
 
-            # 6. render the files into the interactive repo
-            analysis_template.render(repo_dir, asdict(analysis))
+        # 6. render the files into the interactive repo
```
ghickman, line 165:
```suggestion
        # 4. render the files into the interactive repo
```

interactive/submit.py
```diff
@@ -148,60 +143,33 @@ def get_repo_with_token(repo):
...
+        render_analysis(analysis, repo_dir)
 
-            # 7. write a commit to the given interactive repo
-            sha = commit_and_push(repo_dir, analysis)
+        # 7. write a commit to the given interactive repo
```
ghickman, line 168:
```suggestion
        # 5. write a commit to the given interactive repo
```

interactive/submit.py
```diff
@@ -148,60 +143,33 @@ def get_repo_with_token(repo):
...
+        sha = commit_and_push(repo_dir, analysis)
 
-            # 8. return contents of project.yaml (from disk) and sha
-            project_yaml = (repo_dir / "project.yaml").read_text()
+        # 8. return contents of project.yaml (from disk) and sha
```
ghickman, line 171:
```suggestion
        # 6. return contents of project.yaml (from disk) and sha
```

ghickman (approved):
Just some tiny fixes to the comments so the ordering makes sense 👍

Issue #2798, opened 2023-03-13
ADR: pairing fields with check constraints


ghickman:
We have a variety of model fields which pair who did something (eg `created_by`) with when they did it (eg `created_at`).  We want these pairs to stay in sync, and Django's [CheckConstraints](https://docs.djangoproject.com/en/4.1/ref/models/constraints/#checkconstraint) are a great way to ensure this happens even down at the database level.

We should document this choice in an ADR with some examples for common use cases.

PR #2800, opened 2023-03-14
Document why and how to use CheckConstraints for paired fields

ghickman:
Record what and why we're using CheckConstraints with examples on how others can make use of them.

Fix: #2798

### DIFF
--- a/docs/adr/0012-paired-field-state-with-check-constraints.md
+++ b/docs/adr/0012-paired-field-state-with-check-constraints.md
@@ -1,0 +1,19 @@
+# 12. Ensuring paired field state with CheckConstraints
+Date: 2023-03-14
+
+## Status
+Accepted
+
+## Context
+We have fields on many of our database models which are inherently related.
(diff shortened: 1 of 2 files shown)


docs/adr/0012-paired-field-state-with-check-constraints.md
```diff
@@ -0,0 +1,17 @@
...
+## Context
+We have fields on many of our database models which are inherently related.
+The vast majority of these track *who* did something and *when* they did it.
+We want to ensure the state of these fields stays in sync without having to remember to do so at call sites.
+
+## Decision
+We will use [Django's CheckConstraints](https://docs.djangoproject.com/en/4.1/ref/models/constraints/#s-checkconstraint) to codify paired field relationships into the database.
+
+## Consequences
+We will have to put some more thought into our model creation and review to remember to add these constraints.
```
lucyb, line 16: We will also document how we use these constraints by adding some instructions into the developers.md!

PR #2820, opened 2023-03-14
fix: bump to latest templates and revert to output/ID/report.html path

bloodearnest:


tests/unit/jobserver/api/test_releases.py
```diff
@@ -1311,3 +1312,28 @@ def test_workspacestatusapi_success(api_rf):
...
     assert response.status_code == 200
     assert not response.data["uses_new_release_flow"]
+
+
+def test_maybe_create_report_no_match(slack_messages):
+
+    # not 3 parts
+    assert maybe_create_report(ReleaseFileFactory(name="no_match"), None) is None
+    assert len(slack_messages) == 0
+    # starts with output but not endswith report.html
```
ghickman, line 1322: Could you add a blank line before each of these section comments, it really helps denote them as sections.

ghickman (approved):
One small nit to help when scanning down the code, otherwise 👍

tests/unit/jobserver/api/test_releases.py
bloodearnest: done

Issue #2809, opened 2023-03-14
Automatically create the GitHub issue for the output checking request


lucyb:
This is dependent on https://github.com/opensafely-core/job-server/issues/2591 EDIT: we can create the issue in the existing [output-checkers repository](https://github.com/ebmdatalab/opensafely-output-review/issues), which will mean the output-checkers are notified automatically in slack.

We should automatically create an issue for the output checkers to review the outputs of an Interactive request, once the job request has successfully completed. The issue should be given the label "interactive".

The output checking process is slightly different to normal, so we need to make those differences clear in the ticket and include the exact `osrelease` command that the output-checker will need to run to correctly release the file.

See OSI v1 for reference: https://github.com/opensafely-core/interactive.opensafely.org/blob/bb4166cfa86c832f6caaab05719a5078eddf3bfd/services/github.py#L23

PR #2823, opened 2023-03-15
Notify output checkers when the jobs for an AnalysisRequest finish

ghickman:
This replicates the functionality from OSIv1 with a few differences:
* the ticket is created when the JobRequest finishes since we already have a hook for those sorts of notifications
* the body has been updated to match the layout of the current issue template the output checkers use
* it creates the ticket in the output checkers tracker instead of the interactive one

Fix: #2809

### DIFF
--- a/jobserver/api/jobs.py
+++ b/jobserver/api/jobs.py
@@ -12,8 +12,10 @@ from rest_framework.generics import ListAPIView
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
+from interactive.notifications import notify_output_checkers
 from jobserver.api.authentication import get_backend_from_token
 from jobserver.emails import send_finished_notification
+from jobserver.github import _get_github_api
 from jobserver.models import JobRequest, Stats, User, Workspace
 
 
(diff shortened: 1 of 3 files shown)


jobserver/api/jobs.py
```diff
@@ -195,6 +200,9 @@ def handle_alerts_and_notifications(request, job_request, job):
             user_id=job_request.created_by_id,
         )
 
+    if hasattr(job_request, "analysis_request"):
```
bloodearnest, line 203: Hmm, this seems odd, probably a previous decision, but could JobRequest.analysis_request be a property or similar, rather than a possibly missing attribute?

tests/unit/jobserver/api/test_jobs.py
```diff
@@ -555,6 +565,40 @@ def test_jobapiupdate_post_with_flags(api_rf):
...
+    request = api_rf.post(
+        "/",
+        HTTP_AUTHORIZATION=job_request.backend.auth_token,
+        data=data,
+        format="json",
+    )
+
+    response = JobAPIUpdate.as_view(get_github_api=FakeGitHubAPI)(request)
+
+    assert response.status_code == 200
```
bloodearnest, line 599: Shouldn't we also check that a github issue was created? Or the api at least called?

bloodearnest (approved):
oops. forgot to approve

jobserver/api/jobs.py
ghickman: This is an artefact of it being a OneToOneField defined on AnalysisRequest.  When the relationship is missing, the target model (JobRequest in this instance) doesn't have the instance variable defined.  It is _maddening_.

Issue #2804, opened 2023-03-14
Link to create interactive user from an interactive project


ghickman:
When creating an interactive project it is then confusing to add new interactive users.  We can improve this by linking to the `Create interactive user` page from a project page with an interactive workspace.

To improve the UX of that flow we can also:
* set the `project-slug` query arg to pre-fill the project selector
* set and make use of a `next` query arg to take the user back to the project

PR #2824, opened 2023-03-15
Redirect back to add members

ghickman:
This is a two parter:

#### Reworked query arg handling
We've been dabbling in using query args in various places but the functionality was a mixture of furl and QueryDict workarounds.  We know we want to make more use of `next` args, so having an easy way to grab that arg in a view is going to be useful.  I've refactored the existing uses, hopefully making them a little clearer and easier to reason about in the process, and removed a lot of lines in the process.

#### Creating an interactive user now follows the next arg
When coming to the "Create an interactive user" page from the "Add member" page you will now get redirected after form submission to the appropriate project's detail page.  This is done by setting the appropriate `next` arg for the link on the "Add member" page.  It redirects to the detail page because the "Create an interactive user" page already sets a project on the created user, and the project is preselected by dint of the passed `project-slug`.

Fix: #2804


staff/views/orgs.py
```diff
@@ -64,8 +65,11 @@ def form_valid(self, form):
         return HttpResponseClientRedirect(url)
 
     def get_context_data(self, **kwargs):
+        f = furl(reverse("staff:org-create"))
+        f.args.update(self.request.GET)
```
bloodearnest, line 69: I am unclear what this call to update() does.  What is self.request.GET? 

Worth a comment to explain? Or else the reader needs to be familiar with both furl and django specifics, perhaps?

bloodearnest: oh, GET is the dict of original request querystring args. Ok, I was being dumb

bloodearnest (approved):

staff/views/orgs.py
ghickman: This is absolutely me being deeply familiar with both those things, I'll add a comment to explain what we're doing here 👍

Issue #2799, opened 2023-03-13
Improving UX of links to outputs in Workspace


lucyb:
At the moment, on the Releases section of a Workspace the links to "Level 4 outputs", "Released files" and "Published files" only appear if there are files available matching those criteria. This can [cause confusion](https://bennettoxford.slack.com/archives/C0270Q313H7/p1678726719011599?thread_ts=1678367781.695419&cid=C0270Q313H7).

A better solution would be to always display all three links, but to disable/grey out the links that aren't available to be selected and to include a tooltip explaining why (e.g "No files have been released for this workspace")

![Image](https://user-images.githubusercontent.com/3889554/224782171-bf1a796a-78e0-46b5-a98b-a0cfca2d00ab.png)

ghickman:
This would fit nicely with how we handle buttons around the site too 👍

PR #2825, opened 2023-03-16
Always show the outputs links on WorkspaceDetail pages

ghickman:
Matching how we handle buttons that a user doesn't have access to this disables the outputs links when a user doesn't have access to them.

As part of this I've reworked how we expose the aggregated permissions checks to the template so they're easier to work with there.

Fix: #2799

### DIFF
--- a/jobserver/views/workspaces.py
+++ b/jobserver/views/workspaces.py
@@ -249,15 +249,61 @@ class WorkspaceDetail(View):
             and repo_is_private
         )
 
+        can_archive_workspace = has_permission(
+            request.user, "workspace_archive", project=workspace.project
+        )
+        can_run_jobs = has_permission(
+            request.user, "job_run", project=workspace.project
+        )
+        can_toggle_notifications = has_permission(
+            request.user, "workspace_toggle_notifications", project=workspace.project
+        )
+
+        honeycomb_can_view_links = has_role(self.request.user, CoreDeveloper)
+
+        is_interactive_user = has_permission(
+            request.user, "analysis_request_create", project=workspace.project
+        )
+        show_interactive_button = is_interactive_user and workspace.is_interactive
+
+        outputs = self.get_output_permissions(request.user, workspace)
+
+        reports = Report.objects.filter(release_file__workspace=workspace)
+
+        # remove unpublished reports from the list for non-project members
+        if not workspace.project.members.filter(pk=request.user.pk).exists():
+            reports = reports.exclude(publish_request__approved_at=None)
+
+        context = {
+            "first_job": first_job,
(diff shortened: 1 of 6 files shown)


jobserver/views/workspaces.py
```diff
@@ -318,6 +294,44 @@ def get(self, request, *args, **kwargs):
...
+
+        # a user can see backend files if they have access to at least one
+        # backend and the permissions required to see outputs
+        has_backends = (
+            user.is_authenticated and user.backends.exclude(level_4_url="").exists()
+        )
+        can_view_files = is_privileged_user and has_backends
+
+        # are there any releases to show for the workspace?
+        can_view_releases = workspace.releases.exists()
```
lucyb, line 310: When testing this locally, I was able to click on the "Released Outputs" link and go to the releases page, while I was logged out and also as an interactive user.

Clicking on one of the releases produces a 404, but still seems like the link shouldn't be enabled.

jobserver/views/workspaces.py
```diff
@@ -318,6 +294,44 @@ def get(self, request, *args, **kwargs):
...
+            published_at=None
+        ).exists()
+        has_any_snapshots = workspace.snapshots.exists()
+        can_view_outputs = has_published_snapshots or (
+            is_privileged_user and (can_view_releases or has_any_snapshots)
+        )
+
+        return {
+            "level_4": {
+                "disabled": not can_view_files,
```
lucyb, line 325: This might be one for a future PR, but it might be helpful to show the user whether the link is disabled because they don't have permission or because there aren't any files.

lucyb (commented):
There are some useful fixes in this PR that I would happily approve and seem unrelated to the change to the outputs links. If they were in a separate PR, I probably would have reviewed them sooner too. Sorry it took me a while to get around to this one.

ghickman:
@lucyb – updated and fixed a couple of things along the way.  You can ignore commits 2, 3, & 4.  I accidentally merged a dependent PR into this one, sorry for not updating the body!

jobserver/views/workspaces.py
lucyb: I can confirm it's fixed, but the behaviour is inconsistent with published outputs, which doesn't have a link if there are no outputs for unprivileged users (it has a link for privileged users though!)

lucyb: For reference, I'm looking at http://localhost:8000/datalab/bmi-short-data-report/bmi-short-data-report/ while logged in and logged out

ghickman: Fixed and a small tweak handled on the published page too

lucyb (approved):

PR #2839, opened 2023-03-20
Swith out flake8, isort, and pyupgrade for ruff 

ghickman:
Ruff is a linter for python written in rust.  I've been watching it for a while since it fixes my two biggest bugbears with flake8: it's blazing fast, and it's configuration lives with all our other tooling config in pyproject.toml.  It can also fix various rule failures, which is where it takes over from isort.  It's designed to be run in conjunction with black which is why it drops some of flake8's `E` rules which don't come up when using black.

Ruff's ruleset is now a full superset of the ones we've been using in flake8 which means we can look at using it.

Running locally `ruff check --no-cache .` takes ~40ms for me, while running our configurations of isort, flake8, & pyupgrade takes ~2.6s.  Import sorting is noticably quicker for me, and pre-commit also benefits from these speed ups.

#### Reviewing
The first commit contains all the tooling changes to switch over.  The subsequent commits fix errors thrown up by rust being slightly stricter than the existing tooling.

Noteably it caught that 3.11 added an alias for `timezone.utc` which resolves the slightly awkward overlap of Django's timezone (which has the very useful `now()` method allowing us to get a current datetime with[out] a timezone as defined by settings) and a possible-future-footgun with the dual import of `UserManager` where the second one was aliased but still appeared to replace the first!

#### Usage
Ruff ships with a pretty decent (and very well documented!) CLI which we're invoking in our tooling.  However it's always useful to have the ability to integrate tools like this with one's editor, and ruff is no slouch there: https://beta.ruff.rs/docs/editor-integrations/  Between the CLI, official addons, and its very own LSP I think we should cover every editor choice.


justfile
```diff
@@ -125,20 +125,15 @@ test *args:
 black *args=".": devenv
     $BIN/black --check {{ args }}
 
-isort *args=".": devenv
-    $BIN/isort --check-only --diff {{ args }}
-
-flake8 *args="": devenv
-    $BIN/flake8 {{ args }}
+django-upgrade *args="$(find applications jobserver services tests -name '*.py' -type f)": devenv
```
lucyb, line 128: How come the list of django apps has changed here?

lucyb (approved):
Works really well for me locally. It was trivial to get ruff working in vscode. I really like that it's already fixed a few gotchas too.

Can you share this at the tech team meeting (or in #tech) so that others are aware and we can consider using it more widely.

One comment to address before merging though!

bloodearnest (approved):
This is a solid improvement in multiple directions!
 - caught valid issues
 - one tool not 3
 - faster
 - applies isort fixes
 - consolidates config (no more .flake8 \o/)

We're gonna need a bigger Approve button :)

justfile
ghickman: Bad rebase when I pulled the update out into another PR, fixed!

ghickman: And I managed to botch fixing this here 🤦‍♂️ PR incoming!

PR #2841, opened 2023-03-20
Only notify output checkers when the JobRequest has succeeded

ghickman:
We only want to create a ticket for output checkers when the JobRequest enters the successful state, rather than any single job entering a completed state.

We have to do some leg work to look up the status both before and after the Job updates because we're dealing with JobRequests which come from a group_by generator, rather than a QuerySet.

### DIFF
--- a/jobserver/api/jobs.py
+++ b/jobserver/api/jobs.py
@@ -4,6 +4,7 @@ import operator
 
 import sentry_sdk
 import structlog
+from django.db.models import prefetch_related_objects
 from django.http import Http404
 from django.utils import timezone
 from rest_framework import serializers
@@ -94,6 +95,11 @@ class JobAPIUpdate(APIView):
                 )
                 continue
 
+            # grab the status before any updates so we can track updates to it
+            # after saving the incoming changes
+            prefetch_jobs(job_request)
+            initial_status = job_request.status
+
             # bind the job request ID to further logs so looking them up in the UI is easier
             structlog.contextvars.bind_contextvars(job_request=job_request.id)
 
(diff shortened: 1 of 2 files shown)


jobserver/api/jobs.py
```diff
@@ -200,10 +211,22 @@ def handle_alerts_and_notifications(request, job_request, job, github_api):
             user_id=job_request.created_by_id,
         )
 
+
+def handle_job_request_notifications(job_request, github_api):
```
lucyb, line 215: Should this section from `handle_job_notifications` be moved down here?

We'll have to tell people that it's changed, but I think they'd probably thank us?

```
if job_request.will_notify:
        send_finished_notification(
            job_request.created_by.notifications_email,
            job,
        )
        logger.info(
            "Notified requesting user of completed job",
            user_id=job_request.created_by_id,
        )
```

jobserver/api/jobs.py
```diff
@@ -200,10 +211,22 @@ def handle_alerts_and_notifications(request, job_request, job, github_api):
...
     if hasattr(job_request, "analysis_request"):
         notify_output_checkers(job_request, github_api)
```
lucyb, line 217: Would it be possible to add a test to `test_jobs.py` to check that `notify_output_checkers` is called in the same way that we're checking `send_finished_notification`?

lucyb (approved):

jobserver/api/jobs.py
ghickman: Done in #2848

Issue #2844, opened 2023-03-20
Add a field for co-pilot name onto create project page


lucyb:
A co-pilot can be added to a project by editing that project. However, since the co-pilot is usually known at project creation, it would be useful to be able to add them at that point.

This is a requirement for OSIv2, although it would be useful for all projects (I suspect).

PR #2847, opened 2023-03-21
Add copilot name

ghickman:
Fix: #2844


lucyb (approved):
This looks good to me, assuming a co-pilot is mandatory (but that could be changed later if needed).

Issue #2845, opened 2023-03-21
Add purpose of the Analysis Request


lucyb:
So that output checkers can complete their process to release files they need to know why the analysis request was made, so they know why the files have been requested for release. 

This needs to be added to the GitHub Issue that's created for the output checking process.
This should also be displayed alongside the report on Job Server.

PR #2850, opened 2023-03-21
Add AnalysisRequest.purpose

ghickman:
Fix: #2845


lucyb (approved):
I think these changes are fine (good call on removing the unnecessary field on the AnalysisRequest). However, it arguably doesn't complete #2845, because there's no way for the user to enter the purpose at the moment. I assume the plan is to hand it over to Tom O to do that? Would it be worth trying to do that yourself and letting Tom review the change?

ghickman:
@lucyb – updated with some help from Tom (thanks Tom!), do you want to give this a quick run through locally?

interactive/views.py
```diff
@@ -43,6 +43,7 @@ def build_analysis(self, *, form_data, project):
             created_by=self.request.user.email,
             demographics=raw.get("demographics", ""),
             filter_population=raw.get("filterPopulation", ""),
+            purpose=raw.get("purpose", ""),
```
lucyb, line 46: I'm not sure why, but I am not able to submit an analysis request because I'm not pulling in the updated schema from interactive-templates. If it works for you locally though, I'm going to assume there's something weird in my local setup.

assets/src/scripts/interactive/main.jsx
lucyb, line 1: Something I picked up while testing the change, but not related specifically to this change.

I hit this code block when I failed to submit an analysis request (because my version of the schema from interactive-templates is out of date). We should probably do something similar to what we did in v1 and say that their request was not submitted and that they should try again or contact support.

```
<Route>
                <div className="prose">
                  <p className="lead">An error occurred</p>
                  <p>404 - Page not found</p>
                </div>
</Route>
```

lucyb (approved):
A few niggles to pull out into other tickets, but otherwise looks good

interactive/views.py
ghickman: I had to run `pip uninstall interactive_templates && touch requirements.prod.in && just devenv` to get this working and have a note to see if I can work out why.  I should have mentioned on this PR, sorry!

assets/src/scripts/interactive/main.jsx
ghickman: I think we were showing an error on the submission form previously? @tomodwyer – have I messed up the routing?

tomodwyer: Uhhh that's odd 🤔 ok yes I can update that text.

assets/src/scripts/interactive/pages/review-request.jsx
```diff
@@ -123,6 +124,10 @@ function ReviewRequest() {
               ))}
             </ul>
           </ReviewLineItem>
+
+          <ReviewLineItem page="purpose" title="Purpose of this analysis">
+            <p>{formData.purpose}</p>
```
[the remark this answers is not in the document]
tomodwyer: I'll fix that

tomodwyer:
@lucyb @ghickman pushed fixes to the two front-end issues

PR #2873, opened 2023-03-24
chore: add ADR for template repo

### LINKED ISSUE #2636: ADR: GitHub auth token setup for OSI

bloodearnest:
Fixes #2636

### DIFF
--- a/docs/adr/0013-interactive-templates-repo.md
+++ b/docs/adr/0013-interactive-templates-repo.md
@@ -13,8 +13,9 @@ code templates to be rendered with the form data.
 These two parts are tightly coupled - changes to the form data schema needs to
 match changes in the templates.
 
-However, whilst the form itself must live in job-server repo, there are reasons
-having the templates in job-server repo is problematic:
+However, whilst the form itself [must live in job-server
+repo](0005-migrate-interactive-into-job-server.md), there are reasons that make
+having the templates there problematic:
 
 1) job-server would need to run tests for research code, which can be slow.
 
@@ -27,10 +28,13 @@ having the templates in job-server repo is problematic:
 
 ## Decision
 
-We will use a separate `interactive_templates` repo to hold the form data
-schema and templated code, as well as the glue code to perform the rendering
-and committing.  Job-server will have a pinned dependency on this repo as
-a python package.
+We will use a separate [interactive-templates
+repo](https://github.com/opensafely-core/interactive-templates) to hold the
+templated code, and a form data schema, as well as the glue code to perform the
+rendering and committing.
+
+Job-server will have a pinned dependency on this repo as a python package, and
+use the schema and glue code and bundled templates to generate OSI commits.
 
 
 ## Consequences
@@ -42,7 +46,12 @@ a python package.
    researcher to make changes.
 
 3. Explicit pinning means we can manage updates to the templates in job-server
-   as part of the normal dependency update workflow.
+   as part of the normal dependency update workflow. This does mean that making
+   a change to tempaltes for form data schema means committing it to
+   interactive-templates and then bumping the dependency. While slightly
+   onerous, this ensures we have robust synchronisation between job-server and
+   interactive-templates. i.e. committing to interactive-templates won't break
+   job-server.
 
 4. In future, it will be simpler to move the rendering to be in a separate
    secure service. Note that when we do this, job-server will *still* need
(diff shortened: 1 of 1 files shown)


docs/adr/0013-interactive-templates-repo.md
```diff
@@ -0,0 +1,51 @@
+# 13. Use external repo for interactive code schema and templates
+Date: 2023-03-24
+
+## Status
+Accepted
+
+## Context
+
+We need to support dynamically generating code from user input as part of OSI.
```
ghickman, line 9:
```suggestion
We need to support dynamically generated code from user input as part of OSI.
```

docs/adr/0013-interactive-templates-repo.md
```diff
@@ -0,0 +1,51 @@
...
+Functionally, this has two main parts: 1) the form the user fills in, and 2 analysis
```
ghickman, line 10:
```suggestion
Functionally, this has two main parts: 1) the form the user fills in, and 2) analysis
```

docs/adr/0013-interactive-templates-repo.md
```diff
@@ -0,0 +1,51 @@
...
+code templates to be rendered with the form data.
+
+These two parts are tightly coupled - changes to the form data schema needs to
+match changes in the templates.
+
+However, whilst the form itself must live in job-server repo, there are reasons
+having the templates in job-server repo is problematic:
```
ghickman, line 17:
```suggestion
However, whilst the form itself must live in the job-server repo, there are reasons that make doing so problematic:
```

docs/adr/0013-interactive-templates-repo.md
```diff
@@ -0,0 +1,51 @@
...
+2) researchers need to develop the research code using the job-server repo and
+   tooling, which is likey confusing.
+
+3) we want to move the rendering/committing of generated code out to a secure
+   service in future, for security reasons.
+
+
+## Decision
+
+We will use a separate `interactive_templates` repo to hold the form data
```
ghickman, line 30:
```suggestion
We will use a separate `interactive-templates` repo to hold the form data
```

ghickman (changes requested):
This is a great start 👍  I've added a few tweaks.

I think it's worth adding a consequence that adding a field to the data model requires one to work across two repos which is slightly more friction but we don't expect that to be common?

docs/adr/0013-interactive-templates-repo.md
```diff
@@ -0,0 +1,51 @@
...
+## Context
+
+We need to support dynamically generating code from user input as part of OSI.
+Functionally, this has two main parts: 1) the form the user fills in, and 2 analysis
+code templates to be rendered with the form data.
+
+These two parts are tightly coupled - changes to the form data schema needs to
+match changes in the templates.
+
+However, whilst the form itself must live in job-server repo, there are reasons
```
lucyb, line 16: Might be worth linking to [this ADR](https://github.com/opensafely-core/job-server/blob/5197171e19de658f513156c3508b6cf5a1c55ed6/docs/adr/0005-migrate-interactive-into-job-server.md) here.

lucyb (approved):
Looks good, thanks for doing this.

bloodearnest:
> This is a great start +1 I've added a few tweaks.
> 
> I think it's worth adding a consequence that adding a field to the data model requires one to work across two repos which is slightly more friction but we don't expect that to be common?

I thought that was implicit in the fact that the schema is a dependency?

docs/adr/0013-interactive-templates-repo.md
bloodearnest: This changes the meaning? Gonna fix the formatting myself

ghickman: Ah, I missed some words!  It was the dual use of `job-server repo` in the same sentence that I was trying to avoid.  How about:

> However, whilst the form itself must live in the job-server repo, there are reasons that make having the templates there problematic:

ghickman:
> > This is a great start +1 I've added a few tweaks.
> > I think it's worth adding a consequence that adding a field to the data model requires one to work across two repos which is slightly more friction but we don't expect that to be common?
> 
> I thought that was implicit in the fact that the schema is a dependency?

I think that's obvious if you've worked with this type of set up before, but likely not if you haven't.  So I think it's good to call that out explicitly for readers who are in the latter category!

bloodearnest:
@ghickman you happy with this now?

ghickman (approved):
Yep, that's really clear 👍

Issue #2837, opened 2023-03-17
Show Interactive reports on the Workspace page


lucyb:
We've released at least one report for the [internal workspace](https://jobs.opensafely.org/datalab/opensafely-internal/opensafely-internal-interactive/) now and will be releasing and publishing more. At present, it is not possible to see those reports unless you go via the staff area or have the email with the direct link.

It should be possible for anyone with access to the project to view a list of released Interactive reports on the workspace page.

It should be possible for anyone to view a list of published Interactive reports on the workspace page.

PR #2877, opened 2023-03-24
Simplify report published status

ghickman:
As part of #2837 I wanted to denote reports as drafts and realise we're duplicating the published data across two models.  This removes the duplication and adds a property, that I will use in a subsequent PR, to help expose the data simply to a Report.


jobserver/models/reports.py
```diff
@@ -77,6 +77,12 @@ def __str__(self):
     def get_staff_url(self):
         return reverse("staff:report-detail", kwargs={"pk": self.pk})
 
+    @property
+    def is_draft(self):
```
lucyb, line 81: I appreciate this is a term we've probably used elsewhere, but would it make more sense to flip it around and have a property called `is_published` instead?

In keeping with the terminology shown on the site for outputs, we only have released and published reports. Released could be a bit confusing to users (and probably devs too), so just referring to published seems a bit nicer than introducing a third option.

PR #2879, opened 2023-03-27
List workspace reports

ghickman:
This lists the reports built on files released to a given workspace.

**Without reports**

**With reports**

Fix: #2837

### DIFF
--- a/templates/workspace_detail.html
+++ b/templates/workspace_detail.html
@@ -270,6 +270,21 @@
         {% /card %}
       {% endif %}
 
+      {% #card title="Reports" %}
+        {% #list_group small=True %}
+          {% for report in reports %}
+          {% #list_group_item href=report.get_absolute_url %}
+            {{ report.title }}
+            {% if not report.is_published %}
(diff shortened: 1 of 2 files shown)


tomodwyer (approved):

templates/workspace_detail.html
```diff
@@ -270,6 +270,21 @@ <h1 class="mb-2 text-3xl tracking-tight break-words font-bold text-slate-900 sm:
         {% /card %}
       {% endif %}
 
+      {% #card title="Reports" %}
+        {% #list_group small=True %}
+          {% for report in reports %}
+          {% #list_group_item href=report.get_absolute_url %}
+            {{ report.title }}
+            {% if report.is_draft %}
+              {% pill text="Draft" variant="primary" %}
```
lucyb, line 279: Could we change this to "Not published" rather than "Draft" please?

Draft suggests that it could change. It also simplifies the terminology by saying that a report is either published or not. And, since "Not published" uses a negative, it feels like something you would want to change (by publishing!).

PR #2880, opened 2023-03-27
Card and empty list fixes

tomodwyer:
- Create a `list_group_empty` component
- Use `list_group_empty` on the home page
- Use `list_group_empty` on the project detail page
- Move marketing website content to a partial
- Add the option for a button in the card header, and use on the event log


templates/index.html
```diff
@@ -60,17 +61,17 @@ <h3 class="text-xl font-semibold leading-6 text-slate-900">Your OpenSAFELY Dashb
     </div>
 
     <div class="grid gap-4 lg:grid-cols-2 mb-6">
-      {% if analysis_requests %}
+      {% if user.is_interactive_only or analysis_requests %}
```
ghickman, line 63: I think I'm missing a use case for this one, what does `is_interactive_only` give us that `analysis_requests` doesn't?

ghickman (approved):

templates/index.html
tomodwyer: I was trying to get the empty state to display for users who have no yet filled out a request, but could end up on the home page.

I think that's the best way of doing it...? I could be missing something easier though.

ghickman: aha, that's the state I was missing! Carry on 😁

Issue #2876, opened 2023-03-24
Change the time scale to 10 years


lucyb:
The validation rule on the timescale in the Analysis Request form should be extended to a maximum of 10 years. This will allow people to more accurately query for long term conditions before the "any time" option is implemented. It was originally restricted to five years because of fears around performance, but Louis's tests suggest there is no noticeable difference.

Issue #2868, opened 2023-03-24
Choose demographic breakdown help text


tomodwyer:
> These should only be selected if they align with your approved project purpose. Selecting more options will increase report processing time.

PR #2892, opened 2023-03-28
Various fixes for the Interactive application and report view

tomodwyer:
- Remove success page from the React app as Django handles this
- Change time frame to 10 years and fix validation, so that users are aware of the limitations without having to `blur` the field
- Hide the title on the output report, as the title is above at the top of the page
- Use heroicon for arrow on codelist link, as the unicode icon displays differently on different operating systems
- Update the generated text for creating the request
- Closes https://github.com/opensafely-core/job-server/issues/2868
- Closes https://github.com/opensafely-core/job-server/issues/2876


lucyb (approved):
:tada:

PR #2902, opened 2023-03-30
login with token

bloodearnest:
- chore: add rf_messages fixture.
- chore: add basic login token functionality to User model
- feat: add form to login with token


jobserver/urls.py
```diff
@@ -273,9 +273,8 @@
     ),
     path("jobs/<identifier>/", JobDetailRedirect.as_view(), name="job-redirect"),
     path("login/", Login.as_view(), name="login"),
-    path(
-        "login-with-token/<str:token>/", LoginWithURL.as_view(), name="login-with-url"
-    ),
+    path("login-with-url/<str:token>/", LoginWithURL.as_view(), name="login-with-url"),
```
bloodearnest, line 276: Note - I renamed this url, which means any emails generated before we deploy won't work. But they only last an hour anyway...

ghickman (approved):

jobserver/models/core.py
```diff
@@ -838,6 +839,10 @@ class User(AbstractBaseUser):
     pat_token = models.TextField(null=True, unique=True)
     pat_expires_at = models.DateTimeField(null=True)
 
+    # single use token login
+    login_token = models.TextField(null=True)
```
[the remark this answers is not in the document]
bloodearnest: So set to empty string by default?

Hmm, not sure about that. This token should either be set or not.  The pat_token above does have null=True, for comparison.

I think it's more confusing that it's an empty string than whether its None or empty string?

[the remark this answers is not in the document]
ghickman: > This token should either be set or not.

The problem is that `login_token = ""` and `login_token = None` are very similar and IME can lead to confusion if say you're always expecting a string but it's been set to None.  So I've tried to have us stick to TextFields always containing strings.

However…

> The pat_token above does have null=True

I missed that when I was searching so I think it's probably fine as is 😁

jobserver/views/users.py
```diff
@@ -156,6 +165,81 @@ def login_invalid(self):
...
+
+
+@method_decorator(login_required, name="dispatch")
+class GenerateLoginToken(View):
+    template = "generate-login-token.html"
+
+    def get(self, request):
+        return TemplateResponse(self.request, self.template)
+
+    def post(self, request):
```
ghickman, line 232: Can you add `, *args, **kwargs` to these handler invocations?  Keeps them consistent with the rest of the CBVs 🙂 (path of least surprise etc)

jobserver/views/users.py
```diff
@@ -156,6 +165,81 @@ def login_invalid(self):
...
+        if not request.user.social_auth.exists():
+            return TemplateResponse(self.request, self.template)
+
+        token = self.request.user.generate_login_token()
+        formatted = f"{token[:4]} {token[4:]}"
```
ghickman, line 237: Took me a second to realise what this is doing, it's really nice! 👍

ghickman (approved):

jobserver/models/core.py
```diff
@@ -782,6 +783,11 @@ class UserManager(BaseUserManager.from_queryset(UserQuerySet), BaseUserManager):
     pass
 
 
+def human_memorable_token(size=8):
```
madwort, line 786: tiny nit: as I understandable, it's for reading & transcribing not memorisation

```suggestion
def human_readable_token(size=8):
```

jobserver/views/users.py
```diff
@@ -81,21 +101,10 @@ def form_valid(self, form):
     def get_context_data(self, **kwargs):
         return super().get_context_data(**kwargs) | {
             "next_url": self.next_url,
+            "token_form": TokenLoginForm(),
+            "token_form_submit_url": reverse("login-with-token"),
```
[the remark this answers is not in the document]
bloodearnest: done

jobserver/views/users.py
bloodearnest: done

jobserver/models/core.py
```diff
@@ -1033,6 +1043,37 @@ def rotate_token(self):
         # return the unhashed token so it can be passed to a consuming service
         return token
 
+    class BadLoginToken(Exception):
```
lucyb, line 1046: This section reminds me of a conversation @ghickman and I were having yesterday where we were talking about how and when to refactor parts of job-server. I wonder if the "business logic" of generating and validating a login token (and arguably a PAT token) would be better off in a separate module? The User class has seen a few changes recently and is just starting to look a bit too big.

bloodearnest: If we have a convention/location of where that should go, I can move it. 

But possilby as a follow up?

jobserver/models/core.py
bloodearnest: I specifically called it memorable because the human cannot look it at the same time they are typing it, due to how the VPN works. They may write it down, but I wanted to make it clear that we want to be able for a human to potentially hold in memory whilst they switch to VPN/RDP.

jobserver/models/core.py
bloodearnest: `commands` for me clashes with django management commands, conceptually.

I've previously used a `logic` package name for this?

jobserver/models/core.py
```diff
@@ -1049,10 +1053,15 @@ class BadLoginToken(Exception):
...
 
+    def _strip_token(self, token):
+        return token.strip().replace(" ", "")
+
     def generate_login_token(self):
         """Generate, set and return single use login token and expiry"""
         token = human_memorable_token()
-        self.login_token = make_password(token, salt="login-with-token")
+        self.login_token = make_password(
+            self._strip_token(token), salt="login-with-token"
```
lucyb, line 1063: Would it be better to not provide a salt here to add more protection against rainbow table attacks? As I understand, django generates a random salt if one is not provided.

templates/settings.html
```diff
@@ -24,16 +24,52 @@ <h1 class="md:col-span-3 text-3xl break-words md:text-4xl font-bold text-slate-9
...
+                {{ token }}
+              {% /code %}
+            </div>
+          {% endif %}
+          <p>
+            You can use this to login to OpenSAFELY from within secure Level 4 environments.
+          </p>
+          <p>A Single Use Token is valid for logging in for:</p>
+          <ul>
+            <li>one login attempt, or</li>
```
lucyb, line 55: This might be a clearer way of phrasing the requirements (people can get confused with the use of logical-or)

> A Single Use Token is only valid:
> * for one login attempt
> * for one hour from the time you generate it
> * until you generate a new Single Use Token

bloodearnest: done

PR #2903, opened 2023-03-30
Add releasing instructions into issue template

lucyb:
This should make it easier for output checkers to release the Interactive report correctly, as the process is slightly different to the standard releasing of outputs.


LFISHER7 (approved):
Thanks, this looks good!

Issue #2894, opened 2023-03-28
Viewing the amr-uom-brit repo in the staff area doesn't work


lucyb:
See https://jobs.opensafely.org/staff/repos/https%253A%252F%252Fgithub.com%252Fopensafely%252Famr-uom-brit/ (or rather don't, because it eventually returns an error)

The repo appears to be okay on github https://github.com/opensafely/amr-uom-brit

ghickman:
Locally this is doing 43 queries in 16469ms locally which explains why we're hitting a gunicorn timeout in prod.

It's the workspace building function triggering the `users` query on each iteration of the workspaces.  I'll switch this to cache the users via a lookup table for the workspace function.

ghickman:
Closing this one as I don't think we're likely to tackle the O(N) queries left but the page is now performant.

PR #2905, opened 2023-03-30
Fix RepoDetail pages with many workspaces

ghickman:
This improves the render time of the staff area's RepoDetail page.  Locally it goes from ~16800ms (43 queries in ~15900ms) to ~875ms (43 queries in ~127ms).  As the final commit details it's double filtering of the users which created a more complex query plan which added the extra time.

We're still doing some O(N) queries when we look up the users for a workspace but I've yet to find a way around those.  Given our data model (arrows are the many side of the relationships):

                    ┌─────────┐
                    │         │
                    │ Project │
                    │         │
                    └────┬────┘
                         │
                         │
                         │
                         │
                         │
     ┌──────┐      ┌─────▼─────┐
     │      │      │           │
     │ Repo ├──────► Workspace │
     │      │      │           │
     └──────┘      └─────┬─────┘
                         │
                         │
                         │
                         │
                         │
                   ┌─────▼──────┐     ┌──────┐
                   │            │     │      │
                   │ JobRequest ◄─────┤ User │
                   │            │     │      │
                   └────────────┘     └──────┘

We're trying to get a set of users who have created a job request for a given workspace, in the QuerySet of all workspaces for the given repo.  We can prefetch this relationship with `prefetch_related("job_requests__created_by")`, but when we come to iterate the workspaces we're still the wrong side of the JobRequest M-1 User relationship.

I've tried to get fancy with something like `Prefetch("job_requests__created_by", to_attr="job_creators")` but the lack of `job_creators` on the workspaces makes me think that isn't working.

Are we happy with a middle ground of speed up and coming back to this later?

Ref: #2894

### DIFF
--- a/staff/views/repos.py
+++ b/staff/views/repos.py
@@ -57,6 +57,22 @@ class RepoDetail(View):
 
         return "; ".join({build_contact(w.created_by) for w in workspaces})
 
+    def build_dates(self, api_repo, projects):
+        jobs = Job.objects.filter(
+            job_request__workspace__project__in=projects
+        ).annotate(first_run=Min(Least("started_at", "created_at")))
+
+        first_job_ran_at = ran_at(jobs.order_by("first_run").first())
+        last_job_ran_at = ran_at(jobs.order_by("-first_run").first())
+        twelve_month_limit = first_job_ran_at + timedelta(days=365)
+
+        return {
+            "first_job_ran_at": first_job_ran_at,
+            "last_job_ran_at": last_job_ran_at,
+            "repo_created_at": api_repo["created_at"],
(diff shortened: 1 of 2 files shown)


lucyb (approved):
Nice!

Thank you for explaining the final commit with the very small, yet no doubt very complicated, change. I think you've also made it easier to read the code by removing the extra query. I also like the refactoring bits; they make it easier to see what's happening in the get() method.

I think it was sensible to stop where you did and I hope this change is enough to allow the page to load in production. I don't think it should ever need to be particularly fast, so O(n) should be fine for a reasonable size of n.

PR #2906, opened 2023-03-30
Interactive PDF changes and form wording tweaks

tomodwyer:
- PDF interactive request:
  - Fix the size of the margins
  - Improve the file size by switching to JPEG for the canvas screenshot
- Remove the watermark from the "Generating your request" section
- Add read and agree wording to the submission page
- Add a modal with information regarding the download

### DIFF
--- a/assets/src/scripts/interactive/pages/review-request.jsx
+++ b/assets/src/scripts/interactive/pages/review-request.jsx
@@ -139,8 +139,8 @@ function ReviewRequest() {
         <h2 className="sr-only">Read and agree</h2>
         <ul>
           <li>
-            Your analyses will be run on patient records for approximately 44%
-            of the English population
+            Your analyses will be run on primary care patient records for
+            approximately 44% of the English population
           </li>
           <li>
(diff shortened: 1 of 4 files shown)


assets/src/scripts/interactive/pages/review-request.jsx
```diff
@@ -135,6 +135,40 @@ function ReviewRequest() {
 
       {error ? <InputError>{error}</InputError> : null}
 
+      <section className="prose prose-blue mt-8 pt-6 border-t max-w-full">
+        <h2 className="sr-only">Read and agree</h2>
+        <ul>
+          <li>
+            Your analyses will be run on patient records for approximately 44%
```
lucyb, line 142: Should we be making it clear here that it's "primary care patient records"? We haven't really explained what the dataset is anywhere, yet, so might be worth tackling in another ticket.

tomodwyer: Good point, maybe also on the Interactive About page.

PR #2912, opened 2023-03-31
Add application URL to staff area's ProjectEdit

ghickman:
This adds the application URL field to a Project's edit page in the staff area if it's not already linked to an application, including a note on what to do if a user is looking to link the project to an application:

There's several housekeeping changes here to make the form a little more useable with so many fields, handle duplicate numbers, etc.

Fix: #2888

### DIFF
--- a/templates/staff/project_edit.html
+++ b/templates/staff/project_edit.html
@@ -45,6 +45,14 @@
       <form class="mb-3" method="POST">
         {% csrf_token %}
 
+        {% if form.non_field_errors %}
+        <ul>
+          {% for error in form.non_field_errors %}
+          <li class="text-danger">{{ error }}</li>
+          {% endfor %}
+        </ul>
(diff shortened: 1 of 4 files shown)


templates/staff/project_edit.html
```diff
@@ -80,6 +80,23 @@ <h1 class="display-4">Edit: {{ project.title }}</h1>
...
+        <fieldset>
+          <legend>Application</legend>
+
+          {% if project.applications.exists %}
+          <p>This project is linked to an application, which cannot be edited here.</p>
+          {% else %}
+          <div class="alert alert-warning" role="alert">
+            If you are looking to link to an application created on this site
+            use the "Find and Link" button on the
+            <a href="{{ project.get_absolute_url }}">detail page</a>.
```
lucyb, line 92: This should be the staff_url not the main detail page.

PR #2914, opened 2023-04-03
Add a command to search for missing repos

ghickman:
The goal of this command is to help Liam know which rows in the audit spreadsheet need to be talked through with the tech team.  I've run it on a local copy of job-server to ensure it works, but it will need running against a recent copy of the production data (or just in production) once Liam has finished the audit.


lucyb (approved):
I don't have a way to test this, but I assume it works.

One minor comment to address, but otherwise is fine.

PR #2940, opened 2023-04-05
fix: send email for token generation and log in

bloodearnest:


jobserver/emails.py
```diff
@@ -112,3 +112,23 @@ def send_welcome_email(user):
         template_name="emails/welcome.txt",
         context=context,
     )
+
+
+def send_token_login_generated_email(user):
+    send(
+        to=user.notifications_email or user.email,
```
ghickman, line 119: You can rely on `notifications_email` here.

jobserver/emails.py
```diff
@@ -112,3 +112,23 @@ def send_welcome_email(user):
...
+        subject="New OpenSAFELY login token generated",
+        sender="notifications@jobs.opensafely.org",
+        reply_to=["OpenSAFELY Team <team@opensafely.org>"],
+        template_name="emails/login_token_generated.txt",
+    )
+
+
+def send_token_login_used_email(user):
+    send(
+        to=user.notifications_email or user.email,
```
ghickman, line 129: Same here.

ghickman (approved):

PR #2948, opened 2023-04-06
chore: add more ip restriction tests

madwort:


tests/unit/jobserver/test_middleware.py
```diff
@@ -85,6 +90,16 @@ def test_client_ip_middleware_proxied(rf):
     middleware(request)
     assert request.backend is None
 
+    # test spoofing header doesn't work
+    # attacker attempts to spoof both the TPP address & the proxy address
+    request = rf.get(
```
bloodearnest, line 95: Good call, this was designed to defend against this, but didn't include the test.

For just a single proxy, the last address wouldn't be there normally

tests/unit/jobserver/test_middleware.py
```diff
@@ -77,6 +77,11 @@ def test_client_ip_middleware_proxied(rf):
     middleware(request)
     assert request.backend == backend
 
+    # tpp client via proxy
```
bloodearnest, line 80: This wouldn't normally happen without spoofing - the immeadiate proxy address wouldn't be in X-Forwarded-For.

So I think this case is covered in the other test.

bloodearnest (approved):

tests/unit/jobserver/test_middleware.py
madwort: removed the last address

tests/unit/jobserver/test_middleware.py
madwort: yep understand now

PR #2949, opened 2023-04-06
Report tweaks

ghickman:
A handful of tweaks around reports.

### DIFF
--- a/jobserver/models/reports.py
+++ b/jobserver/models/reports.py
@@ -74,6 +74,15 @@ class Report(models.Model):
     def __str__(self):
         return self.title
 
+    def get_absolute_url(self):
+        if hasattr(self, "analysis_request"):
+            return self.analysis_request.get_absolute_url()
+        else:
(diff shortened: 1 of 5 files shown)


jobserver/models/reports.py
```diff
@@ -74,9 +74,22 @@ class Meta:
...
+            # we only have reports tied to analyses currently but we expect
+            # that to change in the future so rather than error out send the
+            # user _somewhere_.
+            return "/"
+
     def get_staff_url(self):
         return reverse("staff:report-detail", kwargs={"pk": self.pk})
 
+    @property
+    def is_draft(self):
```
lucyb, line 90: Would `is_released` be more consistent terminology or would it just lead to confusion?

ghickman: Confusion I think.  I see where you're coming from but releasing is specific to outputs and is the act of uploading filies from an L4 to job-server, where their contents can be published (and can thus be in a draft state!).

Issue #2786, opened 2023-03-10
Set up HTML emails


ghickman:
OSI shipped HTML emails with a conversion function to generate text emails.  We'd like to provide support for HTML emails in job-server as well.  Incuna mail handles HTML emails simply by passing it a template it can render with the same context. 

While adding HTML emails it would make sense to also provide a base template to all the text email templates so we can add a footer and/or header to them all.

I'd like us to get to a state where:
* every email has both an HTML and text template
* HTML emails use a base template to wrap content
* Text emails use a base template to wrap content

PR #2952, opened 2023-04-07
Add HTML emails

ghickman:
This both adds HTML versions of all emails and a base template for the text emails to help them match the HTML emails.

A few things to check during review:
* Do the base templates' footers make sense/could they contain more information
* Have I typo'd anything in the HTML emails?  I coverted the text emails into HTML but have modified various link sections to use hyperlinks to better suit the medium.

I've tested this locally with my datalab email, do we think that's enough?

Fix: #2786


tomodwyer (approved):
Happy with the way the HTML code looks. I think it's best we roll this out and then take feedback based on real user clients/browsers. If they find any issues, we can amend the templates.

PR #2959, opened 2023-04-10
Condense auth pipeline

ghickman:
This takes our auth pipeline, predominently provided by PSA's social_core, and replaces it with a single function which only does the bits we used.

## Methodology
I was aware from previous deep dives of the pipeline that it works by looping the functions defined in settings.py and combining any dict outputs into a general purpose dict, from which any future method can pull values.  So for example returning `{"foo": "bar"}` would allow a subsequent function to be defined like so `def frobnicate(foo):`.

Knowing that I walked through each function down to `create_user` working out what exactly it extracted from the args/kwargs and returned, building up an understanding of the various parts it created to get to a new user.  After `create_user` I realised the functions were almost no-ops (`load_extra_data` did actually construct UserSocialAuth.extra_data but there's so little extra data from the response for us that it's practically negligible).

Once I had a good idea of what each part was doing, and where it was doing it, I recreated various parts in our new `pipeline` function and manually tested.

To speed up manual testing I added a login button directly on the unauthenticated out home page, and a logout button on the authenticated homepage just for a quicker feedback loop.

## Local testing
There are two parts to test with this, both will need you to check the data on your relevant User and UserSocialAuth records.  I found a useful way to do this, before testing run:
```
python manage.py dumpdata social_django.UserSocialAuth --indent=2 --pks=<your social id> > user-social-auth.json
python manage.py dumpdata jobserver.User --indent=2 --pks=<your user id> > user.json
```

Printing these out with something that pretty prints them like `bat` to get a nice, quick reference to the bits you need to care about.

Then you will need to test both:
* existing user login
* registration

For both of these you should be able to logout, and then log in and have the same data as before you used this branch. 
The fields to check post-login are:

```
User
* username
* email
* fullname
* notifications_email

UserSocialAuth
* user -> should be the PK of your user
* provider -> should be "github"
* uid -> should be consistent with the data you captured before the change
* extra_data.id -> should match uid
* login -> should be your username
* access_token -> should exist
* token_type -> should be "bearer"
```

To remove your local user you can run `User.objects.get(username="<your username>").delete()` and the relevant UserSocialAuth object will also be removed.

## Why?
I've explained some of this in the main commit, but it mainly comes down to having to walk through many layers of subclass for the backend (`GitHubOAuth2`), strategy (`DjangoStrategy`), and storage (`DjangoStorage`) each time I need to make changes.  Each time I've done so I've had to dig into the library just to work out which parts of the code we actually use.  Having done this a few times and knowing that we've added code to stop PSA from overwriting fields on our User models on login it felt like a good time to eject from the pipeline.

Along the way I found various code paths that were completely unused by us, eg `associate_user` never fired for either usage type.  One subtle bug in our configuration where our switch of User.USERNAME to `email` meant PSA's DjangoUserMixin was [correctly!] using that field when calling User.objects.create_user.  This has little bearing on our auth story for OAuth users since PSA logins use the UID from GitHub to look up the user so worst case scenario we possibly have some OAuth users with an email address for their username.

My goal with this change in general is to make how OAuth users are looked up or created explicit to the casual reader, and easy to change in the future.

Ref: #1962


jobserver/auth_pipeline.py
```diff
@@ -1,38 +1,69 @@
...
+            "login": response["login"],
+            "access_token": response["access_token"],
+            "token_type": response["token_type"],
+        }
+
+        social = UserSocialAuth.objects.create(
+            user=user, uid=uid, provider="github", extra_data=extra_data
+        )
+
+    notify_new_user(user)
```
lucyb, line 63: Should this be within the transaction? If there's a problem and the transaction rolls back the `notify_new_user` function will not handle that gracefully.

jobserver/settings.py
```diff
@@ -341,19 +341,8 @@
...
-    "social_core.pipeline.social_auth.social_user",
-    "social_core.pipeline.user.get_username",
-    "social_core.pipeline.user.create_user",
-    "social_core.pipeline.social_auth.associate_user",
-    "social_core.pipeline.social_auth.load_extra_data",
-    "social_core.pipeline.user.user_details",
-    "jobserver.auth_pipeline.set_fullname",
-    "jobserver.auth_pipeline.set_notifications_email",
-    "jobserver.auth_pipeline.notify_on_new_user",
+    "jobserver.auth_pipeline.pipeline",
```
lucyb, line 344: Wow, that's simplified things.

tests/unit/jobserver/test_auth_pipeline.py
```diff
@@ -1,22 +1,45 @@
...
-
-
-def test_notify_on_new_user_with_new_user(slack_messages):
-    user = UserFactory()
+from jobserver.auth_pipeline import pipeline
+
+from ...factories import UserFactory, UserSocialAuthFactory
+
+
+def test_pipeline():
```
lucyb, line 6: Might be more expressive to call this something like:

`test_pipeline_with_existing_user()`

tests/unit/jobserver/test_auth_pipeline.py
```diff
@@ -1,22 +1,45 @@
...
-    assert len(slack_messages) == 0
-
-
-def test_notify_on_new_user_with_new_user(slack_messages):
-    user = UserFactory()
+from jobserver.auth_pipeline import pipeline
+
+from ...factories import UserFactory, UserSocialAuthFactory
+
+
```
lucyb, line 5: Are there any failure modes that we should check for in the tests? For example, what happens if the atomic transaction fails?

lucyb:
I'm trying to test using the instructions above, however I am getting an error when running

```
python manage.py dumpdata jobserver.User --indent=2 --pks=<your user id> > user.json
```

This is the output I receive:

```
[
{
  "model": "jobserver.user",
  "pk": 187,
  "fields": {
    "password":  [redacted],
    "last_login": "2023-04-11T11:42:11.619Z",
    "username": "lucyb",
    "email":  [redacted],
    "fullname": "Lucy B",
    "is_active": true,
    "is_staff": false,
    "is_superuser": false,
    "date_joined": "2022-03-08T11:53:13.463Z",
    "notifications_email":  [redacted],
    "is_approved": false,
    "pat_token": null,
    "pat_expires_at": null,
    "login_token": [redacted],
    "login_token_expires_at": "2023-04-11T12:39:10.728Z",
    "created_by": null,
    "roles": [
      CommandError: Unable to serialize database: Object of type type is not JSON serializable
```

tests/unit/jobserver/test_auth_pipeline.py
ghickman: It also tests with a new user too! This was originally two tests, one with that name, and I realised the existing user test was duplicating the entirety of the first test.

jobserver/auth_pipeline.py
ghickman: I don't think so.  An transaction.atomic() block doesn't catch the exception, it makes sure we execute a `ROLLBACK` before letting the exception propogate up so we shouldn't hit this line in that case.

tests/unit/jobserver/test_auth_pipeline.py
lucyb: Are you thinking of `test_pipeline_new_user`, which seems to do what you're describing? I'm not sure I see the advantage of having both tests check the behaviour for an existing user. My preference would be to have it in a separate test rather than have one test check both things.

lucyb (approved):
This looks like a big improvement and less "magic" than before. :tada: 

The comments on the tests are small nitpicks, so feel free to ignore or tidy, then merge.

tests/unit/jobserver/test_auth_pipeline.py
ghickman: I think I meant to combine them into one test and must have gotten distracted half way 😂  I'll split them up 👍

Issue #2961, opened 2023-04-10
Consolide User email fields


ghickman:
Assuming #2595 is merged we will no longer need two email fields because PSA won't be overwriting our email field on each login.

PR #2962, opened 2023-04-10
Consolidate email fields

ghickman:
Fix: #2961


bloodearnest (approved):
LGTM. 

Might be worth taking a dump of the db just before you deploy, so we can recreate?

Or even splitting the RemoveField migration out into a separate PR for safety?

Issue #2958, opened 2023-04-10
Split up homepage templates


ghickman:
We currently have index.html for the homepage which has a couple of big conditionals for users being logged in, with the unauthenticated path loading a different template altogether.

To simplify this lets create two templates and load them in the view where we're already forking on authentication.

PR #2964, opened 2023-04-10
Switch to two templates for the homepage

ghickman:
This splits the homepage into two templates, based on a viewers current authentication state, making it easier to jump to those templates when coming from the view.

This also pulls the job request list structure fully into the partial for rendering it since that's the only common part of the two pages.

Fix: #2958


templates/index.html
lucyb, line 1: I'm not sure what this file is for now?

Would it make more sense to use this instead of the separate index-unauthenticated.html template?

lucyb (approved):
This is a very sensible change. Thank you!

I've added one question, but happy for you to merge if the answer is straightforward.

PR #2971, opened 2023-04-11
Remove fullname middleware

ghickman:
This updates our pipeline to take advantage of Python Social Auth's [partial pipeline feature](https://python-social-auth.readthedocs.io/en/latest/pipeline.html#partial-pipeline), kicking out registering users who don't have a name set in GitHub to a form page where we can capture it, before continuing in the pipeline.

This lets us drop the RequireNameMiddleware which is firing on every request, not a great use of resources.  We currently have 28/216 users with a `fullname` set, but I suspect that isn't going to change if the users haven't logged in in the last 9 months that this middleware has been active.  We'll have to chase those users manually instead. 

Fix: #1962

### DIFF
--- a/jobserver/middleware.py
+++ b/jobserver/middleware.py
@@ -1,60 +1,6 @@
 from django.conf import settings
-from django.shortcuts import redirect
-from django.urls import reverse
 
 from jobserver.models.backends import Backend
-
-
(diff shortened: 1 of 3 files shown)


lucyb (approved):
I'm learning about python social auth! Thank you :)

PR #2990, opened 2023-04-17
Print slack messages to the console in debug mode

ghickman:
A follow up to opensafely-core/interactive.opensafely.org#410 using the suggested implementation.


bloodearnest (approved):
noice

PR #2991, opened 2023-04-17
Implement nginx ratelimiting

madwort:
* rate limits need to be configured outside of the `server` block, which means we cannot use the existing `~dokku/job-server/nginx.conf.d` mechanism & we need to copy dokku's `nginx.conf.sigil` into this repository
* use tighter limits for the login page. This requires copy-pasting the proxy settings in an unsatisfactory way, because they apparently do not inherit from the outer location (as tested in production on `dokku4`)
* fixes https://github.com/ebmdatalab/sysadmin/issues/278
* ratelimiting was testing in production using siege during development

### DIFF
--- a/nginx.conf.sigil
+++ b/nginx.conf.sigil
@@ -1,0 +1,188 @@
+{{ range $port_map := .PROXY_PORT_MAP | split " " }}
+{{ $port_map_list := $port_map | split ":" }}
+{{ $scheme := index $port_map_list 0 }}
+{{ $listen_port := index $port_map_list 1 }}
+{{ $upstream_port := index $port_map_list 2 }}
+
+{{ if eq $scheme "http" }}
+server {
+  listen      [{{ $.NGINX_BIND_ADDRESS_IP6 }}]:{{ $listen_port }};
+  listen      {{ if $.NGINX_BIND_ADDRESS_IP4 }}{{ $.NGINX_BIND_ADDRESS_IP4 }}:{{end}}{{ $listen_port }};
+  {{ if $.NOSSL_SERVER_NAME }}server_name {{ $.NOSSL_SERVER_NAME }}; {{ end }}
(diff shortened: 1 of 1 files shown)


nginx.conf.sigil
```diff
@@ -0,0 +1,227 @@
+{{ range $port_map := .PROXY_PORT_MAP | split " " }}
+{{ $port_map_list := $port_map | split ":" }}
+{{ $scheme := index $port_map_list 0 }}
+{{ $listen_port := index $port_map_list 1 }}
+{{ $upstream_port := index $port_map_list 2 }}
+
+limit_req_zone $binary_remote_addr zone=one:10m rate=1r/s;
```
bloodearnest, line 7: Given these are global to nginx, can I suggest more descriptive names?

e.g.

```
limit_req_zone $binary_remote_addr zone=jobserver_login:10m rate=1r/s;
limit_req_zone $binary_remote_addr zone=jobserver_default:10m rate=20r/s;
```

bloodearnest (commented):
Arg, having to dup the sigil config is really not nice.

Could we define the limit zones in the global nginx config? i.e. use ansible to drop a file into /etc/nginx.conf/conf.d?

That's also not ideal, in that if those zones are not defined, the dokku app's nginx config is probably invalid. But I wonder, given we've got ansible for ensuring the zones are set, then maybe that's cleaner than forking/duplicating the nginx config template just for job-server?

nginx.conf.sigil
bloodearnest: Hmm, AIUI, we can't really reuse limit zones, or else you'd be linking the limits across multiple locations?

so if another url use `one` like /login does, I think that those limits would be shared?

madwort: ah, yeah, you're right, the two urls would share the limit, let's not do that.

madwort:
abandoned in favour of https://github.com/ebmdatalab/sysadmin/pull/283

PR #2992, opened 2023-04-17
Add an "at any time" option to the OSI form

tomodwyer:
So that users do not need to pick a specific window of time for the events to have occurred.

This PR also makes a few tweaks to the SVG image to fix styling issues.


assets/src/scripts/interactive/pages/review-request.jsx
```diff
@@ -34,28 +33,48 @@ function ReviewRequest() {
...
       codelistA: {
         label: codelistA.label,
         type: codelistA.type,
         value: codelistA.value,
       },
       codelistB: {
-        label: codelistB.label,
-        type: codelistB.type,
-        value: codelistB.value,
+        label: codelistA.label,
```
lucyb, line 54: Should these three be codelistB rather than codelistA?

lucyb (approved):
Just a couple of very minor points, otherwise looks good, thank you!

assets/src/scripts/interactive/pages/review-request.jsx
tomodwyer: Great spot! Copy and paste fail there, I'll fix now.

Issue #2976, opened 2023-04-12
Submitting an analysis request should notify a slack channel


lucyb:
This is so Catherine and others can easily keep track of how Interactive is used.

- [x] Notification when a new request is made
- [x] Notification when files are released

PR #3001, opened 2023-04-17
Add OSI slack notifications

ghickman:
Fix: #2976

### DIFF
--- a/interactive/commands.py
+++ b/interactive/commands.py
@@ -1,10 +1,11 @@
 import textwrap
 
 from django.conf import settings
+from django.db import transaction
 
 from jobserver.authorization import InteractiveReporter
 from jobserver.github import _get_github_api
-from jobserver.models import OrgMembership, ProjectMembership, Repo, User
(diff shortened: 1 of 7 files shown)


lucyb (approved):

ghickman:
@lucyb – do you want to merge this one while I'm away? it's probably a useful one to have sooner rather than later

interactive/commands.py
```diff
@@ -41,6 +42,23 @@ def create_repo(*, name, get_github_api=_get_github_api):
     return repo["html_url"]
 
 
+@transaction.atomic()
```
bloodearnest, line 45: There's a change in semantics in this refactor.  The was a reason we originally put the slack/email notifications inside a transaction - if we failed to notify slack, the db would rollback.

As it now stands, the db would commit, the slack/email notification could fail, and we would have no indication of a problem, and the user would be left hanging.

I think the transaction belongs as with statement around the create report and sending.

Issue #3031, opened 2023-04-20
interactive_templates package is not versioned


bloodearnest:
Currently interactive_templates is not installed with an explicit version.

This is because we install from a zip archive, in order to support pips require-hashes arguments.

However, this zip file has not version metadata, so pip-compile cannot bump the explicit package version when it installs it.

Which results in interactive_templates package not being updated when running `just devenv`, even thought the url and hash have changed in the requirements.prod.txt file.

This is foot gun that it would be good to fix. Assuming we want to keep require hashes, we can maybe add a github action that runs after merge to main that does either:

a) auto-updates pyproject.yaml with a version.
b) adds a version, and creates a GH release that is used as installable target, rather than the current repo zip

I think b is a more elegant solution.

CLStables:
Tech debt issue - there is a hack workaround. We could fix if we have time, but if necessary move to a future OSI initiative

lucyb:
I updated the justfile a few weeks ago so that the interactive_templates package is updated when running `just devenv`. This ticket would create a much better, permanent solution but I think what we have in place is good enough for OSI:V2, so I'm removing it from this initiative.

PR #3043, opened 2023-04-21
Forceably reinstall interactive templates package

lucyb:
Locally, changes to the interactive_templates package are not picked up automatically because we are usng a zipfile rather than providing an explicit version. This means we need to uninstall and reinstall the package to pick up any changes.

This is not a problem in CI because it is building from a fresh environment each time.

Note: this change should be removed when the package has proper versions that pip can handle appropriately. This is not intended to be a long term fix.

### DIFF
--- a/justfile
+++ b/justfile
@@ -57,8 +57,15 @@ requirements-dev *args: requirements-prod
 prodenv: requirements-prod
     #!/usr/bin/env bash
     set -eux
+
     # exit if .txt file has not changed since we installed them (-nt == "newer than', but we negate with || to avoid error exit code)
     test requirements.prod.txt -nt $VIRTUAL_ENV/.prod || exit 0
+
+    # Locally, changes to the interactive_templates package are not picked up automatically, because we are usng a zipfile rather
+    # than providing explicit versions. This means we need to uninstall and reinstall the package to pick up any changes.
+    # Note: this is not a problem in CI because it is building from a fresh environment each time.
+    $PIP uninstall -y interactive_templates
+    touch requirements.prod.in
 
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -76,6 +76,12 @@ devenv: prodenv requirements-dev && install-precommit
     # exit if .txt file has not changed since we installed them (-nt == "newer than', but we negate with || to avoid error exit code)
     test requirements.dev.txt -nt $VIRTUAL_ENV/.dev || exit 0
 
+    # Locally, changes to the interactive_templates package are not picked up automatically, because we are usng a zipfile rather
+    # than providing explicit versions. This means we need to uninstall and reinstall the package to pick up any changes.
+    # Note: this is not a problem in CI because it is building from a fresh environment each time.
+    pip uninstall -y interactive_templates
```
bloodearnest, line 82:
```suggestion
    $PIP uninstall -y interactive_templates
```

bloodearnest (approved):
This works around the problem #3031.

Left a fix for using the correct pip invocation (or else it will break if you run outside the venv, like in CI)


Arguably, properly versioning the package would be a better solution, but this is good enough workaround for now.

lucyb:
> Arguably, properly versioning the package would be a better solution, but this is good enough workaround for now.

I agree. This is a hack, but since it tripped me up again this afternoon, I figured it was better to do something imperfect sooner.

lucyb:
@bloodearnest I realised that I'd put this in the `devenv` section but it needs to be in the `prodenv` section instead, otherwise the package isn't going to get reinstalled.

PR #3044, opened 2023-04-21
Add ability to edit Report title and description.

bloodearnest:
* Stores the values in the already existing `Report.title` and
  `Report.description` fields
* Fix: set `Report.title` to be `AnalysisRequest.report_title` that was
  captured via the create form.
* Displays the user supplied title and description at the top of the
  report, and provides new button to edit

### DIFF
--- a/interactive/urls.py
+++ b/interactive/urls.py
@@ -5,7 +5,7 @@ from jobserver.views.reports import (
     ReportPublishRequestUpdate,
 )
 
-from .views import AnalysisRequestCreate, AnalysisRequestDetail
+from .views import AnalysisReportUpdate, AnalysisRequestCreate, AnalysisRequestDetail
 
 
 app_name = "interactive"
@@ -14,6 +14,7 @@ urlpatterns = [
     path("", AnalysisRequestCreate.as_view(), name="analysis-create"),
     path("<slug:slug>/", AnalysisRequestDetail.as_view(), name="analysis-detail"),
     path("<str:slug>/", AnalysisRequestDetail.as_view(), name="analysis-detail"),
+    path("<str:slug>/update", AnalysisReportUpdate.as_view(), name="report-update"),
     path(
         "<str:slug>/publish/",
         ReportPublishRequestCreate.as_view(),
--- a/interactive/views.py
+++ b/interactive/views.py
@@ -1,13 +1,16 @@
 import json
 
 from django.core.exceptions import PermissionDenied
+from django.forms import modelform_factory
+from django.http import Http404
 from django.shortcuts import get_object_or_404, redirect
 from django.template.response import TemplateResponse
+from django.urls import reverse
 from django.views.generic import DetailView, View
 from interactive_templates.schema import Codelist, v2
 
 from jobserver.authorization import has_permission
-from jobserver.models import Backend, Project
+from jobserver.models import Backend, Project, Report
 from jobserver.reports import process_html
 from jobserver.utils import build_spa_base_url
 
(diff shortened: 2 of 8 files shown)


interactive/views.py
```diff
@@ -191,7 +194,66 @@ def get_object(self, queryset=None):
...
+            )
+        else:
+            return self.render_to_response(request, form)
+
+    def render_to_response(self, request, form):
+        return TemplateResponse(
+            request,
+            template="interactive/report_update.html",
+            context={"form": form, "analysis_request": self.analysis_request},
+        )
```
ghickman, line 259: Rather than building the view around an AnalysisRequest we can focus it on the Report and lean on UpdateView:

```
class ReportEdit(UpdateView):
    fields = [
        "title",
        "description",
    ]
    model = Report
    template_name = "interactive/report_edit.html"

    def get_object(self, **kwargs):
        # keep hold of the AnalysisRequest object so we can easily use it for
        # the success_url
        self.analysis_request = get_object_or_404(
            AnalysisRequest, slug=self.kwargs["slug"]
        )

        if not has_permission(
            self.request.user,
            "analysis_request_view",
            project=self.analysis_request.project,
        ):
            raise PermissionDenied

        if not self.analysis_request.report:
            raise Http404

        return self.analysis_request.report

    def get_success_url(self):
        return self.analysis_request.get_absolute_url()
```

I'm returning the report, after the relevant checks, so the form is handled by the view for us.  I'm also keeping a reference on the AnalysisRequest so we don't have to put checks around `report.analysis_request` in get_success_url.

Could we also rename it and the template to use the edit suffix?  It's what we've done for all the other views (except the review request which I must have missed in the spike!).

interactive/views.py
```diff
@@ -191,7 +194,66 @@ def get_object(self, queryset=None):
...
     def get_context_data(self, **kwargs):
         report = process_html(self.object.report_content)
+        update_url = reverse(
+            "interactive:report-update",
+            kwargs=dict(
+                slug=self.object.slug,
+                project_slug=self.object.project.slug,
+                org_slug=self.object.project.org.slug,
+            ),
+        )
```
ghickman, line 204: I think for OSI we can put this on the Report model as `get_edit_url` if we have a note that it might need changing for more generic reports?

interactive/urls.py
```diff
@@ -14,6 +14,7 @@
     path("", AnalysisRequestCreate.as_view(), name="analysis-create"),
     path("<slug:slug>/", AnalysisRequestDetail.as_view(), name="analysis-detail"),
     path("<str:slug>/", AnalysisRequestDetail.as_view(), name="analysis-detail"),
+    path("<str:slug>/update", AnalysisReportUpdate.as_view(), name="report-update"),
```
ghickman, line 17: Same point about renaming to `edit` here please.  Also, the URL needs a trailing slash.

ghickman (approved):

interactive/views.py
bloodearnest: I tried to make this work originally, but it seemed a bit messy. But your way looks cleaner than what I got to.

I also had a concern here about normal Reports vs OSI Reports. Are we going to allow editing title/description on normal reports?

We can maybe punt on that decision. I guess its fine to have multiple urls that update Report objects.

interactive/urls.py
bloodearnest: Why is the slash needed? That seems semantically odd?

interactive/views.py
bloodearnest: So, this was the crux of my uncertainty - is editing an OSI report the same thing as a general Report?

If it's just title/description, then I think yes? As that's exactly the kind of metadata we have on the current reports site.

But if its more, we might to specialise urls/views for OSI report editing?

Should we assume the same for now? In which case, should we have this view in jobserver/views/reports.py, and the url based on the report?

interactive/urls.py
ghickman: It's consistent with how we do the rest of our page URLs :)

bloodearnest:
Me neither!

interactive/views.py
ghickman: I think punt on the OSI vs normal report bits until we have some time to think about how reports will exist in job-server 👍

interactive/views.py
ghickman: I think that's definitely a reasonable concern to have!

I suspect we might revert this once reports in general are all in job-server, but for now I was just thinking this would be a convenience wrapper like our other `get_FOO_url` ones.  Maybe that's a little confusing though 🤔  What about putting a comment above this to point out this exact discussion, just something short like:

> TODO: decide if this should be the canonical way to edit reports once reports arrive in this project

?

PR #3046, opened 2023-04-21
Week of latest extract

lucyb:
This calculates the Monday of the week with the most recent complete extract of
data from TPP. This is passed into the Analysis and used to calculate the
number of people with matching events in that week. This number is then shown
on the report (with low number redaction).

Fixes https://github.com/opensafely-core/interactive-templates/issues/17


bloodearnest (approved):
lgtm

PR #3061, opened 2023-04-25
Make next action explicit in issue text

evansd:
This was [explained][1] in the team manual, but it would have helped resolve the uncertainty on my part if it was explicit in the issue text itself.

Incidentally, the ability to do an org-wide Github search for the text I wanted to change without having any idea what system generated it felt like a super-power.

[1]: https://bennettinstitute-team-manual.pages.dev/tech-group/tech-support/#making-research-repos-public


ghickman (approved):
Thanks Dave!

Issue #3025, opened 2023-04-19
Bug? OSI analysis requests marked as "succeeded" but report isn't yet available


CLStables:

bloodearnest:
Hmm, yeah, the job-request is succeeded.

I think we actually need to use the release of the html file as the status for the user

PR #3078, opened 2023-04-27
Add AnalysisRequest.status

ghickman:
This adds `AnalysisRequest.status`, a wrapper around `JobRequest.status`, to add in intermediary statuses relevant to the Analysis.

As part of this I've made a sweeping change to how `JobRequest.status` works since we only ever pushed a user to use `.prefetch_related()`, but given we have `prefetch_related_objects` at our disposal I think this is a better path in terms of what someone might expect when calling the property.

Fix: #3025


bloodearnest (approved):
Love getting rid of requiring prefetch, nice!

PR #3080, opened 2023-04-28
Add channel to the log context of failed slack messages

ghickman:
When we fail to send a slack message Sentry picks up on the exception log we emit, but it doesn't currently have any useful context around what triggered the message or any config.  We expect knowing which channel it was trying to post to will solve that.

### DIFF
--- a/services/slack.py
+++ b/services/slack.py
@@ -28,7 +28,9 @@ def post(text, channel):
         client.chat_postMessage(channel=channel, text=text)
     except SlackApiError:
         # failing to slack is never fatal, so log and do not error
-        logger.exception("Failed to notify slack")
+        logger.exception(
+            f"Failed to notify slack in channel: {channel}", channel=channel
+        )
 
 
 def link(url, text=None):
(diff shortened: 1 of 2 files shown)


services/slack.py
```diff
@@ -28,7 +28,7 @@ def post(text, channel):
         client.chat_postMessage(channel=channel, text=text)
     except SlackApiError:
         # failing to slack is never fatal, so log and do not error
-        logger.exception("Failed to notify slack")
+        logger.exception("Failed to notify slack", channel=channel)
```
bloodearnest, line 31: will this show up in the log output?

bloodearnest: I guess it will because of our log config, but I'd be inclined to include it in the exception message text?

ghickman: `logger` is a structlog logger here so it'll appear in our logs as `channel=some_string`. I'm not sure what we gain by duplicating it in the log message?

bloodearnest: It's disjointed - the channel will be at the top of the traceback, and the message at the bottom.

Also, in sentry, the log lines are separate from the exception. I think both is good: in the exception message for ease of use/obviousness, and in the log message for grepping/filtering

bloodearnest (approved):

PR #3081, opened 2023-04-28
Stop touching prod requiremnts on every invocation

ghickman:
This removed caching from our stack.  This was triggering our _compile target inside requirements-prod on every invocation.  However without it we will still do the right thing with interactive_templates (and other deps) because any changes to requirements.prod.in will trigger a recompile, updating requirements.prod.txt, meaning we run the uninstall line above this change and then run install on the pinned deps.


lucyb (approved):
Sorry about that. Thanks for doing the fix!

Issue #3014, opened 2023-04-18
Co-pilots can approve making reports (and underlying outputs) public


CLStables:
To approve, co-pilots hit “approve” button, which automatically makes the final report and the underlying output file public 

- [ ] Permissions for any staff user to do this - probably from the analysis request detail page (?) - staff users get extra button

PR #3101, opened 2023-05-02
Approve publish requests

ghickman:
This moves the approval of requests to publish reports into the staff area, removing the user-facing view entirely.  It attempts to make it easy for a staff user to get from a report to a request to publish quickly in the staff area by listing publish requests on various pages.

The publish request pages in the staff area have the usual base tools, some filter, some searching, etc.

Fix: #3014

### DIFF
--- a/staff/views/reports.py
+++ b/staff/views/reports.py
@@ -15,6 +15,22 @@ class ReportDetail(DetailView):
     model = Report
     template_name = "staff/report_detail.html"
 
+    def get_queryset(self):
+        return (
+            super()
+            .get_queryset()
+            .select_related(
+                "analysis_request",
+                "analysis_request__project",
+                "analysis_request__project__org",
+                "created_by",
+                "publish_request",
+                "publish_request__created_by",
+                "release_file__workspace__project",
+                "release_file__workspace__project__org",
+            )
+        )
+
 
 @method_decorator(require_role(CoreDeveloper), name="dispatch")
 class ReportList(ListView):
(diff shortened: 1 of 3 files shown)


templates/staff/report_publish_request_detail.html
```diff
@@ -0,0 +1,97 @@
+{% extends "staff/base.html" %}
```
lucyb, line 1: I'm not sure that this needs to be a separate page. Could it be a card on the Report staff detail page instead? I appreciate there will be more than a single button needed at some point, but I think Tom O could help figure out how to make that work. Similarly, on the list view, the publish request could be a filter option, so that someone could view all outstanding (or cancelled) requests.

I also don't think that the Analysis Request staff detail page should have the Publish Request on it either. 

It might be helpful to think of this in terms of an aggregate root, in DDD speak (https://martinfowler.com/bliki/DDD_Aggregate.html).

ghickman: #3108 deals with several of these points!

It adds a second button (reject), wrapping both of the buttons up so we can have a disabled state with hover to explain.

It switches the two publish request models to have FKs to their relevant target model so it's possible (and I think likely with reports) that we'll have N publish requests for each report.

It adds more states to the reports list page's filter-by-published-state now.

I'd be happy to look at consolidating these pages once all the core publishing features are in and we've got some user feedback but lets keep them around while we're developing it.  They're really useful during dev atm!

ghickman:
Combined into #3108

Issue #3059, opened 2023-04-25
Add the email address of the person who downloaded the report into the watermark (“downloaded by X”) to discourage inappropriate sharing


CLStables:
There is an outstanding question about whether this will work ok for co-pilots downloading reports to send to NHSE for review. If emails cause issues (non-work emails are linked to GitHub accounts) we could use name instead?

tomodwyer:
Closed by https://github.com/opensafely-core/job-server/pull/3103

Issue #2802, opened 2023-03-14
Add styling to Interactive Reports


lucyb:
Look at whether we can embed CSS into the generated report, so that the styling is linked to a particular version of the of the report.

See this for a starter: https://bennettoxford.slack.com/archives/C0441RRL3RC/p1678791262404799

Assume that the report will be available on job server and available to download via print-to-PDF. Aim for best practice HTML and CSS, but don't worry about having the report viewable on other platforms.

tomodwyer:
Closed by https://github.com/opensafely-core/job-server/pull/3103

PR #3103, opened 2023-05-02
Update Interactive Report styles and fix PDF styling

tomodwyer:
This PR covers the following issues:

- https://github.com/opensafely-core/job-server/issues/2802
- https://github.com/opensafely-core/job-server/issues/2670
- https://github.com/opensafely-core/job-server/issues/3059


tomodwyer:
> * when clicking download the page behind the modal resizes to much smaller then back to normal, I grabbed a video to show it in action: https://share.getcloudapp.com/yAuJw2Xn

This is unfortunately a limitation of doing the styles client side to capture the PDF. I could blur it more, but I felt it looked a bit odd if the modal totally blocked the background.

> * my email address in the watermark is truncated, can we resize that text to fit the width or wrap it maybe?

That's a good point, I'll fix that before merging.

PR #3105, opened 2023-05-02
Email the creator of a publish request when it's approved

ghickman:
Do we need to include anything else in this email?

Fix: #3016


ghickman:
Combined into #3108

PR #3106, opened 2023-05-03
Detect OSI Reports created via management command

lucyb:
This is still a work in progress, but happy to get feedback at this point.

A few questions still outstanding
* I couldn't see a way to get reports from osi_run into osi_release without copying them into a local directory. I don't know if there's a better way? If not, then the directory probably needs renaming to something more sensible.
* Is the sample report, `outputs/interactive_report.html`, still needed if we're keeping reports from osi_run around? I think I need to understand more about people's current development process..

### DIFF
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -75,12 +75,14 @@ def is_interactive_report(rfile):
     """
 
     path = Path(rfile.name)
-    if len(path.parts) != 3:
+    if (
+        len(path.parts) == 3
+        and path.parts[0] == "output"
+        and path.parts[2] == "report.html"
+    ):
+        identifier = path.parts[1]
+    else:
         return
-    if not (path.parts[0] == "output" or path.parts[2] == "report.html"):
-        return
-
-    identifier = path.parts[1]
 
     return AnalysisRequest.objects.filter(pk=identifier).first()
 
--- a/jobserver/management/commands/osi_release.py
+++ b/jobserver/management/commands/osi_release.py
@@ -1,6 +1,7 @@
 import hashlib
 import sys
 from datetime import UTC, datetime
+from pathlib import Path
 
 import requests
 from django.conf import settings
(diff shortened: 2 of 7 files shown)


jobserver/api/releases.py
```diff
@@ -74,13 +74,15 @@ def is_interactive_report(rfile):
     """
 
     path = Path(rfile.name)
-    if len(path.parts) != 3:
-        return
-    if not (path.parts[0] == "output" or path.parts[2] == "report.html"):
+    if (
+        len(path.parts) == 3
+        and path.parts[0] == "output"
```
lucyb, line 79: Do we care if the directory is called output?

ghickman:
> Is the sample report, outputs/interactive_report.html, still needed if we're keeping reports from osi_run around? I think I need to understand more about people's current development process..

I would say so, I often work with reports that have already been released in the prod data because I know everything is in the correct state so I just want a reasonable looking file on disk.

lucyb:
> I would say so, I often work with reports that have already been released in the prod data because I know everything is in the correct state so I just want a reasonable looking file on disk.

Given that running `osi_run_and_release` will give you an guaranteed up-to-date file on disk now, is that enough? Having the file in `outputs/interactive_report.html` adds quite a bit of complexity, especially around the need to keep it up-to-date.

jobserver/management/commands/osi_release.py
```diff
@@ -83,10 +90,8 @@ def handle(self, *args, **options):
 
         tpp = Backend.objects.get(slug="tpp")
 
+        local_file = Path(options["report"])
```
ghickman, line 93: You can tell the parser to make this option a path using the `type` kwarg!  It's an `argparse` parser under the hood if that helps :)

ghickman (approved):
👍

jobserver/management/commands/osi_release.py
lucyb: Ooh, yes that does. Thank you!

Issue #3107, opened 2023-05-04
Reworking publish requests to handle rejections


ghickman:
#3015 covers the high level parts we need for OSIv2 but there are some underlying bits of architecture that need shuffling around now.  I've pulled them into this ticket just 

The two big changes are:
* A Report (and thus the related Snapshot) can have multiple requests to publish
* A publish request can be rejected (which is how a user can make subsequent requests)

To handle these we need to:
* Remove the approved fields and replace them with decision fields, covering the usual at/by pairing but also adding one without a suffix where we store the decided state (eg approved/rejected).
* The target objects, Report and ReleaseFile, can be in one of four states: approved, rejected, pending decision, no request.  The staff area filters will need to handle these.
* The 1-1 relationships between the request and target objects will need to become FKs.  Luckly a 1-1 is just an FK with an extra constraint so that migration is the easy part.  We'll need to handle the switch from one to many in a handful of places though (eg `is_published` and `is_draft` properties on the target models)
* Add a `reject` method to both request objects which the staff area view can call.

Related to all this we also want to straighten out how we handle ReleaseFilePublishRequests and Snapshots.  Currently we track the requested files on the publish request, then create a Snapshot from them but we're going to switch to creating the Snapshot up front (similar to how we create a report then ask to publish it), meaning we can drop the `files` field.  This matches the original design

PR #3108, opened 2023-05-04
Approve and reject publish requests

ghickman:
This covers a variety of changes, all driving the approval and rejection of publish requests.

**Staff area**
The approval of requests to publish reports has moved into the staff area, removing the user-facing view entirely.  It attempts to make it easy for a staff user to get from a report to a request to publish quickly in the staff area by listing publish requests on various pages.

The publish request pages in the staff area have the usual base tools, some filter, some searching, etc.

Rejection buttons have been added along side the approve one to keep the process all in the staff area.

**Models**
Both publish request models have been moved from approved_[at|by] fields to using decision fields which track approved/rejected.

**Notifications**
An email is sent to uses when their request is approved or rejected.

Fix: #3014
Fix: #3016
Ref: #3107


jobserver/migrations/0029_convert_reportpublishrequest_report_to_an_fk.py
```diff
@@ -0,0 +1,32 @@
...
+                on_delete=django.db.models.deletion.SET_NULL,
+                related_name="publish_requests",
+                to="jobserver.snapshot",
+            ),
+        ),
+        migrations.AlterField(
+            model_name="reportpublishrequest",
+            name="report",
+            field=models.ForeignKey(
+                on_delete=django.db.models.deletion.CASCADE,
```
lucyb, line 27: Just to confirm my understanding, `CASCADE` here means that if a PublishRequest happened to be deleted then the Report would be deleted too? Should this actually be set to `PROTECT` instead?

lucyb (commented):
Seems good! Happy to approve this once the comments have been addressed.

lucyb:
Not something for this PR, but will inform the next steps.. I noticed that after rejecting a publish request I was unable to update the title&summary and unable to make another publish request.

Screenshot from 2023-05-11 17-46-38

jobserver/migrations/0029_convert_reportpublishrequest_report_to_an_fk.py
ghickman: Yes but otherway around!  If the target object of an FK (report here) is deleted then it's related objects (the publish requests) will also be deleted.

jobserver/models/reports.py
```diff
@@ -92,27 +92,25 @@ def is_draft(self):
 
     @property
     def is_published(self):
-        return hasattr(self, "publish_request") and self.publish_request.approved_at
+        return self.publish_requests.filter(
+            decision=ReportPublishRequest.Decisions.APPROVED
+        ).exists()
```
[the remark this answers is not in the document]
ghickman: That's a good point, I'll flip this so we only look at the most recent.  I think that fits well with the idea of having multiple requests too, so only the most recent one is the one we consider 👍

lucyb (approved):

PR #3117, opened 2023-05-11
Move outputs-viewer to current publishing flows

ghickman:
This moves the output-viewer APIs to use our current underlying publishing models and flow.

The outputs-viewer should still be able to publish outputs in exactly the same way, with the UI being entirely unaffected by this change.

**UI testing**
A user, with the OutputPublisher role, should be able to view a released output and request it is published.  This will generate a snapshot and redirect the user to that snapshot where there is a button to confirm the files are published.

**Backend**
When the snapshot is generated it should have a related `SnapshotPublishRequest` with a NULL decision field, and the published_at/by fields on the Snapshot should also be NULL.

Once you've confirmed the publish request the publish request should have a decsion of `approved`, and the Snapshot's published_at/by fields should still be NULL.


jobserver/models/outputs.py
```diff
@@ -346,6 +346,18 @@ def approve(self, *, user, now=None):
         self.decision = self.Decisions.APPROVED
         self.save(update_fields=["decision_at", "decision_by", "decision"])
 
+    @classmethod
+    @transaction.atomic()
+    def create_from_files(self, *, files, workspace, user):
```
bloodearnest: I think that convention for a classmethod constructor would be to use `cls` rather than `self` as first parameter, then use `cls.objects.create(...)`, to make it work with inheritance.

Now, if we're inheriting from models, we've probably done something wrong, tbh.  But a `@classmethod` with `self` just looks a bit odd to me?

bloodearnest (approved):
This look ok to me.

Left a minor comment about classmethod conventions.

There's a lot of migrations that will be in this deploy - have we tried running all the migrations at once locally from a current copy of the prod db?

ghickman:
> have we tried running all the migrations at once locally from a current copy of the prod db?

Yep!

PR #3118, opened 2023-05-11
Remove ReportPublishRequest

ghickman:
This removes the ReportPublishRequest model, instead inferring requests to publish Reports via their ReleaseFile, via the related Snapshot.

This essentially merges ReportPublishRequest and SnapshotPublishRequest into PublishRequest.  I've maintained a nullable FK from PublishRequest to Report because a property was just too much leg work and doing any kind of querying for "reports which are published" (ie in the staff area) was really painful so now we have an FK and a check in the save() method to ensure the Report we set is the correct one.  It's not completely foolproof since there are situations where save() doesn't fire but it should be pretty reliable.

As part of this I noticed that PublishRequest.Snapshot is nullable and shouldn't be but I'll keep that for another PR to avoid adding yet more to this monster.

The first 3 commits are some housekeeping to set Snapshot up to be ready for this refactor.

The main commit is unfortunately _chonky_.  It might be easier to test by using a local copy of the site.  Some things one could check:
* creating a publish request from an analysis page
* publish request(s) are listed on the relevant report page in the staff area
* approving/rejecting publish requests in the staff area
* adding subsequent publish requests to a report when a previous one is rejected
* the publish request button is active on the analysis page when there is are not existing publish requests or the latest one was rejected
* the publish request button is disabled on the analysis page (with hovers) when there is a pending or approved published request
* there can only be one pending publish request for a report
* a published report can be seen without being logged in
* you can filter reports on the staff area's reports page


bloodearnest (approved):
A lot of careful plumbing work here, thanks!

PR #3119, opened 2023-05-11
Resubmit analysis

bloodearnest:


interactive/submit.py
```diff
@@ -50,18 +53,70 @@ def submit_analysis(
...
+    if repo_url.startswith("http"):
+        return _get_existing_commit_github(analysis_id, repo_url, github_api)
+    else:  # support local repositories for testing
+        return _get_existing_commit_local(analysis_id, repo_url)
+
+
+def _get_existing_commit_github(analysis_id, repo_url, github_api):
+    """Used for github repos, in production."""
+    url = furl(repo_url)
+    org, repo = url.path.segments[:2]
```
ghickman, line 98: If you pass a Repo instance to this function instead of the URL, it's got the `name` and `owner` properties to do this for you :)

interactive/submit.py
```diff
@@ -50,18 +53,70 @@ def submit_analysis(
...
-        created_by=creator,
+        created_by=analysis_request.created_by,
         sha=sha,
         project_definition=project_yaml,
         force_run_dependencies=True,
         requested_actions=["run_all"],
     )
-
     analysis_request.job_request = job_request
     analysis_request.save()
```
ghickman, line 75: Can you add a comment above this block to explain why we're updating the analysis_request here instead of at the call sites?

interactive/submit.py
```diff
@@ -50,18 +53,70 @@ def submit_analysis(
         force=force,
     )
 
-    job_request = project.interactive_workspace.job_requests.create(
+    create_job_request(analysis_request, backend, sha, project_yaml)
+
+    analysis_request.save()
```
ghickman, line 58: I think this is a dupe as `create_job_request` is also calling it?

ghickman (approved):
Couple of small suggestions but this is pleasingly short!

interactive/submit.py
bloodearnest: So we don't forget?

Would you prefer it to be the callers responsibility?

bloodearnest: I've made it the callers responsibility

Issue #3155, opened 2023-05-17
Block access to edit report metadata


ghickman:
When a report has a pending or approved publish request users should no longer be able to edit any of its metadata.

We'll need to disable the button for this on the ReportDetail page with an appropriate hover explaining why and also put the same check into the Edit view.

PR #3158, opened 2023-05-18
Block report editing

ghickman:
Fix: #3155


interactive/views.py
```diff
@@ -225,6 +225,18 @@ def dispatch(self, request, *args, **kwargs):
         if not self.analysis_request.report:
             raise Http404
 
+        # we can only edit reports if they don't have any publish requests or
+        # the latest one is rejected
+        latest_publish_request = self.analysis_request.report.publish_requests.order_by(
+            "-created_at"
+        ).first()
+        if latest_publish_request and (
+            latest_publish_request.decision is None
```
benbc, line 234: What is the semantics of `decision` being `None`? Is it an outstanding request? It might be helpful to make all the cases clear in the comment.

interactive/views.py
```diff
@@ -225,6 +225,18 @@ def dispatch(self, request, *args, **kwargs):
...
+            or latest_publish_request.decision
+            == ReportPublishRequest.Decisions.APPROVED
+        ):
+            raise PermissionDenied
```
benbc, line 238: Is `PermissionDenied` the right error? It would be nice if the error indicated that it's the state of the resource that's wrong, not the permissions of the user.

tests/unit/interactive/test_views.py
```diff
@@ -321,52 +375,55 @@ def test_analysisreportupdate_unauthorized(rf):
...
+    )
+
+    ProjectMembershipFactory(project=project, user=user, roles=[InteractiveReporter])
+
+    ReportPublishRequestFactory(
+        report=report,
+        decision=ReportPublishRequest.Decisions.APPROVED,
+        decision_at=timezone.now(),
+        decision_by=UserFactory(),
     )
```
benbc, line 394: It would be nice to test for the `REJECTED` case too.

jobserver/models/reports.py
```diff
@@ -90,6 +90,18 @@ def get_staff_url(self):
...
+    def is_locked(self):
+        latest = self.publish_requests.order_by("-created_at").first()
+
+        if not latest:
+            return False
+
+        if latest.decision == ReportPublishRequest.Decisions.REJECTED:
+            return False
+
+        return True
```
benbc, line 103: Could the logic in the view use this too?

benbc (approved):
Some questions and comments -- maybe indicative of my ignorance.

Am I right in understanding that `FormView` is a lower-level API than `UpdateView` and friends?

ghickman:
> Am I right in understanding that FormView is a lower-level API than UpdateView and friends?

Pretty much.  `FormView` is for views with a form which maybe doesn't directly use a model, while `UpdateView` is the higher level/more magic version which takes a model and some fields to update.

interactive/views.py
ghickman: I've got a nicer API for this (it'll be `latest_publish_request.is_pending`) coming in a subsequent PR since it's popped up a few times now!  I'll update that PR to cover this case 👍

interactive/views.py
ghickman: Maybe…  We've disabled the button used to get here so this is to catch users who go directly to the URL. And now that I write it out, of course someone is going to be on that page or have a URL somehow.  I'll put a template together for this 👍

interactive/views.py
```diff
@@ -204,15 +204,11 @@ def get_context_data(self, **kwargs):
...
-        "title",
-        "description",
-    ]
-    model = Report
+class ReportEdit(FormView):
+    form_class = AnalysisRequestEditForm
     template_name = "interactive/report_edit.html"
 
-    def get_object(self, **kwargs):
+    def dispatch(self, request, *args, **kwargs):
```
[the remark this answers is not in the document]
ghickman: None, I just missed it in the refactor, fixed!

Issue #3150, opened 2023-05-17
Merge ReportPublishRequests into ReportDetail


ghickman:
Using publish requests on a separate page to the report they target is confusing when you're coming to the report without much context.

Similar to what [Lucy explained in this bit of feedback](https://github.com/opensafely-core/job-server/pull/3101#discussion_r1186052243) we can move them from their own pages to the ReportDetail page as a list in the left-hand column so there's enough room to display their metadata, and have the approve/deny buttons.

PR #3161, opened 2023-05-18
Move ReportPublishRequests to their Report pages

### LINKED ISSUE #3153: Remove email from rejection flow
We're currently emailing (with the wrong email) users who created a publish request.  We're going to create a manual process for this until we have a better idea of how this will work.

ghickman:
It was confusing for a user to find the ReportPublishRequests when arriving in the staff area with no context other than a report.  This lists them, with their metadata, on the relevant report page, and provides users with approve/reject buttons when a decision hasn't already been made.

#### Testing
1. Grab a copy of the production db and load it locally
2. Pick an analysis request in the staff area, choose one which already has a report
3. Grab the analysis request's slug and generate a report file for it locally with `python manage.py osi_release <slug> <your username>` (the contents of the report is unimportant for this fix)
4. Open the analysis request page the user would see and request it's report is published
5. Go to the report in the staff area (its linked from the analysis request page)
6. Try to approve/reject the request

#### With a decision

---

I've also fixed #3153 here since it's a tiny fix but relies on changes in this PR.

Fix: #3150 
Fix: #3153


benbc (approved):
Without having tested this it looks fine.

Issue #3013, opened 2023-04-18
Allow staff users to view and download unpublished OSI reports


CLStables:
The co-pilot must be able to download a copy of the report including Report Summary, to send to Amir (who will then send to NHSE)

lucyb:
@ghickman can you check that the permissions are as Catherine expects for viewing the Analysis Detail page (and for being able to browse to that page), please?

ghickman:
This works if the staff user has the `InteractiveReporter` role assigned globally, which matches what we've done in general for staff users around the site.

@CLStables – you've got that role assigned globally.  Can you try viewing (and downloading) any AnalysisRequest from the staff area that you didn't create and see if it works as expected?

lucyb:
I suspect that in reality, most of our existing "staff" users will want/need to view released reports and the associated analysis request. So would it make sense to allow anyone with permissions to view the released outputs to also be able to view reports/analysis requests, but restrict the ability to create an analysis request to users with the `InteractiveReporter` role?

PR #3206, opened 2023-06-08
Let staff view unpublished reports

ghickman:
Fix: #3013


lucyb (approved):
I think this makes sense. A Report is a fancy wrapper around a ReleaseFile, so basing it off the underlying permissions seems reasonable.

I like the commit message too :)

Issue #2620, opened 2023-02-10
ADR: Reports on OpenSAFELY


ghickman:
* What is a report on OpenSAFELY?

This is related to both OSI, and the Reports' site.

lucyb:
You've now written the ADR for publishing reports. Is that anything else you think should be in an ADR at the moment?

ghickman:
Yes, I want to define what exactly a report is since it's a wrapper object.

PR #3221, opened 2023-06-19
Rework reports-related ADRs

ghickman:
Rather than try to define a report after an ADR which defined publishing the same thing this creates the reoprt-defining ADR as the first one of the pair and edits down the publishing one to remove the duplicate content.

Having them defined as 

1. A Report is defined as […]
2. how we publish a Report

should, I hope, give a clearer picture of reports and publishing.

Fix: #2620


docs/adr/0016-reports.md
```diff
@@ -0,0 +1,21 @@
...
+
+
+## Decision
+We will support reports in job-server by mirroring the implementation of those in the Reports service, with the exception that the underlying ReleaseFile is related to a Report via a ForeignKey.
+
+We are defining a Report on the OpenSAFELY platform as:
+
+    A released file wrapped with some extra metadata and styling.
+
+
```
lucyb, line 18: We also made some decisions about the styling that it might be worth adding a sentence about here. IIRC, we expect the report to be decorated with the standard CSS classes, with Job Server providing the stylesheet and therefore providing a mechanism for some consistent styling across the reports (although of course, this can be overridden with inline styles if needed, I believe?).

docs/adr/0017-publishing-reports.md
```diff
@@ -1,22 +1,18 @@
...
 Accepted
 
 ## Context
-As part of OSIv2 we have added reports to job-server.
-Their implementation aligns with the Reports service in that they are an output from a pipeline job, rendered with some extra styles.
-In this service this means they are also ForeignKey'd to the underlying ReleaseFile.
-
 We already have a process for publishing outputs via the Snapshot models and the outputs-viewer.
 However for a report the rendering of a given HTML file matters.
 We also want to better model the process of a user requesting an object on our platform is made public.
```
lucyb, line 10:
```suggestion
We also want to better model the process of a user requesting that an object on our platform is made public.
```

lucyb (commented):
This looks really good. I think it makes sense to split the ADRs the way you have and to order them like that.

docs/adr/0016-reports.md
ghickman: Updated to add this without being prescriptive on the classes so we don't feel we're tied to the current set.

lucyb: I think that's sensible, it was the division of responsibility that I thought was important to mention, which I can see you've now done. Thank you :)

PR #3231, opened 2023-06-20
Test the publishing flow

ghickman:
Now that publishing is just about considered done and we've documented it, this adds an integration test to codify the flow.

While adding this test I discovered the PublishRequestCreate page doesn't handle a report being locked.  I've mirrored the ReportEdit page's flow, showing a locked template in that situation:

### DIFF
--- a/tests/integration/test_interactive.py
+++ b/tests/integration/test_interactive.py
@@ -1,22 +1,31 @@
 import json
 import tempfile
+from unittest.mock import patch
 
 import pytest
+from django.utils.formats import date_format
 from interactive_templates import git
 from opensafely._vendor.jobrunner.cli import local_run
+from pytest_django.asserts import assertInHTML
 
 from interactive.views import AnalysisRequestCreate
-from jobserver.authorization import InteractiveReporter
+from jobserver.authorization import CoreDeveloper, InteractiveReporter
+from jobserver.models import PublishRequest
+from jobserver.views.reports import PublishRequestCreate
 
 from ..factories import (
+    AnalysisRequestFactory,
     BackendFactory,
+    JobFactory,
+    JobRequestFactory,
     ProjectFactory,
     ProjectMembershipFactory,
     RepoFactory,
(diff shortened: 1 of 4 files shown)


tests/integration/test_interactive.py
```diff
@@ -95,3 +118,110 @@ def test_interactive_submission_success(rf, local_repo, enable_network):
...
+    # request it's published
+    assert analysis.status == "succeeded"
+
+    # user can view their analysis page and the report is rendered there
+    with patch.object(
+        AnalysisRequestCreate, "get_opencodelists_api", new=FakeOpenCodelistsAPI
+    ):
+        response = client.get(project.get_interactive_url())
+        assert response.status_code == 200
+        # TODO: check rendered response has the report details
```
lucyb, line 155: This TODO is confusing. Can this be implemented or removed, please?

tests/integration/test_interactive.py
```diff
@@ -95,3 +118,110 @@ def test_interactive_submission_success(rf, local_repo, enable_network):
...
+    with patch.object(PublishRequestCreate, "get_github_api", new=FakeGitHubAPI):
+        response = client.post(analysis.get_publish_url(), follow=True)
+        assert response.status_code == 200
+        assert response.redirect_chain == [(analysis.get_absolute_url(), 302)]
+
+    # check ticket is requested
+    requests = PublishRequest.objects.filter(report=report)
+    assert requests.count() == 1
+
+    # grab the publish request to make some of the tests below a little easier
```
lucyb, line 168: YMMV, but I'm not sure this comment is needed because the code is clear enough without it, I think.

lucyb (approved):

tests/integration/test_interactive.py
ghickman: It referenced the commented out line below which I forgot to fix 😂

ghickman: And exposed a broken part of the test 🎉

PR #3237, opened 2023-06-22
Move OSI commands into the interactive app

ghickman:
We were previously keeping all commands inside the jobserver app, however these commands are OSI specific so the `interactive` app makes more sense following the principle of least surprise.


bloodearnest (approved):
+1, but should the test be moved from tests/unit/jobserver to tests/unit/interactive too?

PR #3242, opened 2023-06-26
Try to avoid dangling repos from verification tests

ghickman:
We've been building up testing repos in opensafely-testing over time.  It's unclear exactly where these have come from, but this PR adds some changes to hopefully avoid creating them:

The `testing_repo` fixture now suffixes its repos with a unique ID to match the `test_create_repo` test but also changes the repo prefix to `testing-repo` so should we still have this problem we can at least track it to a code path.

`test_create_repo` now deletes its repo when done and doesn't create an extra repo via the `testing_repo` fixture.

### DIFF
--- a/tests/verification/test_github.py
+++ b/tests/verification/test_github.py
@@ -85,30 +85,34 @@ def test_create_issue(enable_network, github_api):
 
 
 def test_create_repo(enable_network, github_api, testing_repo):
-    # create a unique ID for this test using our existing ULID function.
-    # we can have multiple CI runs executing concurrently which means this
-    # test can be running in different CI jobs at the same time.  Given the
-    # test talks to an external API we can't use the same repo name for every
-    # execution.  Instead we use our new_ulid_str since it's already available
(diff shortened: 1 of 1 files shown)


tests/verification/test_github.py
```diff
@@ -110,6 +92,9 @@ def test_create_repo(enable_network, github_api, testing_repo):
     with pytest.raises(RepoAlreadyExists):
         github_api.create_repo(*args)
 
+    # clean up after the test
+    github_api.delete_repo(*args)
```
lucyb, line 96: I think this change is an improvement overall. It's much easier to see where the repo is being created (it's easy to miss it in a fixture). Just one minor thing, which is if there's an error after the initial repo creation, but before the call to delete the repo, then the repo will never get deleted. Wrapping it in a `try..finally` would stop this, it's a bit of a pain but would reduce the chances of having ~2000 repos again.

lucyb (approved):

Issue #3244, opened 2023-06-26
Squash migrations


ghickman:
We're into the 40s with our migrations, so it's time to flatten them again.

Issue #1165, opened 2021-10-18
Switch RolesField to use Arrays


ghickman:
Once #1033 is done we can move RolesField to use Postgres' ArrayField instead of slightly misusing JSON fields.

We'll need to do this in multiple PRs/deploys to handle changing a Field definition in our code and migrations needing those to stick around as long as migrations which reference them do.

PR #3246, opened 2023-06-27
Convert roles fields to native Postgres array

ghickman:
We've been using a JSONField to store Roles assigned to various objects in the database as strings containing dotted paths.  While this has worked relatively well it adds some slightly odd edge cases because the field can be queried in unexpected ways, and the default value of a list is only enforced in the Python code.

This set of changes switches us over to two custom fields:
1. `RolesArrayField`, a subclass of `ArrayField` which handles uniquing the input of values to the field.
2. `RoleField`, a subclass of `TextField` which handles converting roles to dotted path strings and back.

RolesArrayField is the "public" API to this field because ArrayFields require a base field for the data inside the array.

Once this PR is merged we can remove the old `RolesField`, rename `RolesArrayField` to `RolesField`, and fix the references in the migrations.  Doing this after #3244 would minimise the amount of work.

#### Testing
You should be able to test this by exercising various places where we use roles:
* Filtering users by role in the staff area
* Adding uses to projects in the staff area
* Viewing the run job as a user with permissions to do so

Ref: #1165


lucyb (approved):
This is a really nice improvement. I've not used ArrayField before, so I appreciated the explanations in various places. Thank you!

PR #3247, opened 2023-06-27
Enforce publish request snapshot

ghickman:
This fixes some leftovers from #3118 and tightens up the relationship between PublishRequest and Snapshot.

The main thing to look at for this is if we should be cascading Snapshot deletions to PublishRequest or blocking them with protect?


jobserver/models/outputs.py
```diff
@@ -29,7 +29,7 @@ class Decisions(models.TextChoices):
     )
     snapshot = models.ForeignKey(
         "Snapshot",
-        on_delete=models.SET_NULL,
+        on_delete=models.CASCADE,
```
lucyb, line 32: Would we delete a snapshot if the wrong file was requested or published? Having the audit trail of who made the request and what decision was made sounds like it'll be important even if the file is no longer there.

lucyb: Thank you. That makes sense. It sounds like we'll have an audit trail if files are deleted, which is my only concern. I don't think we'll need that extra step either, so I think CASCADE is the right choice here.

lucyb (approved):

PR #3301, opened 2023-07-20
Increase production workers to 5

bloodearnest:


madwort (approved):

madwort:
This was tested in production & observed not to make a significant memory impact.

Issue #3276, opened 2023-07-17
Manually publish a single file


ghickman:
Currently we can only publish all files released from a workspace.  Over time we've found a need to publish a subset of files released for a workspace.

Another request has come in for this, to publish [this released file](https://jobs.opensafely.org/datalab/pincer/pincer-measures/outputs/56/output/report_updating.html).

There are two parts to this ticket:
* [ ] create a draft Snapshot for the linked file above, checking with Louis&Rose that it's correct before making it public.
* [x] Building out a function in the relevant commands.py to release an iterable of the given ReleaseFiles, updating the OSI report-creation code to use it, and wrapping it in a management command so we can try out publishing subsets of files before updating the UI to do it too.

lucyb:
@ghickman has this been documented anywhere?

ghickman:
@lucyb – not yet, we're going to try it out the next time a regular report comes up (there are at least 2 monthly ones so should be soon) and document if we're still happy with the process after that.

PR #3303, opened 2023-07-21
Script publishing a list of files

ghickman:
This adds a check to ensure that Snapshots are made with files from the same workspace.  We're creating all Snapshots via the PublishRequest.create_from_files method so we can rely on the check there.

Fix: #3276

### DIFF
--- a/jobserver/management/commands/publish_files.py
+++ b/jobserver/management/commands/publish_files.py
@@ -1,0 +1,35 @@
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+from furl import furl
+
+from jobserver import models
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument("username", type=str)
+        parser.add_argument(
+            "ids", nargs="+", type=str, help="ReleaseFile IDs to add to Snapshot"
+        )
+
+    def handle(self, *args, **options):
+        ids = set(options["ids"])
+        username = options["username"]
+
+        try:
+            user = models.User.objects.get(username=username)
+        except models.User.DoesNotExist:
+            raise CommandError(f"Unknown username: {username}")
+
+        rfiles = models.ReleaseFile.objects.filter(pk__in=ids)
+
+        found = set(rfiles.values_list("pk", flat=True))
+        if unknown := ids - found:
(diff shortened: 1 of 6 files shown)


tests/unit/jobserver/models/test_outputs.py
```diff
@@ -103,9 +97,16 @@ def test_publishrequest_create_from_files_with_existing_snapshot():
...
-        PublishRequest.create_from_files(
-            files=files, user=UserFactory(), workspace=workspace
-        )
+        PublishRequest.create_from_files(files=files, user=UserFactory())
+
+
+def test_publishrequest_create_from_files_with_multiple_workspaces():
+    user = UserFactory()
+
+    rfiles = ReleaseFileFactory.create_batch(2)
```
benbc, line 106: I take it that the workspace here defaults to a unique thing that's created for each file. It would be easier to understand this test if we explicitly created files with different workspaces, since that's cruicial to the correctness of the test.

jobserver/management/commands/publish_files.py
```diff
@@ -0,0 +1,35 @@
...
+        found = set(rfiles.values_list("pk", flat=True))
+        if unknown := ids - found:
+            raise CommandError(f"Unknown ReleaseFiles: {', '.join(unknown)}")
+
+        publish_request = models.PublishRequest.create_from_files(
+            files=rfiles, user=user
+        )
+
+        url = furl(settings.BASE_URL) / publish_request.snapshot.get_absolute_url()
+        print(url)
```
benbc, line 35: What does this display? Is this a pattern you're using in other commands?

ghickman: It's the full URL to the snapshot created inside of Publish.create_from_files.  BASE_URL is configured to the appropriate domain for a given env, `/` has been overloaded by furl to let us use it to piece together URLs, and `get_absolute_url` is a Django-ism to get the detail page of an object.

I'm not sure I'd go as far as to call it a pattern, but since I'll need the resulting URL to give to whomever requested the files be published, it's very helpful!

tests/unit/jobserver/models/test_outputs.py
ghickman: Yes, by default all our factories create unique items for their FKs, but point taken 👍

PR #3308, opened 2023-07-21
Replace Alpine.js

### LINKED ISSUE #3184: Replace AlpineJS with Web Components JS
It appears that the CSP build of Alpine is no longer officially supported (well, maybe it never was): https://github.com/alpinejs/alpine/pull/2619

We should consider other CSP compliant options for Job Server.

## Todo

- [ ] Replace the Header code
- [ ] Replace the account drop down

tomodwyer:
Closes #3184

- Use separate nav bars for desktop and mobile views
- Replace dropdown code with a custom vanilla JS driven solution based on data attributes

### DIFF
--- a/assets/src/scripts/_expand-list.js
+++ b/assets/src/scripts/_expand-list.js
@@ -1,0 +1,12 @@
+const expandBtns = [...document.querySelectorAll(`[data-expander-button]`)];
+
+expandBtns.map((btn) =>
+  btn.addEventListener("click", () => {
+    const id = btn.getAttribute("data-expander-button");
+    const list = document.querySelector(`[data-expander-list=${id}]`);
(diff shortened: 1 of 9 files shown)


ghickman (approved):
Looks great!

I couldn't find a nice way to put this into the commit view… there looks to be a fair bit of duplicated content in header template with the two menu navs now.  Would there be any benefit to pulling the two headers out into different files?  I can see an argument either way!

tomodwyer:
@ghickman I did think about this, but figured it might be more confusing than helpful, as the two menus are different enough that someone might change one, but forget to change the other.

I think we could do some improvements here when we take a look at the UX work around Job Server though, which would simplify the code.

Issue #3322, opened 2023-07-25
Add JobRequests to the staff area


ghickman:
As part of #3321 we need some of the functionality for staff to stick around, so we should move those bits into the staff area.

PR #3323, opened 2023-07-25
Add JobRequests to the staff area

ghickman:
We can't lean on our typical filtering UI for JobRequests since we'd had to list nearly every User, Project, and Workspace.  I suspect we can add those in later with searchable dropdowns, however I didn't want this change to be bogged down by working out how to get those working in the staff area.

Fix: #3322

### DIFF
--- a/staff/views/job_requests.py
+++ b/staff/views/job_requests.py
@@ -36,6 +36,7 @@ class JobRequestList(ListView):
                 "created_by__fullname",
                 "created_by__username",
                 "identifier",
+                "jobs__identifier",
                 "pk",
                 "workspace__name",
                 "workspace__project__name",
--- a/staff/views/job_requests.py
+++ b/staff/views/job_requests.py
@@ -2,11 +2,18 @@ import functools
 
 from django.db.models import Q
 from django.utils.decorators import method_decorator
-from django.views.generic import ListView
+from django.views.generic import DetailView, ListView
 
 from jobserver.authorization import CoreDeveloper
 from jobserver.authorization.decorators import require_role
 from jobserver.models import JobRequest
+
+
+@method_decorator(require_role(CoreDeveloper), name="dispatch")
+class JobRequestDetail(DetailView):
+    context_object_name = "job_request"
+    model = JobRequest
+    template_name = "staff/job_request_detail.html"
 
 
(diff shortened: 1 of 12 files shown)


staff/views/job_requests.py
```diff
@@ -0,0 +1,56 @@
...
+            ]
+
+            # build up Q objects OR'd together.  We need to build them with
+            # functools.reduce so we can optionally add the PK filter to the
+            # list
+            qwargs = functools.reduce(
+                Q.__or__, (Q(**{f"{f}__icontains": q}) for f in fields)
+            )
+
+            qs = qs.filter(qwargs)
```
lucyb, line 51: Filter by identifier isn't working for me for some reason

lucyb (commented):
As I understand it, Job Runner provides the Job ID, so this gets searched for a lot when debugging issues. Can this be added to the filter in the staff area too please?

templates/staff/job_request_detail.html
```diff
@@ -0,0 +1,113 @@
...
+
+{% block jumbotron %}
+<div class="jumbotron jumbotron-fluid jumbotron--danger pt-md-2">
+  <div class="container">
+    <h1 class="display-4">{{ job_request.pk }}</h1>
+
+    <ul class="list-unstyled lead">
+      <li>
+        <strong>Identifier:</strong> <code>{{ job_request.identifier }}</code>
+      </li>
```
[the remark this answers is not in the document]
lucyb: Oh sorry, that comment was supposed to be about the Job Request List page.. so we can quickly see more information about the job request without having to click into each one. The workspace name is more useful to me than the person who made the request.

templates/staff/job_request_list.html
lucyb, line 1: One final thing... should the filters that have been selected be visible on the left hand side of this page? At the moment, it's not obvious that I'm viewing a filtered list, except for a "clear all" option

ghickman: There are no filter controls currently.  We don't have any dropdowns-with-search that I could throw into this page.  The search queries aren't displayed on the page in any of our list views, but thinking about it that would be a nice addition, captured in #3326

ghickman:
I've also added Job identifiers to the search now too.

lucyb (approved):

Issue #3327, opened 2023-07-26
Add searchable dropdowns to the staff area


ghickman:
As part of adding JobRequests to the staff area (#3323) we've hit having filters that can't be easily displayed alongside the content.  We've paginated JobRequestList to 25 items, but having orgs, projects, or workspace UI filters would display all of those, of which there are many.

Instead we should add select2-style dropdowns so filters can selected with fuzzy searching, and then displayed next to those dropdowns when the page reloads.

This will have the added benefit of helping out #2827 which will the same (or a very similar) UI control everywhere that orgs are selected for projects.

PR #3330, opened 2023-07-26
Add searchable filters to the JobRequestList page

ghickman:
To use select2, or any dropbox element, we need to wrap the entire page in a form so we're also capturing the search box in the right-hand column.  This allows us to keep the ability increase/reduce filters without them overriding each other, and to search within any given set of filters.

Fix: #3327

### DIFF
--- a/staff/views/job_requests.py
+++ b/staff/views/job_requests.py
@@ -1,12 +1,13 @@
 import functools
 
 from django.db.models import Q
+from django.db.models.functions import Lower
 from django.utils.decorators import method_decorator
 from django.views.generic import DetailView, ListView
 
 from jobserver.authorization import CoreDeveloper
 from jobserver.authorization.decorators import require_role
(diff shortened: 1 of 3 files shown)


staff/views/job_requests.py
```diff
@@ -23,6 +24,31 @@ class JobRequestList(ListView):
     paginate_by = 25
     template_name = "staff/job_request_list.html"
 
+    def get_context_data(self, **kwargs):
+        orgs = {
+            "is_active": "orgs" in self.request.GET,
+            "items": Org.objects.order_by(Lower("name")),
```
bloodearnest, line 30: Worth eagerly evaluated these queries in the view rather than pushing them in to the template rendering? Given our recent funs...

bloodearnest (approved):

tomodwyer (approved):

Issue #3311, opened 2023-07-24
Replace model field validators


ghickman:
Model field validators are [only run as part of form validation](https://docs.djangoproject.com/en/4.2/ref/validators/#how-validators-are-run), so they're not a great fit for us to ensure models are always created correctly.

We can either replace them with constraint(s) or wrap creation for a given model in a classmethod depending on the validator we're replacing.

PR #3331, opened 2023-07-26
Replace model field validators with constraints

ghickman:
Model field validators are only run when used in conjunction with ModelForms.  While this is useful, it doesn't bring with it the guarantees one might expect when reading a model definition.

Moving to database-level constraints also brings the benefit of ensuring db rows created manually are correct.

I've left User.username alone since that field and its validator were taken from Django's own `auth.User` and I don't want to dig through all of the auth stuff before I fix the two fields I've done in this PR.

Fix: #3311


redirects/models.py
```diff
@@ -65,6 +66,18 @@ class Redirect(models.Model):
...
+                name="old_url_is_not_empty",
+            ),
+            models.CheckConstraint(
+                check=Q(old_url__endswith="/"),
+                name="old_url_endswith_slash",
+            ),
+            models.CheckConstraint(
+                check=Q(old_url__startswith="/"),
+                name="old_url_startswith_slash",
+            ),
```
rebkwok, line 80: Is it any more efficient to put these into a single constraint with `&`'d `Q`s?

rebkwok (approved):

redirects/models.py
ghickman: Not sure about efficiency but definitely seems better to have them in one 👍

ghickman: Also added some tests specifically for the constraints that I'd missed

Issue #3326, opened 2023-07-26
Display search queries on staff area list pages


ghickman:
The list pages in the staff area don't currently show what's been searched for.  We should add this into filters sidebar below the "Clear filters" link.

PR #3335, opened 2023-07-26
Display search query in the filters sidebar of staff area list pages

ghickman:
This adds the current search query for a staff area list page into the sidebar above any possible filters.

Along the way I've added any missing bits to various list pages to try and bring them in line with each other.

Fix: #3326

### DIFF
--- a/staff/views/redirects.py
+++ b/staff/views/redirects.py
@@ -1,7 +1,4 @@
-import functools
-
 from django.contrib import messages
-from django.db.models import Q
 from django.shortcuts import get_object_or_404, redirect
 from django.utils.decorators import method_decorator
 from django.views.generic import DetailView, ListView, View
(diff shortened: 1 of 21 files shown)


staff/views/redirects.py
```diff
@@ -48,15 +50,26 @@ def get_queryset(self):
...
+                "org__name",
+                "project__name",
+                "workspace__name",
+            ]
+
+            # build up Q objects OR'd together.  We need to build them with
+            # functools.reduce so we can optionally add the PK filter to the
+            # list
+            qwargs = functools.reduce(
+                Q.__or__, (Q(**{f"{f}__icontains": q}) for f in fields)
```
bloodearnest, line 67: I get this, but also :scream:

Slight suggestion to improve future readability
```suggestion
                Q.__or__, (Q(**{f"{field}__icontains": q}) for field in fields)
```

bloodearnest (approved):

staff/views/redirects.py
madwort: this is amazing

ghickman:
@bloodearnest – fixed the single letter variable, and did you more: pulled it out into a function so they can all use the same code path!

Issue #2827, opened 2023-03-16
Convert Project/Org relationship to an M2M


ghickman:
We expected each project to be owned by a single organisation when we added both objects, however in reality it's possible for two or more orgs to collaborate on a project.

We can model this by converting the existing ForeignKey to a ManyToManyField.

~We will also need to remove orgs from the main URL structure.  This _should_ be trivial to manage by setting up the apppriate redirect instances on all projects.~ (done in #3312)

We also need to handle various staff area controls switching to a multiple item selection UI.

madwort:
@bloodearnest to review this week

PR #3348, opened 2023-07-31
Replace org breadcrumbs

ghickman:
This drops the Organisation breadcrumb from all pages.

This related to #2827 which is turning out to be somewhat onerous, and this is the first of probably a few chunks to make reviewing the changes manageable.


templates/job_request_detail.html
```diff
@@ -36,6 +35,11 @@ <h1 class="text-3xl tracking-tight break-words font-bold text-slate-900 sm:text-
             Job request: {{ job_request.pk }}
           </h1>
           <dl class="flex flex-col gap-2 text-sm text-slate-600 items-center sm:flex-row sm:gap-x-6 sm:justify-center lg:justify-start">
```
tomodwyer, line 37: Tiny styling tweak:

- Add class `leading-snug` to the `h1`
- Add class `mt-1` to the `dl`

ghickman: `leading-snug` is overridden by `sm:text-4xl` at all widths, is that expected?

PR #3388, opened 2023-08-11
fix update-interactive-templates bash snippet to handle -e

bloodearnest:


justfile
```diff
@@ -118,7 +118,7 @@ update-interactive-templates ref="": && prodenv
     #!/usr/bin/bash
     set -euo pipefail
 
-    test "$ref" == "" && ref=$(git ls-remote https://github.com/opensafely-core/interactive-templates HEAD | awk '{print $1}')
+    test "{{ ref }}" == "" && ref=$(git ls-remote https://github.com/opensafely-core/interactive-templates HEAD | awk '{print $1}')
     prefix="interactive_templates@https://github.com/opensafely-core/interactive-templates/archive"
     sed -i "s#${prefix}.*#${prefix}/${ref}.zip#" requirements.prod.in
```
madwort, line 123: should this be
```suggestion
    sed -i "s#${prefix}.*#${prefix}/{{ ref }}.zip#" requirements.prod.in
```

madwort (approved):

PR #3424, opened 2023-08-21
Bump pydantic to v2 (via pipeline)

ghickman:
TODO: bump cli&job-runner to pydantic v2 to save on PRs here.


ghickman:
Closing this because we can't use dependencies with .so files in our stack at present, a la opensafely-core/pipeline#216

Issue #3347, opened 2023-07-31
Track number of queries in tests


ghickman:
We're starting to see a few performance issues recently, and particularly with the removal of an implicit query (#3323), and the removal of orgs from urls (#3312).

It would be useful to have some query counts in our tests, but it's not clear how many pages we should do this for.  Lets err on the side of adding them to the success test case for each page to start with.

The homepage has various objects in it so it's an obvious first target to try this out.

Templates:
- [ ] _job_request_pagination.html
- [ ] _job_requests.html
- [ ] index-authenticated.html
- [ ] index-unauthenticated.html
- [ ] interactive/analysis_request_create.html
- [ ] interactive/analysis_request_detail.css
- [ ] interactive/analysis_request_detail.html
- [ ] interactive/publish_request_create.html
- [ ] interactive/publish_request_create_locked.html
- [ ] interactive/report_edit.html
- [ ] interactive/report_edit_locked.html
- [ ] job_detail.html
- [ ] job_request_create.html
- [ ] job_request_detail.html
- [ ] job_request_list.html
- [ ] job_request_pick_ref.html
- [ ] login.html
- [ ] org_detail.html
- [ ] org_list.html
- [ ] outputs-spa.html
- [ ] pipeline/require_name.html
- [ ] project_detail.html
- [ ] project_edit.html
- [ ] project_release_list.html
- [ ] project_repo_list.html
- [ ] project_report_list.html
- [ ] release_detail.html
- [ ] settings.html
- [ ] sign_off_repo.html
- [ ] snapshot_detail.html
- [ ] staff/analysis_request_detail.html
- [ ] staff/analysis_request_list.html
- [ ] staff/application_approve.html
- [ ] staff/application_detail.html
- [ ] staff/application_edit.html
- [ ] staff/application_list.html
- [ ] staff/backend_create.html
- [ ] staff/backend_detail.html
- [ ] staff/backend_edit.html
- [ ] staff/backend_list.html
- [ ] staff/dashboards/copiloting.html
- [ ] staff/dashboards/index.html
- [ ] staff/dashboards/projects.html
- [ ] staff/dashboards/repos.html
- [ ] staff/dashboards/repos_with_multiple_projects.html
- [ ] staff/index.html
- [ ] staff/job_request_detail.html
- [ ] staff/job_request_list.html
- [ ] staff/org_add_github_orgs.htmx.html
- [ ] staff/org_create.form.html
- [ ] staff/org_create.html
[...]

PR #3437, opened 2023-08-22
Pull queries for JobRequestDetail from template

ghickman:
This is an initial attempt to pull queries out of template rendering and into the view.  It uses [django-zen-queries](https://github.com/dabapps/django-zen-queries), from the ever wonderful folks over at dabapps, a nice set of utilities built around Django's [database instrumentation hooks](https://docs.djangoproject.com/en/4.2/topics/db/instrumentation/).

#### Implementation thoughts
zen-queries provides a great toolbox for opting in or out of render-time querying.  Getting clear errors from Django when we unexpectedly do it was our goal here and this gives it to us, with the offending query in the exception.

Some parts of zen-queries that I liked in particular:
* TemplateResponse is a drop-in replacement for Django's own response, allowing us to configure our CBVs ore return it directly.
* There are opt-outs allowing us to keep rendering in the template if the alternative is onerous (ie moving JobRequest's `status` and `completed_at` properties into QuerySet/Manager annotation methods)
* It includes the `fetch` wrapper, to force QuerySet evaluation.

#### Thoughts on the theory
There's been a clear tension while doing this work.  On the one hand we have some complex logic in places that's very non-trivial to move into the database (JobRequest.status and JobRequest.completed_at are the obvious candidates here), yet they're very necessary to keep template or view logic to a minimum.

Given the small, hard to track bugs we've had because of render-time queries I think we're headed in the right direction with this, despite the tensions.

Ref: #3347

### DIFF
--- a/jobserver/views/job_requests.py
+++ b/jobserver/views/job_requests.py
@@ -5,11 +5,11 @@ from django.db import transaction
 from django.db.models import Q
 from django.http import Http404
 from django.shortcuts import get_object_or_404, redirect
-from django.template.response import TemplateResponse
 from django.utils.safestring import mark_safe
 from django.views.generic import CreateView, ListView, RedirectView, View
 from django.views.generic.edit import FormMixin
 from pipeline import load_pipeline
+from zen_queries import TemplateResponse
 
 from .. import honeycomb
 from ..authorization import CoreDeveloper, has_permission, has_role
--- a/jobserver/views/job_requests.py
+++ b/jobserver/views/job_requests.py
@@ -197,6 +197,12 @@ class JobRequestDetail(View):
 
         jobs = fetch(job_request.jobs.order_by("started_at"))
 
+        # we encode errors raised by job-runner when processing a JobRequest by
+        # misusing a Job, since that's our only method of returning messages
+        # from job-runner to job-server.  These misused jobs use the __error__
+        # action to differentiate themselves.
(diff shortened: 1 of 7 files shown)


jobserver/views/job_requests.py
```diff
@@ -195,6 +195,14 @@ def get(self, request, *args, **kwargs):
         except (JobRequest.DoesNotExist, MultipleObjectsReturned):
             raise Http404
 
+        jobs = fetch(job_request.jobs.order_by("started_at"))
+
+        # we encode errors raised by job-runner when processing a JobRequest by
+        # misusing a Job, since that's our only method of returning messages
+        # from job-runner to job-server.  These misused jobs use the __error__
+        # action to differentiate themselves.
+        is_invalid = jobs.filter(action="__error__").exists()
```
lucyb, line 204: This line isn't really being tested at the moment, except as a side effect of fewer database queries being made. Is there an easy way to test that we're handling invalid jobs correctly? I know the tests are quite long at the moment..

lucyb (approved):
One small recommendation, otherwise it looks like a step in the right direction. Zen queries looks good but does add some complexity, so we should just keep an eye on how it goes over time.

The ideal next step would be to do something similar with the pages that Tom is about to change. I think he's starting to look at the more complex staff area pages, so having these guardrails in place before he does that would be ideal.

PR #3460, opened 2023-08-25
Restyle Applications templates in Staff Area

tomodwyer:
- Update templates in Staff Area: Applications
- Use input filter for sidebar filtering of results
- Add pagination to the applications


tests/unit/staff/views/test_applications.py
```diff
@@ -465,7 +464,7 @@ def test_applicationlist_search(rf, core_developer):
     assert response.status_code == 200
 
     assert len(response.context_data["object_list"]) == 2
-    assert set_from_qs(response.context_data["object_list"]) == {app1.pk, app2.pk}
+    assert set(response.context_data["object_list"]) == {app2, app1}
```
tomodwyer, line 467: Managed to get the test working with this code, but no idea why it worked 😅

Issue #3464, opened 2023-08-25
Remove the backend warning banner


lucyb:
See 🧵 [here](https://bennettoxford.slack.com/archives/C03ELD5FEAU/p1692979011797009) and [here](https://bennettoxford.slack.com/archives/C03ELD5FEAU/p1692962353584169?thread_ts=1692961827.384149&cid=C03ELD5FEAU).

I believe that we have good enough alerts in place to be able to remove the general warning banner ([here](https://github.com/opensafely-core/job-server/blob/2e5f1de43b4b7175c39d6e5aff8e82b5e0ca7c10/jobserver/context_processors.py#L24)). 

However, as part of this change we should also add the job specific banner to the job request page, so that it's more likely to be seen and to be consistent with the information we're providing users, since that page would also be out of date.

As a bonus, this removes the `if debug` statement that causes prod to act differently to development.

PR #3474, opened 2023-08-29
Remove site-wide banner when backends are missing

ghickman:
Fix: #3464


jobserver/views/job_requests.py
```diff
@@ -215,9 +221,24 @@ def get(self, request, *args, **kwargs):
             )
         )
 
+        # build up is_missing_updates to define if we've not seen the backend
```
lucyb, line 224: Is it worth moving the logic to calculate `is_missing_updates` to a method on the model?

jobserver/views/job_requests.py
```diff
@@ -215,9 +221,24 @@ def get(self, request, *args, **kwargs):
...
+        # running this JobRequest for a while.
+        # it's completed, we don't expect updates now
+        incomplete = not job_request.is_completed
+
+        # was the last update more than our threshold ago?  The last_updated_at
+        # annotation uses a default of `now` so we don't have to deal with None
+        # here.
+        delta = timezone.now() - job_request.last_updated_at
+        threshold = timedelta(minutes=30)
```
lucyb, line 233: We've now used "30" in both Job Requests and Jobs. Might be worth extracting into a constant in core.py as `LAST_UPDATE_THRESHOLD` is similar?

tests/unit/jobserver/views/test_job_requests.py
```diff
@@ -666,6 +667,7 @@ def test_jobrequestdetail_with_permission(rf, django_assert_num_queries):
 
     assert response.status_code == 200
     assert not response.context_data["is_invalid"]
+    assert response.context_data["is_missing_updates"]
```
lucyb, line 670: If I understand this correctly, we don't have a test for a job that's currently running on a backend that's been contacted recently (i.e not missing). Might be worth adding, just to check all the logic.

lucyb (approved):
Thanks for doing this, it looks good. I've made some suggestions for improvements, I hope they're useful.

PR #3478, opened 2023-08-29
Display only active backends

lucyb:
We now have quite a few backends that are for testing, staging or that are no longer used. We don't need to display these to users looking for the current status of a backend that's actively being used.


tests/unit/jobserver/views/test_status.py
```diff
@@ -172,3 +172,13 @@ def test_status_counts_all_pending_jobs(rf):
...
+
+
+def test_status_shows_only_active_backends(rf):
+    BackendFactory.create_batch(2, is_active=True)
+    BackendFactory(is_active=False)
+
+    request = rf.get("/")
+    response = Status.as_view()(request)
+
+    assert len(response.context_data["backends"]) == 2
```
ghickman, line 184: I've just added `set_from_list` to `jobserver/utils.py` as part of pulling querying out of rendering.  So you can improve this assertion to check the backends are the ones you expected.

Assign the output of `create_batch()` to a variable called `backends`, then this line can become:

```
assert set_from_qs(response.context_data["backends"]) == set_from_list(backends)
```

ghickman (approved):

PR #3489, opened 2023-08-31
Remove new release flow

lucyb:


jobserver/migrations/0006_move_to_new_release_flow.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 4.2.4 on 2023-08-31 16:13
+
+from django.db import migrations
+
+
+def move_workspaces_to_new_release_flow(apps, schema_editor):
+    Workspace = apps.get_model("jobserver", "Workspace")
+    Workspace.objects.all().update(uses_new_release_flow=True)
```
ghickman, line 8: Not worth changing for this but you don't need the `.all()` call here, it's a noop :)

jobserver/migrations/0006_move_to_new_release_flow.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("jobserver", "0005_add_job_run_command"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            move_workspaces_to_new_release_flow,
```
ghickman, line 19: Without a `reverse_code` entry this migration isn't reversible, however `migrations.RunPython.noop` should be fine here as there's no meaningful choice for the reverse.

ghickman (changes requested):

jobserver/migrations/0006_move_to_new_release_flow.py
lucyb: Good point. Thank you.

jobserver/migrations/0006_move_to_new_release_flow.py
lucyb: Good to know for the future!

ghickman (approved):

PR #3524, opened 2023-09-06
test out a larger GitHub runner

ghickman:


madwort:
the slow github actions (`test` and `docker test`) ran in ~5m, compared to ~6-7m for other PRs around the same time. Not a huge speed-up, but I don't see why not take it.

madwort (approved):

PR #3529, opened 2023-09-06
Abstract order user by name

ghickman:
Found while convering Project.org to an M2M.  I suspect there are more places but didn't want to look too hard while working on the M2M transition.

Ref: #2827


jobserver/models/core.py
```diff
@@ -767,6 +769,32 @@ def human_memorable_token(size=8):
...
 
+class UserQuerySet(models.QuerySet):
+    def order_by_name(self):
+        """
+        Order Users by their "name"
+
+        we don't have fullname populated for all users yet and having some
+        users at the top of the list with just usernames looks fairly odd.
+        We've modelled our text fields in job-server such that they're not
+        nullable because we treat empty string as they only empty case.  The
```
madwort, line 780:
```suggestion
        nullable because we treat empty string as the only empty case.  The
```

madwort (changes requested):
looks good, one typo though

madwort (approved):

PR #3535, opened 2023-09-08
Move `workspace.db` API field to `database_name`

evansd:
The fact that this attribute lives on `workspace` is an internal Job Server implementation detail that we don't want to expose in the API it presents to Job Runner (specifically because we want to change it shortly).

The corresponding PR in Job Runner, which must be deployed first, is:
 * https://github.com/opensafely-core/job-runner/pull/650


jobserver/models/core.py
```diff
@@ -365,6 +365,10 @@ def status(self):
 
         return "unknown"
 
+    @property
+    def database_name(self):
+        return self.workspace.db
```
ghickman, line 370: Could you add a comment to explain why this property exists, given it's not obviously when you're looking at this file in isolation!

Alternatively, did you look at making this a field directly on `JobRequestAPIList.serializer_class`?  You can tell DRF how to source a field when you define one manually.  That would keep the implementation all on the serializer then.

ghickman (approved):

jobserver/models/core.py
evansd: Ah yes, good point. This is pretty temporary code (I'm working on a separate PR to remove `workspace.db` altogether but I should have made that clear in the comments.

ghickman: Perfect.  We've been undiligently using TODO/FIXME to mark temporary code, but anything grep-able that's consistent across the features/transitional code would be ideal if you wouldn't mind? 🙂

Issue #3419, opened 2023-08-18
Trial the use of larger runners for Job Server Actions


lucyb:
Job Server CI is currently taking almost 10 minutes to complete. As this needs to be run on a PR and on merge to main, it means we have a nearly 20 min minimum time to deploy and we'd like to make that faster.

GitHub have enabled [larger runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners) for the opensafely-core organisation. We could use these to see if it improves the time it takes the GitHub Actions to run and therefore speed up the time to deployment.

📔 This comes with a financial cost, so we should carefully monitor the impact of any change on our monthly bills as part of this ticket. If it's a success, then we might be able to ask GitHub to waive the cost.

See also: #3297

ghickman:
I've added a single larger runner with 4 cores, added it to a group called "Default larger runners", and configured the job-server workflow to use it (the runner, not the group) in #3524.  As @madwort pointed out in #3524 we're shaving ~1m off the test time with this larger runner which isn't stellar but it's definitely meaningful progress, and with a deploy represents 2m.

lucyb:
That's something at least. Does this mean we can use xdist to run our tests in parallel now that we have multiple cores?

ghickman:
Definitely something to test out 👍

lucyb:
It does look like the tests are now running 1-2 minutes faster than a couple of weeks ago. We've spent $25 so far and I expect this to be around $50 by the end of the month. However, we could potentially move down to a two-core runner, which would be much cheaper. It would be worth testing to see if that makes any significant difference.

I think the key thing is how the two-core and four-core runners perform with running the tests in parallel (#3545). So I'm moving this ticket into blocked until we've tested that.

ghickman:
Smallest larger runner they do is 4 cores: https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#machine-sizes-for-larger-runners

lucyb:
In an interesting twist, GitHub have made 4-CPU runners available for free for opensource projects
https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

I'll have a look into it and see if we need to take any action and whether the free version is the same as the paid for ones.

lucyb:
This was implemented in https://github.com/opensafely-core/job-server/pull/4004

[...]

PR #3545, opened 2023-09-12
Run tests in parallel in CI

ghickman:
Ref: #3419


bloodearnest:
This was done in #3735

PR #3553, opened 2023-09-12
Retry deleting repos with backoff and jitter

ghickman:
We're still seeing intermittent errors from GitHub when trying to delete a repo.  Instead of a single fixed pause before retry the stamina library lets us do multiple attempts with an exponential backoff and jitter.  This uses the defaults with the expectation that we'll modify them if we still see problems.


madwort (approved):
yes please

PR #3554, opened 2023-09-13
Update interactive templates to latest version

bloodearnest:
Update interative-templates to v2023.09.12.100407


ghickman:
A test that went wrong.

PR #3558, opened 2023-09-13
Set `database_name` based on project permissions

evansd:
We've taken the deliberate decision here to use a hardcoded list of project numbers committed to the codebase, rather than storing a new set of permissions in the database. In the short-term we don't anticipate needing to add to this list. We can revist the implementation as and when that proves necessary.

As well as simplicity this approach has the advantage of creating a natural public audit trail in the form of the commit history, where changes to the database would be opaque.

### DIFF
--- a/tests/unit/jobserver/models/test_job_request.py
+++ b/tests/unit/jobserver/models/test_job_request.py
@@ -7,7 +7,13 @@ from django.utils import timezone
 
 from jobserver.models import JobRequest
 
-from ....factories import JobFactory, JobRequestFactory, RepoFactory, WorkspaceFactory
+from ....factories import (
+    JobFactory,
+    JobRequestFactory,
+    ProjectFactory,
+    RepoFactory,
+    WorkspaceFactory,
+)
 from ....utils import minutes_ago, seconds_ago
 
 
@@ -378,3 +384,29 @@ def test_jobrequestmanager_with_started_at():
     JobRequestFactory(workspace=workspace)
 
     assert JobRequest.objects.with_started_at().filter(workspace=workspace).count() == 2
+
+
+def test_jobrequest_database_name_with_no_project_number(build_job_request):
+    job_request = build_job_request(project_number=None)
+    assert job_request.database_name == "default"
+
+
+def test_jobrequest_database_name_without_t1oo_permission(build_job_request):
+    job_request = build_job_request(project_number=1)
+    assert job_request.database_name == "default"
+
+
+def test_jobrequest_database_name_with_t1oo_permission(build_job_request):
+    # This is one of the project numbers hardcoded into `permissions/t1oo.py`
+    job_request = build_job_request(project_number=2)
+    assert job_request.database_name == "include_t1oo"
+
+
+@pytest.fixture
+def build_job_request():
+    def _build_job_request(project_number):
+        project = ProjectFactory(number=project_number)
+        workspace = WorkspaceFactory(project=project)
+        return JobRequestFactory(workspace=workspace)
+
+    return _build_job_request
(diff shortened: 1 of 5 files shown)


tests/unit/jobserver/models/test_job_request.py
```diff
@@ -378,3 +384,25 @@ def test_jobrequestmanager_with_started_at():
...
+def test_jobrequest_database_name_with_t1oo_permission():
+    # This is one of the project numbers hardcoded into `t1oo_permissions.py`
+    job_request = _make_job_request(project_number=2)
+    assert job_request.database_name == "include_t1oo"
+
+
+def _make_job_request(project_number):
+    project = ProjectFactory(number=project_number)
+    workspace = WorkspaceFactory(project=project)
+    return JobRequestFactory(workspace=workspace)
```
ghickman, line 408: Could you wrap this function in a fixture?  That's how we expose combinations of factories (as per our [testing docs](https://github.com/opensafely-core/job-server/blob/main/TESTING.md#tooling)) and how we expose these builder functions as well.  There's some prior art in there, for example `build_release()`.

ghickman (approved):
Looks great!  One small bit of code org feedback 🙂

tests/unit/jobserver/models/test_job_request.py
evansd: Thanks. It would need to be fixture which returns a function though, wouldn't it? Otherwise I can't pass the `project_number` in. Or am I missing something?

ghickman: Yep!  That's what `build_release()` does too :)

evansd: OK, fine. It just looked like a pointless extra level of indirection to me:
```py
def test_jobrequest_database_name_without_t1oo_permission(build_job_request):
    job_request = build_job_request(project_number=1)
    assert job_request.database_name == "default"


@pytest.fixture
def build_job_request():
    def _build_job_request(project_number):
        project = ProjectFactory(number=project_number)
        workspace = WorkspaceFactory(project=project)
        return JobRequestFactory(workspace=workspace)
    return _build_job_request
```

vs

```py
def test_jobrequest_database_name_without_t1oo_permission():
    job_request = build_job_request(project_number=1)
    assert job_request.database_name == "default"

def build_job_request(project_number):
    project = ProjectFactory(number=project_number)
    workspace = WorkspaceFactory(project=project)
    return JobRequestFactory(workspace=workspace)
```

But if it's more idiomatic I'm very happy to keep things consistent with what's done elsewhere.

jobserver/t1oo_permissions.py
```diff
@@ -0,0 +1,110 @@
...
+    152,  # https://jobs.opensafely.org/digital-access-to-primary-care-for-older-people-during-covid/
+    153,  # https://jobs.opensafely.org/what-was-the-impact-of-covid-19-on-mortality-related-to-venous-thromboembolism-in-england/
+    154,  # https://jobs.opensafely.org/validation-of-isaric-sus-phosp-data-for-covid-related-hospital-admissions/
+    155,  # https://jobs.opensafely.org/the-effect-of-long-covid-on-quality-adjusted-life-years-using-openprompt/
+}
+
+
+def project_is_permitted_to_use_t1oo_data(project):
+    if not project.number:
+        return False
```
inglesp, line 109: Note that the internal project should have access to T1OO data, but doesn't have a project number ([Slack](https://bennettoxford.slack.com/archives/C03ELD5FEAU/p1698846374848329)) so will need to be checked for here.

inglesp: I'm wrong.

inglesp (approved):
Have checked list against DPIA :tick:

PR #3560, opened 2023-09-14
Staff area fixes

tomodwyer:


templates/staff/project/list.html
```diff
@@ -66,22 +66,26 @@
...
 
     {% #card %}
-      {% #list_group small=True %}
-        {% for project in object_list %}
-          {% #list_group_item type="Projects" href=project.get_staff_url %}
-            {{ project.title }}
-          {% /list_group_item %}
-        {% endfor %}
-      {% /list_group %}
+      {% if object_list|length > 0 %}
```
ghickman, line 69: You shouldn't need the `|length > 0` part here, empty lists are falsey in Python!

ghickman (approved):

templates/staff/project/list.html
tomodwyer: Good catch, I was trying to debug why it wasn't showing, turns out I was editing the wrong template!

Issue #3406, opened 2023-08-16
Improvements for logs view


LFISHER7:
I have just been using the jobs log view (screenshot below) to try and find the runtime for an action i've previously run, but found it frustrating.  This used to be quite easy, with collapsible boxes for all of the jobs within a job request. Now I have to go to another page to see the jobs within a job request. To see the runtime for an individual job I have to go one page further.

I could use the search box, but:

- I can't always remember the exact name of the action I'm looking for - it would be useful to have suggested actions from the workspace as you type.
- After I search, i don't know if i'm capturing the actions I'm looking for until i've clicked through to another page
- There's nothing in the UI to show the search term or remove it if it's wrong.

![jobs_logs](https://github.com/opensafely-core/job-server/assets/37254732/03f30a0d-f702-41a5-9ece-408a01a9cecd)

What I think would make this easiest, is if there was a view that showed all the current actions, as on the run jobs page, where I could click through to the logs for individual actions. Perhaps this could go on the run jobs page (below)

![jobs](https://github.com/opensafely-core/job-server/assets/37254732/f9b746b9-44aa-4dfd-b1d3-ebc5115e43c6)

madwort:
@tomodwyer to split up this ticket

tomodwyer:
- I can't always remember the exact name of the action I'm looking for - it would be useful to have suggested actions from the workspace as you type.
- After I search, i don't know if i'm capturing the actions I'm looking for until i've clicked through to another page
- There's nothing in the UI to show the search term or remove it if it's wrong.
  - [ ] #3561 
- What I think would make this easiest, is if there was a view that showed all the current actions, as on the run jobs page, where I could click through to the logs for individual actions. Perhaps this could go on the run jobs page (below)

tomodwyer:
@LFISHER7 I've split this in to a few tickets above, if you wish to subscribe to them for updates on when these issues will be resolved.

PR #3564, opened 2023-09-14
Various workspace log fixes

tomodwyer:
Ref: #3406

- Allow searching the workspace logs by request ID
  - Not sure if users use this, but it seems like missing functionality
- Full width container to match the other event log pages
- View button seems superfluous when users could click the request ID
- Using `details` to show all job actions
  - From a UX perspective it's not beautiful, but it requires no JS or additional SQL queries and helps until we do the HTMX solution


ghickman (approved):
👌 

One Weird Edgecase™, requests have no jobs until they're acknowledged by the backend, so we can get cases like this:

PR #3578, opened 2023-09-18
SPIKE: Grafana

lucyb:
**Don't merge**

This is an experiment to see if Grafana can be used to easily get useful information from the Job Server database. This uses the default, official Grafana docker image, using docker-compose to provide some configuration.

To get up and running, do:

`cd docker && docker-compose up grafana`

Then access grafana at http://localhost:3000/.

You should have a datasource configured and be able to use that to create a dashboard using the `dashboard-example.json` file.

Screenshot from 2023-09-18 16-02-23

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -12,6 +12,22 @@ services:
     volumes:
       - postgres_data:/var/lib/postgresql/data/
 
+  grafana:
+    image: "grafana/grafana:latest"
+    environment:
+      GF_DATABASE_TYPE: postgres
+      GF_DATABASE_HOST: db:5432
+      GF_DATABASE_NAME: jobserver
+      GF_DATABASE_USER: user
+      GF_DATABASE_PASSWORD: pass
+      GF_DATABASE_SSL_MODE: disable
+    depends_on:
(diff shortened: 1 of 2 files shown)


docker/docker-compose.yaml
```diff
@@ -12,6 +12,22 @@ services:
     volumes:
       - postgres_data:/var/lib/postgresql/data/
 
+  grafana:
+    image: "grafana/grafana:latest"
+    environment:
+      GF_DATABASE_TYPE: postgres
+      GF_DATABASE_HOST: db:5432
+      GF_DATABASE_NAME: jobserver
```
madwort, line 20: this is being used to store grafana's own data - should it be `grafana`?

madwort, line 20: this is being used to store grafana's own data - should we have created another service e.g. `db-grafana` and use db_name `grafana` here?

lucyb: (For anyone reading this in future) Yep, you're absolutely right, I should have done that or let it default to a sqlite db! The env vars to configure the postgres datasource are very similar to this and I think I copied the wrong ones.

Issue #3555, opened 2023-09-13
Limit release file size in API


bloodearnest:
We should limit the uploaded file size the release API.

This probably should be limited in both the release request creation (returning a 400 error) *and* also the upload api (returning 413 - possibly implemented in nginx?)

A sensible starting default might be 16mb, and then see if there are any complaints with that. We should look at existing released file size to check.

Care needs to be taken to communicate this change in output-checking process.

ghickman:
We have released 10409 files so far.

Which are distributed like so:

| Size range | Count |
| --- | --- |
| <1Mb | 9812 |
| >=1<2 | 309 |
| >=2<3 | 77 |
| >=3<4 | 19 |
| >=4<5 | 16 |
| >=5<6 | 14 |
| >=6<7 | 12 |
| >=7<8 | 17 |
| >=8<9 | 10 |
| >=9<10 | 5 |
| >=10<11 | 4 |
| >=11<12 | 0 |
| >=12<13 | 2 |
| >=13<14 | 4 |
| >=14<15 | 3 |
| >=15<16 | 3 |
| >=16<17 | 2 |
| >=16Mb | 19 |

The files over 16 Mb are a bit of a mixed bag:

```
88.27Mb
49.54Mb
39.73Mb
29.12Mb
25.21Mb
24.0Mb
23.97Mb
23.94Mb
23.92Mb
23.28Mb
22.78Mb
20.04Mb
19.72Mb
17.91Mb
17.88Mb
17.62Mb
17.41Mb
16.91Mb
16.89Mb
```

Based on these numbers 16Mb seems reasonable.

PR #3579, opened 2023-09-19
Limit release file sizes in the API

ghickman:
This limits the size of files both when a release is created and when the actual file is uploaded.

The limit is configurable in settings.  I've wrapped it such that we can set an env var but I don't plan to use that in production by default, it's there so we can temporarily raise the limit as an escape hatch.

Fix: #3555


jobserver/settings.py
```diff
@@ -349,7 +349,10 @@
 # interactive functionality
 LOCAL_GIT_REPOS = BASE_DIR / "repos"
 
-# Releases storage location.
+# Released files per-file size limit
+RELEASE_FILE_SIZE_LIMIT = env.int("RELEASE_FILE_SIZE_LIMIT", default=16777216)  # 16Mb
```
bloodearnest, line 353:
```suggestion
RELEASE_FILE_SIZE_LIMIT = env.int("RELEASE_FILE_SIZE_LIMIT", default=16*1024*1024)  # 16Mb
```

Encoding it like this makes it easy to adjust later?

bloodearnest (approved):

jobserver/settings.py
ghickman: ooh yeah, fewer magic numbers 👍

PR #3590, opened 2023-09-22
Restyle staff area redirect templates

tomodwyer:


templates/staff/redirect/detail.html
```diff
@@ -0,0 +1,78 @@
+{% extends "staff/base-tw.html" %}
+
+{% load humanize %}
+{% load static %}
```
ghickman, line 4: I don't think we need this tag any more?

ghickman (approved):

PR #3591, opened 2023-09-22
Reduce number of queries in staff area

lucyb:
We recently added a call to `user.all_roles` in the template, which increased the number of queries from 7 to 57. This change reduces the number of queries down to 9. The same approach has been applied to a couple of other views that had a similar issue.


ghickman (approved):
Brilliant query detective work 👍

You can ditch the `model` variables now that we're using `queryset`.  `model` is just a shorthand for the view to call the default manager on it and setting `queryset` overrides that.

PR #3594, opened 2023-09-25
Use Q objects to build up a [does not] have roles query

ghickman:
This lets us have one construction that we either filter or don't. Django's .filter() and .exclude() have different semantics for how they produce AND/OR queries.

.filter() uses AND regardless of multiple kwargs or chaining, while .exclude() produces OR statements with multiple kwargs and AND statements when chaining, making using these methods in combination confusing to the reader.

Since we know we want the same query as either a filter or an exclude this change builds up the query with Q objects (so we can explicitly control the AND/OR statements generated) and then calls the appropriate QuerySet method based on the input value.

This also removes the __exact lookup because it doesn't change the resulting SQL.  It's unclear if this is because we're querying an array field or because Postgres has optimised away the need for using IS.

### DIFF
--- a/staff/views/users.py
+++ b/staff/views/users.py
@@ -3,7 +3,7 @@ import inspect
 import structlog
 from django.core.exceptions import FieldError
 from django.db import transaction
-from django.db.models import Exists, OuterRef
+from django.db.models import Exists, OuterRef, Q
 from django.db.models.functions import Lower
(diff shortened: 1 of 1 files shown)


lucyb (approved):
Thanks for the tidy-up. Using Q objects does make that easier to parse.

PR #3618, opened 2023-10-03
Restyle staff area researcher edit template

tomodwyer:


templates/staff/application/researcher_edit.html
```diff
@@ -0,0 +1,107 @@
...
+      {% form_input field=form.name label="Researcher name" name="name" %}
+      {% form_input field=form.job_title label="Researcher job title" name="job_title" %}
+      {% form_input field=form.email label="Researcher work email address" name="email" type="email" %}
+      {% form_input field=form.telephone label="Mobile phone number" name="telephone" %}
+      {% form_input field=form.github_username label="GitHub username" name="github_username" %}
+      {% form_input field=form.daa label="DAA" name="daa" %}
+    {% /form_fieldset %}
+  {% /card %}
+
+  {% #card title="Secure results server" class="max-w-3xl" container=True %}
```
lucyb, line 45: If there's a validation error somewhere on the page when I press submit, then the page refreshes and loses this value.

I can then submit without answering this question - I'm not sure if that's intentional or not.

templates/staff/application/researcher_edit.html
```diff
@@ -0,0 +1,107 @@
...
+        {% endwith %}
+      {% endfor %}
+    {% /form_fieldset %}
+  {% /card %}
+
+  {% #card title="Safe researcher training" class="max-w-3xl" container=True container_class="grid gap-y-6 max-w-prose" %}
+    {% #form_fieldset class="grid gap-y-2 !min-w-0 w-full max-w-prose" %}
+      {% form_legend text="Has this researcher undertaken safe researcher training?" %}
+
+      {% for value, label in form.has_taken_safe_researcher_training.field.choices %}
```
lucyb, line 87: If there's a validation error somewhere on the page when I press submit, then the page refreshes and loses this value.

I can then submit without answering this question - I'm not sure if that's intentional or not?

templates/staff/application/researcher_edit.html
tomodwyer: Huh, it seems like when the if statement is `False == False` the `isChecked` variable is not being set.

tomodwyer:
@lucyb I've fixed that issue my string-ifying the values, not sure if there's a better way to fix the issue?

lucyb:
> @lucyb I've fixed that issue my string-ifying the values, not sure if there's a better way to fix the issue?

Hmm, I'm not sure. There probably is, but it would require some more digging and I'm not sure it's worth it.

I've noticed that the applications edit form has one less field than the staff area edit form. Is that intentional?

Screenshot from 2023-10-06 17-01-23
Screenshot from 2023-10-06 17-01-12

tomodwyer:
@lucyb I believe that is because the DAA is added in the staff area by the IG Team, rather than by users when filling in the form.

lucyb (approved):

PR #3624, opened 2023-10-06
Select org for job requests in SQL queries

lucyb:
This removes 10 additional queries and reduces page load by 100ms for both the authenticated and unauthenticated versions of the index page.


ghickman (approved):
🚀

PR #3626, opened 2023-10-06
Restyle various staff templates

tomodwyer:
Restyle various remaining staff area templates, leaving only Dashboard tables to be restyled.

~Requires a rebase once #3618 is merged.~

[Diff of restyling without file moves commits](https://github.com/opensafely-core/job-server/pull/3626/files/d9f5ccb9529192143ea9e1aad56b8c3617decc3b)


templates/staff/project/create.html
```diff
@@ -0,0 +1,45 @@
...
+        {% #alert variant="danger" class="mb-6" %}
+          {{ error }}
+        {% /alert %}
+      {% endfor %}
+    {% endif %}
+
+    {% #card title="Create an Interactive Project" class="max-w-3xl" container=True %}
+      <div class="flex flex-col items-stretch gap-y-6 w-full max-w-3xl">
+        {% form_select field=form.fields.org choices=form.fields.org.choices selected=form.org.value id="id_org" name="org" label="Select an org" %}
+        {% form_select field=form.fields.copilot choices=form.fields.copilot.choices selected=form.copilot.value id="id_copilot" name="copilot" label="Select a user to co-pilot this project" %}
```
ghickman, line 34: These fields used to used select2, how difficult would it be to keep that for this page?  The org selector is fine now, but the users selector is pretty long already!

templates/staff/project/create.htmx.html
ghickman, line 1: I can't comment on the lines directly… the order of the Project name and Application URL in this form are reversed compared to the non-HTMX form, was that intended?

ghickman (approved):

templates/staff/project/create.htmx.html
tomodwyer: Nope, most likely just copying what was in the old template, I'll fix that up.

templates/staff/project/create.html
tomodwyer: I'm going to add this in a separate PR, as it requires a few changes to the multiselect component (maybe a new component).

Issue #3617, opened 2023-10-03
Python 3.12


ghickman:
Python 3.12 is out.  This ticket tracks blockers for us switching to it.

Known blockers:
* [x] psycopg-binary hasn't shipped 3.12 wheels yet
* [x] GHA hasn't shipped 3.12 support yet
* [x] otel deps use deprecated utcnow/utcfromtimestamp
* [x] freezegun [doesn't look interested in future maintenance](https://github.com/spulec/freezegun/pull/511#issuecomment-1748219731)

PR #3629, opened 2023-10-09
Bump to python 3.12

ghickman:
Fix: #3617


iaindillingham (approved):
I also found references to 3.11 in `runtime.txt` and `DEVELOPERS.md`. I'd be interested to know what uses the former.

ghickman:
@iaindillingham – thanks for catching those!  `runtime.txt` was an attempt to tell dependabot which version of python to run updates with, since `.python-version` is so tied into pyenv, it doesn't work, so I've removed it.

PR #3643, opened 2023-10-10
Staff area list page improvements

ghickman:
Catherine noticed we had some duplicate users on the user list page and what looked like a simple fix turned into a mildly-sisyphean task.

Duplicate users was because we were missing a `distinct()` call, an easy fix! The other bug will be just as easy! Dear reader, let me assure you, it was not.

Note: I checked for and fixed the other staff area list pages missing the `distinct()` call.

The "has any roles: y/n" filter was giving some rather confusing results, some users were being included when filtering on `no` when they clealry had roles.  After more time than I'd like to admit, and a full repro with raw SQL, I tracked this down to users who had any membership with `roles=[]`.  So even when they had other memberships of the same type with roles they were still getting included.

### DIFF
--- a/staff/views/orgs.py
+++ b/staff/views/orgs.py
@@ -178,7 +178,7 @@ class OrgList(ListView):
             ]
             qs = qs.filter(qwargs(fields, q))
 
-        return qs
+        return qs.distinct()
 
 
 @method_decorator(require_role(CoreDeveloper), name="dispatch")
--- a/tests/unit/staff/views/test_users.py
+++ b/tests/unit/staff/views/test_users.py
@@ -485,7 +485,7 @@ def test_userlist_filter_by_invalid_role(rf, core_developer):
 
 def test_userlist_filter_by_any_roles_yes_includes_global_roles(rf, core_developer):
     UserFactory(roles=[OutputPublisher])
-    UserFactory(roles=[])
+    UserFactory()
 
     request = rf.get("/?any_roles=yes")
     request.user = core_developer
@@ -498,8 +498,9 @@ def test_userlist_filter_by_any_roles_yes_includes_global_roles(rf, core_develop
 
 
 def test_userlist_filter_by_any_roles_yes_includes_project(rf, core_developer):
-    UserFactory(roles=[])
-    user_with_project = UserFactory(roles=[])
+    UserFactory()
+
+    user_with_project = UserFactory()
     ProjectMembershipFactory(user=user_with_project, roles=[ProjectDeveloper])
 
     request = rf.get("/?any_roles=yes")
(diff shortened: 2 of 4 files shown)


tests/unit/staff/views/test_users.py
```diff
@@ -484,102 +484,119 @@ def test_userlist_filter_by_invalid_role(rf, core_developer):
...
 
     request = rf.get("/?any_roles=yes")
     request.user = core_developer
 
     response = UserList.as_view()(request)
 
-    assert (
-        len(response.context_data["object_list"]) == 2
-    )  # this includes the core_developer
+    assert set_from_qs(response.context_data["object_list"]) == {
```
lucyb, line 495: Thank you for converting these to use `set_from_qs` it's much easier to see what's going on now, especially with the extra assertions in later tests.

tests/unit/staff/views/test_users.py
```diff
@@ -560,6 +560,12 @@ def test_userlist_filter_by_any_roles_no_excludes_project_roles(rf, core_develop
     user_with_project_and_no_roles = UserFactory()
     ProjectMembershipFactory(user=user_with_project_and_no_roles)
 
+    user_with_mixture_of_project_roles = UserFactory()
```
lucyb, line 563: Can you add the similar for the `test_userlist_filter_by_any_roles_no_excludes_org_roles` and create a `user_with_mixture_of_org_roles`? This should check the change you've made to count the number of org roles.

staff/views/orgs.py
```diff
@@ -178,7 +178,7 @@ def get_queryset(self):
             ]
             qs = qs.filter(qwargs(fields, q))
 
-        return qs
```
lucyb, line 181: If it's fairly straightforward to do it would be good to have tests for these `qs.distinct` changes, as it would be good to avoid doing it again in future.

lucyb (commented):
Excellent detective work! :detective: 

Sorry for introducing the queryset bug. It's an annoyingly complex piece of logic, but I do think you've managed to write some nice, clear code for the query.

staff/views/orgs.py
ghickman: I would love to have this but I couldn't find a nice solution.  There's several ways that each list view can produce duplicates so then we'd need to either have a test that ran through all the use cases (most of which we're already testing), or add extra objects to the existing tests.  Neither of which are _terrible_, but they both feel really prone to a developer forgetting to add them, which is exactly how we ended up here!

I'll keep it in mind and see if anything stands out as a nice, obvious fix to this, because it would definitely be desireable to have 🤔

ghickman:
> Sorry for introducing the queryset bug

Happens to all of us and I didn't notice it when converting to Q objects!  It only captured a tiny handful of users unexpectedly so it was really tricky to find, and it was surprising to me that the generated SQL performed as it did.  I count the TIL as the most useful outcome of this.

lucyb (approved):

PR #3653, opened 2023-10-13
Switch freezegun for time machine

ghickman:
Fix: #3649


tests/unit/services/test_logging.py
```diff
@@ -3,11 +3,14 @@
 from services.logging import timestamper
 
 
-def test_timestamper_with_debug(monkeypatch, freezer):
+def test_timestamper_with_debug(monkeypatch, time_machine):
     monkeypatch.setattr("services.logging.DEBUG", True)
 
+    dt = datetime.now()
+    time_machine.move_to(dt, tick=False)
```
madwort, line 10: is there a reason you used `dt` here, rather than `now` as in other tests?

```suggestion
    now = datetime.now()
    time_machine.move_to(now, tick=False)
```

tests/unit/services/test_logging.py
```diff
@@ -3,11 +3,14 @@
...
+
     log = timestamper(None, None, {"event": "derp"})
-    assert log == {"event": "derp", "timestamp": datetime.now().isoformat() + "Z"}
+    assert log == {"event": "derp", "timestamp": dt.isoformat() + "Z"}
```
madwort, line 13:
```suggestion
    assert log == {"event": "derp", "timestamp": now.isoformat() + "Z"}
```

madwort (approved):
LGTM, one nit you can take or leave

tests/unit/services/test_logging.py
ghickman: Genuinely can't remember! Consistency ftw 👍

PR #3714, opened 2023-10-23
Add sortable tables to staff dashboards

tomodwyer:
Use `simple-datatables` to build the staff area dashboard tables.


ghickman (approved):
This looks so good!  I've poked at each of the boards and nothing looks out of place 👍

Issue #3723, opened 2023-10-24
Check and fail jobs if codelists are stale


rebkwok:
Part of the plan to [address acute codelist rot in ehrQL]https://github.com/opensafely-core/ehrql/issues/1581.

Proposed plan:
- Add an endpoint on OpenCodelists to check a study repo's codelists (https://github.com/opensafely-core/opencodelists/issues/1702)
- Have job-server call the endpoint
    - On EVERY job request creation
    - If the job request explicitly runs an ehrQL/cohort-extractor action, fail (by form validation error) if codelists are stale. This should involve minimal change to the job-server UX 
    - Otherwise, allow the job request to be created, and add a flag in the job request data with the result of the codelist check
    - At the point where the job's actions have been compiled (by job-runner), the job/job request should fail if it includes an ehrql/cohort-extractor job and the codelist check failed). This means there will be some lag before the job fails if a cohort-extractor/ehrql action is a dependency, rather than an explictly requested action, but it means that the check only has to be done in one place.

PR #3738, opened 2023-10-30
Check for stale codelists on job request creation

rebkwok:
Part of the [plan for dealing with acute codelist rot](https://github.com/orgs/opensafely-core/projects/13/views/1?pane=issue&itemId=42536167) - the first bit of #3723 

OpenCodelists now has an endpoint that takes the content of a study repo's codelists.txt and codelists.json, and checks that the data that would be downloaded for the specified codelists is the same as the sha in the codelists.json file. This is specifically intended to deal with changing VMPs in dm+d codelists, in case new mappings have been introduced since the codelist was downloaded into the study repo.

The plan is for job-server to prevent jobs being created if codelists are stale, but only for those that involve the database (and codelists), i.e. generate-cohort or generate-ehrql actions.  As a first step, this PR adds the mechanism for checking for stale codelists, and adds a warning message to the user.  It doesn't yet actually prevent a job request from being created.


jobserver/github.py
```diff
@@ -353,13 +353,14 @@ def get_tag_sha(self, org, repo, tag):
 
         return r.json()["object"]["sha"]
 
-    def get_file(self, org, repo, branch):
+    def get_file(self, org, repo, branch, filepath=None):
+        filepath = filepath or "project.yaml"
```
ghickman, line 357: Any reason not to use `"project.yaml"` instead of `None` in the kwargs?

jobserver/views/job_requests.py
```diff
@@ -112,6 +117,23 @@ def dispatch(self, request, *args, **kwargs):
...
 
         self.actions = list(get_actions(data))
+
+        # Find the status of the codelists in this workspace and branch
+        # Codelist status is either "ok" or "error"
+        self.codelists_status = get_codelists_status(
+            self.workspace.repo.owner,
+            self.workspace.repo.name,
+            ref,
+        )
```
ghickman, line 127: Should this go in the try/except above so we can handle the exceptions `get_codelists_status` can raise?

jobserver/pipeline_config.py
```diff
@@ -32,6 +34,34 @@ def get_project(org, repo, branch, get_github_api=_get_github_api):
...
+        codelists_check = opencodelists_api.check_codelists(*codelists_content)
+        return codelists_check["status"]
+
+    if github_api.get_branch(org, repo, branch) is None:
+        raise Exception(f"Missing branch: '{branch}'")
+
+    if github_api.get_file(org, repo, branch, filepath="codelists") is not None:
+        raise Exception("Could not find codelists.txt or codelists.json")
+
+    return "ok"
```
ghickman, line 61: Is there a known state for getting to this return? If so could you add it in a comment?

tests/unit/jobserver/test_pipeline_config.py
```diff
@@ -149,6 +150,104 @@ def test_get_project_success():
...
+            "main",
+            get_github_api=BrokenGitHubAPI,
+            get_opencodelists_api=BrokenOpenCodelistsAPI,
+        )
+
+
+def test_get_codelists_status_no_codelist_dir():
+    class BrokenGitHubAPI:
+        def __init__(self):
+            self.call_count = 0
```
ghickman, line 231: This looks unused, do we need to keep it around?

ghickman (approved):
Couple of small nits, but this looks great 👍

jobserver/github.py
rebkwok: No, I guess not...I just default to this pattern due to fear of mutable defaults, which "project.yaml" is not

jobserver/pipeline_config.py
rebkwok: Yes, will do.  If there's no codelists directory in the repo, there are no codelists to check, so we don't care that it couldn't fetch codelists.txt and codelists.json from inside that directory.

tests/unit/jobserver/test_pipeline_config.py
rebkwok: No, leftover copy/paste

PR #3755, opened 2023-11-01
Raise validation error on job request creation if codelists are out-of-date

rebkwok:
This is part 2 of #3723 

#3738 made the JobRequestCreateView check for stale codelists and add a warning message.  This PR adds a method to identify jobs that use the backend database (and therefore codelists) and prevent the job request creation at all if database jobs are explicitly requested in the actions.  We pass the names of database actions, and the status of codelists to the JobRequestCreateForm and raise a standard form validation error.

It does not yet address database jobs that are dependencies of requested actions (that'll be part 3)


jobserver/pipeline_config.py
```diff
@@ -46,6 +46,15 @@ def get_actions(config):
         yield {"name": "run_all", "needs": all_actions}
 
 
+def get_database_actions(config):
+    """
+    Return the names of actions that use the backend database.
+    """
+    for name, action in config.actions.items():
+        if action.is_database_action:
+            yield name
```
ghickman, line 55: Probably not worth it for this PR but I wonder if it would be useful for pipeline to also grow `Config.database_actions` that's just a property returning this list 🤔  do we need to know the same list in job-runner?  Could be a useful deck-scrubbing task if so…

tests/unit/jobserver/test_forms.py
```diff
@@ -11,22 +11,90 @@
...
+        ("do_something", "do_something")
+    ]
 
-    assert form.is_valid, form.errors
+    assert form.is_valid(), (form.errors, form.non_field_errors())
 
 
 def test_jobrequestcreateform_with_multiple_backends():
+    for i in range(3):
+        BackendFactory()
```
ghickman, line 38: Factories all have a `create_batch` method that takes a number and the normal kwargs, so you can express this as:

    BackendFactory.create_batch(3)

tests/unit/jobserver/test_forms.py
```diff
@@ -11,22 +11,90 @@
...
 
     assert "backend" in form.fields
     assert form.fields["backend"].choices == choices
+    assert "requested_actions" in form.fields
+    assert form.fields["requested_actions"].choices == [
+        ("do_something", "do_something")
+    ]
 
-    assert form.is_valid, form.errors
+    assert form.is_valid(), (form.errors, form.non_field_errors())
```
ghickman, line 33: Thank you for catching this 😂

ghickman (approved):
Couple of nits, but feel free to ignore them if you don't think they're useful!

Otherwise it looks great 👍

jobserver/pipeline_config.py
rebkwok: I thought about adding that, but at the time I was thinking of individual actions being failed. But yes, I think job-runner will need to know the same list, especially if it's just going to fail the whole job request.  I'll leave it for now though :)

jobserver/views/job_requests.py
```diff
@@ -174,6 +176,8 @@ def get_form_kwargs(self):
         return super().get_form_kwargs() | {
             "actions": [a["name"] for a in self.actions],
             "backends": backends_to_choices(self.backends),
+            "database_actions": self.database_actions,
```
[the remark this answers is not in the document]
rebkwok: Actually, it needs the output of `load_pipeline(self.project)`, which isn't passed around on `self`. I thought it might be able to use `self.actions`, but that's lost the actual action obj that has the `is_database_action` method.  I think I'd need to make `data` in the dispatch `self.data`, which would then mean it has to go into that slightly icky except block

PR #3762, opened 2023-11-01
Restrict access to cohort-extractor for new projects

### LINKED ISSUE #3639: Restrict access to Cohort Extractor for new projects
This is not for immediate action but will becomes necessary when we re-open to new projects. Cohort Extractor's design does not support it enforcing access controls, which it will need to do in a world of more fine-grained data access approvals.

Projects on the [T1OO authorised](https://github.com/opensafely-core/job-server/blob/evansd/set-database-name/jobserver/t1oo_permissions.py) list can continue to use Cohort Extractor. And we may want to allow specific other projects on a case-by-case basis. But the default for new projects should be ehrQL only.

inglesp:
This takes the same approach to identifying projects with access to cohort-extractor as #3558 does for access to T1OO data.

Fixes #3639

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# The IG team need to review any changes files in this directory
+jobserver/permissions/*    @opensafely-core/ig-team
--- a/jobserver/permissions/cohortextractor.py
+++ b/jobserver/permissions/cohortextractor.py
@@ -1,0 +1,114 @@
+# ****************************************************************************
+# ANY ADDITIONS TO THE LIST BELOW MUST BE EXPLICITLY APPROVED BY THE IG TEAM *
+# ****************************************************************************
+
+# Initial list taken from Appendix 3 of DPIA document linked here:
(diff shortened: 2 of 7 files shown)


ghickman (approved):

jobserver/cohortextractor_permissions.py
```diff
@@ -0,0 +1,114 @@
...
+    151,  # https://jobs.opensafely.org/venous-thromboembolism-and-cardiovascular-events-in-inflammatory-rheumatic-diseases/
+    152,  # https://jobs.opensafely.org/digital-access-to-primary-care-for-older-people-during-covid/
+    153,  # https://jobs.opensafely.org/what-was-the-impact-of-covid-19-on-mortality-related-to-venous-thromboembolism-in-england/
+    154,  # https://jobs.opensafely.org/validation-of-isaric-sus-phosp-data-for-covid-related-hospital-admissions/
+    155,  # https://jobs.opensafely.org/the-effect-of-long-covid-on-quality-adjusted-life-years-using-openprompt/
+}
+
+
+def project_is_permitted_to_use_cohortextractor(project):
+    if project.slug == "opensafely-internal":
```
evansd, line 112: I remember George telling me that slug is mutable (and possibly even user-editable) so I wonder if we should `pk` here?

evansd (approved):

PR #3773, opened 2023-11-03
Add link to docs in the out-of-date codelists warning

rebkwok:
Screenshot from 2023-11-03 15-37-43


ghickman (approved):

ghickman:
@tomodwyer – this looks fine to me, but could you just do a quick sense check to make sure we're not misusing something in this lovely new components-based world we have?

PR #3788, opened 2023-11-10
Change text on output viewer app buttons

tomodwyer:
So that it is clearer what the button does.

- "Publish" changed to "Create a draft publication"
- "Confirm Publish?" changed to "Create a public published output"


tomodwyer:
@LFISHER7 are you happy with the language changes here? Happy to modify it if there's something you would prefer.

LFISHER7:
Thanks for picking this up! The first button is good but I wonder if the second would be clearer as "Approve draft publication"?

tomodwyer:
@LFISHER7 I was considering something like that, but was worried the word "draft" might confuse people. I'll get this merged then we can review and see what staff think after the change.

LFISHER7 (approved):

PR #3841, opened 2023-11-28
Make flaky test more specific

lucyb:
This test failed because it was doing a search on the first name, which was very short, and happened to return back more than one result. This changes it to search for the fullname instead, which should only ever return back a single result.

### DIFF
--- a/tests/unit/staff/views/test_job_requests.py
+++ b/tests/unit/staff/views/test_job_requests.py
@@ -169,13 +169,12 @@ def test_jobrequestlist_search_using_fullname(rf, core_developer):
     user = UserFactory(fullname="Ben Goldacre")
     job_request = JobRequestFactory(created_by=user)
 
-    request = rf.get("/?q=ben")
+    request = rf.get("/?q=ben+goldacre")
     request.user = core_developer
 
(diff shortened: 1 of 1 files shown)


tests/unit/staff/views/test_job_requests.py
```diff
@@ -169,13 +169,12 @@ def test_jobrequestlist_search_using_fullname(rf, core_developer):
     user = UserFactory(fullname="Ben Goldacre")
     job_request = JobRequestFactory(created_by=user)
 
-    request = rf.get("/?q=ben")
+    request = rf.get("/?q=ben goldacre")
```
ghickman, line 172:
```suggestion
    request = rf.get("/?q=ben+goldacre")
```

Spaces get converted to a `+` for this query arg 🙂

ghickman (approved):

tests/unit/staff/views/test_job_requests.py
lucyb: Ah, cool. Thank you. I'd assumed they were being converted, but wasn't sure what to and that seemed to work. I've updated it now

PR #3850, opened 2023-12-05
Update Vite and django-vite

tomodwyer:
Both django-vite and the Vite had major version releases. This PR updates both, and renames the `DJANGO_VITE_DEV_MODE` env var to `ASSETS_DEV_MODE`.

Closes #3808


.github/workflows/main.yml
```diff
@@ -70,6 +70,9 @@ jobs:
       - name: Build assets
         run: just assets-build
 
+      - name: Collect assets
+        run: just collectstatic
```
ghickman, line 74: We can flatten these two into the `just assets-install` step, and use `just assets` there, it does install, build, and collectstatic in one place then?

ghickman (approved):

Issue #3770, opened 2023-11-03
Inactive JobRequests are incorrectly considered active by API endpoint


evansd:
The API endpoint here:
https://jobs.opensafely.org/api/v2/job-requests/

Currently includes inactive JobRequests like this one:
https://jobs.opensafely.org/long-term-kidney-outcomes-after-sars-cov-2-infection/long_term_kidney_outcomes_after_covid/20161/

In one sense, this doesn't do any harm; but it does put unnecessary load on our system as job-runner will keep repeatedly syncing the state of the associated jobs back up to job-server. And this problem will only get worse over time.

The issue is that we're relying on the `job.completed_at` timestamp to identify inactive jobs:
https://github.com/opensafely-core/job-server/blob/6102fea9ffb391eafa0c71e40f365cea1bf24bb0/jobserver/api/jobs.py#L288-L290

And for whatever reason (presumably to do with being cancelled rather than hitting an error) these jobs don't have a `completed_at` time set.

It might be more robust to check that the `job.status` is not one of `pending` or `running`. (It would be probably be good to get `completed_at` set for cancelled jobs as well, but it would be better not to rely on that.)

One thing to watch out for in changing the query is that JobRequests with _no associated jobs at all_ (i.e. ones that are not yet acknowledged by job-runner) need to be considered active as well. The current `jobs__completed_at__isnull=True` check obviously manages to achieve this, and the tests should enforce it anyway, but worth being aware of at the start.

PR #3857, opened 2023-12-07
Remove cancelled JobRequests from API

ghickman:
With a db dump from earlier today this takes the number of JobRequests served by the API locally from 100+ to 2.

Fix: #3770


evansd (approved):
This makes me very happy!

Issue #3781, opened 2023-11-07
The run jobs page is slow to respond


iaindillingham:
The run jobs page is slow to respond. This may be where a study has many actions. For example, [long-covid-symptoms/long-covid-symptoms](https://jobs.opensafely.org/long-covid-symptoms/long-covid-symptoms/) has 200 actions.

Thank you for reporting this, @kevwing.

ghickman:
todo:
* get stats from honeycomb about these pages across various workspaces
* check if this is related to the extra checks added for codelist rot
* check honeycomb for the requests made in the background (honeycomb has telemetry for the requests library)

madwort:
https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/datasets/job-server/result/nwdBZ8xhqQt

tomodwyer:
No current plans to optimise. The job submission flow will eventually be redesigned to fix some of these issues.

PR #3860, opened 2023-12-11
Remove O(N) queries from workspace status API

ghickman:
Previously this was doing O(N) queries, which we accepted when the API was addede because we did not, at the time, have workspaces with more than low-10s of jobs.  We now have workspaces with hundreds of jobs defined.  On a particularly large workspace (long-covid-symptoms) this endpoint could take as long as 5s (and ~300 queries) to calculate.

This change uses raw SQL so we can make use of a GROUP BY in the CTE to get the latest actions and then JOIN to it in the outer query to match those up with their statuses.

Found while investigating #3781.  This isn't the cause of the slow down mentioned in that ticket, but it's running at least once on every page load, and eating away at our resources needlessly.

### DIFF
--- a/jobserver/models/core.py
+++ b/jobserver/models/core.py
@@ -10,7 +10,7 @@ from django.contrib.auth.models import AbstractBaseUser
 from django.contrib.auth.models import UserManager as DjangoUserManager
 from django.contrib.auth.validators import UnicodeUsernameValidator
 from django.contrib.postgres.fields import ArrayField
-from django.db import models, transaction
+from django.db import connection, models, transaction
 from django.db.models import Min, Q, prefetch_related_objects
 from django.db.models.functions import Lower, NullIf
 from django.urls import reverse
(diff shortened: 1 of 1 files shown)


jobserver/models/core.py
```diff
@@ -1363,21 +1363,42 @@ def get_action_status_lut(self, backend=None):
...
+            INNER JOIN jobserver_workspace workspace ON (job_request.workspace_id = workspace.id)
+          WHERE workspace.name = %s
+          GROUP BY action
+        )
+        SELECT
+          latest_actions.name AS name,
+          job.status AS status
+        FROM
+          latest_actions
+          INNER JOIN jobserver_job job ON (
```
Jongmassey, line 1388: nit: I'd indent the `INNER JOIN` the same as `FROM` but YDY if that was your intention!

jobserver/models/core.py
```diff
@@ -1363,21 +1363,42 @@ def get_action_status_lut(self, backend=None):
...
+            latest_actions.name = job.action
+            AND
+            latest_actions.latest = job.created_at
```
Jongmassey, line 1391: Joining back into one of the tables also referenced within your CTE feels like a bit of a smell to me. Did you evaluation the performance of  this approach (aggregate with max in the CTE, join back to source table on result of aggregation) to using `ROW_NUMBER()` within the CTE and the selecting the row from the CTE with the lowest row number? My instinct would be to prefer the latter rather but that might be a mssqlism rather than anything that matters in pgsql.

Jongmassey:
```sql
WITH latest_actions AS (
  SELECT
    job.action as name,
    jon.status as status,
    row_number() OVER (PARTITION BY job.action ORDER BY job.created_at DESC) as row_num
  FROM 
    jobserver_job job
  INNER JOIN jobserver_jobrequest job_request ON (job.job_request_id = job_request.id)
  INNER JOIN jobserver_workspace workspace ON (job_request.workspace_id = workspace.id)
  WHERE workspace.name = %s
)
SELECT 
  name,
  status
FROM cte
WHERE row_num = 1
```

ghickman: I hadn't even thought to consider window functions!  And your version is nearly twice as fast on my local machine 👌

jobserver/models/core.py
ghickman: ooh interesting.  My GUI of choice (Postico) formatted it like this.  Updated 👍

Jongmassey (approved):
🚀

Issue #3702, opened 2023-10-17
Flag that a job request has been cancelled


ghickman:
When a user cancels an entire job request we don't show this to other viewers of the page, until the backend resolves that request.  However we should be able to derive that information from the cancelled_actions field.

Note: we only put actions into that field for jobs which have not already completed, so we'll need to take that into account when trying to derive "did the user click the cancel button"

Issue #3701, opened 2023-10-17
Flag that a job has been requested for cancellation


ghickman:
After cancelling a job we show a success message to the user who requested it, but we don't show any change in state to other viewers.

We're storing that a job has been requested for cancellation on the owning job request so, pending Tomproval, I think we can do two things:
* add a warning alert to the top of the page saying the job has been requested for cancellation but the backend has not yet processed that request
* mark the cancel button as disabled

The alert will need to go away once the job has been cancelled (IIRC we mark it as failed but that should be checked), but the disabled state for the button should persist.

PR #3880, opened 2023-12-14
Show alerts when a job or job request has been cancelled

ghickman:
This adds info alerts to the JobRequestDetail and JobDetail pages when a Job[Request] is still running but a user has requested they are cancelled, and sets the cancel buttons to disabled.

Fix: #3701 
Fix: #3702

### DIFF
--- a/jobserver/views/jobs.py
+++ b/jobserver/views/jobs.py
@@ -103,6 +103,7 @@ class JobDetail(View):
             ] = honeycomb.previous_actions_link(job)
 
         context = {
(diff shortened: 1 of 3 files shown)


lucyb:
I'm not sure where to put this, but the job status (shown on the job detail page when a cancellation request is made) is confusing:
Screenshot from 2024-01-02 16-57-53

ghickman:
@lucyb – I've captured this in #3913 since it's not relevant to the changes in this PR, but certainly a valid point!

tomodwyer (approved):

PR #3888, opened 2023-12-18
Fix smart quotes in reports

ghickman:
Well, this was an absolute doozy.

To recap, we have two views of the same file:
- Reports: quotes are mangled
- outputs-viewer: quotes are fine

Viewing the endpoint directly also rendered mangled quotes.  I'm still not sure why outputs-viewer was displaying them correctly, possibly something to do with the iframe? @tomodwyer – any ideas? (not relevant to this PR as all three contexts still work after this change).

I've kept FileResponse here since it sets a few values for us and recreating that, accurately, with HttpResponse wasn't worth the effort, so instead I've updated the Content-Type header to include an explicit charset.

Interestingly, `response.charset` was already correctly set to utf-8 before this change.

Fix: opensafely-core/reports#550

### DIFF
--- a/jobserver/releases.py
+++ b/jobserver/releases.py
@@ -236,6 +236,14 @@ def serve_file(request, rfile):
         # bypass DRFs renderer framework and just serve bytes
         response = FileResponse(path.open("rb"))
 
+        content_type = response.headers.get("Content-Type")
+        if content_type.startswith("text"):
+            # for text-based files append a charset to the existing
+            # content-type header, being careful just in case the existing
+            # value is empty
+            joiner = "; " if content_type else ""
+            response.headers["Content-Type"] = f"{content_type}{joiner}charset=utf-8"
+
     # set Last-Modified header as per:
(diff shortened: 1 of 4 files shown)


jobserver/releases.py
```diff
@@ -236,6 +236,12 @@ def serve_file(request, rfile):
         # bypass DRFs renderer framework and just serve bytes
         response = FileResponse(path.open("rb"))
 
+        # append a charset to the existing content-type header, being careful
+        # just in case the existing value is empty
+        content_type = response.headers.get("Content-Type")
+        joiner = "; " if content_type else ""
```
bloodearnest, line 242: I think this is buggy - should probably only add a charset if the content-type starts with `text/`, right? We don't want images to be served with a charset?

bloodearnest (approved):
Minor bug fix suggested

jobserver/releases.py
ghickman: oooh good shout 👍

Issue #3946, opened 2024-01-04
Remove references to "datalab" org


lucyb:
We'd like to rename the "datalab" org in Job Server to "Bennett Institute". However, before we can do that we need to ensure that we're not referencing the organisation by name anywhere. I've found one instance, which I believe to be the only instance in Job Server, where we're explicitly referencing the datalab by name. This line should be updated to use the org IDs instead.

https://github.com/opensafely-core/job-server/blob/431fbf65344a2acbc0182b1146cd2e6199a618c7/staff/views/dashboards/copiloting.py#L89

:notebook: Note: this ticket does not include renaming the organisation. That is expected to be done some time after this work.

ghickman:
We should make sure to keep the Org names with the IDs so a reader doesn't need access to production data to check what's being ignored here!  We used slugs here originally for that reason, but having IDs would certainly save on any future work.

Whoever does this should do the rename in a migration so they can create a redirect instance for the Org and not break any URLs, and it saves any other devs from needing to pull an updated database after the change.

ghickman:
We're also looking this org up [when we create an output checking issue](https://github.com/opensafely-core/job-server/blob/main/jobserver/issues.py#L94).

PR #3949, opened 2024-01-04
renamed the datalab org in Job Server to Bennett Institute. issue #3946

aman24deep:


lucyb:
Hi @aman24deep. Thank you for the PR. Unfortunately, while we take pride in working in the open, we are not currently set up to receive external contributions. I appreciate that the use of the `good-first-issue` label might be confusing and we'll review this for any future issues.

Issue #3892, opened 2023-12-20
OpenTelemetry HTTP Attributes Breaking Changes


madwort:
Email from honeycomb notifying about these changes. Not currently planned for the languages that were're currently using, but will be at some point, and I think we do use these params in some of our honeycomb boards.

e.g. `http.target` -> `url.path AND url.query`, `http.status_code` -> `http.response.status_code`, ...

see also https://opentelemetry.io/blog/2023/http-conventions-declared-stable/#migration-plan

madwort:
at some point in the future, the python lib will be updated, and then the dashboards will need to be updated to reflect that

madwort:
https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2002

ghickman:
`just upgrade prod opentelemetry-instrumentation-django` isn't picking up 0.43b0 for the django package, nor the psycopg2, or requests ones.

Running the fully qualified pip-compile command I get the same thing, no changes to `requirements.prod.txt`.

Dropping down to `pip install --upgrade opentelemetry-instrumentation-django` gives a workable error:

```ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
opentelemetry-instrumentation-django 0.41b0 requires opentelemetry-instrumentation==0.41b0, but you have opentelemetry-instrumentation 0.43b0 which is incompatible.
opentelemetry-instrumentation-dbapi 0.41b0 requires opentelemetry-instrumentation==0.41b0, but you have opentelemetry-instrumentation 0.43b0 which is incompatible.
opentelemetry-instrumentation-requests 0.41b0 requires opentelemetry-instrumentation==0.41b0, but you have opentelemetry-instrumentation 0.43b0 which is incompatible.
opentelemetry-instrumentation-psycopg2 0.41b0 requires opentelemetry-instrumentation==0.41b0, but you have opentelemetry-instrumentation 0.43b0 which is incompatible.
opentelemetry-instrumentation-wsgi 0.41b0 requires opentelemetry-instrumentation==0.41b0, but you have opentelemetry-instrumentation 0.43b0 which is incompatible.
```

Running this command with any of the `opentelemetry-instrumentation-*` dependencies gives a similar error, however, `opentelemetry-instrumentation` works.  Trying to upgrade this package via pip-compile does not give any changes.

madwort:
[...]

PR #3953, opened 2024-01-05
feat: emit both old & new opentelemetry http conventions

madwort:
* partly fixes https://github.com/opensafely-core/job-server/issues/3892


madwort:
parking for the time being.

PR #3962, opened 2024-01-09
Remove references to "datalab" org

### LINKED ISSUE #3946: Remove references to "datalab" org

iaindillingham:
This removes references to the "datalab" org, selecting this org with the PK rather than the slug. For consistency, it does the same for two other orgs: "lshtm" and "university-of-bristol".

Closes #3946

### DIFF
--- a/jobserver/settings.py
+++ b/jobserver/settings.py
@@ -392,3 +392,8 @@ BACKEND_IP_MAP = {
     # uncomment to pretend your browser is on tpp
     # "127.0.0.1": "tpp",
 }
+
+# These orgs are not copiloted
+BENNETT_ORG_PK = 3
+LSHTM_ORG_PK = 4
+UNIVERSITY_OF_BRISTOL_ORG_PK = 9
--- a/jobserver/issues.py
+++ b/jobserver/issues.py
@@ -91,7 +91,7 @@ def create_copilot_publish_report_request(report, github_api):
 
 
 def create_output_checking_request(release, github_api):
-    is_internal = release.created_by.orgs.filter(slug="datalab").exists()
+    is_internal = release.created_by.orgs.filter(pk=settings.BENNETT_ORG_PK).exists()
 
     files = release.files.all()
     number = len(files)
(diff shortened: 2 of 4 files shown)


jobserver/settings.py
```diff
@@ -392,3 +392,8 @@ def immutable_file_test(path, url):
     # uncomment to pretend your browser is on tpp
     # "127.0.0.1": "tpp",
 }
+
+# These orgs are not copiloted
+BENNETT_INSTITUTE_ORG_PK = 3
+LSHTM_ORG_PK = 4
+UNIVERSITY_OF_BRISTOL_ORG_PK = 9
```
ghickman, line 399: I don't love putting the PKs into settings like this!  But there isn't much else we can do for now.  Having spoken to Brian it doesn't sound like there's a proper process defined about Orgs "graduating" from copiloting, so eg it's not a simple option to stick a boolean on the Org.

It might be worth dropping `INSTITUTE` from the Bennett name, to save everyone a bit of typing and reading 🙂  I doubt we'll have to differentiate between ourselves and another similarly named org any time soon, YAGNI if you will.

ghickman (approved):

jobserver/settings.py
iaindillingham: Thanks, George. I've dropped the `INSTITUTE` from the Bennett name, which tidies up a line. I'd not considered adding a boolean to the Org; even if we can't do this now, it's useful to know that it's an alternative. Thank you for highlighting it.

PR #3972, opened 2024-01-10
Small refactor jobs api

lucyb:


jobserver/api/jobs.py
```diff
@@ -145,7 +145,12 @@ def post(self, request, *args, **kwargs):
                     defaults={**job_data},
                 )
 
-                if not created:
```
iaindillingham, line 148: This is in the weeds, but I'd be interested to know (a small amount) more about why you rearranged the `if`, here. I read somewhere (possibly Clean Code) that it was better for the `if` to be the default path (and if it was Clean Code, then there would have been a justification for this claim!). Is this your rationale?

iaindillingham (approved):

jobserver/api/jobs.py
```diff
@@ -200,7 +199,7 @@ def post(self, request, *args, **kwargs):
         return Response({"status": "success"}, status=200)
 
 
-def handle_job_notifications(request, job_request, job):
```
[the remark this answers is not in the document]
lucyb: Yes, it was! I did add that to the commit message..

jobserver/api/jobs.py
lucyb: So, I tried to have a look online for a reference but I couldn't find one, I think it's probably Clean Code or similar though. I believe there's a general consensus that a positive statement is easier to read than a negative one. Arguably this statement is very simple and is clear enough that it doesn't need changing, but I thought it was a tiny readability improvement.

jobserver/api/jobs.py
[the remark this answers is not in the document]
iaindillingham: Sorry, I should have stepped through the commit messages. I didn't, but I don't know why.

jobserver/api/jobs.py
iaindillingham: Ah, okay, that's slightly different, then: you're saying that the `if` should be a positive statement, rather than the default (most common) path. I misremembered the advice about the default path. I didn't read it in _Clean Code_, which I don't think I've read, I read it in _Code Complete_:

> Write the nominal path through the code first; then write the unusual cases. Write your code so that the normal path through the code is clear. Make sure that the rare cases don't obscure the normal path of execution. This is important for both readability and performance.[^1]

[^1]: https://learning.oreilly.com/library/view/code-complete-2nd/0735619670/ch15.html#if_statements

Issue #3958, opened 2024-01-08
Rename "DataLab" to "Bennett Institute"


lucyb:
Requires #3946 to be completed first.

Background 🧵 [here](https://bennettoxford.slack.com/archives/C04NHSU5DRB/p1704279098820819).

The DataLab organisation in Job Server should be renamed to Bennett Institute, given that we renamed ourselves a while ago.

We can rename the organisation without any negative effects (I believe), but we should update the URLs as well for consistency (represented by the `slug` field in the Organisation class). It's possible to do this within the staff area; this will automatically create a redirect as well. However, we could consider doing this as a migration, so that all other developers will have the change applied without needing a new copy of the production database.

This has the potential to cause some confusion, so we need to clearly communicate the change to everyone within the Bennett Institute.

iaindillingham:
I was going to do this in a migration as part of #3946. I'll not do this work as part of that issue, and instead create a new branch/PR for this work.

iaindillingham:
I asked for thoughts on this issue[^1]...

> I’m unclear whether I should or shouldn’t create a migration for the rename. On the one hand, using the admin interface creates a redirect. On the other hand, having used the admin interface, everyone developing job-server would have to fetch a new DB dump to see the changes locally. This wouldn’t be the case with a migration.

I think there was slightly more support for the admin interface approach than for the migration approach.

* It would be slightly less work for developers in the short term, but would introduce a small amount of work for a developer in the short term, too.
* Whilst both approaches could be tested locally, a migration wouldn't be as well tested as the code that underpins the admin interface.

[^1]: https://bennettoxford.slack.com/archives/C069SADHP1Q/p1704810336233969

iaindillingham:
> we need to clearly communicate the change to everyone within the Bennett Institute

I think a message in #general would be enough. Something like:

> **job-server announcement:** We have renamed "DataLab" (the organisation) to "Bennett Institute". If you have bookmarked the former, then the URL will redirect to the latter automatically.

iaindillingham:
@ghickman, I've noticed that whilst changing the slug adds a redirect for the main site, it doesn't add a redirect for the staff area. I think this is okay; the staff area isn't public-facing, and shouldn't be crawled. Whilst a user may visit the old URL because it's bookmarked/in their history, hopefully they'll realize that it has changed and navigate to the new URL. Do you agree?

iaindillingham:
[...]

PR #3976, opened 2024-01-10
Add `dump-co-pilot-reporting-data.sh`

iaindillingham:
After completing #3958, Millie asked for a dump of co-pilot reporting data. Whilst the snippet in `DEVELOPERS.md` is almost complete, I realized I needed to remove a pre-existing dump before running the snippet and zip the dump before sharing it. For convenience, then, I decided to move the snippet into a script.

### DIFF
--- a/scripts/dump-co-pilot-reporting-data.sh
+++ b/scripts/dump-co-pilot-reporting-data.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if ! command -v db-to-sqlite &>/dev/null; then
+    echo 'db-to-sqlite was not found'
+    exit 1
+fi
+
+if test -z "${DATABASE_URL:-}"; then
+    echo 'DATABASE_URL was not set'
+    exit 1
+fi
+
+rm -f jobserver.sqlite.zip
+
+db-to-sqlite \
+    --progress \
+    --table applications_application \
+    --table jobserver_project \
+    --table jobserver_projectmembership \
+    --table jobserver_workspace \
+    --table jobserver_user \
+    --table jobserver_org \
+    --table jobserver_orgmembership \
+    --table jobserver_job \
+    --table jobserver_jobrequest \
+    --table jobserver_release \
+    --table jobserver_releasefile \
+    ${DATABASE_URL} \
+    jobserver.sqlite
+
+# `--move` deletes the uncompressed file after compressing it
+zip --move jobserver.sqlite.zip jobserver.sqlite
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ __pycache__
 /docker/staticfiles/*
 /htmlcov
 /jobserver.dump
+/jobserver.sqlite.zip
 /node_modules
 /release-hatch
 /releases
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -376,27 +376,9 @@ Alternatively, the `osi_release` command can be used without running an analysis
 Co-pilots [have a report](https://github.com/ebmdatalab/copiloting/tree/copiloting-report) they run every few months, building on data from this service.
 
 To produce a dump in the format they need you will need to install [db-to-sqlite](https://pypi.org/project/db-to-sqlite/) via pip, pipx, or your installer of choice.
+You will also need to set the `DATABASE_URL` environment variable.
 
-Then run the following, replacing `<database URL>` with the URL to your database, this is likely in the `DATABASE_URL` environment variable.
-
-```
-db-to-sqlite \
-    --progress \
-    --table applications_application \
-    --table jobserver_project \
-    --table jobserver_projectmembership \
-    --table jobserver_workspace \
-    --table jobserver_user \
-    --table jobserver_org \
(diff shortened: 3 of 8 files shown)


scripts/dump-co-pilot-reporting-data.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
ghickman, line 1: Can you use `#!/usr/bin/env bash` here?  It's a little more portable and how we do it elsewhere.

scripts/dump-co-pilot-reporting-data.sh
```diff
@@ -0,0 +1,32 @@
...
+set -euo pipefail
+
+if ! command -v db-to-sqlite &>/dev/null; then
+    echo 'db-to-sqlite was not found'
+    exit 1
+fi
+
+if [ -z ${DATABASE_URL+none} ]; then
```
ghickman, line 9: What does `+none` do here?

ghickman (commented):
This is a great start, thank you for improving the state of this task!

Could you move the script into a just target?  We've worked hard to keep entry points inside just on the principle of least surprise.  It also means you can add a little docstring to the target which shows up when you run `just`.

If you think it's a little long for that you could always invoke the script from just like we do for several other things.

scripts/dump-co-pilot-reporting-data.sh
iaindillingham: I can, but each of the other scripts in `scripts` uses `#!/bin/bash` rather than `#!/usr/bin/env bash`. I agree that the latter is more portable. Should I change them all?

ghickman: Oh yes, that would be great.  IIRC @madwort normalised all our justfile bash invocations to the env-based one, so it would be good to be consistent.

scripts/dump-co-pilot-reporting-data.sh
iaindillingham: If `DATABASE_URL` isn't set, then the braces will evaluate to an empty string. If it is set, then the braces will evaluate to the string `none`. In the former case, [the condition will be true][1] (length is zero). In the latter case, the condition will be false (length is four). I couldn't think of a better token. Possibly `false`? (More on [SO][2].)

Why bother? Well, without the braces, if `DATABASE_URL` isn't set, then the script terminates with

```
./scripts/dump-co-pilot-reporting-data.sh: line 9: DATABASE_URL: unbound variable
```

is displayed, because of `set -eu`.

[1]: https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html
[2]: https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash

ghickman: Well that's no fun, is it?  @bloodearnest – is there a better way you know of to check for an env var existing and also having a value?

bloodearnest: I could totally do a yarn on test vs [ vs [[, thinking about it.

iaindillingham: I'd love a yarn on that! How about Thursday 25th January?

iaindillingham: Thank you for your comments, @bloodearnest. Really useful and really interesting! I've updated the script.

scripts/dump-co-pilot-reporting-data.sh
iaindillingham: I've changed them all.

iaindillingham:
> Could you move the script into a just target? We've worked hard to keep entry points inside just on the principle of least surprise. It also means you can add a little docstring to the target which shows up when you run `just`.
> 
> If you think it's a little long for that you could always invoke the script from just like we do for several other things.

Thanks for the high quality feedback, @ghickman and @bloodearnest. George, I've added a Just target that points to the script. I've also made a couple of changes to the script since you last looked at it. As well as Simon's suggestions, it now removes the uncompressed SQLite database. I've also added the compressed SQLite database to `.gitignore`, so we don't accidentally commit it.

ghickman (approved):
This looks great, thanks Iain 👍

scripts/dump-co-pilot-reporting-data.sh
```diff
@@ -0,0 +1,33 @@
...
+    --table jobserver_user \
+    --table jobserver_org \
+    --table jobserver_orgmembership \
+    --table jobserver_job \
+    --table jobserver_jobrequest \
+    --table jobserver_release \
+    --table jobserver_releasefile \
+    ${DATABASE_URL} \
+    jobserver.sqlite
+
```
bloodearnest, line 31: Minor suggestion - it might be worth running sqlite's VACCUUM command to shrink the db file before then compressing.

```sqlite3 jobserver.sqlite VACUUM```

This might not make much difference, depending on how db-to-sqlite does its thing, however.

iaindillingham: Thanks, Simon. I tried this, and it didn't make any difference.

PR #4009, opened 2024-01-19
Add collaboration created and updated

ghickman:
I forgot to add created_at, created_by, updated_at, and updated_by fields to ProjectCollaboration when I added it.

There are a couple of little extras in here I found along the way, and I ended up pulling creating and editing projects into commands just to make them a little easier to reason about.  I tried refactoring ProjectEditForm into a Form to address the TODO but hit a snag with the `check_number` method because it uses the instance to do the lookup, so I'll come back to that one.

Note: there are no constraints for the created/updated pairs because are non-nullable.


templates/staff/project/create.html
```diff
@@ -34,7 +34,7 @@
       <div class="flex flex-col items-stretch gap-y-6 w-full max-w-3xl">
         {% #form_fieldset %}
           {% form_legend text="Select an org" %}
-          {% multiselect class="w-full" data-max-items=1 field=form.fields.org multiple=False input=form.org %}
+          {% multiselect class="w-full" data-max-items=1 field=form.fields.orgs multiple=False input=form.orgs %}
```
iaindillingham, line 37: Based on the arguments alone, I think this multiselect allows a user to select one item. Don't we want a user to select one or more items?

tests/unit/jobserver/commands/test_projects.py
iaindillingham, line 1: What ensures `test_add_interactive_project` adds an interactive project but `test_add_standard_project` adds a standard project? The tests look almost identical to me.

jobserver/commands/projects.py
iaindillingham, line 1: What do we mean by "command", here? (We don't mean a management command, for example.)

iaindillingham (approved):
Couple of questions, but no blockers.

jobserver/commands/projects.py
ghickman: This is a reference to the C in CQRS.  TLDR we haven't come up with a better name for them, but roughly speaking they're a domain layer.

tests/unit/jobserver/commands/test_projects.py
ghickman: `application_url` is set… it's not the most obvious part :/   I'll go over how these bits fit together in more detail when I talk about OSI in depth!

templates/staff/project/create.html
ghickman: No!  This, and #3571, are all about changing the schema to support multiple orgs but not actually supporting them in the UI, so in practice we still only have one Org per Project, but when we're ready to support them there won't be a mammoth change needed to do so.

staff/views/projects.py
```diff
@@ -101,7 +101,23 @@ def form_valid(self, form):
                     created_by=self.request.user,
                     updated_by=self.request.user,
                 )
-                project.orgs.set(orgs)
+
+                lead, *other = orgs
```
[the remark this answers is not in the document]
ghickman: They can still only select one Org in the UI, so all orgs are technically leads currently, but the goal is for this to chance once we can work out how the UI will handle that… which will answer this question 😁

jobserver/commands/projects.py
iaindillingham: CQRS?

PR #4024, opened 2024-01-24
Display per-job metrics on Job Detail page

### LINKED ISSUE #3998: Display per-job metrics on Job Detail page
We've recently started receiving metrics for jobs from Job Runner (see https://github.com/opensafely-core/job-server/issues/3867). These are stored as JSON in `jobserver_job.metrics`. We want to display these metrics back to the user, so they know what resources their job used. This may also be useful for tech support to aid debugging of problems.

This information should be displayed on the Job Detail page to logged in users only.

At this time, we aren't concerned with showing it in other areas of Job Server.

Some things to consider:
1. We only have data for recent jobs (as of 17th Jan). We don't have historic data.
2. The fields may change over time as we learn more about what's useful and what's not.
3. The data is available while jobs are running and when they've completed, with the values being updated on a regular basis.

Example from a running job:
``` json
{"cpu_mean": 11.665892511771082, "cpu_peak": 103.93, "cpu_cumsum": 1194284.0800000527, "cpu_sample": 0.01, "mem_mb_mean": 15822.789339345383, "mem_mb_peak": 9709.567999839783, "mem_mb_cumsum": 1619842235.8261442, "mem_mb_sample": 4081.663999557495}
```

A completed job (same fields as a running job):
```json
{"cpu_mean": 149.15872093023256, "cpu_peak": 425.93, "cpu_cumsum": 51310.6, "cpu_sample": 99.9, "mem_mb_mean": 22259.211918184923, "mem_mb_peak": 37068.79999923706, "mem_mb_cumsum": 7657168.899855614, "mem_mb_sample": 579.8999996185303}
```

A pending job:
```
{}
```

iaindillingham:
This adds a "Job metrics" card to the Job Detail page for users with one or more roles. This card displays CPU and memory usage statistics, when available (see below).

In addition to mean and peak, which are displayed here, Job Runner collects sample and cumsum for both CPU and memory usage statistics (opensafely-core/job-runner#686). However, I don't think sample and cumsum usage statistics are useful to users: sample, because we (and so they) don't know when the sample was taken; cumsum, because we (and so they) don't know how many samples were taken (and if we did, then all we would be able to compute would be the mean, which Job Runner has computed for us).

As #3998 states, usage statistics are not available for historic jobs (i.e. prior to 17/01/2024). In this case, `job.metrics == None`. They are also not available for some other types of job. In these cases, `job.metrics == {}`. We don't distinguish these cases, displaying "-" to users in both cases.

Closes #3998

---

Thanks, @ghickman and @lucyb for suggesting different approaches to determine whether the "Job metrics" card should be displayed. I've opted for `request.user.all_roles` for simplicity. We do something similar, here:

https://github.com/opensafely-core/job-server/blob/624b6b9c8ebba387c2d3e8c5805ae7c6966009a0/templates/staff/user/list.html#L145-L147

### DIFF
--- a/templates/job/detail.html
+++ b/templates/job/detail.html
@@ -188,6 +188,50 @@
       </dl>
     {% /card %}
 
+    {% if request.user.all_roles %}
+    {% #card title="Job metrics" subtitle="Computed for the time spent running on the backend" %}
+      <dl class="border-t border-slate-200 sm:divide-y sm:divide-slate-200">
+        {% #description_item title="Mean CPU usage" %}
+          {% if job.metrics %}
+          <span class="relative group cursor-pointer">
+          {{ job.metrics.cpu_mean|floatformat:"0" }}%
+          {% tooltip content="Values are reported across all CPU cores" %}
+          </span>
+          {% else %}
+          -
+          {% endif %}
+        {% /description_item %}
+
+        {% #description_item title="Peak CPU usage" %}
+          {% if job.metrics %}
+          <span class="relative group cursor-pointer">
+          {{ job.metrics.cpu_peak|floatformat:"0" }}%
+          {% tooltip content="Values are reported across all CPU cores" %}
+          </span>
+          {% else %}
+          -
+          {% endif %}
+        {% /description_item %}
+
+        {% #description_item title="Mean memory usage" %}
+          {% if job.metrics %}
+          {{ job.metrics.mem_mb_mean|floatformat:"0g" }} MB
+          {% else %}
+          -
+          {% endif %}
+        {% /description_item %}
+
+        {% #description_item title="Peak memory usage" %}
+          {% if job.metrics %}
+          {{ job.metrics.mem_mb_peak|floatformat:"0g" }} MB
+          {% else %}
+          -
+          {% endif %}
+        {% /description_item %}
+      </dl>
+    {% /card %}
+    {% endif %}
+
     {% if honeycomb_links %}
       {% #card title="Monitoring" subtitle="Honeycomb login required" container %}
         <ul class="flex flex-col gap-y-1.5 list-disc text-sm">


sebbacon:
It's not clear to me that this shouldn't be public information.

@iaindillingham asked:

> Could you help me understand when information should be public and when it shouldn’t, 
@seb?

1. You need an account to use it
2. Honeycomb is like an "expert" view. We'd have to explain lots of things for anyone to make use of it. Even if it were free, we're not in a position to make it useful for most users

In contrast, memory and CPU are fairly simple concepts. I don't think they add any extra noise, and I would personally go with the principle that most things should happen in public. You never know what benefits may accrue from exposing the data. 

That said, I'm sure @lucyb had a good reason to suggest it should be for logged in users only in #3998...

lucyb:
> That said, I'm sure @lucyb had a good reason to suggest it should be for logged in users only in

I was taking a precautionary approach. I wasn't sure the data should be public, since we're receiving it straight from the secure environments, and so I thought it would be best if it was for logged in users only. This also reduces the pressure to present it in a way that's meaningful to most people.

That said, I can't think of a reason why showing it to more people would actually cause an issue.

iaindillingham:
Thanks, @lucyb. If I've understood opensafely-core/job-runner#686 correctly, then I don't think there's a significant risk of leaking sensitive information. I also don't think that logged in users would be better at interpreting the metrics than public users; if there's an issue with their interpretation, in other words, then I think the issue would apply to both groups of users.

For these reasons, I'm going to remove the check. Sound good, @sebbacon?

iaindillingham:
Before I :shipit:, @ghickman asked on Slack about `request.user.is_authenticated`.

> This information should be displayed on the Job Detail page to logged in users only.

I took "logged in users" from #3998 at face value: an authenticated user is also an authorized user. However, I think George was going to suggest that an authenticated user may not be an authorized user. Is that correct, George?

ghickman:
> I think George was going to suggest that an authenticated user may not be an authorized user.

Yep, that's correct, [`user.is_authenticated` does not imply permissions](https://docs.djangoproject.com/en/5.0/topics/auth/customizing/#django.contrib.auth.models.AbstractBaseUser.is_authenticated).  This is important for us because we have open registration to support applications, so authenticated is barely a step above unauthenticated (by design) for us.

A better gate here would be to check for a permission in the view and pass that as a boolean out to the template via the context.  The `run_job` permission seems like it would fit well here I think?

lucyb:
Alternatively, if you didn't want to be constrained by a particular permission, there's the `all_roles` [property](https://github.com/opensafely-core/job-server/blob/3bbb7b186ba81580c3b47f3dffe22cc246550c0d/jobserver/models/user.py#L153) which you could do a truthy check on. It's a bit clunky but it's a pattern we've used in at least one other area.

ghickman (approved):

PR #4028, opened 2024-01-25
Fix flaky test

### LINKED ISSUE #4003: Track down intermittent search tests
We very infrequently get test failures for one of the views with a search box in it, [for example](https://github.com/opensafely-core/job-server/actions/runs/7572121851/job/20621208055).

When these pop up we typically fix them by being more specific about our test data to ensure we're not generating test data that could unexpectedly overlap with the search query.

The goal of this ticket would be to fix the linked test so it's not intermittent and also look at all search query tests, in general, to see if any others could be fixed.

iaindillingham:
`test_jobrequestlist_search_using_username` is a flaky test (it can either pass or fail without the code under test, or the test code itself, changing). This flakiness is caused by factories that occasionally create instances with values that match the query string. Most values are obviously not going to match the query string (e.g. values of `Workspace.name` will be `workspace-0`, `workspace-1`, `workspace-2`, etc.). However, `JobRequest.identifier` might match the query string, because by default it contains a string of random, lower case, alphanumeric characters. Instead, we use a string of non-random characters with the same characteristics.

As well as `test_jobrequestlist_search_using_username`, this fixes a couple of other flaky tests.[^1] However, it doesn't fix all tests that call `JobRequestFactory.create_batch`; it fixes only those tests where the query string is four characters or fewer.

Closes #4003 

[^1]: I've reviewed the tests that execute a search (i.e. those that contain `?q=`).

### DIFF
--- a/tests/unit/staff/views/test_job_requests.py
+++ b/tests/unit/staff/views/test_job_requests.py
@@ -238,7 +238,8 @@ def test_jobrequestlist_search_using_fullname(rf, core_developer):
 
 
 def test_jobrequestlist_search_using_identifier(rf, core_developer):
-    JobRequestFactory.create_batch(5)
+    for i in range(5):
+        JobRequestFactory.create(identifier=f"{i:016}")
 
     job_request = JobRequestFactory(identifier="1234abcd")
 
@@ -260,7 +261,8 @@ def test_jobrequestlist_search_using_identifier(rf, core_developer):
 
 
 def test_jobrequestlist_search_using_job_identifier(rf, core_developer):
-    JobRequestFactory.create_batch(5)
+    for i in range(5):
+        JobRequestFactory.create(identifier=f"{i:016}")
 
     job_request = JobRequestFactory()
     JobFactory(job_request=job_request, identifier="1234abcd")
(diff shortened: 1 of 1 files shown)


ghickman:
@iaindillingham – definitely no monkeypatching please!!  Your approach looks great to me, it's clear, explicit, and I would expect anyone setting data like this has a good reason for doing so 👍

One tiny nit with your explanation.  Factories don't create fixtures, they create model instances.  Why does this matter?  [Django uses the term fixtures](https://docs.djangoproject.com/en/5.0/topics/db/fixtures/) to describe blocks of data it can import in tests, which we don't use because they easily get stale, and the data definitions live away from your tests.  [Pytest _also_ uses the term fixture](https://docs.pytest.org/en/6.2.x/fixture.html) to describe functions used to set up a test, which we _do_ use!  I would love it if these two very old projects didn't use the same name for sort of very different things 😂

ghickman (approved):

iaindillingham:
> One tiny nit with your explanation. Factories don't create fixtures, they create model instances.

Thanks for this nit, @ghickman.

PR #4029, opened 2024-01-25
token login api

### LINKED ISSUE #4025: Add job-server token login API
We added a token based login explicitly for logging in from Level 4.

With the releases UI initiative, we now need to move that login form to live in airlock. However, the token authentication needs to be done by job-server.

In addition to username, airlock also needs to know the workspaces they have permission to view, and whether they are an OutputChecker

This ticket is to add a new token validation API that will
 - receive the user supplied token
 - validate it
 - return information about the user:
    - GH username
    - Display name
    - list of workspaces (airlock has no concept of Projects) 
    - whether the user is OutputChecker (to allow request approval)
    - whether they are staff (no planned usage yet, but might come in useful).
    
There may be other useful stuff, perhaps.

The API should log the authentication in the same way other authentications are logged.

bloodearnest:
- Invert check logic in authorization checks.
- Refactor token generation and validation logic
- Add token auth api

Fixes #4025

### DIFF
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -8,6 +8,7 @@ from django.db import transaction
 from django.db.models import Value
 from django.shortcuts import get_object_or_404
 from django.utils import timezone
+from opentelemetry import trace
 from rest_framework import serializers
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.exceptions import (
@@ -27,7 +28,8 @@ from interactive.models import AnalysisRequest
 from interactive.slacks import notify_report_uploaded
 from jobserver import releases, slacks
 from jobserver.api.authentication import get_backend_from_token
-from jobserver.authorization import has_permission
+from jobserver.authorization import OutputChecker, has_permission, has_role
+from jobserver.commands import users
 from jobserver.models import (
     PublishRequest,
     Release,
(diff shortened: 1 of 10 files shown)


jobserver/authorization/utils.py
evansd, line 1: I've got the no objection to the change, but I don't quite understand the comment here:
> but also allows us to query the role for a user who is *not* logged in from Django's perspective, which is need for API logins.

The logic is literally identical in this version, isn't it? If you compare the `BaseUser` implementation of these methods:
https://github.com/django/django/blob/e676f56527e70c4587567b97f9cfd22ab396142c/django/contrib/auth/base_user.py#L77-L91

With the `AnonymousUser` implementation:
https://github.com/django/django/blob/e676f56527e70c4587567b97f9cfd22ab396142c/django/contrib/auth/models.py#L491-L497

jobserver/api/releases.py
```diff
@@ -556,3 +558,73 @@ class WorkspaceStatusAPI(RetrieveAPIView):
...
+        data = serializer.data
+
+        try:
+            # user can be username or email
+            user = users.validate_login_token(data["user"], data["token"])
+        except (
+            User.DoesNotExist,
+            users.InvalidTokenUser,
+            users.BadLoginToken,
+            users.ExpiredLoginToken,
```
evansd, line 596: Minor aside, and I know this is just replicating the existing token auth code, but I found it surprising there wasn't a base exception class we could catch here. It seems odd the caller needs to enumerate the specific ways the login could fail.

jobserver/api/releases.py
```diff
@@ -556,3 +558,73 @@ class WorkspaceStatusAPI(RetrieveAPIView):
...
+        ) as exc:
+            logger.info(f"API Login with token failed for user {data['user']}: {exc}")
+            trace.get_current_span().record_exception(exc)
+            raise NotAuthenticated()
+
+        logger.info(f"User {user} logged in with login token via API")
+
+        workspaces = []
+        # this is 1 or 2 queries per project, not ideal, but we permissions are not stored in the db
```
evansd, line 605: Yeah there's not much we can do about this given the current structure, is there? But as you say, given we only have to do it once per login it's hard to see it causing us too much trouble for a while.

evansd (approved):

jobserver/authorization/utils.py
bloodearnest: Yeah, I think I got my wires crossed.  It works with either.

In my head, is_authenticated meant some kind of session look up, and in my API scenario, the user wouldn't have a session.

I will remove the commit, since its just noise

jobserver/api/releases.py
bloodearnest: Fair point, will add

PR #4030, opened 2024-01-26
Get first template name

ghickman:
`TemplateResponseMixin`, used in several GCBVs, returns a list of templates.  The first template in that list is the one used when rendering the response, so we want to use that for plausible.

### DIFF
--- a/tests/unit/jobserver/test_middleware.py
+++ b/tests/unit/jobserver/test_middleware.py
@@ -1,8 +1,12 @@
+from django.http import HttpResponse
+from django.template.response import TemplateResponse
 from django.test.utils import override_settings
+from django.views.generic import DetailView, View
 
-from jobserver.middleware import ClientAddressIdentification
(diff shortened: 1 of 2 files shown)


tests/unit/jobserver/test_middleware.py
```diff
@@ -84,3 +88,63 @@ def test_client_ip_middleware_get_forwarded_ip_no_ips():
...
+
+    template_name = getattr(response, "template_name", "already set")
+
+    TemplateNameMiddleware(None).process_template_response(request, response)
+
+    # when template_name is already set it should not be changed by the middleware
+    assert response.template_name == template_name
+
+
+def test_template_name_middleware_with_template_name_as_a_list(rf):
```
Jongmassey, line 125: I'd appreciate an ELI5 on how this forms a template name with a list

Jongmassey (approved):

PR #4067, opened 2024-02-01
Correct orgs and projects URLs in user detail area

### LINKED ISSUE #4042: Links on staff area's UserDetailWithEmail are broken
The links for Organisations and Projects for interactive users are broken.

You can see this in action on [Catherine's interactive user](https://jobs.opensafely.org/staff/users/catherine.stables@phc.ox.ac.uk/).

We're constructing `orgs` and `projects` in the context very differently between UserDetailWithEmail and UserDetailWithOAuth, I suspect this is because UserDetailWithEmail wasn't updated in 0035654, so we're calling `get_staff_url` in the template instead of `staff_url`.

StevenMaude:
Fixes #4042.

These were broken in the staff "user with detail" view.

### DIFF
--- a/templates/staff/user/detail_with_email.html
+++ b/templates/staff/user/detail_with_email.html
@@ -51,7 +51,7 @@
   {% #card title="Organisations" button=True button_text="Add to organisation" button_href=staff_user_set_orgs_url %}
     {% #list_group %}
       {% for org in orgs %}
-        {% #list_group_item href=org.get_staff_url %}
+        {% #list_group_item href=org.staff_url %}
           {{ org.name }}
         {% /list_group_item %}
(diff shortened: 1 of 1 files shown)


ghickman:
Apologies, I missed this one going by and in hindsight didn't make this at all clear on the issue - this section of the page is the same as the section on the other user page, but we construct it differently in the view.  We used to do the same thing for the OAuth user page until I added the `projects` and `orgs` M2M relationships on User which simplified getting the data.  I think it would be useful to normalise looking up this data to how the OAuth page does it to 1) avoid confusion and 2) potentially consider combining the pages in the longer term (or even just parts of them).

PR #4069, opened 2024-02-01
Include github username in issue body when requesting repo be made public

### LINKED ISSUE #3749: Tag requester in issues for making repos public
Requests to make repos public trigger an issue to be created, but there is no feedback loop to the requester once the issue is closed. 

Could we tag the requester in the created issue [here](https://github.com/opensafely-core/job-server/blob/33e7a10512f8e65d32d20ed40c88f6c34fc9f18d/jobserver/issues.py#L136), instead of including the name?

Jongmassey:
fixes #3749

### DIFF
--- a/jobserver/issues.py
+++ b/jobserver/issues.py
@@ -133,7 +133,7 @@ def create_switch_repo_to_public_request(repo, user, github_api):
     body = f"""
     The [{repo.name}]({repo.url}) repo is ready to be made public.
 
-    This repo has been checked and approved by {user.name}.
+    This repo has been checked and approved by {user.name} - @{user.username}.
 
     An owner of the `opensafely` org is required to make this change, they can do so on the [repo settings page]({repo.url}/settings).
 


jobserver/issues.py
```diff
@@ -133,7 +133,7 @@ def create_switch_repo_to_public_request(repo, user, github_api):
     body = f"""
     The [{repo.name}]({repo.url}) repo is ready to be made public.
 
-    This repo has been checked and approved by {user.name}.
+    This repo has been checked and approved by {user.name} - @{user.username}.
```
iaindillingham, line 136: I hesitate to be this pedantic, but do we need name _and_ username? Louis suggested the latter instead of the former in the issue. If you think "Yes" then I accept your answer without further comment 🙂

iaindillingham (approved):

jobserver/issues.py
Jongmassey: I can think of a few people who have less obvious usernames than you or I do. I figured it'd save someone the effort of "who on earth is madwort or bloodearnest?" etc.

ghickman: Chiming in to say we have usernames outside the tech team that don't relate to full names in any way is the main reason I started on the journey to enforce full names! (eg who on earth is <alphanumeric string>, they're running jobs so _someone_ must know who they are?!)

iaindillingham: Very reasonable!

PR #4075, opened 2024-02-02
Add create_backend command

bloodearnest:
It prints the auth token. Motivating use case was to set up a test
backend from cli for local tests.

Replaced the janky scripot in the local release-hatch setup.

We want to use to automate setting up a local job-server in airlock.

### DIFF
--- a/jobserver/management/commands/create_backend.py
+++ b/jobserver/management/commands/create_backend.py
@@ -1,0 +1,58 @@
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from jobserver.models import Backend, User
+
+
+class Command(BaseCommand):
+    """
+    Create or update a backend
+
+    Designed to be used to automate development setup.
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument("slug", default="local-dev", help="The id for backend")
+        parser.add_argument("--name", default=None, help="The name will be set to this")
+        parser.add_argument(
+            "--url", default=None, help="level_4_url will be set to this"
+        )
+        parser.add_argument(
+            "--user", default=None, help="Username to add to the backend"
+        )
+        parser.add_argument(
+            "--quiet",
+            action="store_true",
+            help="only print the auth token (useful for scripts)",
+        )
+
+    @transaction.atomic()
+    def handle(self, *args, **options):
+        backend, created = Backend.objects.get_or_create(
+            slug=options["slug"],
+            defaults={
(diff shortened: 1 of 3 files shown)


jobserver/management/commands/create_backend.py
```diff
@@ -0,0 +1,39 @@
...
+        backend, created = Backend.objects.get_or_create(slug=options["slug"])
+        backend.name = options["name"]
+        if options["url"]:
+            backend.level_4_url = options["url"]
+        backend.save()
+
+        if options["user"]:
+            user = User.objects.get(username=options["user"])
+            user.backends.add(backend)
+            user.save()
```
ghickman, line 37: This is entirely a matter of taste, would calling BackendMembership (the through model behind `user.backends`) be clearer to you here?

jobserver/management/commands/create_backend.py
```diff
@@ -0,0 +1,39 @@
+from django.core.management.base import BaseCommand
+
+from jobserver.models.backend import Backend
+from jobserver.models.user import User
```
ghickman, line 4: Nit: We've generally opted to keep model imports in the form `from jobserver.models import MyModel` to avoid the specific location being relevant to the calling code.  We've moved models around a couple of times at this point so it's paid off there.  It's not _necessary_ but consistency is useful, etc.

jobserver/management/commands/create_backend.py
bloodearnest: yes, it would. idempotency iz gud.

jobserver/management/commands/create_backend.py
```diff
@@ -0,0 +1,39 @@
...
+            "--user", default=None, help="Username to add to the backend"
+        )
+
+    def handle(self, *args, **options):
+
+        backend, created = Backend.objects.get_or_create(slug=options["slug"])
+        backend.name = options["name"]
+        if options["url"]:
+            backend.level_4_url = options["url"]
+        backend.save()
```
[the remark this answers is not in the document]
bloodearnest: Ah, I wanted something like this, thanks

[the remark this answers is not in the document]
bloodearnest: > You could also avoid the if block by setting the URL's default to empty string.

That potentially would overwrite existing url?

I've used defaults, that's a nice change, but I still needed to branch on created to avoid the above, and I also added some more output.

PR #4086, opened 2024-02-05
Test release command

### LINKED ISSUE #3986: Test `release` command
We don't typically test commands and aim to make them as thin wrappers as possible around a function call, avoiding the difficulties that running commands directly can bring.

The `release` command has broken, likely due to changes to the underlying `jobserver/releases.py::create_release` function.  This command is very useful to developers when they want to work with releases without having to understand how to construct a working Release object.

We'd like to ensure this command keeps working, and having it create Releases and ReleaseFiles in a test should be fine, so I think we can add tests for it.

We should be able to delete these files after tests have run too.

Jongmassey:
Fixes #3986 

Adds tests and a couple of fixes for bugs revealed by these tests

### DIFF
--- a/tests/unit/jobserver/management/commands/test_release.py
+++ b/tests/unit/jobserver/management/commands/test_release.py
@@ -5,6 +5,27 @@ from django.core.management import call_command
 
 from jobserver.models import Workspace
 
+from .....factories import ProjectFactory, RepoFactory, WorkspaceFactory
+
+
+def assert_expected_output(stdout, expected_files, workspace):
+    a, release_url, workspace_url = stdout.getvalue().splitlines()
+    assert a == "Release created:"
+
+    assert workspace_url.endswith(workspace.get_absolute_url())
+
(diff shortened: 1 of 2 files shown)


tests/unit/jobserver/management/commands/test_release.py
```diff
@@ -0,0 +1,147 @@
+import csv
+from io import StringIO
+
+import pytest
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.management import call_command
+
+from jobserver.models import Workspace
+
+from .....factories import (
```
ghickman, line 10: Elsewhere I started moving to `tests.factories` after ~3 dots as it was getting a bit out of hand!

ghickman (approved):

tests/unit/jobserver/management/commands/test_release.py
Jongmassey: It did feel like a lot of dots.....

PR #4103, opened 2024-02-08
Disable auto-reload middleware on POST

evansd:
In development, this causes the single-use token page to be unexpectedly reloaded, thus generating a new token and causing confusion.

This is a (hopefully) temporary workaround in advance of this being a configurable option in `django-browser-reload`:
https://github.com/adamchainz/django-browser-reload/issues/243


bloodearnest (commented):
LGTM

iaindillingham (approved):

PR #4104, opened 2024-02-08
Add submitted fields to application detail page

iaindillingham:
Add both `submitted_by` and `submitted_at` to the application detail page, just as we have approved, created, and deleted fields.

### DIFF
--- a/templates/staff/application/detail.html
+++ b/templates/staff/application/detail.html
@@ -30,6 +30,20 @@
       </div>
     </div>
     <div>
+      {% if application.submitted_by %}
+      <div class="flex flex-row flex-wrap gap-1">
+        <dt class="font-semibold">Submitted by:</dt>
+        <dd>{% link href=application.submitted_by.get_staff_url text=application.submitted_by.name %}</dd>
+      </div>
+      {% endif %}
+      {% if application.submitted_at %}
+      <div class="flex flex-row flex-wrap gap-1">
+        <dt class="font-semibold">Submitted at:</dt>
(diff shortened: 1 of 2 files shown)


templates/staff/application/detail.html
```diff
@@ -17,8 +17,8 @@
 {% block hero %}
 {% #staff_hero title="Application "|add:application.pk_hash %}
   <dl class="flex flex-col gap-2 mt-1">
-    {% if application.created_by %}
     <div>
+      {% if application.created_by %}
```
lucyb, line 21: This is an aside, but we always seem to have a value for `created_by` in the database. I know the field is nullable, but I wonder if that's something that needs to be changed?

lucyb (approved):
This is a nice, well-considered change. Thank you.

templates/staff/application/detail.html
iaindillingham: I don't think it is nullable. Unless I'm looking at the wrong `Application`?

https://github.com/opensafely-core/job-server/blob/3fabb37bc8e6413bd902394d073e217888a3cba3/applications/models.py#L38-L42

If it isn't, then the `if` is redundant, and I'll remove it in this PR. (I moved it in this PR, after all 🙂.)

lucyb: Ah yes I think you're right. I was looking at the database schema rather than the model definition and hadn't realised that they differ. Try removing it and see what happens :)

iaindillingham: This seems odd: there are many required fields (`null=False`, the default) whose columns aren't labelled as `not null`. I've asked a question on Slack.[^1]

[^1]: https://bennettoxford.slack.com/archives/C63UXGB8E/p1707744068865509

PR #4107, opened 2024-02-09
Replace `clear_all_roles` method with command

### LINKED ISSUE #4093: "Clear all roles" button doesn't trigger an audit entry to be recorded
On the roles page for a user in the staff area, it's possible to remove all global and project roles with the "Clear all roles" button. However, this doesn't currently add an entry into the audit log.

It is expected that whenever project roles are changed that an entry is added into the audit log.

Image

iaindillingham:
This PR replaces the `User.clear_all_roles` method with the `users.clear_all_roles` command in the `UserClearRoles` view. It also removes the method, which is no longer used. Along the way, it introduces a couple of additional commands. Because the `users.clear_all_roles` command calls the existing `project_members.update_roles` command, we can be confident that clearing a user's roles is -- for project-membership roles -- audited.

Fixes #4093

### DIFF
--- a/tests/unit/jobserver/commands/test_users.py
+++ b/tests/unit/jobserver/commands/test_users.py
@@ -3,6 +3,7 @@ from datetime import timedelta
 import pytest
 from django.utils import timezone
 
+from jobserver.authorization import ProjectDeveloper
 from jobserver.commands import users
 
 from ....factories import (
@@ -77,3 +78,15 @@ def test_user_validate_login_token_wrong(token_login_user):
 
     with pytest.raises(users.BadLoginToken):
         users.validate_login_token(token_login_user.username, "bad token")
+
+
+def test_update_roles():
+    updator = UserFactory()
+    user = UserFactory()
+
+    assert user.roles == []
+
+    users.update_roles(user=user, by=updator, roles=[ProjectDeveloper])
+
+    user.refresh_from_db()
+    assert user.roles == [ProjectDeveloper]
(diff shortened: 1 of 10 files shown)


tests/unit/jobserver/commands/test_users.py
```diff
@@ -90,3 +91,21 @@ def test_update_roles():
...
+    assert len(user.roles) == 1
+    assert len(user.org_memberships.get().roles) == 1
+    assert len(user.project_memberships.get().roles) == 1
+
+    users.clear_all_roles(user=user, by=updator)
+
+    user.refresh_from_db()
+    assert len(user.roles) == 0
+    assert len(user.org_memberships.get().roles) == 0
+    assert len(user.project_memberships.get().roles) == 0
```
lucyb, line 111: I'm not completely sure about this, but is it worth asserting that there's an entry in the audit log here too?

lucyb (approved):
Thank you for breaking this down into some nice, well explained commits.

tests/unit/jobserver/commands/test_users.py
iaindillingham: I was in two minds, because we'd be testing that there was an entry in the audit log twice. Once here, and once in `test_project_members.test_update_roles`. I agree, however, that _not_ testing doesn't feel quite right.

I guess it doesn't feel quite right because you'd have to know how `clear_all_roles` worked to know that it added an entry. In the future, we may rewrite `clear_all_roles` but forget to add an entry. `test_clear_all_roles` would pass, but `clear_all_roles` would be defective.

So yes, you're quite right. It's worth asserting that there's an entry in the audit log 🙂 Thank you for the nudge.

PR #4117, opened 2024-02-12
Custom create_superuser

### LINKED ISSUE #4076: `createsuperuser` management command fails due to lack of username
With a fresh install, [this line](https://github.com/opensafely-core/job-server/blob/main/DEVELOPERS.md?plain=1#L180) of teh developers guide fails due to the lack of prompting for the required non-null `username` field

Jongmassey:
Fixes #4076 

django uses `User.USERNAME_FIELD` to derive what fields are required for `manage.py createsuperuser` but not for the built-in `UserManager.create_superuser()` which expects a `username` regardless, so if this is changed to not include `username` then one will encounter a missing argument error.

Since we set the username field to `email` to support OSI user creation we can/must set the username to email for superusers created with the createsuperuser management command.

This also allows us to set the staff and superuser boolean flags (which may soon be retired) and also assign the `CoreDeveloper` role to ensure an adequately permissioned "staff" user is created by this command.

### DIFF
--- a/jobserver/models/user.py
+++ b/jobserver/models/user.py
@@ -15,7 +15,7 @@ from django.utils.functional import cached_property
 from sentry_sdk import capture_message
 from zen_queries import queries_dangerously_enabled
 
-from ..authorization import InteractiveReporter
+from ..authorization import CoreDeveloper, InteractiveReporter
 from ..authorization.fields import RolesArrayField
 from ..hash_utils import hash_user_pat
 
@@ -47,6 +47,12 @@ class UserQuerySet(models.QuerySet):
 
 class UserManager(DjangoUserManager.from_queryset(UserQuerySet)):
     use_in_migrations = True
+
+    def create_superuser(self, email, password, **extra_fields):
+        username = email
+        return super().create_superuser(
+            username, email, password, roles=[CoreDeveloper]
+        )
 
 
 def get_or_create_user(username, email, fullname, update_fields=None):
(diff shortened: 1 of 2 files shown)


jobserver/models/user.py
```diff
@@ -48,6 +48,17 @@ def order_by_name(self):
...
+    def create_superuser(self, email, password, **extra_fields):
+        username = email
+        return super().create_superuser(
+            username,
+            email,
+            password,
+            is_staff=True,
+            is_superuser=True,
+            roles=[CoreDeveloper],
+        )
```
iaindillingham, line 60: Django sets `is_staff` and `is_superuser` to `True` by default and raises a `ValueError` if either are `False`, so we need not set them here.[^1] Whilst setting them here makes no difference, functionally, it did lead me to ask "Why are we setting them here?"

```suggestion
    def create_superuser(self, email, password, **extra_fields):
        username = email
        return super().create_superuser(
            username, email, password, roles=[CoreDeveloper]
        )
```

[^1]: https://github.com/django/django/blob/428d06ccef09e70bcef9869c5a9404863b2fc7d8/django/contrib/auth/models.py#L163-L172

iaindillingham (approved):
Looks good! One comment about not needing to set a couple of arguments to `create_superuser`. If you haven't already, then consider combining the two commits in this PR.

jobserver/models/user.py
Jongmassey: That seems obvious now you mention it!

PR #4130, opened 2024-02-15
Add link to github diff on JobRequest detail page

### LINKED ISSUE #4034: Link to diff between code for JobRequests
We want to help users understand what's changed in their code between jobs.  We store the commit sha for the latest commit on the configured branch of the Workspace as part of the JobRequest a user creates.  When job-runner picks up this request it uses that sha to check out the expected version of the code for a job to run.

Since we know the sha for all jobs via their JobRequest we can show a link to GitHub's compare view on both the JobDetail and JobRequestDetail pages comparing the current JobRequest with the previous one.

As part of this ticket, it would be useful to look into the feasibility of extending the comparison for JobRequests which have failed, back to the most recently successful request.

[Slack conversation](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1706273317250379)

Jongmassey:
fixes #4034


templates/job/detail.html
```diff
@@ -185,6 +185,16 @@ <h1 class="min-w-0 text-3xl break-words md:text-4xl font-bold text-slate-900">
           -
           {% endif %}
         {% /description_item %}
+
+        {% #description_item title="Git commit hash" stacked=True %}
+          {% link href=job.job_request.get_repo_url text=job.job_request.sha|slice:7|default:"-" %}
+        {% /description_item %}
+
+        {% if compare_url %}
+          {% #description_item title="Compare code" stacked=True %}
```
sebbacon, line 194: Does this need to say something more explicit like "show code differences between this run and the last one"? (but put more elegantly!)

Jongmassey:
Job detail screenshot 

image


Job request detail screenshot

image

templates/job/detail.html
Jongmassey: yes, certainly could be made clearer, I'll have a think of some suitable wording

Jongmassey:
>Did you consider this? IMO this would deliver the most value.
> OR a feature that allows you to diff two arbitrary jobrequests rather than just sequential ones.


Yes, I did think about it but I thought I'd try the most basic implementation first just to get my around django and job server.

I think having both "Compare to last run" and "Compare to last successful run" links wouldn't be too hard to do and could provide good value.

The arbitrary commit comparison would require more extensive UI changes.

sebbacon:
> I think having both "Compare to last run" and "Compare to last successful run" links wouldn't be too hard to do and could provide good value.
> 
> The arbitrary commit comparison would require more extensive UI changes.

Yes. 

I think ideally we would add the "compare to last successful run" as part of this work, while we're thinking about it.  If you choose not to, we should capture it in a new issue, as I don't want to lose it as a feature request. I can see it being super useful for tech support, and also gradually educating users that they can do the same.

jobserver/models/job_request.py
```diff
@@ -40,6 +40,19 @@ def with_started_at(self):
         )
 
 
+class JobRequestManager(models.Manager.from_queryset(JobRequestQuerySet)):
+    use_in_migrations = True
```
iaindillingham, line 44: I'm unsure of the pros/cons of using this manager in migrations. Could you check with a Djangonaut? However, if we do use it in migrations, then we should create a migration, which we should include in the same commit. Run

```sh
python manage.py makemigrations --dry-run
```

to see what we would create.

George have migrations more meaningful names than the defaults. I'd suggest we do the same.

templates/job_request/detail.html
```diff
@@ -298,6 +298,14 @@ <h1 class="text-3xl tracking-tight break-words font-bold text-slate-900 sm:text-
         </dl>
       {% /card %}
 
+      {% if compare_url %}
+        {% #card title="Previous Job Request" %}
```
iaindillingham, line 302: For consistency with the page title...

```suggestion
        {% #card title="Previous job request" %}
```

iaindillingham: That said, I think I'd

* Not make the card conditional. (You may wish to add an `else`.)
* Make the card's purpose clearer

```suggestion
      {% #card title="Code comparison" %}
        {% if compare_url %}
          {% #list_group small=True %}
              {% #list_group_item href=compare_url %}
                Compare this job request's code to the previous job request's code
              {% /list_group_item %}
          {% /list_group %}
        {% endif %}
      {% /card %}
```

templates/job/detail.html
iaindillingham: I've made a couple of suggestions in a separate comment. I think that on this page it's especially important to emphasize that _all_ the code is being compared, not just the code that's relevant to a job. (I'm not really sure whether this page should have this link, as we know the relationship between a job and a job request can confuse users.)

tests/unit/jobserver/models/test_job_request.py
```diff
@@ -174,6 +175,34 @@ def test_jobrequest_num_completed_success():
...
+
+
+def test_jobrequest_previous_url():
+    workspace = WorkspaceFactory()
+    backend = BackendFactory()
+
+    a = JobRequestFactory(workspace=workspace, backend=backend)
+    a.sha = "abc1234"
+    b = JobRequestFactory(workspace=workspace, backend=backend)
+    b.sha = "def5678"
```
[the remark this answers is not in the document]
Jongmassey: good point, I don't think I'd initially groked that I could add those to the factory ctor

tests/unit/jobserver/models/test_job_request.py
```diff
@@ -174,6 +175,34 @@ def test_jobrequest_num_completed_success():
...
 
 
+def test_jobrequest_previous_exists():
+    workspace = WorkspaceFactory()
+    backend = BackendFactory()
+
+    a = JobRequestFactory(workspace=workspace, backend=backend)
+    b = JobRequestFactory(workspace=workspace, backend=backend)
+    previous = JobRequest.objects.previous(b)
+    assert previous == a
```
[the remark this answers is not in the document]
Jongmassey: d'oh!

Jongmassey:
The wording of these feels clunky but I'm struggling to make it better. @iaindillingham  and @sebbacon  raised very good points that it needs to be clear on the Job detail page that this is compare the code for the whole workspace, not just the action in question.

jobserver/models/job_request.py
lucyb: I agree with what Iain has suggested.

The `use_in_migrations` option makes the manager available in RunPython operations within a migration ([docs](https://docs.djangoproject.com/en/5.0/topics/migrations/#model-managers)). We don't need to use this manager within a migration right now and don't use RunPython operations that often, but I'd probably go with the principle of least surprise and make it available anyway. Otherwise, it might not be clear to someone in future why it's not available and we probably do want managers to be available in migrations generally.

sebbacon:
I suspect we can help the user understand what will happen when they click partly with a little bit of UX tinkering of the sort @tomodwyer excels at?

Just spitballing but something like:

<green github icon> "View all repo changes since last execution on this backend"
<red github icon> "View all repo changes since last successful execution on this backend"

jobserver/models/job_request.py
Jongmassey: On closer inspection, it seems like something very similar happens in `0007_add_User_order_by_name.py` so I'll press on and add it

templates/job/detail.html
```diff
@@ -186,17 +188,22 @@ <h1 class="min-w-0 text-3xl break-words md:text-4xl font-bold text-slate-900">
...
+        {% #description_item title="Code comparison" %}
           {% if code_compare_url %}
-            {% link href=code_compare_url text="Compare workspace code to when the last time this action ran on this backend" %}
-            {% if code_compare_succeeded_url %}
-              <br/>
-              {% link href=code_compare_succeeded_url text="Compare workspace code to when the last time this action ran successfully on this backend" %}
-            {% endif %}
+            <ul class="list-disc ml-4 space-y-1">
+              <li>
+                {% link href=code_compare_url text="Compare workspace code to the previous action run on "|add:backendName %}
```
Jongmassey, line 199: I'm not 100% on this now I think about it. The "previous action" might not have been this action, and the comparison URL is specifically for the commits for this run of this action vs the last run of **this** action

tomodwyer: > Compare workspace code to the previous run of this action on [...]

Or is that even more confusing? "this action" we could text replace the action name?

Jongmassey: Oh I think you're onto something with putting the action name in there. That feels pretty unambiguous to me.

Jongmassey: @sebbacon  thoughts on  "Compare workspace code to the previous _my_nifty_action_ run on TPP" or similar?

Jongmassey: image


too much text 👎  It's slightly better if I manually fettle the html to make the action name a monospaced code block but can't seem to get that to place nice with the custom `link` component. Even then it's not great. Some action names can be very long

Jongmassey: image

I think I prefer this simple change

Jongmassey:
With thanks to @tomodwyer  I think most of the bikeshedding re wording is resolved. [Latest screenshots](https://github.com/opensafely-core/job-server/pull/4130#issuecomment-1963800900) updated.

jobserver/models/job_request.py
```diff
@@ -43,17 +43,17 @@ def with_started_at(self):
...
+    def previous(self, job_request, succeeded=None):
+        workspace_backend_job_requests = super().filter(
+            workspace=job_request.workspace,
+            backend=job_request.backend,
+            id__lt=job_request.id,
         )
+        if succeeded:
+            workspace_backend_job_requests = workspace_backend_job_requests.annotate(
+                min_status=Min("jobs__status")
+            ).filter(min_status="succeeded")
```
iaindillingham, line 55: You're going to have to explain to me what you're doing with `Min` here, because I've read it, run it in the shell, and I can't shake the feeling that it's a hack.

jobserver/models/job_request.py
```diff
@@ -43,17 +43,17 @@ def with_started_at(self):
...
-        return (
-            super()
-            .filter(
-                workspace=job_request.workspace,
-                backend=job_request.backend,
-                id__lt=job_request.id,
-            )
-            .order_by("created_at")
-            .last()
+    def previous(self, job_request, succeeded=None):
```
iaindillingham, line 46: If I passed `succeeded=False`, what would this method return? Rather than parameterizing this method, consider adding another method.

templates/job_request/detail.html
```diff
@@ -304,6 +304,11 @@ <h1 class="text-3xl tracking-tight break-words font-bold text-slate-900 sm:text-
             {% #list_group_item href=code_compare_url %}
               Compare to the previous Job Request
             {% /list_group_item %}
+            {% if code_compare_succeeded_url %}
+              {% #list_group_item href=code_compare_succeeded_url %}
+                Compare to previous successful Job Request
```
iaindillingham, line 309:
```suggestion
                Compare to the previous successful Job Request
```

jobserver/models/job.py
```diff
@@ -16,21 +16,18 @@
...
-            super()
-            .filter(
-                job_request__workspace=workspace,
-                job_request__backend=backend,
-                action=action,
-                id__lt=job.id,
-            )
-            .order_by("created_at")
-            .last()
+    def previous(self, job, succeeded=None):
```
iaindillingham, line 19: As before, what would this method return if `succeeded=False`?

templates/job/detail.html
```diff
@@ -198,6 +198,11 @@ <h1 class="min-w-0 text-3xl break-words md:text-4xl font-bold text-slate-900">
               <li>
                 {% link href=code_compare_url text="Compare workspace code to this action's previous run on "|add:backendName %}
               </li>
+              {% if code_compare_succeeded_url %}
+                <li>
+                  {% link href=code_compare_succeeded_url text="Compare workspace code to this action's previous successful run on "|add:backendName %}
```
iaindillingham, line 203:
```suggestion
                  {% link href=code_compare_succeeded_url text="Compare workspace code to this action's previous successful run on the "|add:backendName %} backend
```

iaindillingham (approved):
Thank you for reworking this PR. I've made several comments that I'd like you to consider; the most important relate to the use of `Min` and `succeeded=False`. I'm going to approve, because requesting changes feels heavy-handed (and I've requested changes once already). If you'd like me to spend a bit of time on `JobRequestManager.previous`, then please ask.

jobserver/models/job_request.py
```diff
@@ -40,6 +40,22 @@ def with_started_at(self):
...
+class JobRequestManager(models.Manager.from_queryset(JobRequestQuerySet)):
+    use_in_migrations = True
+
+    def previous(self, job_request):
+        return (
+            super()
+            .filter(
+                workspace=job_request.workspace,
+                backend=job_request.backend,
+                id__lt=job_request.id,
```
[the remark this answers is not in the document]
Jongmassey: It's defined as an integer `AutoField` - https://github.com/opensafely-core/job-server/blob/b5b1a9dc3bd2322261a28c338e6b21a2a3d64750/jobserver/migrations/0001_initial.py#L811-L817 under what circumstances would this not be monotonic? I chose it as the sorting key as it's the PK and so should be far faster than using `created_at`

jobserver/models/job_request.py
Jongmassey: the values for `job.status` are ("failed", "pending", "running", "succeeded"). An efficient way of answering the question: "are all of the status values for all of this `JobRequest`'s `Job`s equal to 'succeeded'?" is to take the minimum (by alphabetical sort) of the `Job`s' `status`es. Which is to say: `min("failed", "pending", "running", "succeeded") == "succeeded"`

This could be made more robust by adding a `CheckConstraint` to that field to enforce the fact those are indeed the only permissible values or by 

```python
workspace_backend_job_requests = workspace_backend_job_requests.annotate(
                min_status=Min("jobs__status"), max_status=Max("jobs__status")            
               ).filter(min_status="succeeded", max_status="succeeded")
```

such that if "zzz" was added as a status then `(min("succeeded","zzz")=="succeeded")&(max("succeeded","zzz")=="succeeded")==False`


To use the logic in the `@property` of `Job.status` involves running the query to get all the `JobRequests` prior to this one for this workspace/backend, then filtering in memory for the "succeeded" ones. By using the aggregate(s) above, this filtering can be pushed into the database query.

tests/unit/jobserver/models/test_job_request.py
```diff
@@ -174,6 +175,33 @@ def test_jobrequest_num_completed_success():
...
+    assert previous is None
+
+
+def test_jobrequest_previous_url():
+    workspace = WorkspaceFactory()
+    backend = BackendFactory()
+
+    a = JobRequestFactory(workspace=workspace, backend=backend, sha="abc1234")
+    b = JobRequestFactory(workspace=workspace, backend=backend, sha="def5678")
+    comp_url = workspace.repo.get_compare_url(b.sha, a.sha)
```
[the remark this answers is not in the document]
Jongmassey: comparing latter to previous is how my brain works for this, but clearly that's not universal?

jobserver/models/job_request.py
Jongmassey: see https://github.com/opensafely-core/job-server/pull/4130/commits/35ccbf88e41d832ac13566c7269675e8ef952ed4

PR #4153, opened 2024-02-22
Show lead org as default for projects

### LINKED ISSUE #4136: Show lead org as default for projects
When we moved over to supporting many-to-many organisations and projects we stopped displayed the organisation in various places around the site, including on https://jobs.opensafely.org/projects/.

We can get around this by adding an `org` property onto the Project model, so that templates referencing `{{ project.org }}` will return the expected result.

Note: check the performance of a few pages when making this change (the query count and timings in the django debug toolbar can help here).

iaindillingham:
`Project.org` was replaced by `Project.orgs` when we moved from a one-to-many to a many-to-many relationship between orgs and projects. However, we need `Project.org` in several locations. For convenience, it makes sense to show the lead org in these locations.

We expose `Project.org` as a property, meaning that an extra query is executed each time a project's lead org is requested. Hooking properties into [`prefetch_related`][1], which would reduce the number of extra queries, is non-trivial. Should we bother?

The answer depends on how many extra queries we expect. Let's take the `yours.ProjectList` view as an example
(https://jobs.opensafely.org/projects/). At present, the maximum number of projects per user is 20 - held jointly by @alexwalkerepi and @LFISHER7. For these users, this view executes 26 queries: five for page setup, one for the list of projects, and 20 for the lead org for each project. Locally, these queries execute in roughly 40ms. If we don't expose the property, then this view executes six queries in roughly 10ms. So, the worst case for this view is a four-fold increase in execution time.

20 projects per user is the maximum, though. At present, the median is two projects per user (for users that are associated with at least one project). So, hooking the property into `prefetch_related` wouldn't change much for many users. We shouldn't bother!

[1]: https://docs.djangoproject.com/en/5.0/ref/models/querysets/#prefetch-related

Fixes #4136

### DIFF
--- a/jobserver/models/project.py
+++ b/jobserver/models/project.py
@@ -2,9 +2,10 @@ import structlog
 from django.db import models
 from django.db.models import Q
 from django.urls import reverse
-from django.utils import timezone
+from django.utils import functional, timezone
 from django.utils.text import slugify
 from furl import furl
+from zen_queries import queries_dangerously_enabled
 
 
 logger = structlog.get_logger(__name__)
@@ -197,3 +198,14 @@ class Project(models.Model):
             return self.name
 
         return f"{self.number} - {self.name}"
+
+    @functional.cached_property
+    def org(self):
+        # zen-queries wants to stop Django's ORM executing queries "dangerously"; that
+        # is, executing one query for a set of elements, and then one extra query for
+        # each element. However, it's hard to hook properties into prefetch_related and,
+        # hence, to avoid making queries dangerously. We feel the costs of executing
+        # queries dangerously for this property are less than those of hooking this
+        # property into prefetch_related, so we ignore zen-queries for this property.
+        with queries_dangerously_enabled():
+            return self.orgs.filter(collaborations__is_lead=True).first()
--- a/tests/unit/jobserver/models/test_project.py
+++ b/tests/unit/jobserver/models/test_project.py
@@ -5,7 +5,12 @@ from django.utils import timezone
 
 from jobserver.models import Project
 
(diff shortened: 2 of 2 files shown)


iaindillingham:
Thank you for spotting this, @lucyb. I'd not considered zen-queries 🙁 

I've used the `queries_dangerously_enabled` context manager within `Project.org` to allow the ORM to execute queries "dangerously" for this property. I've added a comment to explain the use of this context manager. The commit message explains why -- at present -- I think the costs of this approach are less than the costs of the alternative approach (hooking into `prefetch_related`). The _at present_ is why I feel this information is better located in the commit message than in the comment.

lucyb:
>I've used the queries_dangerously_enabled context manager 

I think this is a good, pragmatic decision. I appreciate that you've documented it (in particular the "why") for whoever comes along to it in the future too.

jobserver/models/project.py
```diff
@@ -197,3 +198,14 @@ def title(self):
             return self.name
 
         return f"{self.number} - {self.name}"
+
+    @property
```
lucyb, line 202: Just to be a pain (I should have thought about this on Friday), but I think you can improve the performance in some cases by using the Django `cached_property` [decorator](https://docs.djangoproject.com/en/5.0/ref/utils/#module-django.utils.functional) instead. It doesn't seem to make a difference in many cases, but for the [event log](http://localhost:8000/event-log/) page for an unauthenticated user it takes the number of queries down from 105 to 14 (approximately).

lucyb (approved):
I think this is good to go, but I've made one suggestion that would be worth considering before merging.

PR #4162, opened 2024-02-27
Explain what the audit log is to users

### LINKED ISSUE #4109: Add a message to the audit pages about what the log is (and isn't)

lucyb:
This provides explanatory text to users with context about what the audit log is and how it's expected to be used I've also included the date that the audit log went fully live, as I expect that will be useful to people in future.

Screenshot from 2024-02-27 18-45-40

Screenshot from 2024-02-27 18-45-35

Fixes #4109

### DIFF
--- a/templates/staff/project/audit_log.html
+++ b/templates/staff/project/audit_log.html
@@ -18,7 +18,11 @@
 {% endblock breadcrumbs %}
 
 {% block hero %}
-  {% staff_hero title="Audit log: "|add:project.title %}
+  {% fragment as hero_text %}
+      <p>The audit log shows changes to project membership from <date>12th Feb 2024</date>. It includes project members added, removed and any change to roles. </p>
+      <p>It's only visible to staff members and is used for internal record keeping purposes only.</p>
+  {% endfragment %}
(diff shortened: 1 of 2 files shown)


CLStables:
Could we change "It includes members added..." to "It includes project members added..."?

iaindillingham (approved):
This is certainly an improvement! If I could make one suggestion, it would be that "at present and is intended" is evasive. Will the audit log be visible to non-staff members in the future? Will it be used for unintended, but still agreed upon, purposes?

lucyb:
> This is certainly an improvement! If I could make one suggestion, it would be that "at present and is intended" is evasive. Will the audit log be visible to non-staff members in the future? Will it be used for unintended, but still agreed upon, purposes?

That's a good spot. Thank you. I'll change it. For the "at present" part I was thinking that we might show a list of people who were previously a member of a project on the project detail page. That might come from the audit log (or be populated initially from there). That wouldn't be the full audit log though and I believe the updated wording is still fair and accurate?

iaindillingham (approved):
LGTM 🚀

PR #4163, opened 2024-02-29
Explain how min/max is used to derive JobRequest status

Jongmassey:
Quick post-merge addendum to #4130  based comments from @inglesp

### DIFF
--- a/jobserver/models/job_request.py
+++ b/jobserver/models/job_request.py
@@ -54,7 +54,7 @@ class JobRequestManager(models.Manager.from_queryset(JobRequestQuerySet)):
             # if the minimum of the related Job's statuses is "succeeded" then they
             # all must be. There are no constraints on this field, however, so
             # checking the maximum as well ensure this is robust to new statuses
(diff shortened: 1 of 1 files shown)


inglesp (approved):
Thanks, but yuck!

jobserver/models/job_request.py
```diff
@@ -50,6 +50,11 @@ def previous(self, job_request, filter_succeeded=None):
             id__lt=job_request.id,
         )
         if filter_succeeded:
+            # "succeeded" is the last in an alphabetical sort of the statuses, so
+            # if the minimum of the related Job's statuses is "succeeded" then they
+            # all must be. There are no constraints on this field, however, so
+            # checking the maximum as well ensure this is robust to new statuses
+            # which would sort later thatn "succeeded"
```
madwort, line 57:
```suggestion
            # which would sort later than "succeeded"
```

Jongmassey: 🤦‍♂️

PR #4164, opened 2024-02-29
Update some UI components

rebkwok:
3 updates to UI components that have come up while using these in Airlock:

1) Pass through id for a button component that is a link type
It was surprising that the docs said `id` was an available attribute, but i couldn't get it to appear (because my button was of type=link)

2) There were a couple of places in the docs for form components where a component was documenting using a django form field, but didn't quite work with a real django form. The main one was using `choices=<field>.choices` for a `form_select` component, whereas with a real django form field it should be `choices=<field>.field.choices`.  I've replaced the dicts that pretended to be django form fields with an actual django form.

3) Updated `form_radios` - only did this one because there was a heading for it in the docs without any content. It needed some minimal changes to make it more generic. We're currently stealing all of job-server's components for Airlock, so if we wanted to use a radio select element, the hard-coded ids would be a problem. Have checked that the create job request page is the only place it's used in job-server, and that everything looks the same after these changes.

### DIFF
--- a/templates/_components/index.html
+++ b/templates/_components/index.html
@@ -332,6 +332,14 @@
       <h3>Checkbox</h3>
       <ul>
         <li><code>checked</code>: [boolean] - if true, checkbox is checked</li>
+        <li><code>field</code>: a Django form field
+          <ul>
+            <li><a href="https://devdocs.io/django~4.1/ref/forms/api#django.forms.Form.auto_id"><code>auto_id</code></a></li>
+            <li><a href="https://devdocs.io/django~4.1/ref/forms/api#django.forms.BoundField.html_name"><code>html_name</code></a></li>
+            <li><a href="https://devdocs.io/django~4.1/ref/forms/api#django.forms.BoundField.id_for_label"><code>id_for_label</code></a></li>
+            <li><a href="https://devdocs.io/django~4.1/ref/forms/api#django.forms.BoundField.label"><code>label</code></a></li>
+          </ul>
+        </li>
         <li><code>custom_field</code>: [boolean] - set to true if field does not directly map to a Django Form field</li>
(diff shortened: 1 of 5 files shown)


templates/_components/index.html
```diff
@@ -447,13 +455,36 @@ <h3>Radio</h3>
...
+        <li><code>id</code>: [string] - set the HTML id for the select element</li>
+        <li><code>label</code>: [string] - overwrite the label passed from Django</li>
+        <li><code>name</code>: [string] - set the HTML name for the select element</li>
+        <li><a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required"><code>required</code></a></li>
+        <li><code>selected</code>: [string] - matches value of choice for preselected option</li>
+      </ul>
+      <div class="not-prose flex flex-col gap-y-8 relative w-screen max-w-5xl left-1/2 -translate-x-1/2 p-8 bg-slate-200 border border-slate-500 rounded">
+        {% form_radios class="w-full max-w-lg mx-auto" label="Choose a language" required=True field=example_form.example_radios choices=example_form.example_radios.field.choices selected=example_form.example_radios.value %}
+        <pre><code class="language-django">{% form_radios class="w-full max-w-lg mx-auto" label="Choose a language" required=True field=example_form.example_radios choices=example_form.example_radios.field.choices selected=example_form.example_radios.value %}{% endverbatim %}</code></pre>
```
tomodwyer, line 480:
```suggestion
        <pre><code class="language-django">{% verbatim %}{% form_radios class="w-full max-w-lg mx-auto" label="Choose a language" required=True field=example_form.example_radios choices=example_form.example_radios.field.choices selected=example_form.example_radios.value %}{% endverbatim %}</code></pre>
```

Missing the opening tag here

tomodwyer (approved):

PR #4165, opened 2024-02-29
Show most recently active workspaces on home/yours page

### LINKED ISSUE #4125: Show most recently active workspaces on yours page
I want to be able to quickly navigate to the workspace(s) i'm currently running jobs from. There is no guarantee that these appear in the 5 workspaces displayed on the home page. I currently do this by clicking on the recent job requests and navigating to the workspace in the breadcrumbs. 

A better solution would be to display the 5 most recently active workspaces (ie workspaces with most recently run jobs)

Example below. I'm currently running jobs exclusively in the flucats workspace, but there is no direct link to that workspace from the home page (except via the parent project, but its the same number of clicks as going via a recent job request).

Screenshot 2024-02-13 111031

Jongmassey:
Fixes #4125


Jongmassey:
SQL query stats from debug toolbar on authenticated homepage


2ecfcda5 (before changes) - 29 queries in 60.33ms
7beaef9 (sort by most recent `JobRequest`) - 30 queries in 73ms
705368b (sort by most recent `JobRequest`, exclude those without)- 30 queries in 99ms
87b9844 (sort by most recent activity) - 30 queries in 76ms

tests/unit/jobserver/models/test_workspace.py
```diff
@@ -416,3 +416,31 @@ def test_workspace_is_interactive():
...
+
+    w = Workspace.objects.with_most_recent_activity_at().get(pk=workspace.pk)
+
+    assert w.most_recent_activity_at == w.updated_at
+
+
+def test_workspace_with_most_recent_activity_at_updated_and_jobrequest():
+    workspace = WorkspaceFactory()
+    JobRequestFactory(workspace=workspace)
+    workspace.purpose = "test"
```
lucyb, line 441: This is a really small nit, but this:
`workspace.purpose = "test"` 
could be changed to something like this, to make the intent of the test a bit clearer:
`workspace.purpose = "update the workspace"`

lucyb (approved):
Nice Django-ing!

tests/unit/jobserver/models/test_workspace.py
Jongmassey: :+1:

PR #4168, opened 2024-03-01
Assert on num db queries for views and templates

lucyb:
Our index page, whether logged in or not, makes quite a few calls to the database, yet we want it to remain fairly performant. This adds some assertions into the existing unit tests that check the view. It also adds some tests using the Django Client to render the template and assert on the number of database queries we're making using both the view and the template. That way, if we accidentally change this behaviour by updating the template, we'll know. We may want to expand these tests over time to ensure we're rendering the template correctly or to check the number of db queries in different scenarios.

### DIFF
--- a/tests/unit/jobserver/views/test_index.py
+++ b/tests/unit/jobserver/views/test_index.py
@@ -12,7 +12,11 @@ from ....factories import (
 
 
 def test_index_authenticated(
-    rf, complete_application, project_membership, project_memberships
+    rf,
+    django_assert_num_queries,
+    complete_application,
+    project_membership,
+    project_memberships,
 ):
     user = UserFactory()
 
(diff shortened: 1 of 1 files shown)


iaindillingham (approved):
I think this is a step in the right direction. If we increase the number of queries, then we see a test failure.

As this is an example of the experiment you discussed in Slack,[^1] I thought about it in the context of removing zen-queries. Clearly, it's _still_ a step in the right direction: not only do we see a test failure, but we also see it before a user sees a 500 error.

With this approach, it's hard to relate 1. the number of queries we're asserting are made, to 2. the associated views and models. Consequently, I wondered whether zen-queries offered some help. It doesn't 🙁 To put that another way, whichever approach we take, it's still hard to work back from a test failure to a change to a view or model. (Of course, we'd expect the test to fail after we made the change, so cause and effect should be clear, but that isn't always the case.) However, this approach involves less code.

[^1]: https://bennettoxford.slack.com/archives/C069SADHP1Q/p1709043464199309

lucyb:
> I think this is a step in the right direction. 

Thank you. This is what I'd hoped someone would say. Hopefully we can start to build on this approach in other areas of the Codebase.

> it's still hard to work back from a test failure to a change to a view or model.

That's a good observation. It's quite a coarse-grained test. I hope it's fast enough that people will be running it frequently though, so I hope that won't be a problem. If it does prove to be difficult to understand why a test failure is happening, we should stop and ask why, and then figure out what needs to change.

PR #4176, opened 2024-03-05
Add djhtml to format templates

Jongmassey:


iaindillingham:
Could we introduce this whilst also maintaining `git blame`? Black's documentation has a section entitled "[Introducing Black to your project](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html)". We should do something similar, here.

justfile
```diff
@@ -156,7 +156,7 @@ ruff *args=".": devenv
 
 # run the various dev checks but does not change any files
 check: black django-upgrade ruff
-
+    $BIN/djhtml --tabwidth 2 --check templates
```
iaindillingham, line 159: For consistency with the `black` and `ruff` recipes, we should have a `djhtml` recipe

```
djhtml *args="templates": devenv
    $BIN/djhtml --tabwidth 2 --check {{ args }}
```

which we then call from the `check` recipe

```suggestion
check: black django-upgrade ruff djhtml
```

templates/_components/article/header.html
iaindillingham, line 1: It would be good if @tomodwyer could cast his eye over the changes in this commit. Is `djhtml` doing the right thing? Do we have a different understanding of the right thing to `djhtml`?

Assuming we're okay with the changes, then we should add a file called `.git-blame-ignore-revs` to the root directory, which should contain the full SHA of this commit as well as a comment to explain why we're asking `git blame` to ignore this commit. We should then let everyone in the team know that they should configure `git` to ignore SHAs listed in this file. More details in "[Introducing Black to your project](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html)".

iaindillingham (commented):
Couple more in-place comments. I think there's a bit more work to do before merging. I'll ask the team on Slack to consider the `git blame` question.

justfile
Jongmassey: excellent suggestion

templates/_components/article/header.html
Jongmassey: I've added the ignore file, set my local config and tested and it appears to work - could you check with the latest commits I've just pushed?

tomodwyer: So for clarity, we should get everyone to run `git config blame.ignoreRevsFile .git-blame-ignore-revs` after this PR is merged? That works locally for me.

iaindillingham: That's correct, @tomodwyer. (Alternatively, you may want to enable this setting globally with `--global`.)

@evansd suggested on Slack[^1] that he felt the incremental approach was better for job-server, though. I'm also unclear whether you feel the templates are better having been reformatted, or just different.

[^1]: https://bennettoxford.slack.com/archives/C031L3FTFC0/p1709801952150089?thread_ts=1709801688.008049&cid=C031L3FTFC0

Jongmassey: One thing I'm uncertain of re the incremental approach is: if we want to use this formatter for all new templates/changed templates going forward it makes sense to add it to the `check` recipe to ensure this; surely then all the unchanged templates would fail this check?

iaindillingham: We'd have to pass an include list to `djhtml`. We do something similar -- although it's an exclude list rather than an include list -- in the `test-dev` recipe.

Issue #4128, opened 2024-02-14
Spike: Rationalize logging event-types


iaindillingham:
It's likely that we'll be asked to log more event-types than we do at present. (We're talking specifically about the audit log, rather than logging in general.)

https://github.com/opensafely-core/job-server/blob/935ae263364a2bf0af34f05ff7c3375b5706ef4b/jobserver/models/auditable_event.py#L14-L19

Before we do, we should consider whether:

* we can reduce the amount of code that's needed to log more event-types
* we can make logging more event-types safer (i.e. whether we can make it harder to bypass `Model.save`, `QuerySet.create`, etc.)
* we can be more consistent with our use of fixtures and factories in our tests (we have a `project_membership` fixture but an `OrgMembershipFactory`)

@lucyb and I feel that having a `commands` module makes sense. We also feel that calling it `actions`, to avoid confusion with Django's management commands and for consistency with opencodelists (see [`opencodelists/opencodelists/actions.py`][1]) makes sense, too. @inglesp has written up some [design decisions][2], which will be helpful, as will [this Slack thread][3], where @bloodearnest points out the similarity with CQRS.

[1]: https://github.com/opensafely-core/opencodelists/blob/main/opencodelists/actions.py
[2]: https://docs.google.com/document/d/1Pln5DR3qeDKCOJm9vmMj0T8-BWnA_e1DENcUZVX8a_I/edit?pli=1
[3]: https://bennettoxford.slack.com/archives/C069YDR4NCA/p1707830780559509

iaindillingham:
I've spent some time thinking about how best to structure the commands in `jobserver.commands`.

One idea was to replace each module with a class, such that the functions in each module became the methods of each class. The classes would inherit from an ABC (an [Abstract Base Class][2]), meaning they would implement a common set of methods. If they didn't, then a `TypeError` would be raised at runtime. Having read "[Interfaces, Protocols, and ABCs][3]" in Fluent Python, I felt that some of these methods could be concrete, which would reduce the repetition within `jobserver.commands`.[^1]

For example, the `org_members` module, which has an `update_roles` function, would be replaced by an `org_members` class, which would have an `update_roles` method. Because this method would be static, and with a small change to `jobserver.commands.__init__`, client code would be unaware of the replacement: calling `org_members.update_roles` (module/function) would be the same as calling `org_members.update_roles` (class/method).

[...]

PR #4179, opened 2024-03-06
Rationalize logging event-types

iaindillingham:
This is part one of a two-part attempt to rationalize logging event-types (#4128). Here, we add `ImmutableModelMixin` and `ImmutableManager` to prevent us from accidentally mutating `ProjectMembership` outside `jobserver.commands.project_members` and, hence, outside an audited workflow.

Adding these models has already highlighted one location where we mutated `ProjectMembership` in this way. This location was removed in #4178.

The tests should help us add, delete, and update existing models safely. It is for this reason that we add `ImmutableError`: it's hard to test that an existing model has been updated from a mutable to an immutable model without it. `ImmutableError` is a child of `TypeError` because

> This exception may be raised by user code to indicate that an attempted operation on an object is not supported, and is not meant to be.[^1]

I've got a follow up PR in mind that will complete #4128. It will involve some refactoring, to make the code in `jobserver.commands` easier to generalize to other models and other logging event-types.

[^1]: https://docs.python.org/3.12/library/exceptions.html#TypeError

### DIFF
--- a/jobserver/model_utils.py
+++ b/jobserver/model_utils.py
@@ -1,4 +1,19 @@
+from django.apps import apps
 from django.db import models
+
+
+_SKIP_APP_LABELS = {"auth", "contenttypes", "sessions", "social_django"}
+
+
+def get_models():
+    """
+    Return a list of all installed, project-specific models.
+    """
+    return [
+        model
+        for model in apps.get_models()
+        if model._meta.app_label not in _SKIP_APP_LABELS
+    ]
 
 
 class ImmutableError(TypeError):
--- a/jobserver/model_utils.py
+++ b/jobserver/model_utils.py
@@ -1,0 +1,19 @@
+class ImmutableError(TypeError):
+    pass
+
+
+def _raise_error(model):
+    model_name = model._meta.label.split(".")[1]
+    raise ImmutableError(f"Direct access to {model_name} is disabled")
+
+
+class ImmutableModelMixin:
+    def delete(self, *args, override=False, **kwargs):
+        if not override:
+            _raise_error(self)
+        super().delete(*args, **kwargs)
+
+    def save(self, *args, override=False, **kwargs):
+        if not override:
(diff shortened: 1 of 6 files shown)


jobserver/model_utils.py
```diff
@@ -0,0 +1,65 @@
...
+        super().delete(*args, **kwargs)
+
+    def save(self, *args, override=False, **kwargs):
+        if not override:
+            _raise_error(self)
+        super().save(*args, **kwargs)
+
+
+class ImmutableQuerySet(models.QuerySet):
+    # We don't override get_or_create, because we don't want to disable the get part.
```
lucyb, line 42: This is an extremely useful comment!

lucyb (approved):
This is a significant improvement.

iaindillingham:
I've omitted a small, but important, aspect of this PR @lucyb. Given that the only immutable model at present is `ProjectMembership`, the following function will direct a user to `commands.projectmembership`:

https://github.com/opensafely-core/job-server/blob/3af19e42e07bddf1180990f0c0ae501639b6ac81/jobserver/model_utils.py#L23-L26

However, the module is actually `commands.project_members`. The naming convention for commands modules is something like "the snake-case name of the participants in the model". It's hard to derive the name of a commands module from the name of a model, so I'd like to change the convention. I'd like to name commands modules from their models as per the previous function.

This would mean:

| From                 | To.                    |
|----------------------|------------------------|
| `org_members.py`     | `orgmembership.py`     |
| `project_members.py` | `projectmembership.py` |
| `projects.py`        | `project.py`           |
| `users.py`           | `user.py`              |

An alternative would be to name commands modules as follows:

```py
Model._meta.verbose_name_plural.replace(" ", "_")
```

`OrgMembership` would give `org_memberships`, and we would rename `org_members.py` to `org_memberships.py`. However, I don't think this adds much, other than slightly more code.

PR #4192, opened 2024-03-12
Correct naming for custom Job Manager migration

Jongmassey:
Fixes previous copy-paste error


StevenMaude:
This comment is maybe more useful for clarifying my own understanding than anything else: I don't know Django well.

From searching around, Django tracks applied migrations through a `django_migrations` database table, which apparently stores the name of the migration.

So is there any issue with renaming the migration? I'll refer to the version in `main` as the "old" migration, and the one in this PR as the "new" migration, though they are identical.

I'm thinking of a scenario where the "old" migration's already been run, and then we update the migration filenames. It's possible that the migration might already have been applied in production, and some developers may have already run it locally.

For local development: if we *create* a "new" migration after merging this PR, and the "old" has already been run, which filename will the "new" migration state its `dependencies` on? The filename for the "old" migration that may exist in the database already, or the filename for the "new" migration on disk?

For deployment: if we try and *apply* a new migration after merging this PR, and the old migration has already been run, the "new" won't be in the database. Will that cause any problems? Or will it just re-run the "new" migration with null effect, and leave rows referring to the "old" migration in the `django_migrations` table?

***

For what it's worth, [this post](https://stackoverflow.com/a/30427833) seems to suggest that it's OK to do so, although may require additional steps:

> This works, with a minor caveat: Django will no longer know that the renamed migration is applied.
>
> So the steps to renaming a migration are:
>
> 1. Rename the file.
> 1. Repoint any dependencies to the new file.
> 1. If the renamed migration was already applied, apply it again using `--fake`.
>
> If it's a brand new migration, 2 and 3 won't apply, and it's perfectly fine to rename the

PR #4199, opened 2024-03-14
Allow `OLD_SECRET_KEY` for secret key rotation

StevenMaude:
This PR allows the use of Django's `SECRET_KEY_FALLBACKS` by configuring the `OLD_SECRET_KEY` environment variable to use the previous key.

This avoids users all getting logged out at once. Instead, we can:

1. Leave the previous secret key active, side-by-side with the new key.
1. Give some time for active users to access the site running the new secret key which updates their sessions.
1. Remove the previous secret key entirely.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -70,7 +70,6 @@ There are three steps to this process.
 ### 1. Duplicate the existing secret key
 
 ```bash
-dokku config:get job-server SECRET_KEY
 dokku config:set job-server OLD_SECRET_KEY="$(dokku config:get job-server SECRET_KEY)"
 ```
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -56,3 +56,44 @@ dokku$ sudo mkdir -p /var/log/journal
 dokku$ dokku enter job-server
 container$ python manage.py sendtestemail me@myemail.org
 ```
+
+## Rotating the Django secret key
+
+If this is a *routine rotation* of the secret key
+— and not a compromised secret key —
+you can temporarily leave the secret key to be rotated alongside the new secret key.
+
+This allows for changing the secret key without logging out all users at once.
+
+There are three steps to this process.
+
+### 1. Duplicate the existing secret key
+
+```bash
+dokku config:get job-server SECRET_KEY
+dokku config:set job-server OLD_SECRET_KEY="$(dokku config:get job-server SECRET_KEY)"
+```
+
+### 2. Replace the existing secret key
+
+```bash
+# Prefix the command with a space to avoid saving the input to shell history:
+ dokku config:set job-server SECRET_KEY='xxx'
+```
+
+### 3. Remove the previous secret key
+
+A suitable expiry time is given by adding the `SESSION_COOKIE_AGE` duration
+to the time at which the secret key was replaced.
+Once that expiry time is reached,
+any sessions created before the secret key was replaced will have then expired anyway.
+
+If this is not set in the code,
+Django's default [`SESSION_COOKIE_AGE`](https://docs.djangoproject.com/en/dev/ref/settings/#session-cookie-age) is *two weeks*.
+
+Set a Slack reminder for this time,
+and then remove `OLD_SECRET_KEY` as soon as possible when reminded:
+
+```bash
+dokku config:unset job-server OLD_SECRET_KEY
+```
--- a/jobserver/settings.py
+++ b/jobserver/settings.py
@@ -34,6 +34,14 @@ BASE_DIR = Path(__file__).resolve().parent.parent
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = env.str("SECRET_KEY")
 
+# Optional fallback for rotating the secret key.
+# Any OLD_SECRET_KEY that is added should then be removed
+# after the time in SESSION_COOKIE_AGE elapses.
+# Refer to DEVELOPERS.md for guidance.
+OLD_SECRET_KEY = env.str("OLD_SECRET_KEY", default=None)
+if OLD_SECRET_KEY is not None:
+    SECRET_KEY_FALLBACKS = [OLD_SECRET_KEY]
+
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = env.bool("DEBUG", default=False)
 


jobserver/settings.py
```diff
@@ -34,6 +34,21 @@
...
 
+
+def configure_secret_key_fallbacks():
+    """Return a list containing a single secret key for SECRET_KEY_FALLBACKS."""
+    old_secret_key = env.str("OLD_SECRET_KEY", default=None)
+    if old_secret_key is not None:
+        return [old_secret_key]
+
+    # Empty list is the default Django value.
+    return []
```
StevenMaude, line 45: I'm not completely sure if the extra complexity is worth the test here.

The simplest way of doing this may be sufficient:

```diff
diff --git a/jobserver/settings.py b/jobserver/settings.py
index b4fad626..96e35836 100644
--- a/jobserver/settings.py
+++ b/jobserver/settings.py
@@ -34,6 +34,12 @@ BASE_DIR = Path(__file__).resolve().parent.parent
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = env.str("SECRET_KEY")
 
+# Optional fallback for rotating the secret key.
+# Any OLD_SECRET_KEY that is used should eventually be removed.
+OLD_SECRET_KEY = env.str("OLD_SECRET_KEY", default=None)
+if OLD_SECRET_KEY is not None:
+    SECRET_KEY_FALLBACKS = [OLD_SECRET_KEY]
+
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = env.bool("DEBUG", default=False)
```

That said, with it being a security configuration, it may make sense to test it behaves as expected.

The reason for the function is to make the change easier to test: otherwise, you'd have to import the whole `settings.py`. Even if you import the module `importlib` to avoid reloading the existing `settings.py` in the global namespace, there are still other side-effects, because `settings.py` doesn't only set module-level variables. It also runs `initialise_sentry()`.

INSTALL.md
```diff
@@ -56,3 +56,19 @@ dokku$ sudo mkdir -p /var/log/journal
...
+as opposed to a compromised secret key,
+you can temporarily leave the secret key to be replaced active:
+
+```bash
+# Store the existing SECRET_KEY as OLD_SECRET_KEY.
+dokku config:get job-server SECRET_KEY
+dokku config:set job-server OLD_SECRET_KEY='xxx'
+# Update the SECRET_KEY to a new, secure value.
+dokku config:set job-server SECRET_KEY='xxx'
+# When sufficient time has elapsed, remove the previous key.
```
iaindillingham, line 72: This troubles me. How much time is sufficient time? How would we know when it had elapsed? I think we need to pick a period in answer to the first question; what does the session expiry policy say? Could we set a reminder in Slack in answer to the second question?

iaindillingham (commented):
I'm going to ask an experienced Djangonaut like @inglesp to help review this PR.

jobserver/settings.py
inglesp: I agree.

This seems to be in a function to help with testing, but our tests don't actually test the behaviour we care about, which is that a user who's previously logged in when the key was X does not have to log in now that the key is Y.

INSTALL.md
```diff
@@ -56,3 +56,19 @@ dokku$ sudo mkdir -p /var/log/journal
...
+## Rotating the Django secret key
+
+If this is a routine rotation of the secret key,
+as opposed to a compromised secret key,
+you can temporarily leave the secret key to be replaced active:
+
+```bash
+# Store the existing SECRET_KEY as OLD_SECRET_KEY.
+dokku config:get job-server SECRET_KEY
+dokku config:set job-server OLD_SECRET_KEY='xxx'
```
inglesp, line 69: I bet you could even do something like `dokku config:set job-server OLD_SECRET_KEY=$(dokku config:get job-server SECRET_KEY)` and then there's no risk of secrets ending up in bash history.

jobserver/settings.py
iaindillingham: An excellent point; not one I had considered 🙂

INSTALL.md
StevenMaude: I wasn't sure of this either, which is why I originally didn't state anything.

The answer is actually given by the setting for [`SESSION_COOKIE_AGE`](https://docs.djangoproject.com/en/5.0/ref/settings/#session-cookie-age) — which we've not overridden, so it's the default of two weeks.

In case I missed anything, I also confirmed this by locally logging out, logging back in, and checking the most recent `expiry_date` in the `django_session` database table.

jobserver/settings.py
StevenMaude: > This seems to be in a function to help with testing, but our tests don't actually test the behaviour we care about, which is that a user who's previously logged in when the key was X does not have to log in now that the key is Y.

We'd then be testing Django's behaviour even more so, so the simple option is preferable.

INSTALL.md
evansd: FWIW, I think it's absolutely acceptable to roll this over sooner, if you prefer. We don't want to log people out if they're in the middle of doing something, but I think it's perfectly reasonable to log out anyone who hasn't interacted with the site in, say, 24 hours. On the other hand, there's not a lot a downside to keeping the fallback key around a bit longer if you'd prefer to do that.

INSTALL.md
```diff
@@ -56,3 +56,44 @@ dokku$ sudo mkdir -p /var/log/journal
...
+
+This allows for changing the secret key without logging out all users at once.
+
+There are three steps to this process.
+
+### 1. Duplicate the existing secret key
+
+```bash
+dokku config:get job-server SECRET_KEY
+dokku config:set job-server OLD_SECRET_KEY="$(dokku config:get job-server SECRET_KEY)"
```
iaindillingham, line 74: I may have misunderstood, but doesn't the get in the second line make the get in the first line unnecessary?

INSTALL.md
```diff
@@ -56,3 +56,44 @@ dokku$ sudo mkdir -p /var/log/journal
...
+A suitable expiry time is given by adding the `SESSION_COOKIE_AGE` duration
+to the time at which the secret key was replaced.
+Once that expiry time is reached,
+any sessions created before the secret key was replaced will have then expired anyway.
+
+If this is not set in the code,
+Django's default [`SESSION_COOKIE_AGE`](https://docs.djangoproject.com/en/dev/ref/settings/#session-cookie-age) is *two weeks*.
+
+Set a Slack reminder for this time,
+and then remove `OLD_SECRET_KEY` as soon as possible when reminded:
```
iaindillingham, line 95: Thank you for including this.

iaindillingham (approved):
Thank you for writing up the notes in `INSTALL.md`. They're going to be really helpful.

INSTALL.md
StevenMaude: Good spot; that's correct and I'll remove the first of the two lines.

PR #4220, opened 2024-03-20
Replace cli arg with env vars

lucyb:
This means we're no longer providing the password as part of the connection string on the command line and are instead passing it in as an environment variable. This stops the password from being accidentally reported in error logs, etc.

### DIFF
--- a/jobserver/jobs/hourly/dump_db.py
+++ b/jobserver/jobs/hourly/dump_db.py
@@ -2,18 +2,15 @@ import pathlib
 import subprocess
 import sys
 
+from django.conf import settings
 from django_extensions.management.jobs import HourlyJob
-from environs import Env
-
-
-env = Env()
 
 
 class Job(HourlyJob):
     help = "Dump the database to storage for copying to local dev environments"  # noqa: A003
 
     def execute(self):
-        database_url = env.str("DATABASE_URL")
+        db = settings.DATABASES["default"]
         output = pathlib.Path("/storage/jobserver.dump")
 
         if not output.exists():
@@ -30,6 +27,7 @@ class Job(HourlyJob):
                 "--no-acl",
                 "--no-owner",
                 f"--file={output}",
-                database_url,
+                f"postgresql://{db['USER']}@{db['HOST']}:{db['PORT']}/{db['NAME']}",
             ],
+            env={"PGPASSWORD": db["PASSWORD"]},
         )


dotenv-sample
```diff
@@ -4,7 +4,12 @@ ADMIN_USERS=
 
 # Note: this assumes docker dev db. You will need to change it if you want to
 # point another postgres
-DATABASE_URL=postgres://user:pass@localhost:6534/jobserver
+DATABASE_USER="user"
+DATABASE_PASSWORD="pass"
+DATABASE_DB="jobserver"
+DATABASE_HOST="localhost"
+DATABASE_PORT="6543"
+DATABASE_URL=postgres://${DATABASE_USER}:${DATABASE_PASSWORD}@${DATABASE_HOST}:${DATABASE_PORT}/${DATABASE_DB}
```
bloodearnest, line 12: Rather than split these out into separate environments, we could easily parse them out when we need them for this commnad?

urllib.parse.urlparse does this fine e.g.

```
from urllib.parse import urlparse

db = urlparse(settings.DATABASE_URL)
# db.username, db.password, db.hostname, db.port, db.path, etc
```

psycopg2 also has [Connection.dsn_parameters](https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.ConnectionInfo.dsn_parameters), but it looks like it doesn't include the password.

jobserver/jobs/hourly/dump_db.py
```diff
@@ -30,6 +29,12 @@ def execute(self):
                 "--no-acl",
                 "--no-owner",
                 f"--file={output}",
-                database_url,
             ],
+            env={
+                "PGUSER": env.str("DATABASE_USER"),
```
bloodearnest, line 34: Possibly only the password needs to be in the environment? Thats how I've [done this before](https://github.com/canonical-ols/talisker/blob/master/talisker/postgresql.py#L72) - its actually useful to have the rest of connection string visible when we error.

I think you can then do, using urlparse as above, something like:

```
from urlib.parse import urlparse

db = urlparse(config.DATABASE_URL)

subprocess.check_output(
  [
    "pg_dump"
    f"postgresql://{db.user}@{db.netloc}{db.path}"
    ...
  ],
  env={"PGPASS": db.password},
)
```

lucyb: Thanks, that's a much nicer solution! I didn't like what I'd done, but it was late so I was hoping that something better would occur today. I didn't realise that urlparse would work, I was thinking it would have to be a regexp, which I wanted to avoid. I'll give that solution a try..

bloodearnest (approved):

PR #4223, opened 2024-03-21
Clean up the otel config

bloodearnest:
The value PYTHONPATH made no sense. And its only required to be present
not set.  So I incorporated that in the code

Likewise, set a default DJANGO_SETTINGS_MODULE in the right place.


gunicorn.conf.py
```diff
@@ -26,6 +26,10 @@
 
 
 def post_fork(server, worker):
+    # opentelemetry initialisation needs these env vars to be set, so ensure they are
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "airlock.settings")
```
lucyb, line 30:
```suggestion
    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "jobserver.settings")
```

bloodearnest: :facepalm:

lucyb (commented):
For the env vars that have been removed, like `OTEL_EXPORTER_OTLP_ENDPOINT` do they also need to be removed in production or are they just not needed in development?

lucyb (approved):

bloodearnest:
> For the env vars that have been removed, like `OTEL_EXPORTER_OTLP_ENDPOINT` do they also need to be removed in production or are they just not needed in development?

Just not needed in developement. I removed it because a) it's set to that value by default anyway, and b) you shouldn't need to point it anywhere else in local development.

gunicorn.conf.py
bloodearnest: FTR, this didn't fail because it was already set correctly

Issue #4255, opened 2024-04-02
Remove the `org_invite_members` permission


iaindillingham:
The `org_invite_members` isn't used. We should remove it.

https://github.com/opensafely-core/job-server/blob/7de6a7d4843745131e335ede6f1b6e5fbda891f2/jobserver/authorization/permissions.py#L8

PR #4256, opened 2024-04-02
Strings to symbols

iaindillingham:
This PR replaces permissions encoded as strings with their corresponding symbols. (I hope that's the right word for them; if not, then sorry computer scientists 🙁.) For example, it replaces `"workspace_toggle_notifications"` with `permissions.workspace_toggle_notifications`. Doing so makes it much easier to look up where, and how, permissions are used.

I've not replaced strings with symbols when the strings are expected values in the test suite. This gives me greater confidence that I've not made an error, and also avoids having to come up with another name for the concept of permissions, as a variable with that name often clobbers the module with that name.

One permission is unused. For more information, see #4255.

### DIFF
--- a/applications/views.py
+++ b/applications/views.py
@@ -8,7 +8,7 @@ from django.utils import timezone
 from django.utils.decorators import method_decorator
 from django.views.generic import CreateView, RedirectView, UpdateView, View
 
-from jobserver.authorization import CoreDeveloper, has_permission, has_role
+from jobserver.authorization import CoreDeveloper, has_permission, has_role, permissions
 from jobserver.hash_utils import unhash_or_404
 from jobserver.slacks import notify_application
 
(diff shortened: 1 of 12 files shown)


StevenMaude:
Looking at the [19 permissions](https://github.com/opensafely-core/job-server/blob/7de6a7d4843745131e335ede6f1b6e5fbda891f2/jobserver/authorization/permissions.py) and the 17 updated in this PR, and the unused `org_invite_member` permission, that leaves `backend_manage`. `backend_manage` is used, but not compared as the other permissions changed here.

StevenMaude (approved):

StevenMaude:
While I'm here, I wonder if there's a reason the permissions aren't stored as an `Enum`, but that's not really a concern for this PR.

iaindillingham:
> While I'm here, I wonder if there's a reason the permissions aren't stored as an `Enum`, but that's not really a concern for this PR.

I thought the same, but 🤷🏻‍♂️

Issue #4261, opened 2024-04-05
Decouple production roles from tests


iaindillingham:
We often test production roles, when we want to test permissions. This is because permissions are assigned to production roles. However, doing so couples our tests to production roles. We should instead create test roles -- one per permission -- for testing permissions.

iaindillingham:
I've a draft PR that I'd like some feedback on before continuing with this issue. (I asked for feedback in a Slack thread,[^1] but only because comments on PRs can become unwieldy.) I'm going to move this issue to "In Review" for now, with the expectation that it will be moved to "Next" or "Later" when we've more time.

[^1]: https://bennettoxford.slack.com/archives/C069SADHP1Q/p1712842544456739

Issue #4260, opened 2024-04-05
Apply permissions to views consistently


iaindillingham:
We don't have a consistent approach for applying permissions to views. Buckle up.

In some cases, we use the `require_permission` function as a decorator. For example:

https://github.com/opensafely-core/job-server/blob/5c809f65b9dd916737e919a83bb6639deaa5e13d/staff/views/users.py#L100-L103

In other cases, we similarly use the `require_role` function. (We use this function even though it has been asserted that we shouldn't test roles, but we should test permissions.) For example:

https://github.com/opensafely-core/job-server/blob/5c809f65b9dd916737e919a83bb6639deaa5e13d/staff/views/job_requests.py#L40-L45

Then there are the many other cases when we use neither of these functions, and instead use the `has_permission` function, raising `PermissionDenied`. For example:

https://github.com/opensafely-core/job-server/blob/5c809f65b9dd916737e919a83bb6639deaa5e13d/jobserver/views/workspaces.py#L54-L59

And the one other case when we override the `get_object` method, although this is when the "permission" isn't a permission in the "roles and permissions" sense (should it be?).

https://github.com/opensafely-core/job-server/blob/5c809f65b9dd916737e919a83bb6639deaa5e13d/jobserver/views/projects.py#L211-L217

There may be more cases. I didn't exhaustively consider every time we raise `PermissionDenied`.

Not having a consistent approach makes it time-consuming and error-prone (not to mention frustrating) to apply a permission to a view.

Issue #4236, opened 2024-03-25
Restrict editing of project status and description


lucyb:

iaindillingham:
To generate the list of users that have been removed from projects:

```py
events = AuditableEvent.objects.filter(type=AuditableEvent.Type.PROJECT_MEMBER_REMOVED).order_by("target_user", "parent_id").values_list("target_user", "parent_id")

with open("project_member_removed.csv", "w", newline="") as f:
    fieldnames = ["username", "project_id"]
    writer = csv.writer(f)
    writer.writerow(fieldnames)
    writer.writerows(events)
```

iaindillingham:
I'm meeting with Juliet and Liam to discuss this change. Details below.

> Very soon, editing a project (i.e. the status and the status description) will be restricted to users with the Project Developer role either at the global level or at the project level.
>
> In preparation for this change, I'd like to show you what a user who has and doesn't have this role will see, if they try to edit a project. I'd also like to tell you about which users should be re-added to which projects, such that they appear in the list of researchers (spoiler: only Scott Walter and George Hickman).
>
> Following this meeting, and when this change is released, I'd like you to:
> * Re-add users to projects. (It's important that you do this, because doing this is audited.)
> * Check that your leavers processes are up to date. (Essentially: don't remove users from projects; do clear their roles.)
> * Be prepared to edit projects on users' behalf in the Staff Area
>
> Similarly, I will:
> * Supply you with a list of users and projects
> * Communicate this change to our colleagues in the Bennett Institute

iaindillingham:
I've let Juliet and Liam know that I've released this change and sent them a list of project memberships that should be reinstated.[^1]

[^1]: https://bennettoxford.slack.com/archives/C04NHSU5DRB/p1712835432055429

iaindillingham:
And I've let copilots know,[^1] so I'm going to close this issue.

[^1]: https://bennettoxford.slack.com/archives/C028EJH752A/p1712836881520099

PR #4262, opened 2024-04-05
Restrict editing of project status and description

iaindillingham:
This implements the solution for #4236, but doesn't close the issue. I hope the issue, the sequence of commits, and the commit messages describe this PR. You may also be interested in the following, which I feel are worthwhile, but beyond the scope of this issue:

* #4261
* #4260


lucyb (commented):
This is a nice set of very clear commits and some good refactorings.

Just to check my understanding though.. this change means that someone needs to be a member of a project and have the global project developer role to edit a project. There are very few people who have the global project developer role at the moment, as it's typically assigned per project. Is this an intentional stepping stone or did you intend to check for the role on the project instead?

iaindillingham:
Thank you for your thorough review, @lucyb.

You are correct: as the PR stood, a user would have to be a member of a project and have the `project_manage` permission associated with a _global_ role, not a (local) project role or a (local) organization role.

I'd forgotten that `has_permission` checks global roles for permissions, but without a context doesn't check project or organization roles for permissions.

The following (`project=project`) ensures that global roles and project roles are checked for permissions:

https://github.com/opensafely-core/job-server/blob/c9ba8ab5c0f642506f6f2a61e0f77b42e46a4e7f/jobserver/views/projects.py#L216-L218

We can be confident that it's working as expected, because the tests now have project roles and not global roles. For example:

https://github.com/opensafely-core/job-server/blob/c9ba8ab5c0f642506f6f2a61e0f77b42e46a4e7f/tests/unit/jobserver/views/test_projects.py#L233-L244

jobserver/views/projects.py
```diff
@@ -211,7 +211,13 @@ def form_valid(self, form):
...
         project = get_object_or_404(Project, slug=self.kwargs["project_slug"])
 
-        if not project.members.filter(username=self.request.user.username).exists():
+        user = self.request.user
+        user_is_member = project.members.filter(username=user.username).exists()
+        user_has_project_manage = has_permission(
+            user, permissions.project_manage, project=project
+        )
+
+        if not user_is_member or not user_has_project_manage:
```
lucyb, line 220: At the moment, this raises if someone has the global ProjectDeveloper role but isn't a member of the project. However, the intent of the global role is that you don't need to be a member of the project.

I think you can remove the `not user_is_member` half of this if-statement and it will work as expected. Although you should probably add another test; something like `test_projectedit_get_success_with_global_role`.

iaindillingham: Originally, only members of the project could edit this view. The issue calls for restricting editing this view to a user with the Project Developer role. Arguably, that's less restrictive, although easier to reason about.

iaindillingham: Thanks for talking this through, @lucyb. I've updated the `get_object` method, which now restricts only using the `project_manage` permission. I've added an extra test for users that have global `project_manage`.

PR #4274, opened 2024-04-09
Fix JS tests

tomodwyer:
Issues caused by upgrading the nested dependencies means the tests needed to be modified.

As part of this, I switched all tests involving clicks to `userEvents`.

- Remove `await` from `waitFor`
- Use `await` and `findBy[X]` instead of `getBy[X]`
- Replace `textContent` lookups with accessible name checks


tomodwyer:
Replaced by #4275

PR #4275, opened 2024-04-10
Fix for testing library version mismatch

tomodwyer:
...and fix incorrect use of `@testing-library/user-event` dependency.


Jongmassey (approved):
looks great -  the vitest syntax is interesting and surprisingly readable as a newcomer

PR #4277, opened 2024-04-11
Decouple production roles from tests

### LINKED ISSUE #4261: Decouple production roles from tests

iaindillingham:
Closes #4261

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -19,6 +19,7 @@ from opentelemetry.sdk.trace.export import SimpleSpanProcessor
 from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
 from structlog.testing import LogCapture
 
+import jobserver.authorization.roles
 import services.slack
 from applications.form_specs import form_specs
 from jobserver.authorization.roles import CoreDeveloper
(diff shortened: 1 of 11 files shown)


lucyb (commented):
This is nice. The creation of a Role class is very clear, which isn't necessarily easy to achieve since it's doing something quite complicated.

In the commits here you've shown how this can significantly improve the tests and I think it's a change worth introducing.

tests/conftest.py
```diff
@@ -333,3 +334,34 @@ def func(count, **kwargs):
         return [project_membership(**kwargs) for i in range(count)]
 
     return func
+
+
+@pytest.fixture
+def role_factory():
+    """A fixture for dynamically creating a role with a given set of permissions."""
+
+    def _role_factory(*, permissions):
```
lucyb, line 343: The role_factory takes a list of permissions, but in all cases only a single permission is provided. I'm not sure if it's worth continuing accepting a list because it makes semantic sense and might be useful, or if it would be worth removing it and simplifying the code slightly, what do you think?

iaindillingham: I wasn't sure whether I was going to encounter any cases where we needed a role with more than one permission. I'm still not 🙂

lucyb (approved):

StevenMaude (approved):
I haven't really spent any time looking at how job-server's permissions and roles work yet. However, I could understand the intent behind these changes.

PR #4292, opened 2024-04-22
Simplify the `role_factory` fixture

iaindillingham:
Rather than associating a role with zero or many permissions, this associates a role with zero or one permission, making the implementation less error-prone. In the handful of cases where a function/method under test requires a user to have two permissions, we now assign the user two roles with one permission each. Previously, we would have assigned them one role with two permissions.

Thank you, @lucyb for suggesting this change when reviewing #4277.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -352,16 +352,20 @@ def role_factory():
 
         name = f"Role_{permission}"
         assert name.isidentifier(), f"{name} is not a valid Python identifier"
-        Role = type(
-            name,
-            (object,),
-            {
-                "__module__": jobserver.authorization.roles.__name__,
-                "models": [],
-                "permissions": [permission],
-            },
-        )
-        setattr(jobserver.authorization.roles, name, Role)
+
+        Role = getattr(jobserver.authorization.roles, name, None)
+        if Role is None:
+            Role = type(
+                name,
+                (object,),
+                {
+                    "__module__": jobserver.authorization.roles.__name__,
+                    "models": [],
+                    "permissions": [permission],
+                },
+            )
+            setattr(jobserver.authorization.roles, name, Role)
+
         return Role
 
     return _role_factory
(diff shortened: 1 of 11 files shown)


tests/unit/jobserver/views/test_workspaces.py
```diff
@@ -167,7 +167,7 @@ def test_workspacebackendfiles_with_permission(rf, project_membership, role_fact
     project_membership(
         project=workspace.project,
         user=user,
-        roles=[role_factory(permissions=[permissions.unreleased_outputs_view])],
+        roles=[role_factory(permission=permissions.unreleased_outputs_view)],
```
lucyb, line 170: The number of lines you've had to update in this file for such a simple change suggests there's something that can be improved here, either with the UserFactory / project_membership fixture or the tests or workspaces.py (is it doing too much?). I don't think you should do anything more now, but there's maybe a hint towards a future refactoring that's possible here. Same with releases and job_requests above, too.

tests/conftest.py
```diff
@@ -352,16 +352,20 @@ def _role_factory(*, permission):
...
-            (object,),
-            {
-                "__module__": jobserver.authorization.roles.__name__,
-                "models": [],
-                "permissions": [permission],
-            },
-        )
-        setattr(jobserver.authorization.roles, name, Role)
+
+        Role = getattr(jobserver.authorization.roles, name, None)
```
lucyb, line 356: Thank you for doing this. It's a nuance that I hadn't realised before.

lucyb (approved):

tests/unit/jobserver/views/test_workspaces.py
iaindillingham: You're right, but that something isn't immediately clear (to me, at least). When working through the tests, I was struck by how many very similar tests we have. Possibly we could move the permissions check into a decorator, and then just test the decorator?

tests/conftest.py
iaindillingham: I can't remember when it popped into my head, but I think it's worth making it explicit.

PR #4355, opened 2024-06-03
Clarify `ProjectDeveloper` permissions

### LINKED ISSUE #4331: Clarify `ProjectDeveloper` permissions
`ProjectDeveloper.description` states that a project developer can publish released outputs:

https://github.com/opensafely-core/job-server/blob/7904dd3862143fd3306a1cdf32deb723ab701110/jobserver/authorization/roles.py#L135-L137

The ability to publish released outputs is governed by the `snapshot_publish` permission. However, `ProjectDeveloper.permissions` doesn't include that permission:

https://github.com/opensafely-core/job-server/blob/7904dd3862143fd3306a1cdf32deb723ab701110/jobserver/authorization/roles.py#L142-L151

Either `description` is wrong or `permissions` is wrong 🙂.

Thanks for spotting this, @evansd.[^1]

[^1]: https://bennettoxford.slack.com/archives/C069SADHP1Q/p1715333536043939

iaindillingham:
`ProjectDeveloper.description` stated that a project developer can publish released outputs. The ability to publish released outputs is governed by the `snapshot_publish` permission. However, `ProjectDeveloper.permissions` didn't include that permission.

We will assume that `ProjectDeveloper.permissions` is correct and `ProjectDeveloper.description` is incorrect, and remove the statement that a project developer can publish released outputs from `ProjectDeveloper.description`.

Closes #4331

### DIFF
--- a/jobserver/authorization/roles.py
+++ b/jobserver/authorization/roles.py
@@ -132,9 +132,7 @@ class ProjectDeveloper:
     """
 
     display_name = "Project Developer"
-    description = (
-        "Run and cancel Jobs, publish released outputs, and manage workspaces."
-    )
+    description = "Run and cancel Jobs, and manage workspaces."
     models = [
         "jobserver.models.project_membership.ProjectMembership",
         "jobserver.models.user.User",


lucyb (approved):
I wonder if this was intended to be "view released outputs and request they're published" or something similar, since that's what a Project Developer can do. I hope it won't be a problem that it's left out of the description. Overall, I agree that being less specific in the descriptions is a good thing though and I imagine we'll revisit this at some point fairly soon.

iaindillingham:
Thanks for your comment, @lucyb. I think we're due a rethink of how we describe roles and permissions. I think it would be better to attach a human-readable description to each permission, and stitch these together in their associated roles.

PR #4372, opened 2024-06-13
Allow configuration of org/repo for airlock issues by organisation

rebkwok:
This allows  output-checking issues to be opened in custom repos, depending on organisation. Airlock can also send configuration to use, which allows an entire backend to use a different repo (e.g. the test backend, which we don't want to use a real output checking repo).

Note: the opensafely-readonly user will need to be added as a collaborator on Bristol's output checking repo so it has permissions to open/edit issues.


madwort (approved):
code looks good, although I'd like to see a test that passes the repo as a get param

PR #4373, opened 2024-06-14
Update description-help_text.md

JulietUnderdown1:
Word count updated as users did not have room to add URL links into the lay desc.


iaindillingham:
Thank you for creating a PR, @JulietUnderdown1. I've bundled this change with one other, in #4395. As such, I'm going to close this PR.

PR #4384, opened 2024-06-24
Replace lxml_html_clean with nh3

### LINKED ISSUE #4254: Consider replacing lxml_html_clean with nh3
Currently, we use both [lxml_html_clean](https://github.com/fedora-python/lxml_html_clean) and [nh3](https://github.com/messense/nh3) for cleaning HTML. Assuming that both packages perform similar functions, using them means we have an unnecessary dependency.

I'd suggest replacing lxml_html_clean with nh3, rather than the other way around, [because](https://lxml-html-clean.readthedocs.io/en/latest/usage.html):

> Note: the HTML Cleaner in `lxml_html_clean` is not considered appropriate for security sensitive environments. See e.g. [bleach](https://pypi.org/project/bleach/) for an alternative.

However, bleach is deprecated, which is why we added nh3 in the first place.

## Usage

We use lxml_html_clean to prepare OS Interactive reports for display on OS Reports:

https://github.com/opensafely-core/job-server/blob/7de6a7d4843745131e335ede6f1b6e5fbda891f2/jobserver/reports.py#L13-L15

We use nh3 to prepare a project's status description:

https://github.com/opensafely-core/job-server/blob/7de6a7d4843745131e335ede6f1b6e5fbda891f2/jobserver/views/projects.py#L38

https://github.com/opensafely-core/job-server/blob/7de6a7d4843745131e335ede6f1b6e5fbda891f2/staff/views/projects.py#L218-L220

## Related

* #4251 
* #4155

iaindillingham:
Closes #4254


lucyb (approved):
This a really nice piece of refactoring. I like that you've kept the commits as a series of isolated small steps.

PR #4392, opened 2024-06-26
Speed up file compression and decompression in CI

### LINKED ISSUE #4354: Speed up CI by improving the Docker save and upload steps
Steve made some nice improvements to the Docker save and upload steps in the CI for the research-template-docker repo. We should copy that to improve the Job Server build times.

https://github.com/opensafely-core/research-template-docker/blob/7ed414d39919fc857fbe6a77706cd723719e39c0/.github/workflows/main.yml#L34-L44

In particular, we should:
* Use pigz rather than gzip (EDIT [Dave recommends](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1717067259731309) `zstd` instead, which should be preinstalled)
* Add compression-level: 0 to the upload step

## Acceptance criteria
* The updated CI workflow runs faster on average than the existing CI workflow.
* The CI workflow isn't meaningfully more complex or brittle.

Jongmassey:
Replace gzip with ZStandard in fast mode.

Pass compressed file directly to `docker image load` removing a redundant separate decompression and pipe.

Fixes #4354

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -136,13 +136,14 @@ jobs:
 
       - name: Save docker image
         run: |
-          docker save job-server | gzip > /tmp/job-server.tar.gz
+          docker save job-server | zstd --fast -o /tmp/job-server.tar.zst
 
       - name: Upload docker image
         uses: actions/upload-artifact@v4
         with:
             name: job-server-image
-            path: /tmp/job-server.tar.gz
+            path: /tmp/job-server.tar.zst
+            compression-level: 0
 
   deploy:
     needs: [check, test, docker-test, lint-dockerfile]
@@ -170,7 +171,7 @@ jobs:
             path: /tmp/image
 
       - name: Import docker image
-        run: gunzip -c /tmp/image/job-server.tar.gz | docker load
+        run: docker image load --input /tmp/image/job-server.tar.zst
 
       - name: Test image we imported from previous job works
         run: |


Jongmassey:
"Save Docker Image" step:

Before: [21s](https://github.com/opensafely-core/job-server/actions/runs/9648849435/job/26610987230)

After: [4s](https://github.com/opensafely-core/job-server/actions/runs/9677811848/job/26700258666?pr=4392)

Not sure how to compare the speed of the deploy step without actually doing a deploy

.github/workflows/main.yml
```diff
@@ -136,13 +136,13 @@ jobs:
...
         run: |
-          docker save job-server | gzip > /tmp/job-server.tar.gz
+          docker save job-server | zstd --fast -o /tmp/job-server.tar.zst
 
       - name: Upload docker image
         uses: actions/upload-artifact@v4
         with:
             name: job-server-image
-            path: /tmp/job-server.tar.gz
+            path: /tmp/job-server.tar.zst
```
StevenMaude, line 145: thought: You might get a further improvement from:

```suggestion
            path: /tmp/job-server.tar.zst
            # Disable compression; the file is already compressed
            compression-level: 0
```

StevenMaude (approved):
I'll approve as this is working as it is.

I added one other suggestion to try (which improved things considerably for the `research-template-docker` repository, although that research-template image may be much larger).

.github/workflows/main.yml
Jongmassey: No harm in adding it, so I will!

Issue #4182, opened 2024-03-07
Squash Job Server migrations


lucyb:

Jongmassey:
This has been split into two phases per the [django docs](https://docs.djangoproject.com/en/5.0/topics/migrations/#squashing-migrations)

- [ ] Squash migraitons #4393 
- [ ] Transition squashed migration to regular #TBD

PR #4393, opened 2024-06-26
Squash migrations

Jongmassey:
Following [process from django docs](https://docs.djangoproject.com/en/5.0/topics/migrations/#migration-squashing)

First step of #4182


Jongmassey:
>Once you’ve squashed your migration, you should then commit it alongside the migrations it replaces and distribute this change to all running instances of your application, making sure that they run migrate to store the change in their database.

```bash
$ python manage.py migrate
Operations to perform:
  Apply all migrations: applications, auth, contenttypes, interactive, jobserver, redirects, sessions, social_django
Running migrations:
  No migrations to apply.
```

on a freshly downloaded prod database dump

Jongmassey:
>You must then transition the squashed migration to a normal migration by:
> * Deleting all the migration files it replaces.
> * Updating all migrations that depend on the deleted migrations to depend on the squashed migration instead.
> * Removing the `replaces` attribute in the `Migration` class of the squashed migration (this is how Django tells that it is a squashed migration).

```bash
$python manage.py migrate
Traceback (most recent call last):
  File "/home/jmassey/opensafely-core/job-server/manage.py", line 21, in <module>
    main()
  File "/home/jmassey/opensafely-core/job-server/manage.py", line 17, in main
    execute_from_command_line(sys.argv)
  File "/home/jmassey/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/home/jmassey/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/jmassey/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/core/management/base.py", line 413, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/jmassey/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/core/management/base.py", line 459, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jmassey/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/core/management/base.py", line 107, in wrapper
    res = handle_func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jmassey/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/core/management/commands/migrate.py", line 117, in handle
    executor = MigrationExecutor(connection, self.migration_progress_callback)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jmassey/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/db/migrations/executor.py", line 18, in __init__
    self.loader = MigrationLoader(self.connection)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jmassey/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/db/migrations/loader.py", line 58, in __init__
    self.build_graph()
  File "/home/jmassey/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/db/migrations/loader.py", line 276, in build_graph
    self.graph.validate_consistency()
  File "/home/jmassey/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/db/migrations/graph.py", line 198, in validate_consistency
    [n.raise_error() for n in self.node_map.values() if isinstance(n, DummyNode)]
     ^^^^^^^^^^^^^^^
  File "/home/jmassey/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/db/migrations/graph.py", line 60, in raise_error
    raise NodeNotFoundError(self.error_message, self.key, origin=self.origin)
django.db.migrations.exceptions.NodeNotFoundError: Migration applications.0001_initial dependencies reference nonexistent parent node ('jobserver', '0001_initial')
```

Jongmassey:
Ah, I think it's because the applications and interactive applications' migrations reference the old initial 0001 migration!

iaindillingham (changes requested):
Thanks for this, @Jongmassey. It's great to see the task broken down into a logical sequence of commits. The commit messages are also very clear.

Although it's been a while since I worked with Django migrations, I can see that we can't merge and release this PR, because the release will fail:

```
django.db.migrations.exceptions.InconsistentMigrationHistory: Migration applications.0001_initial is applied before its dependency jobserver.0001_squashed_0027_add_custom_manager_to_JobRequest on database 'default'
```

The solution is to break the task into two PRs. The first PR would take us to 653079b, which we'd merge and release. The second PR would take us to 5089645, which we'd merge and release. It's important that the first PR is merged and released before the second PR. From the "[Squashing migrations](https://docs.djangoproject.com/en/5.0/topics/migrations/#squashing-migrations)" section:

> The recommended process is to squash, keeping the old files, commit and release, wait until all systems are upgraded with the new release ... and then remove the old files, commit and do a second release.

You can recreate the error yourself locally by running `python manage.py migrate` at `HEAD`. Now try checking out 653079b and migrating; and checking out 5089645 and migrating. No error!

I see that you've opted to copy the functions that are called by the [`RunPython`](https://docs.djangoproject.com/en/5.0/ref/migration-operations/#django.db.migrations.operations.RunPython) instances. There's an alternative: to mark these instances as elidable.

Whilst I think it's valuable in the short-term to be able to reverse a migration, I don't think the same is true in the medium-term. To put that another way, can you imagine rolling back to July last year, when migration 0003 was added? (the earliest migration to use a `RunPython` instance) Alternatively, to January this year, when migration 0021 was added? (the latest)

I can't, which is why I think we should mark these -- and the intervening `RunPython` instances -- as elidable. Doing so means we need not copy the functions, and makes squashing migrations more mechanical.

One final thought: Consider naming the squashed migration:

```sh
python manage.py squashmigrations jobserver 0027 --squashed-name squashed
```

The above command would result in `jobserver/migrations/0001_squashed.py`. Check with @lucyb first, because I think this name may mean we need to prune migrations later -- we may choose to add a timestamp, by convention, to avoid doing so. I think that's probably okay, but as I say, it's been a while since I worked with Django migrations.

Jongmassey:
>Consider naming the squashed migration:

I agree, I've re-run with a datestamped name as I think that might be helpful over just "squashed" in the future

Jongmassey:
>I think we should mark these -- and the intervening RunPython instances -- as elidable

Is my understanding correct that this requires editing the Migrations that contain `RunPython` instances (those from which I copied the functions per the instructions in the comment at the head of the output from the `sqaushmigration` command) such that they are marked as elidable e.g. [00003](https://github.com/opensafely-core/job-server/blob/27553a2d957442482674b64c7a050b2c07df26c3/jobserver/migrations/0003_create_redirects_to_remove_org_from_all_urls.py#L49C1-L59C6) would become

```python
class Migration(migrations.Migration):
    dependencies = [
        ("jobserver", "0002_ensure_project_slugs_have_a_value"),
    ]

    operations = [
        migrations.RunPython(
            add_redirects_to_remove_org_from_orgs,
            reverse_code=remove_redirects_for_orgs,
            elidable=True,
        )
    ]
```


Then run the `squashmigrations` command which would automatically handle the situation I addressed manually?

iaindillingham:
> Is my understanding correct...

Your understanding is correct!

Jongmassey:
@iaindillingham  you mentioned in the meeting yesterday that there was more to squashing migrations than you'd recalled - are there any more changes required on this PR? Ta

Jongmassey:
@iaindillingham  you mentioned in the meeting yesterday that there was more to squashing migrations than you'd recalled - are there any more changes required on this PR? Ta

jobserver/migrations/0001_0001_squashed_2024_06.py
iaindillingham, line 1: Notice the name of the file, @Jongmassey. Django adds the prefix. The command should be:

```sh
./manage.py squashmigrations jobserver 0027 --squashed-name squashed_2024_06
```

iaindillingham:
> @iaindillingham you mentioned in the meeting yesterday that there was more to squashing migrations than you'd recalled - are there any more changes required on this PR? Ta

There's one more change required: I've commented on the name of the squashed migration. My comment in yesterday's meeting was more "I'd forgotten how fiddly squashing migrations is".

jobserver/migrations/0001_0001_squashed_2024_06.py
Jongmassey: Ah good spot, my mind totally elided the double `0001`!

iaindillingham (approved):
Thanks for updating this, @Jongmassey. Looking forward to seeing part two. Before you merge, I'd suggest removing the "Fixes" line from the description, as this PR is only part one.

PR #4407, opened 2024-07-02
Fix project redirect regression

### LINKED ISSUE #4158: Renaming a project caused a redirect to the same URL
## Steps to reproduce

1. Go to a project in the Staff Area
2. Rename the project
3. Update the project slug
4. A redirect is created which has the same "to" and "from" URLs

Jongmassey:
7ba76cf refactored project editing into a command
and in the process the `old` Project instance in
this edit method was changed from `form.get_object()`
to `form.object`.

The latter is the same as the `new` Project instance
and so redirects were being created with the same URL
for both old and new.

fixes #4158

### DIFF
--- a/staff/views/projects.py
+++ b/staff/views/projects.py
@@ -230,7 +230,7 @@ class ProjectEdit(UpdateView):
 
     @transaction.atomic()
     def form_valid(self, form):
-        new = projects.edit(old=self.object, form=form, by=self.request.user)
+        new = projects.edit(old=self.get_object(), form=form, by=self.request.user)
 
         return redirect(new.get_staff_url())
 


StevenMaude (approved):

iaindillingham:
It would be great to see a test for this, @Jongmassey -- ideally, committed before the fix. We'd be testing that we passed the correct arguments to `projects.edit`, rather than that `projects.edit` worked correctly.

If it helps, then this line is exercised by `test_projectedit_post_success` and `test_projectedit_post_success_when_not_changing_slug`.

Jongmassey:
Sorry @iaindillingham  I'd already merged based on @StevenMaude 's approval.

[Here's ](https://github.com/opensafely-core/job-server/pull/4416) a PR with some such tests in

PR #4413, opened 2024-07-03
Move hard-coded slack channels and github repos to settings

rebkwok:
Instead of hard-coding slack channels and github org/repos in multiple places, move them to settings and allow them to be overridden by environment variables. This makes it much easier to  set up local testing that includes testing slack/github integrations without posting to production channels/repos.

### DIFF
--- a/jobserver/settings.py
+++ b/jobserver/settings.py
@@ -407,6 +407,33 @@ BACKEND_IP_MAP = {
     # "127.0.0.1": "tpp",
 }
 
+
+# SLACK CHANNELS
+RELEASES_SLACK_CHANNEL = env.str(
+    "RELEASES_SLACK_CHANNEL", default="opensafely-releases"
+)
+REGISTRATIONS_SLACK_CHANNEL = env.str(
+    "REGISTRATIONS_SLACK_CHANNEL", default="job-server-registrations"
+)
+APPLICATIONS_SLACK_CHANNEL = env.str(
+    "RELEASES_SLACK_CHANNEL", default="job-server-applications"
+)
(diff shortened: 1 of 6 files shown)


jobserver/settings.py
```diff
@@ -407,6 +407,33 @@ def immutable_file_test(path, url):
...
+)
+
+# OUTPUT_CHECKING_REPOS
+DEFAULT_OUTPUT_CHECKING_GITHUB_ORG = env.str(
+    "DEFAULT_OUTPUT_CHECKING_GITHUB_ORG", "ebmdatalab"
+)
+DEFAULT_OUTPUT_CHECKING_REPO = env.str(
+    "DEFAULT_OUTPUT_CHECKING_REPO", "opensafely-output-review"
+)
+
```
iaindillingham, line 436: I don't understand why some of these settings are prefixed with `DEFAULT` and others aren't. What have I missed?

iaindillingham (approved):
I have a question about naming things! Otherwise, 👍🏻

jobserver/settings.py
rebkwok: The DEFAULT ones come from the Airlock app (airlock/config.py), and are default values that can be overidden per org. i.e. these are the ones for any org that Bennett output checks, but can be different for orgs that do their own output-checking. Currently that's just Bristol, and they have a different repo set in airlock/config.py.  (They don't have a different slack channel atm, because we haven't worked out what it should be for them, if anything).
The non-DEFAULT ones are (at least currently) expected to be a single value anywhere this thing is needed.

iaindillingham: Ah, that's really useful. Thank you for clarifying!

PR #4419, opened 2024-07-04
Transition squashed migration to normal migration

### LINKED ISSUE #4182: Squash Job Server migrations

Jongmassey:
Fixes #4182 (second step)

From [django migrations docs](https://docs.djangoproject.com/en/5.0/topics/migrations/#squashing-migrations):

>You must then transition the squashed migration to a normal migration by:
>
>* Deleting all the migration files it replaces.
> * Updating all migrations that depend on the deleted migrations to depend on the squashed migration instead.
> * Removing the `replaces` attribute in the `Migration` class of the squashed migration (this is how Django tells that it is a squashed migration).


iaindillingham (approved):
When this is released, could you pop a message in [#team-rex](https://bennettoxford.slack.com/archives/C069SADHP1Q) suggesting we fetch a new database dump? If we don't, then we'll encounter a `InconsistentMigrationHistory` error as we won't have gone through the release-migrate-release-migrate process.

PR #4420, opened 2024-07-04
Include project status in user details endpoint

rebkwok:
This sends a boolean "ongoing" status rather than the specific Project status. This data is expected to be consumed by Airlock, which just needs to know whether a project is ongoing or not in order to determine whether release requests can be created. In case the specific statuses change in job-server (and there's some comments in the model that suggest that they might in future), we don't want Airlock to be looking for exact job-server controlled strings.

### DIFF
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -31,6 +31,7 @@ from jobserver.api.authentication import get_backend_from_token
 from jobserver.authorization import OutputChecker, has_permission, has_role, permissions
 from jobserver.commands import users
 from jobserver.models import (
+    Project,
     PublishRequest,
     Release,
     ReleaseFile,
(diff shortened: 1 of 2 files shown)


Jongmassey (approved):

jobserver/api/releases.py
```diff
@@ -576,13 +577,22 @@ class Level4AuthenticatedUser(serializers.Serializer):
 
 
 def build_level4_user(user):
+    ongoing_project_statuses = [
+        Project.Statuses.ONGOING,
+        Project.Statuses.ONGOING_LINKED,
+    ]
```
iaindillingham, line 583: Making this a set, rather than a list, would be more consistent with the later test for containment.

iaindillingham (approved):

Issue #4391, opened 2024-06-25
Replace Zen Queries with `django_assert_num_queries` on rendered template


lucyb:
We've been bitten by Zen Queries a few times (including with a production outage). I think we can replace this dependency with unit tests that use the Django test client and the `django_assert_num_queries` fixture. This will tell us about errors when we run our tests rather than at runtime. It should also provide developers with a more obvious error message that they can debug and fix ([example one](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1708035538382269), [example two](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1708946787197529)).

We are already doing this in some places, such as the unauthenticated index page:
https://github.com/opensafely-core/job-server/blob/327fbd971c28d8b19bcc6f96fa39502c1afa8589/tests/unit/jobserver/views/test_index.py#L70-L78

As another example, a test for the JobDetail page might look like this ([slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1709043464199309) :thread: ):

```python
def test_jobrequestdetail_count_queries_in_rendered_template(
    client, rf, django_assert_num_queries
):
    job_request = JobRequestFactory()
    JobFactory(job_request=job_request, updated_at=minutes_ago(timezone.now(), 31))
    request = rf.get("/")
    request.user = UserFactory()

    with django_assert_num_queries(8):
        response = JobRequestDetail.as_view()(
            request,
            project_slug=job_request.workspace.project.slug,
            workspace_slug=job_request.workspace.name,
            pk=job_request.pk,
        )

    with django_assert_num_queries(10):
        response = client.get(job_request.get_absolute_url())
        assert response.status_code == 200
        assert job_request.identifier in response.rendered_content
```

## TODO

- [x] Write an ADR
   - [x] Update the ADR
- [x] Remove zen queries from INSTALLED_APPS and requirements.prod.txt (final step)

Remove zen queries from:

- [x] Context processors
- [x] Applications in staff area
- [x] Job Requests
- [x] Projects
- [x] Users (must be done last because the header.html references user.users_social_auth and so anything using zen queries breaks)
- [x] Projects in staff area
- [x] Users in staff area

PR #4432, opened 2024-07-08
Remove staff_nav function

lucyb:
As part of [#4391](https://github.com/opensafely-core/job-server/issues/4391) to remove the use of zen-queries, I noticed that the only use of this function was removed in 4dd6c3 and so it's no longer needed.


iaindillingham (approved):
Nice!

PR #4439, opened 2024-07-10
Update interactive-templates to v2024.07.09.130747

StevenMaude:
This is a manual update, because the machinery for generating an automatic PR for that update is fragile.

See opensafely-core/interactive-templates#377.


StevenMaude:
I may try moving the GitHub Actions update workflow to this repository; in which case, it's useful to leave the update as is, so that we can see it take effect with any new workflow.

StevenMaude:
We could try creating these same changes via workflow once #4442 is merged, instead of merging this one.

StevenMaude:
#4458 covered this.

PR #4442, opened 2024-07-11
Add interactive-templates update workflow

StevenMaude:
This is a proposal to move the [interactive-templates update workflow](https://github.com/opensafely-core/interactive-templates/blob/b8f742c2ecea2de223aa74ec8c8773f953e5a2d9/.github/workflows/create-job-server-pr.yml) from the interactive-templates repository to here.

The workflow was added in the interactive-templates repository to run on a new release of interactive-templates, and then open a pull request to job-server. The workflow is run daily and can be run manually on request.

This is based on the version of the workflow from opensafely-core/interactive-templates#377 which has fixes necessary for the workflow to work. The workflow operates like the "check resource for update and create a PR if so" [workflows in ehrQL](https://github.com/opensafely-core/ehrql/blob/5f36cc0c99397ab07e1bc0844c6ebeb3f2ab0871/.github/workflows/update-external-studies.yml).

See opensafely-core/interactive-templates#118 which proposed this.

The main motivation for moving it here is to remove the need for the personal access token, which appears to have expired. With the low frequency of updates, it also seems reasonable to simplify this.

### DIFF
--- a/.github/workflows/update-interactive-templates.yml
+++ b/.github/workflows/update-interactive-templates.yml
@@ -1,25 +1,17 @@
 ---
-name: "Create PR to update job-server"
+name: "Create PR to update interactive-templates"
 
 on:
   workflow_dispatch:
-  workflow_run:
-    workflows:
-      - CI
-    branches:
-      - main
-    types:
-      - completed
+  schedule:
+    - cron: "25 13 * * *"
 
 jobs:
   create_job_server_pr:
-    if: ${{ (github.event.workflow_run.conclusion == 'success') || (github.event_name == 'workflow_dispatch') }}
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v4
-        with:
-          repository: opensafely-core/job-server
 
       - uses: opensafely-core/setup-action@v1
         with:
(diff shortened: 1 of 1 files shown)


StevenMaude:
Two notes:

* opensafely-core/interactive-templates#118 also mentioned possibly notifying the Slack code review channel on a release, which we could look at now or in future.
* like lots of GitHub Actions changes, it's not easy to just try this workflow out without merging. It is very much based on existing workflows, both from the [interactive-templates PR](https://github.com/opensafely-core/interactive-templates/pull/377) (which worked up to the point of needing the personal access token), and from ehrQL.

.github/workflows/update-interactive-templates.yml
```diff
@@ -0,0 +1,54 @@
...
+        id: create_pr
+        uses: peter-evans/create-pull-request@c5a7806660adbe173f04e3e038b0ccdcd758773c # v6.1.0
+        with:
+          token: "${{ secrets.JOB_SERVER_PR_TOKEN }}"
+          add-paths: requirements.*
+          base: main
+          branch: bot/update-interactive-templates
+          author: "opensafely-github-bot <opensafely-github-bot@users.noreply.github.com>"
+          committer: "opensafely-github-bot <opensafely-github-bot@users.noreply.github.com>"
+          commit-message: "feat: Update interactive templates to ${{ env.UPDATED_VERSION }}"
```
iaindillingham, line 37: We don't need the "feat" prefix, do we? job-server doesn't use conventional commits. Also, how would you feel about making the commit messages consistent with Dependabot's commit messages?

```suggestion
          commit-message: "Bump interactive-templates to ${{ env.UPDATED_VERSION }}"
```

iaindillingham (approved):
Couple of comments, but neither are deal-breakers.

PR #4443, opened 2024-07-12
Airlock notification updates

rebkwok:
Airlock now sends notifications for status changes only, with potential optional update messages that are now included in emails.

### DIFF
--- a/airlock/emails.py
+++ b/airlock/emails.py
@@ -43,24 +43,6 @@ def send_request_rejected_email(airlock_event):
     )
 
 
-def send_request_updated_email(airlock_event):
-    context = {
-        "release_request_id": airlock_event.release_request_id,
-        "request_author": airlock_event.request_author.name,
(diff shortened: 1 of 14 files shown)


airlock/emails.py
```diff
@@ -5,16 +5,24 @@
...
     f = furl(settings.BASE_URL)
     f.path = release.get_absolute_url()
 
-    context = {
-        "url": f.url,
-        "request_author": airlock_event.request_author.name,
-        "workspace": airlock_event.workspace.name,
-    }
+    context = get_email_context(airlock_event)
+    context["url"] = (f.url,)
```
evansd, line 25: Is this meant to be a tuple? Might be a stray trailing comma left over after the refactor.

evansd (approved):
Looks good, just one question below.

airlock/emails.py
rebkwok: It is, thanks

PR #4456, opened 2024-07-16
Fix up the "update interactive-templates" workflow

StevenMaude:
Fixes a few small issues with the newly added workflow, that were, of course, discovered after trying to run the workflow.


StevenMaude:
This finally, after a manual run of the "create PR" workflow, created a pull request correctly (#4458).

iaindillingham (approved):

PR #4460, opened 2024-07-16
Test `ApplicationList` view's context

iaindillingham:
This test corresponds to the fix introduced by #4459. Here, we test that `ApplicationList` view's context contains `object_list`, which is provided by `BaseListView`. More importantly, we test that it contains `application_list`, which is used by `staff/application/list.html`.

### DIFF
--- a/tests/unit/staff/views/test_applications.py
+++ b/tests/unit/staff/views/test_applications.py
@@ -476,6 +476,17 @@ def test_applicationedit_without_core_dev_role(rf):
         ApplicationEdit.as_view()(request, pk_hash=application.pk_hash)
 
 
+def test_applicationlist_context_object_name(rf, core_developer):
+    ApplicationFactory()
+
+    request = rf.get("/")
+    request.user = core_developer
+
+    response = ApplicationList.as_view()(request)
+    assert "object_list" in response.context_data
+    assert "application_list" in response.context_data
+
+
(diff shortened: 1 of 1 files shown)


tests/unit/staff/views/test_applications.py
```diff
@@ -476,6 +476,17 @@ def test_applicationedit_without_core_dev_role(rf):
...
 
+def test_applicationlist_context_object_name(rf, core_developer):
+    ApplicationFactory()
+
+    request = rf.get("/")
+    request.user = core_developer
+
+    response = ApplicationList.as_view()(request)
+    assert "object_list" in response.context_data
+    assert "application_list" in response.context_data
```
lucyb, line 487: I think this is fine, but I wonder if it makes more sense to change the tests below instead? The template only uses application_list. We don't actually care what's in object_list (I think), so we should only assert on the things we do care about (in this case, that the context data we use is present as we expect).

iaindillingham: Actually, I don't think this is fine! I think I'd like to back out of this PR and also out of #4459. I think that `ApplicationList.get_queryset` should return a queryset, not a list. If returning a queryset means that `test_applicationlist_num_queries` fails, then I think I need to make judicious use of `select_related` and `prefetch_related`.

I also think the tests below should replace `object_list` with `application_list`, as you say. However, by backing out of #4459, Django would be setting the context object name.

PR #4461, opened 2024-07-16
Replace zen queries in the staff area

lucyb:
This removes zen queries from the remaining places that it's used in the staff area.

Refs #4391.

### DIFF
--- a/staff/views/projects.py
+++ b/staff/views/projects.py
@@ -15,8 +15,6 @@ from django.views.generic import (
 from django_htmx.http import HttpResponseClientRedirect
 from furl import furl
 from markdown import markdown
-from zen_queries import TemplateResponse as zTemplateResponse
-from zen_queries import fetch
 
 from applications.models import Application
 from interactive.commands import create_repo, create_workspace
@@ -95,7 +93,6 @@ class ProjectAddMember(FormView):
 class ProjectAuditLog(ListView):
     paginate_by = 25
     template_name = "staff/project/audit_log.html"
-    response_class = zTemplateResponse
 
     def dispatch(self, request, *args, **kwargs):
         self.project = get_object_or_404(Project, slug=self.kwargs["slug"])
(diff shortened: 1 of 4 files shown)


StevenMaude (approved):

staff/views/users.py
```diff
@@ -84,7 +81,7 @@ def get_queryset(self):
         if types := self.request.GET.getlist("types"):
             qs = qs.filter(type__in=types)
 
-        return fetch(qs.distinct())
+        return list(qs.distinct())
```
iaindillingham, line 84: If we return a queryset (i.e. remove `list`), then we actually _reduce_ the number of queries from 14 to 13.

iaindillingham (commented):
Thanks for highlighting that we don't need to set `context_object_name` because of `.get_context_data`. I'm uneasy about returning a list from `.get_queryset`. Would it be helpful to talk through doing so?

lucyb:
>Thanks for highlighting that we don't need to set context_object_name because of .get_context_data. I'm uneasy about returning a list from .get_queryset. Would it be helpful to talk through doing so?

Thank you for this. After looking at your changes and the explanation, I agree. I've now removed the use of list(..) and weirdly that hasn't caused any detrimental performance impact. Maybe Knuth was right about premature optimisation!

staff/views/users.py
lucyb: Good spot. That's the case for both of them infact.

staff/views/projects.py
```diff
@@ -126,7 +123,7 @@ def get_queryset(self):
         if types := self.request.GET.getlist("types"):
             qs = qs.filter(type__in=types)
 
-        return fetch(qs.distinct())
+        return list(qs.distinct())
```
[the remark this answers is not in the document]
lucyb: You've convinced me. Thank you.

PR #4463, opened 2024-07-17
Fix `ApplicationList` view

iaindillingham:
This attempts to do the right thing, making up for #4459 (a quick fix) and #4460. The commit messages contain my reasoning, so I'll not repeat them here.

### DIFF
--- a/staff/views/applications.py
+++ b/staff/views/applications.py
@@ -182,7 +182,6 @@ class ApplicationList(ListView):
     ordering = "-created_at"
     paginate_by = 25
     template_name = "staff/application/list.html"
-    context_object_name = "application_list"
 
     def get_context_data(self, **kwargs):
         # sort in Python because `User.name` is a property to pick either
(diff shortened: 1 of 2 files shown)


lucyb (approved):
Thanks for this slow fix after the quick one. I think this is a definite improvement and it no longer feels like we're fighting against Django. I really appreciate the detailed commit messages with your reasoning and links to the relevant code. I also feel un-zen about using an internal Django method, but forgot to mention it when I made the original change.

For completeness, I want to point out in other places we are setting the result of calling `get_queryset` explicitly in the context data. So, doing the equivalent of:

```
        application_list = self.get_queryset()
        ...
        return super().get_context_data(..., **kwargs) | {
            "application_list": application_list,
        }
``` 
I don't think we should do that here, because your change hasn't caused any problems as you've demonstrated (and as I've tested by loading the page and checking Django Debug Toolbar) and I think it's a better approach anyway.

PR #4466, opened 2024-07-18
Update study purpose help text

iaindillingham:
This updates the study purpose help text, which is part of the "Study information" page, following a request from @LiamHart-hub.[^1]

The "Study information" page before the change is as follows:

Study information before

The same page after the change is as follows:

Study information after

The same page after the change, but with the change highlighted in red, is as follows:

Study information after with highlight

[^1]: https://bennettoxford.slack.com/archives/C068NDYALSF/p1721295082531219


JulietUnderdown1 (approved):
Thank you - the update is as per the instruction.

PR #4480, opened 2024-07-22
Make django-debug-toolbar a development dependency

### LINKED ISSUE #4441: Make django-debug-toolbar a development dependency
django-debug-toolbar is a production dependency. This is unexpected.[^1] We should make it a development dependency. Doing so will involve moving it from production requirements to development requirements, and also ensuring that its settings are conditional on running in development.

[^1]: https://bennettoxford.slack.com/archives/C069SADHP1Q/p1719993204062649

iaindillingham:
Although making django-debug-toolbar (DDT) a development dependency introduces branches in `settings.py` and `urls.py`, which we have tried to avoid in the past (e.g. bfdc66f), it nevertheless makes our production deployment more secure. Depending on DDT only in development is also clearer than depending on it in production and in development, but indirectly disabling it in production via `settings.INTERNAL_IPS`.

This commit takes inspiration from opensafely-core/airlock#131 and isolates specific DDT-related settings from general debug-related settings. It does this with the `DJANGO_DEBUG_TOOLBAR` environment variable and the `DEBUG_TOOLBAR` setting (the `DJANGO_` prefix is dropped from the setting, as it's unnecessary). For safety, the setting is `False` by default. Because seeing DDT is a good reminder that the browser tab you're looking at is pointing to development and not production, DDT is enabled in development by hard-coding the environment variable in `just run`.

Closes #4441

### DIFF
--- a/jobserver/settings.py
+++ b/jobserver/settings.py
@@ -45,6 +45,8 @@ if OLD_SECRET_KEY is not None:
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = env.bool("DEBUG", default=False)
 
+DEBUG_TOOLBAR = env.bool("DJANGO_DEBUG_TOOLBAR", default=False)
+
 BASE_URL = env.str("BASE_URL", default="http://localhost:8000")
 
 ALLOWED_HOSTS = ["*"]
@@ -59,7 +61,6 @@ INSTALLED_APPS = [
     "redirects",
     "staff",
     "anymail",
-    "debug_toolbar",
     "django_extensions",
     "django_htmx",
     "django_vite",
@@ -80,7 +81,6 @@ MIDDLEWARE = [
     "django.middleware.security.SecurityMiddleware",
     "django_permissions_policy.PermissionsPolicyMiddleware",
     "whitenoise.middleware.WhiteNoiseMiddleware",
-    "debug_toolbar.middleware.DebugToolbarMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
@@ -95,6 +95,10 @@ MIDDLEWARE = [
     "jobserver.middleware.ClientAddressIdentification",
     "jobserver.middleware.TemplateNameMiddleware",
 ]
+
+if DEBUG_TOOLBAR:
+    INSTALLED_APPS.append("debug_toolbar")
+    MIDDLEWARE.append("debug_toolbar.middleware.DebugToolbarMiddleware")
 
 ROOT_URLCONF = "jobserver.urls"
 
--- a/jobserver/urls.py
+++ b/jobserver/urls.py
@@ -1,4 +1,3 @@
-import debug_toolbar
 import social_django.views as social_django_views
 from django.conf import settings
 from django.conf.urls.static import static
@@ -294,6 +293,11 @@ user_urls = [
     path("<str:username>/logs/", UserEventLog.as_view(), name="user-event-log"),
 ]
 
+if settings.DEBUG_TOOLBAR:  # pragma: no cover
+    debug_toolbar_urls = [path("__debug__/", include("debug_toolbar.urls"))]
+else:
+    debug_toolbar_urls = []
+
 urlpatterns = [
     path("", Index.as_view(), name="home"),
     path(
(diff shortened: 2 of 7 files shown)


iaindillingham:
I'm trying to understand the failure of `docker-test`. I don't _think_ it's transient.

iaindillingham:
> I'm trying to understand the failure of `docker-test`. I don't _think_ it's transient.

It wasn't transient! However, I had to...

```sh
just docker-build prod
just docker-serve prod -d
docker logs job-server-prod-1
```

...to find out more. Originally, the call to `include` passed `debug_toolbar.urls` (a module) rather than `"debug_toolbar.urls"` (a string).

https://github.com/opensafely-core/job-server/blob/aadb57b01d1aa3b82e038858df1a1a02d7381b79/jobserver/urls.py#L341-L344

Whilst the `debug_toolbar.urls` module exists in development, it doesn't exist in production. The production container was starting and stopping, with...

```sh
ModuleNotFoundError: No module named 'debug_toolbar'
```

...because of the import at the top of `urls.py`.

jobserver/urls.py
```diff
@@ -337,9 +336,17 @@
...
     path("workspaces/", yours.WorkspaceList.as_view(), name="your-workspaces"),
-    path("__debug__/", include(debug_toolbar.urls)),
-    path("<str:project_slug>/", include(project_urls)),
-] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
+]
+
+if settings.DEBUG:  # pragma: nocover
+    urlpatterns += [
+        path("__debug__/", include("debug_toolbar.urls")),
+    ] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```
lucyb, line 344: I think we want the static files in both production and development?

evansd (commented):
Commenting after being summoned for my opinions on `DEBUG`!

I do feel that hanging loads of behaviour of the `DEBUG` setting is an anti-pattern which Django encourages. I think it's fine to have a single flag which changes the amount of logging detail but doesn't (in theory) change any behaviour. But I think keeping all behavioural changes individually switchable is really important for being able to reason about the system when debugging. And DDT _is_ a behavioural change as it injects a whole bunch of stuff into the page.

In Airlock, we've handled this by having a default environment file which is configured with all the relevant development features enabled (`DJANGO_DEBUG`, `AIRLOCK_DEV_USERS_FILE` etc):
https://github.com/opensafely-core/airlock/blob/eda259ed1b2004de9bed5d932ec758106b2c475f/dotenv-sample

I guess this is similar to Lucy's suggestion of setting the env var by default in the Justfile.

jobserver/urls.py
iaindillingham: Of course we do! Thank you for spotting this.

jobserver/settings.py
```diff
@@ -45,6 +45,8 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = env.bool("DEBUG", default=False)
 
+DEBUG_TOOLBAR = env.bool("DJANGO_DEBUG_TOOLBAR", default=False)
```
iaindillingham, line 48: In Airlock, we do the following:

```py
if DEBUG:  # pragma: no cover
    DJANGO_DEBUG_TOOLBAR = os.environ.get("DJANGO_DEBUG_TOOLBAR", "false") == "True"
else:  # pragma: no cover
    DJANGO_DEBUG_TOOLBAR = False
```

@evansd, you mentioned that we should avoid making lots of behaviour conditional on `DEBUG`. For this reason, I've not done in job-server what we do in Airlock. Have I missed something?

evansd: No, I think you're right.

The [default behaviour](https://django-debug-toolbar.readthedocs.io/en/stable/configuration.html#show-toolbar-callback) of DDT is to disable itself when `DEBUG = False` so in one sense there's not much point allowing it to be enabled when `DEBUG = False`. But I think it's better to keep things independent.

jobserver/urls.py
```diff
@@ -337,7 +341,7 @@
     path("ui-components/", components),
     path("users/", include(user_urls)),
     path("workspaces/", yours.WorkspaceList.as_view(), name="your-workspaces"),
-    path("__debug__/", include(debug_toolbar.urls)),
+    *debug_toolbar_urls,
     path("<str:project_slug>/", include(project_urls)),
 ] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```
lucyb, line 346: Following Dave's suggestion, you could change this line for consistency, so we aren't using two different ways to add to a list. However, I can understand not wanting to make an unrelated change right now.

i.e turn:
```
] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```

into:
```
    *static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
]
```

lucyb (approved):
This is a really nice solution and a significant improvement. Thanks for responding so positively to various pieces of feedback.

jobserver/urls.py
iaindillingham: Thanks, Lucy. I considered doing so, but I didn't want to introduce an unrelated change into the diff.

PR #4487, opened 2024-07-25
Exclude floats from sensitive token checking

Jongmassey:
Sentry cron check-in events (#4421) contain float-typed fields which the `parse()` method would try to iterate whilst searching for sensitive tokens, causing an error.

The tokens listed in this method are all strings so safe to exclude floats for now.

### DIFF
--- a/tests/unit/services/test_sentry.py
+++ b/tests/unit/services/test_sentry.py
@@ -48,6 +48,10 @@ def test_parse_without_envvar(event):
 
     data1 = event("sentry_event.json")
     data2 = event("sentry_event2.json")
+    data3 = event("sentry_event_cron_in_progress.json")
+    data4 = event("sentry_event_cron_error.json")
 
     assert parse(data1) == data1
     assert parse(data2) == data2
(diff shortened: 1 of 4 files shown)


tests/unit/services/test_sentry.py
```diff
@@ -48,6 +48,10 @@ def test_parse_without_envvar(event):
 
     data1 = event("sentry_event.json")
     data2 = event("sentry_event2.json")
+    data3 = event("sentry_event3.json")
```
lucyb, line 51: Would you mind renaming the two json files you've added to something that's more expressive, please? The existing names aren't very clear; I've no idea what sentry_event2.json is for and I'm sure I'll ask the same about the ones you've added in a few months too!

lucyb (approved):

tests/unit/services/test_sentry.py
Jongmassey: Fair comment - I've got no idea about 1 or 2 so I'll leave them alone

PR #4499, opened 2024-07-30
Move health check to `app.json`

### LINKED ISSUE #4489: Dokku CHECKS file no longer supported
Dokku have removed support for a CHECKs file in a recent version (see [migration guide](https://dokku.com/docs/appendices/0.32.0-migration-guide/)). This means we're no longer using the settings we've defined in that file as Dokku is expecting it to be in the app.json file instead.

We should check which of those settings we still need and move them to the app.json.

StevenMaude:
Fixes #4489.

`WAIT` was originally the time to wait before running the first check as explained, for example, in this old Dokku documentation page:

https://dokku.com/docs~v0.3.26/checks-examples/

This now corresponds to `initialDelay` in the `app.json` health check:

> Number of seconds to wait after a container has started before triggering the healthcheck.

(`wait` in `app.json` corresponds to the time between check attempts, which doesn't, from my reading, seem to have been configurable in the old `CHECKS` configuration. That is, a subsequent check would occur immediately after a failed check.)

### DIFF
--- a/app.json
+++ b/app.json
@@ -21,6 +21,7 @@
     "web": [
       {
         "name": "Web health check",
+        "type": "startup",
         "description": "Check if the app responds to the /health-check endpoint",
         "path": "/health-check",
         "port": 8000,
--- a/app.json
+++ b/app.json
@@ -16,5 +16,18 @@
       "command": "python manage.py runjobs monthly",
       "schedule": "@monthly"
     }
-  ]
+  ],
+  "healthchecks": {
+    "web": [
+      {
+        "name": "Web health check",
+        "description": "Check if the app responds to the /health-check endpoint",
+        "path": "/health-check",
+        "port": 8000,
+        "attempts": 5,
+        "initialDelay": 30,
+        "timeout": 60
+      }
+    ]
+  }
 }
--- a/CHECKS
+++ b/CHECKS
@@ -1,5 +1,0 @@
-WAIT=30
-TIMEOUT=60
-ATTEMPTS=5
-
-/health-check/


app.json
```diff
@@ -16,5 +16,18 @@
...
     }
-  ]
+  ],
+  "healthchecks": {
+    "web": [
+      {
+        "name": "Web health check",
+        "description": "Check if the app responds to the /health-check endpoint",
+        "path": "/health-check",
+        "port": 8000,
```
StevenMaude, line 26: thought: We didn't have to configure the port before. The [current Dokku documentation](https://dokku.com/docs/deployment/zero-downtime-deploys/#customizing-checks) says the health check defaults to `5000` in this new health check configuration. We expose port 8000 via the Dockerfile.

I'm not sure if there's some other magic that would remove the need for specifying the port.

StevenMaude:
I'm also unsure of is whether there's a nice way to test this configuration out without just merging it.

app.json
Jongmassey: If it works then I think it's fine for us to specify it here

Jongmassey (approved):

Jongmassey:
Spinning up a local dokku and [manually running the check](https://dokku.com/docs/deployment/zero-downtime-deploys/#manually-invoking-checks) might be a way?

StevenMaude:
Haven't completely got Dokku working yet.

But I did manage to at least get it to parse the health check from app.json; it complained about missing `type` key. I'm not sure if that's sufficient to break the health check, but maybe worth adding.

StevenMaude:
This is what I have for Dokku running job-server in Codespaces:

```json
{
  "cron": [
    {
      "command": "python manage.py runjobs hourly",
      "schedule": "@hourly"
    },
    {
      "command": "python manage.py runjobs daily",
      "schedule": "@daily"
    },
    {
      "command": "python manage.py runjobs weekly",
      "schedule": "@weekly"
    },
    {
      "schedule": "@monthly"
    }
  ],
  "healthchecks": {
    "web": [
      {
        "name": "Web health check",
        "type": "startup",
        "description": "Check if the app responds to the /health-check endpoint",
        "path": "/health-check",
        "port": 8000,
        "attempts": 5,
        "initialDelay": 30,
        "timeout": 60
      }
    ]
  }
}
```

StevenMaude:
The format of the listening check log is the same as the one that currently runs in production:

```
Running healthcheck name='port listening check' attempts=3 port=8000 retries=2 timeout=5 type='listening' wait=5
```

StevenMaude:
It wasn't actually necessary to go all the way of running job-server in Dokku completely to see the health check failure. However, I did, so have an idea of how to do that now. I've made notes and can document this later.

StevenMaude:
If someone merges this one, I'd recommend checking the deploy logs and seeing the health check functions as expected in production.

PR #4503, opened 2024-07-30
Remove `first` dependency

### LINKED ISSUE #4491: Remove `first`
[`first`](https://pypi.org/project/first/) returns the first true value from an iterable, or `None`. It's implemented in [10 lines](https://github.com/hynek/first/blob/main/first.py#L69-L78) of Python, and it's imported 12 times; in middleware, views, models, and tests. In each case, it would be useful to know whether we expect the iterable to contain a true value. In other words, whether we expect `first(items)` to return `item` or `None`.

As an example, consider `TemplateNameMiddleware`:

https://github.com/opensafely-core/job-server/blob/ece78bb83e928de911a8d2437c81bae89e564544/jobserver/middleware.py#L19-L45

I'm fairly sure we want to reduce many template names to one template name, so we don't expect `first(template_name)` to return `None`.

https://github.com/opensafely-core/job-server/blob/ece78bb83e928de911a8d2437c81bae89e564544/jobserver/middleware.py#L40-L42

Setting aside how we've chosen to name our variables, wouldn't the following be clearer?

```py
template_name = next(name for name in template_name if name)
```

Of course, if `template_name` didn't contain a true value, then the above would raise a `StopIteration` exception. Again, wouldn't handling that exception be clearer?

StevenMaude:
Fixes #4491.

For context, `first` is a function in a [package also called `first`](https://github.com/hynek/first/). Briefly, from the way we use `first()` in job-server, it's useful to know that:

* `first` returns the first truthy value from an iterable.
* `first` can also take an optional `default` argument to provide if no such value is found. If no default is specified, the default is `None`.

This PR replaces this dependency with inline code. This reduces our dependency count by 1, and makes it more immediately clear to the reader what the code is doing by removing the indirection.


StevenMaude:
TODO:
- [ ] ensure the removal is correct and tests pass
- [ ] reconsider use of a `None` default; some of the code has an implicit assertion that there will be a valid `first` value, perhaps we could make this explicit?

tests/integration/test_interactive.py
```diff
@@ -196,7 +195,10 @@ def test_interactive_publishing_report_success(
         string=re.compile(".*Created by.*")
     )
     spans = [s.parent for s in strings]
-    span = first(spans, key=lambda span: "on" in span.text)
+    try:
+        span = next(sp for sp in spans if "on" in sp.text)  # pragma: no branch
+    except StopIteration:  # pragma: no cover
+        pytest.fail("This test currently asserts a span will be found")
```
StevenMaude, line 201: thought: Previously `first()` would default to `None` as a value if no match is found.

For the tests, I've favoured explicitly adding failure states where we shouldn't ever fail.

This then needs:

* `no branch` because `coverage` complains about [partial branch coverage](https://github.com/nedbat/coveragepy/blob/2a7d9779add1e8fa3425a784347316f5d28a510d/doc/branch.rst#L120-L131); in these cases, the branch where the generator is exhausted is not reached in these cases, as we do find a match
* `no cover` because we should never get the failure state occurring if the tests pass

jobserver/middleware.py
```diff
@@ -39,7 +38,9 @@ def process_template_response(self, request, response):
 
         # flatten template_name to a string
         if isinstance(template_name, list):
-            template_name = first(template_name)
+            template_name = next(  # pragma: no branch
+                (t for t in template_name if t), None
+            )
```
StevenMaude, line 43: thought: This is the existing behaviour.

Should this fail more loudly if there is no template name found? Is that ever possible?

jobserver/views/workspaces.py
```diff
@@ -183,7 +182,7 @@ def dispatch(self, request, *args, **kwargs):
             )
 
         org = self.request.user.orgs.first()
-        gh_org = first(org.github_orgs)
+        gh_org = next((o for o in org.github_orgs if o), None)
```
StevenMaude, line 185: thought: This feels OK to leave as is, with the default set. We explicitly check for that default in the next line.

redirects/models.py
```diff
@@ -132,7 +131,8 @@ def get_staff_delete_url(self):
     @property
     def obj(self):
         """Work out which object we're pointing to"""
-        return first([getattr(self, f.name, None) for f in self.targets()])
+        field_names = [getattr(self, f.name, None) for f in self.targets()]
+        return next((f for f in field_names if f), None)  # pragma: no branch
```
StevenMaude, line 135: thought: This could perhaps complain more loudly, instead of returning `None`.

Do we ever expect it could not find a match?

(In other code, we access attributes of `.obj()`. For example, [in `type()` below](https://github.com/opensafely-core/job-server/pull/4503/files#diff-b4dca3e75ff654d499fcd477fafbf5671808e7180271f9f3bbd1f6e91172c75cR152-R153).)

jobserver/forms.py
```diff
@@ -136,7 +141,10 @@ def clean(self):
         if not (repo_url and branch):
             return
 
-        repo = first(self.repos_with_branches, key=lambda r: r["url"] == repo_url)
+        repo = next(
+            (r for r in self.repos_with_branches if r["url"] == repo_url),
+            None,
+        )
```
StevenMaude, line 147: thought: This feels OK to leave as is, with the default set. We explicitly check for that default in the next line.

tests/unit/redirects/test_models.py
```diff
@@ -18,7 +17,9 @@ def get_factory(model):
     classes = [value for name, value in inspect.getmembers(factories, inspect.isclass)]
     all_factories = [c for c in classes if c.__name__.endswith("Factory")]
 
-    return first(all_factories, key=lambda m: m._meta.model == model)
+    return next(  # pragma: no branch
+        (f for f in all_factories if f._meta.model == model), None
+    )
```
StevenMaude, line 22: thought: Could this complain more loudly with `StopException`? Would we ever expect a model object doesn't have a factory?

jobserver/forms.py
```diff
@@ -114,12 +113,18 @@ def __init__(self, project, repos_with_branches, *args, **kwargs):
...
-            repo = first(
-                self.repos_with_branches,
-                key=lambda r: r["url"] == kwargs["data"]["repo"],
+            repo = next(  # pragma: no branch
+                (
+                    r
+                    for r in self.repos_with_branches
+                    if r["url"] == kwargs["data"]["repo"]
+                ),
+                None,
```
StevenMaude, line 122: thought: `None` is not a valid value for the subsequent lines, which check attributes of the value. We could replace these defaults with a `StopException` which should never be reached and warns as such. 

Note that we don't do any validation of inputs in the [code above](https://github.com/opensafely-core/job-server/pull/4503/files#diff-8a8820228bf2d257ba3f9c3c703d55f1247e7be2a2d8d5f7ff61352d19103739R105-R108). Therefore, I'm not sure whether to start adding such validation here.

staff/views/dashboards/repos.py
```diff
@@ -72,7 +71,7 @@ def enhance(repo):
                 w for w in all_workspaces if repo["url"].lower() == w.repo.url.lower()
             ]
             workspaces = sorted(workspaces, key=lambda w: w.name.lower())
-            workspace = first(workspaces)
+            workspace = next((w for w in workspaces if w), None)
```
StevenMaude, line 74: thought: This feels OK to leave as is, with the default set.

StevenMaude (commented):
question: My main two questions for these changes are:

* *For consistency, should we always use a `try: next( … except StopException:…` explicitly when replacing `first`?*
  * This would make the format of the code that replaces `first` *consistent*.
  * Or is it preferable to use a default value — currently `None` in every case — instead where we need a default value to be specified. That can be expressed more *concisely* as `next(iterable, default)` without the `try… except…`.
* There are a few places in the code I've highlighted where the behaviour of the existing code was masked by `first()`. We currently have to use `pragma: no branch` in many of the cases where we've replaced `first`. This flags that our tests don't currently execute the branch where we don't find a value from the provided iterator. *What should the behaviour be in these cases?*
  * I don't know enough about Django/job-server to make the most sensible choice. Should these replacements of `first` raise `StopException` instead, where the `except` does something akin to `assert False` to imply this state is assumed to never be reached, or perhaps do more graceful handling of a missing value?
    * With the good test coverage in this project, my reading for the lack of tests for these cases is that the "no value found" state is never expected to be reached.
    * For the uses of `first` in the tests, I've explicitly marked the `StopException` as a test failure. We know the input values that we provide there and these shouldn't result in that happening.

StevenMaude:
note: This PR currently includes a copied over version of `first` that is then subsequently removed. The license of that code permits it. We could remove those commits before merging.

jobserver/forms.py
lucyb, line 1: I'd like to comment on a commit, not on a particular file, but I don't think GitHub will let me... anyway, I would have been tempted to rewrite the imported first function to something like:
```
def first(iterable, default=None, key=None):
    if key is None:
        return next((i for i in iterable if i), default)
    else:
        return next((i for i in iterable if key(i)), default)
```
in one commit and then doing the refactoring step to inline the function separately. This would have made it easier for the reader to see that what you're replacing it with is equivalent because the tests will have passed. I expect you effectively did this locally anyway.

lucyb:
I hope you don't mind, but I've merged the first half of this PR [here](https://github.com/opensafely-core/job-server/pull/4514) and then rebased this one to remove the commits that have already been merged.

jobserver/forms.py
```diff
@@ -115,12 +113,18 @@ def __init__(self, project, repos_with_branches, *args, **kwargs):
...
             )
         else:
-            repo = first(self.repos_with_branches)
+            repo = next(  # pragma: no branch
+                (r for r in self.repos_with_branches if r), None
```
lucyb, line 126: From looking at the code above, it doesn't look like we're expecting `r` to be falsy here, so can we just write `repo = self.repos_with_branches[0]` instead?

tests/integration/test_interactive.py
StevenMaude: Have removed the commit with these changes.

tests/unit/redirects/test_models.py
StevenMaude: I've removed the defaults from all the tests.

jobserver/forms.py
StevenMaude: I think this is reasonable.

jobserver/forms.py
StevenMaude: To make this clear, I've added a commit that does exactly that, but retains the `first` tests. The `first` tests still pass.

jobserver/views/workspaces.py
StevenMaude: I've simplified this code.

redirects/models.py
StevenMaude: Still not entirely sure about this. No big harm in leaving it, but not sure.

staff/views/dashboards/repos.py
StevenMaude: I think this is fair reasoning.

redirects/models.py
StevenMaude: I think the default's removable; the only reference to this code is in `Redirect.type()` which accesses attributes, as I said, so the implicit assumption is that it's a non-falsy value.

redirects/models.py
```diff
@@ -133,7 +131,8 @@ def get_staff_delete_url(self):
     @property
     def obj(self):
         """Work out which object we're pointing to"""
-        return first([getattr(self, f.name, None) for f in self.targets()])
+        field_names = [getattr(self, f.name, None) for f in self.targets()]
+        return next(f for f in field_names if f)  # pragma: no branch
```
lucyb, line 135: I'm a bit concerned that we have `# pragma: no branch` in production code, especially when that's code that's just been changed. However, it might well be that it's a pragmatic solution and less risky than removing the `if f` (to use this code snippet as an example). I wouldn't want people to think it was okay these lines weren't tested in future though.

I would recommend either removing the conditional (and manually testing that's okay) or adding in an extra unit test so that coverage doesn't complain. How and when to do that is up to you though.

lucyb (approved):
This is a really nice improvement. I also really like some of the code simplification that you've been able to do as a result. It's actually much clearer now, in some instances, what the code is doing.

redirects/models.py
StevenMaude: I'll merge this now, and opened #4562 for fixing up this issue.

From memory, I _think_ some of the issue here may also be the code branch reached at the end of the iterator. And not just the `if` branch.

Without testing, it's also possible that this case was buried by virtue of `coverage` not recognising it, but it's made explicit now. See [this comment](https://github.com/nedbat/coveragepy/issues/475#issuecomment-697375303).

It's possible there's some debate on how best to solve that, which is maybe best in a new PR.

PR #4504, opened 2024-07-31
Remove ability to run interactive reports

Jongmassey:


Jongmassey:
@ 15201d8c
image

@ a865200

image

lucyb:
Hello. This is a drive-by comment, so apologies if it's unhelpful, but I noticed the test failure and thought that you may want to take a look at how we're writing some of our other tests. Rather than creating a user with a particular role, we're manufacturing a role with the exact permission we want to test. It makes the test more explicit and I think will help you with this failure.

For example:
https://github.com/opensafely-core/job-server/blob/84d04a2df6e6a63abc20e6ad88dd3a3f57ee4742/tests/unit/interactive/test_views.py#L39-L43

Jongmassey:
After these changes, `InteractiveReporter` retains the `release_file_view` permission that is required to view `AnalysisRequestDetail` and this test still passes

https://github.com/opensafely-core/job-server/blob/84d04a2df6e6a63abc20e6ad88dd3a3f57ee4742/tests/unit/interactive/test_views.py#L155-L175


Ditto the `analysis_request_view` permission which is required to edit a report's metadata:
https://github.com/opensafely-core/job-server/blob/84d04a2df6e6a63abc20e6ad88dd3a3f57ee4742/tests/unit/interactive/test_views.py#L360-L388

lucyb (approved):
Thanks for doing this and for updating the tests to ensure they're showing the correct behaviour.

PR #4505, opened 2024-07-31
Warn if templates are missing variables

### LINKED ISSUE #4462: Add unit test to fail on warnings in Django Templates
We currently ignore errors in our Django templates ([example](https://bennettoxford.slack.com/archives/C01SE1WB6AW/p1721142918550459)). This is good because then users don't see an error, but it's bad because we don't see the error either. In Airlock, they've configured their app to produce warnings if there are missing variables and added a unit test to fail if there are any warnings. We should do the same in Job Server.

In Airlock, the main bit of code is this logging filter defined here:
https://github.com/opensafely-core/airlock/blob/e4b686debd5738ff639ec469db2396c78cd55fb1/airlock/settings.py#L363-L399

Which is then used in the logging config here:
https://github.com/opensafely-core/airlock/blob/e4b686debd5738ff639ec469db2396c78cd55fb1/airlock/settings.py#L421-L424

Finally, there is a hook in the main conftest.py file which fails the test run if any warnings are present:
https://github.com/opensafely-core/airlock/blob/e4b686debd5738ff639ec469db2396c78cd55fb1/tests/conftest.py#L35-L40

Warnings that they don't care about are configured in pyproject.toml:
https://github.com/opensafely-core/airlock/blob/e4b686debd5738ff639ec469db2396c78cd55fb1/pyproject.toml#L28-L31

iaindillingham:
Previously, messages about missing context variables were ignored. This was because the `django.template` logger logged them as `DEBUG` messages ([1][]), but the `django` logger (the parent of the `django.template` logger) sent only `INFO` messages (or higher) to the console.

Now, the `django.template` logger sends `DEBUG` messages (or higher) to the `missing_variable_error` filter. This filter always returns `False`, meaning that messages are (still) not sent to the console. However, it also converts messages about missing context variables to warnings.

pytest catches warnings during, and displays them at the end of, test sessions ([2][]). So, if a template is missing a context variable, and the template response associated with the template is tested, then pytest should tell us, shouldn't it? Sadly, it's not that simple. Messages about missing context variables are only generated when a template response is rendered. If a test constructs, but doesn't render, a template response, then pytest won't tell us about missing context variables because it won't know about them either. In these cases, because warning messages are written to `sys.stderr`, we would most likely find out about missing context variables by inspecting the output of Django's development server.

Notice that the `missing_variable_error` filter only converts *some* messages about missing context variables to warnings. Whilst some are beyond our control, others are not cost- effective to fix.

[1]: https://docs.djangoproject.com/en/5.1/ref/logging/#django-template
[2]: https://docs.pytest.org/en/latest/how-to/capture-warnings.html#warnings

Closes #4462

### DIFF
--- a/services/logging.py
+++ b/services/logging.py
@@ -80,6 +80,12 @@ class MissingVariableErrorFilter(logging.Filter):
         # are missing when pages are requested by a `RequestFactory` instance.
         "csp_nonce",  # csp.middleware.CSPMiddleware
         "template_name",  # jobserver.middleware.TemplateNameMiddleware
+        # And some template variables are just a pain, because of how we use them.
+        "name",  # jobserver.context_processors.nav
+        "q",
+        # The following is a property on User but not AnonymousUser. It's non-trivial to
+        # extend the latter to provide it, so we ignore it.
+        "all_roles",
     }
 
     def filter(self, record):  # pragma: no cover
--- a/services/logging.py
+++ b/services/logging.py
@@ -1,3 +1,6 @@
+import logging
+import warnings
+
 import structlog
 from environs import Env
 
@@ -50,6 +53,52 @@ structlog.configure(
     wrapper_class=structlog.stdlib.BoundLogger,
 )
 
+
+class MissingVariableErrorFilter(logging.Filter):
+    """
+    Convert "missing template variable" log messages into warnings.
+
+    Heavily inspired by Adam Johnson's work here:
+    https://adamj.eu/tech/2022/03/30/how-to-make-django-error-for-undefined-template-variables/
+    """
+
+    ignored_prefixes = (
+        # Some of Django's internal templates rely on the silent missing template
+        # variable behaviour
+        "admin/",
+        "auth/",
+        "django/",
+        # As does our internal component library
+        "_components",
+        "_partials",
+        # As do two of our internal applications
+        "applications/",
+        "interactive/",
+    )
+    ignored_variable_names = {
+        # Some template variables are added by middleware, and so
+        # are missing when pages are requested by a `RequestFactory` instance.
+        "csp_nonce",  # csp.middleware.CSPMiddleware
+        "template_name",  # jobserver.middleware.TemplateNameMiddleware
+    }
+
+    def filter(self, record):  # pragma: no cover
+        if record.msg.startswith("Exception while resolving variable "):
+            variable_name, template_name = record.args
+            if (
+                not template_name.startswith(self.ignored_prefixes)
+                # This shows up when rendering Django's internal error pages
+                and template_name != "unknown"
+                and variable_name not in self.ignored_variable_names
+            ):
+                # We use `warn_explicit` to raise the warning at whatever location the
+                # original log message was raised
+                warnings.warn_explicit(
+                    record.getMessage(), UserWarning, record.pathname, record.lineno
+                )
+        return False
+
+
 logging_config_dict = {
     "version": 1,
     "disable_existing_loggers": False,
@@ -68,11 +117,20 @@ logging_config_dict = {
         }
     },
     "root": {"handlers": ["console"], "level": "WARNING"},
+    "filters": {
+        "missing_variable_error": {
+            "()": f"{MissingVariableErrorFilter.__module__}.{MissingVariableErrorFilter.__name__}"
+        }
+    },
     "loggers": {
         "django": {
             "handlers": ["console"],
             "level": env.str("DJANGO_LOG_LEVEL", default="INFO"),
             "propagate": False,
+        },
+        "django.template": {
+            "level": "DEBUG",
+            "filters": ["missing_variable_error"],
         },
         "gunicorn.access": {
             "handlers": ["console"],
--- a/tests/unit/services/test_logging.py
+++ b/tests/unit/services/test_logging.py
@@ -1,5 +1,8 @@
 from datetime import datetime
 
+import pytest
+from django.template import Context, Template
+
 from services.logging import timestamper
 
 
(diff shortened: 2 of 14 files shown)


tests/unit/services/test_logging.py
```diff
@@ -17,3 +20,25 @@ def test_timestamper_without_debug(monkeypatch):
     monkeypatch.setattr("services.logging.DEBUG", False)
 
     assert timestamper(None, None, {"event": "derp"}) == {"event": "derp"}
+
+
+def test_missing_variable_error_filter():
```
lucyb, line 25: These tests are really nice. I tried changing the `django.template` logging settings because I thought it might be possible to use some better options and I broke one of the tests as a result, which immediately told me I'd done the wrong thing.

services/logging.py
```diff
@@ -80,6 +80,13 @@ class MissingVariableErrorFilter(logging.Filter):
         # are missing when pages are requested by a `RequestFactory` instance.
         "csp_nonce",  # csp.middleware.CSPMiddleware
         "template_name",  # jobserver.middleware.TemplateNameMiddleware
+        # And some template variables are just a pain, because of how we use them.
+        "name",  # jobserver.context_processors.nav
+        "q",
+        "get_staff_url",
```
lucyb, line 86: As far as I can see, `get_staff_url` is currently only producing a warning from one test:
```
FAILED tests/integration/test_interactive.py::test_interactive_publishing_report_success - UserWarning: Exception while resolving variable 'get_staff_url' in template 'staff/report/detail.html'.
```

If that's correct, then it would be worth just editing that template (I think it should be a simple fix).

lucyb (commented):
I'm noticing some exceptions in the logs when manually checking the site. They aren't being reported as errors though (seems to be `info` level), which is good because it won't trigger sentry. I think they might be a side effect of some changes you made earlier, so might be worth looking at now (happy for them to be fixed later if they're unrelated):

```
/home/lucy/datalab/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/template/base.py:925: UserWarning: Exception while resolving variable 'yes_active_class' in template 'staff/user/list.html'.
  logger.debug(
/home/lucy/datalab/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/template/base.py:925: UserWarning: Exception while resolving variable 'no_active_class' in template 'staff/user/list.html'.
  logger.debug(
/home/lucy/datalab/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/template/base.py:925: UserWarning: Exception while resolving variable 'value' in template 'staff/user/list.html'.
  logger.debug(
```

And 

```
/home/lucy/datalab/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/template/base.py:925: UserWarning: Exception while resolving variable 'value' in template 'staff/index.html'.
  logger.debug(
```

And 

```
/home/lucy/datalab/opensafely-core/job-server/.venv/lib/python3.12/site-packages/django/template/base.py:925: UserWarning: Exception while resolving variable 'value' in template 'staff/report/list.html'.
  logger.debug(
```

lucyb:
> Thanks for talking this through earlier, @lucyb. Since we spoke, I have made three changes.

This all sounds very sensible. Thank you again for walking me through this. It's not easy to make a change like this in a mature application without getting lost in the weeds and I think you've found some good, pragmatic solutions to the inevitable snags that have come up.

iaindillingham:
Thanks for your comments, @lucyb.

> I'm noticing some exceptions in the logs when manually checking the site.

I _think_ you're noticing warnings, rather than exceptions (although the text is confusing!). The warnings are created from log records by `MissingVariableErrorFilter.filter`:

https://github.com/opensafely-core/job-server/blob/b1b58e883cb2c26bdcb0ed4f90d6437d718da8f1/services/logging.py#L103-L105

I'm afraid I was mistaken when I predicted that we would see warnings in Sentry. We won't, because warnings are not logs, and Sentry handles logs. `MissingVariableErrorFilter.filter` returns `False`, meaning it filters out the log records. And even if it didn't, then they're at the wrong level. (I'm confused by your comment about the info level. Do you mean the debug level? As you say, though, Sentry's `DjangoIntegration` captures at the error level.)

I'm concerned that you (and I) can see warnings when manually checking the site. At first, I thought this was because the templates weren't covered by tests. However, this doesn't seem to be the case. Take the following example:

```
UserWarning: Exception while resolving variable 'value' in template 'staff/index.html'
```

This template is associated with `staff.views.index.Index`, which is tested by `tests.unit.staff.views.test_index.test_index_without_search`. Why isn't the test failing? Because the test doesn't render the response! 🤦🏻‍♂️

Because we won't see warnings in Sentry and because we can see warnings when manually checking the site, I don't think this PR gives us the confidence we need. I think we should render the responses (and fix the templates), although I imagine we're not doing that now for performance. I think we should also consider not filtering out the log records, and promoting them to errors, such that we see them in Sentry (see "[With a logging filter that promotes messages to errors](https://adamj.eu/tech/2022/03/30/how-to-make-django-error-for-undefined-template-variables/#with-a-logging-filter-that-promotes-messages-to-errors)"). Doing both should ensure that we have more confidence in our tests and, should they be deficient, then Sentry will remind us (when a user visits the page).

lucyb:
FYI I've rebased this commit against main as I incorporated one of your commits into [this PR](https://github.com/opensafely-core/job-server/pull/4513).

iaindillingham:
@lucyb, I've addressed the `get_staff_url` issue. I haven't fixed the cases where warnings are displayed on stderr, but the tests for the corresponding views pass (because we don't render the template responses). As you said when we last spoke, this is a step -- but not all the steps -- in the right direction 🙂

lucyb (approved):

PR #4506, opened 2024-08-02
Rename "GITHUB_TOKEN"

### LINKED ISSUE #4318: Rename `GITHUB_TOKEN`
job-server uses an environment variable named `GITHUB_TOKEN` to undertake various tasks, such as fetching the status of repos (public or private) for the co-pilot dashboard. There is a placeholder `GITHUB_TOKEN` in `dotenv-sample`, from which `.env` is created. It is often the case that `.env` is sourced, such that `GITHUB_TOKEN` is present in a local development environment.

GitHub CLI (`gh`) also uses an environment variable named `GITHUB_TOKEN` (see "[gh environment][1]").

> `GH_TOKEN`, `GITHUB_TOKEN` (in order of precedence): an authentication token for github.com API requests. Setting this avoids being prompted to authenticate and takes precedence over previously stored credentials.

Consequently, if a developer uses `gh` within their local development environment, then it is likely that they will authenticate to github.com using it, rather than an alternative method (see "[gh auth login][2]"). Furthermore, the OAuth scopes will be that of `GITHUB_TOKEN` rather than that of the alternative method. The developer will likely not expect the change in authentication method and OAuth scopes 😕.

[1]: https://cli.github.com/manual/gh_help_environment
[2]: https://cli.github.com/manual/gh_auth_login

Jongmassey:
Fixes #4318

This is to distinguish from any GITHUB_TOKEN env var that might already exist on a developers machine.

Name chosen to mirror the existing GITHUB_WRITEABLE_TOKEN and the description given for this token in DEVELOPERS.md


Jongmassey:
Before deploying to production, the dokku config should be updated

`dokku config:set job-server GITHUB_READONLY_TOKEN='[value of pre-existing GITHUB_TOKEN setting]'`

services/sentry.py
```diff
@@ -70,7 +70,7 @@ def parse(data):
 
     # tokens with values we want to strip
     tokens = [
-        "GITHUB_TOKEN",
```
lucyb, line 73: It might be worth keeping this value in just in case. I don't often say that, but in this instance I think we want to be very sure this is always stripped even if we shouldn't be using a token with that name.

.github/workflows/main.yml
```diff
@@ -76,7 +76,7 @@ jobs:
       - name: Run tests
         env:
           DATABASE_URL: postgres://user:password@localhost/jobserver
-          GITHUB_TOKEN: empty
```
lucyb, line 79: The GITHUB_TOKEN used in CI is different to the one we're using locally and in prod, so these two yaml files shouldn't be changed.

Jongmassey: Ah thanks for letting me know, I was a bit unsure of this change

lucyb (commented):
This looks good, thanks for taking onboard my feedback.

Would you like to add the new env vars in production and then I'll approve the PR? I don't want it to be merged accidentally. You'll need to remove the old ones after a successful deployment too, obviously.

Jongmassey:
OK, those env vars are now set in prod

lucyb (approved):

PR #4516, opened 2024-08-07
Improve wording shown to the public

lucyb:
This clarifies the use of highly senstive data and improves the wording in a few other places. In particular, I think it's important that we're consistent with the nouns we use to describe concepts such as jobs, actions and environments.


madwort (approved):
thanks for tweaking

PR #4522, opened 2024-08-08
Move verification tests to own workflow

Jongmassey:


Jongmassey:
7d05bfa was added just to give me confidence the github workflow works. It should be removed before merging to `main`

.github/workflows/verification-tests.yml
```diff
@@ -0,0 +1,86 @@
...
+  workflow_dispatch:
+  schedule:
+    - cron:  "47 0 * * *"
+
+jobs:
+  test:
+    runs-on: ubuntu-22.04
+
+    services:
+      postgres:
```
lucyb, line 21: Technically, these tests don't require the use of a database. However, we've created a fixture called `enable_db_access_for_all_tests` and set `autouse=True` on it, which means that currently every test does require the database. This hasn't been an issue before, but it now seems unnecessary.

We can make that fixture optional (according to stackoverflow) by creating a pytestmark called `disable_db` and changing the fixture to this:

```python
@pytest.fixture(autouse=True)
def enable_db_access_for_all_tests(request):
    disable_db = request.node.get_closest_marker("disable_db")
    if disable_db:
        yield
    else:
        yield request.getfixturevalue("db")
```

.github/workflows/verification-tests.yml
```diff
@@ -0,0 +1,86 @@
...
+          install-just: true
+
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: ".node-version"
+          cache: "npm"
+          cache-dependency-path: package-lock.json
+
+      - name: Install node_modules
```
lucyb, line 49: I don't think we need to run any set up for the assets or to run the js tests, because I don't think our verification tests require them.

.github/workflows/verification-tests.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Run verification tests
+        env:
+          DATABASE_URL: postgres://user:password@localhost/jobserver
+          GITHUB_TOKEN: empty
+          GITHUB_TOKEN_TESTING: ${{ secrets.OPENSAFELY_GITHUB_TESTING_ORG_PAT }}
+          SECRET_KEY: 12345
+          SOCIAL_AUTH_GITHUB_KEY: test
+          SOCIAL_AUTH_GITHUB_SECRET: test
+        run: |
+          just check-migrations
```
lucyb, line 73: We won't need to check migrations, if we remove the need for the database.

.github/workflows/verification-tests.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Build assets
+        run: just assets-build
+
+      - name: Install venv
+        run: just devenv
+
+      - name: Run verification tests
+        env:
+          DATABASE_URL: postgres://user:password@localhost/jobserver
+          GITHUB_TOKEN: empty
```
lucyb, line 67: Can we remove the `GITHUB_TOKEN` now?

justfile
```diff
@@ -134,7 +134,17 @@ run-telemetry: devenv
...
+    export COVERAGE_PROCESS_START="pyproject.toml"
+    export COVERAGE_REPORT_ARGS="--omit=jobserver/github.py,jobserver/opencodelists.py,tests/fakes.py,tests/verification/*"
+    ./scripts/test-coverage.sh -m "not verification" {{ args }}
+
+
+test-verification *args: assets
+    #!/bin/bash
+    export COVERAGE_PROCESS_START="pyproject.toml"
+    export COVERAGE_REPORT_ARGS="--include=jobserver/github.py,jobserver/opencodelists.py,tests/fakes.py,tests/verification/*"
+    ./scripts/test-coverage.sh -m "verification" {{ args }}
```
lucyb, line 147: Looking into a bit more detail about these tests, I think we have some integration tests as well as verification tests, so enabling coverage seems like it's the right thing to do. Although, as you say, we can always revisit that idea in future.

docs/adr/0021-move-verification-tests.md
```diff
@@ -0,0 +1,25 @@
...
+These verification tests made requests to the GitHub API.
+
+Lots of CI runs in close succession could result in this API returning 403 errors (see #4434), causing test failures.
+
+## Decision
+
+Verification tests will be moved to their own GitHub workflow, which will be run on its own nightly schedule.
+
+## Consequences
+
```
lucyb, line 22: Another (good) consequence is that our deployment pipeline is no longer dependent on the availability of the GitHub or Open Codelists APIs.

docs/adr/0021-move-verification-tests.md
```diff
@@ -0,0 +1,25 @@
+# 1. Avoid listing file counts unless user is authenticated
```
lucyb, line 1: The title of the ADR needs to be updated!

docs/adr/0021-move-verification-tests.md
```diff
@@ -0,0 +1,25 @@
...
+
+Accepted
+
+## Context
+
+Regular CI runs (on push, PR etc.) ran all tests, including the verification tests.
+
+These verification tests made requests to the GitHub API.
+
+Lots of CI runs in close succession could result in this API returning 403 errors (see #4434), causing test failures.
```
lucyb, line 15: I would be a bit more assertive here and say that it `does result` in the API returning 403 errors rather than `could`. 

It's also caused some test failures when two workflows are run concurrently (i.e two of us push to different PRs as the same time as happened with Steve and I recently) because we share state in the opensafely-testing GitHub org.

justfile
```diff
@@ -134,7 +134,17 @@ run-telemetry: devenv
 
 
 test-all *args: assets
```
lucyb, line 136: I agree about renaming this to `test-ci` as it's no longer testing everything. I haven't managed to think of a better name yet!

lucyb (commented):
I know I've written quite a few comments, but this does look good and I'm looking forward to less flaky tests.

The other thing to consider as part of this change is that we'd want to be notified in Slack when the workflow fails. I think there's prior art from ehrql that you can look at to figure out how to do that.

docs/adr/0021-move-verification-tests.md
Jongmassey: 🤦

.github/workflows/verification-tests.yml
Jongmassey: Thanks - that looks like it's worked :)

.github/workflows/verification-tests.yml
Jongmassey: We can!

Jongmassey:
Thanks, @lucyb  for you excellent comments! I've added what I think should be the right slack integration, I'll induce a verification test failure tomorrow and see if it posts to slack :)

.github/workflows/verification-tests.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Run verification tests
+        env:
+          GITHUB_TOKEN_TESTING: ${{ secrets.OPENSAFELY_GITHUB_TESTING_ORG_PAT }}
+          SECRET_KEY: 12345
+          SOCIAL_AUTH_GITHUB_KEY: test
+          SOCIAL_AUTH_GITHUB_SECRET: test
+        run: |
+          # hardcode n because auto=2 in CI for some reason
+          just test-verification --migrations -n 4
```
lucyb, line 32: We probably don't need the `--migrations` flag here anymore.

Jongmassey: ah yes I missed that one - thanks!

docs/adr/0021-move-verification-tests.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Regular CI runs (on push, PR etc.) ran all tests, including the verification tests.
+
+These verification tests make requests to the GitHub API (and the OpenCodelists API).
```
lucyb, line 13: It might be worth explaining what the verification tests are used for, so something like:

> We use verification tests to ensure that our understanding of external APIs is still correct and that our fake API objects, used in our unit tests, are still valid. These verification tests make requests to the GitHub API (and the OpenCodelists API).

lucyb (approved):
Just one tiny comment, but please remove the two `temp` commits before merging!

This looks really nice overall. Thanks for updating the various pieces of documentation and good spot with the docker tests, I'd forgotten about those.

docs/adr/0021-move-verification-tests.md
Jongmassey: 👍

PR #4539, opened 2024-08-14
Don't store user GitHub access-tokens in db

### LINKED ISSUE #4222: Stop storing GitHub access_tokens for users
We get this information when someone logs in via GitHub, from Django Social Auth and store it in the database. To the best of my knowledge we aren't using the token. Having it presents an extra security risk that we could easily avoid, so we should stop saving it to the database.

Jongmassey:
We don't subsequently use these tokens and their unnecessary storage in the database represents a potential security risk.

Fixes #4222

### DIFF
--- a/tests/unit/jobserver/test_auth_pipeline.py
+++ b/tests/unit/jobserver/test_auth_pipeline.py
@@ -84,7 +84,7 @@ def test_pipeline_with_new_user(slack_messages, strategy):
 
     social = output["social"]
     assert social.user == user
-    assert social.extra_data["access_token"] == "sekret"
+    assert "sekret" not in social.extra_data.values()
 
     url = f"http://localhost:8000{user.get_staff_url()}"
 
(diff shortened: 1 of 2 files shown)


Jongmassey:
Removing `access_token` from the `extra_data` passed to `UserSocialAuth.objects.create()`  in the auth pipeline doesn't cause the tests of this pipeline for new or existing users to fail.

tests/unit/jobserver/test_auth_pipeline.py
```diff
@@ -84,7 +84,6 @@ def test_pipeline_with_new_user(slack_messages, strategy):
 
     social = output["social"]
     assert social.user == user
-    assert social.extra_data["access_token"] == "sekret"
```
StevenMaude, line 87: thought: Perhaps it's worth checking that `sekret` _doesn't_ appear as a value anywhere in the extra data? It's a property that we presumably wish to ensure remains so for the foreseeable future.

StevenMaude (approved):
summary: I had one thought to consider, but otherwise looks fine.

tests/unit/jobserver/test_auth_pipeline.py
Jongmassey: Good point, that would make the test as a better expression of "we don't want to be storing this token"

PR #4540, opened 2024-08-14
Remove ProjectDeveloper global role

Jongmassey:
Remove global assignment of the `ProjectDeveloper` role and replace with a dedicated Deployment Administrator role.

### DIFF
--- a/jobserver/authorization/roles.py
+++ b/jobserver/authorization/roles.py
@@ -121,7 +121,6 @@ class ProjectDeveloper:
     description = "Run and cancel Jobs, and manage workspaces."
     models = [
         "jobserver.models.project_membership.ProjectMembership",
-        "jobserver.models.user.User",
     ]
     permissions = [
         job_cancel,
--- a/jobserver/authorization/roles.py
+++ b/jobserver/authorization/roles.py
@@ -146,3 +146,24 @@ class SignOffRepoWithOutputs:
     permissions = [
         repo_sign_off_with_outputs,
     ]
+
+
+class DeploymentAdministrator:
+    """
+    Run and cancel Jobs on any project, for development and maintenance purposes
+    including technical support for Approved Projects.
+    See Developer Permissions Log for the list of individuals who are approved for this role.
+    """
+
+    display_name = "Deployment Administrator"
+    description = """
+    Run and cancel Jobs on any project, for development and maintenance purposes including technical support for Approved Projects.
+    See Developer Permissions Log for the list of individuals who are approved for this role.
+    """
+    models = [
+        "jobserver.models.user.User",
+    ]
+    permissions = [
+        job_cancel,
+        job_run,
+    ]
--- a/jobserver/migrations/0003_remove_global_projectdeveloper.py
+++ b/jobserver/migrations/0003_remove_global_projectdeveloper.py
@@ -1,0 +1,23 @@
+from django.db import migrations
+
+from jobserver.authorization.roles import DeploymentAdministrator, ProjectDeveloper
+
+
+def replace_user_projectdeveloper_with_deploymentadministrator(apps, schema_editor):
+    User = apps.get_model("jobserver", "User")
+    for user in User.objects.filter(roles__contains=ProjectDeveloper):
+        user.roles.remove(ProjectDeveloper)
+        user.roles.append(DeploymentAdministrator)
+        user.save()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("jobserver", "0002_remove_user_github_access_tokens"),
+    ]
+
+    operations = [
+        migrations.RunPython(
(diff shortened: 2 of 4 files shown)


iaindillingham (approved):
Thanks for writing up the rationale for this change.

I've made a couple of comments. Nitpicks aside, it's worth checking with Lucy about the pros and cons of reverse migrations.

You'll have to rework the migration slightly because #4543 will change its dependencies.

iaindillingham (changes requested):
Actually, sorry, I'm going to withdraw my approval to prevent accidental merge. `python manage.py migrate` fails with:

```
AttributeError: 'User' object has no attribute 'roles__contains'
```

I've made a suggestion that should fix this.

jobserver/migrations/0003_remove_global_projectdeveloper.py
```diff
@@ -0,0 +1,32 @@
+from django.db import migrations
+
+from jobserver.authorization.roles import DeploymentAdministrator, ProjectDeveloper
+
+
+def change_user_role(apps, from_role, to_role):
+    User = apps.get_model("jobserver", "User")
+    for user in User.objects.filter(roles__contains=from_role):
+        user.roles.remove(from_role)
+        user.roles.add(to_role)
```
lucyb, line 10: This doesn't work, it needs to be `append` not `add`.

```
  File "/home/lucy/datalab/opensafely-core/job-server/jobserver/migrations/0003_remove_global_projectdeveloper.py", line 10, in change_user_role
    user.roles.add(to_role)
    ^^^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'add'
```

jobserver/migrations/0003_remove_global_projectdeveloper.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("jobserver", "0002_remove_user_github_access_tokens"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            replace_user_projectdeveloper_with_deploymentadministrator,
+            reverse_code=replace_user_deploymentadministrator_with_projectdeveloper,
```
lucyb, line 30: Iain asked for my opinion on reversible migrations. I think in general we want all migrations to be reversible because we want to be able to easily move forward and backwards through different versions of our database. However, this can also be achieved using `reverse_code=migrations.RunPython.noop`.

If I've understood the behaviour correctly, this is running in a transaction. So if the code throws an exception, no change is made and the deployment is stopped. We have no need to rollback at that point. Therefore, I don't think this is particularly useful from the point of view of production (I could be wrong though). However, we might want to rollback locally to move between different versions of the app and in this case, I think we do want to change the user roles back rather than doing a `noop`. For example, to spot and fix the error above (`add` v `append`) I needed to change the data back in order to execute the broken line of code, merely relying on `noop` to move between migrations wasn't enough.

That said, I don't think we should be keeping migrations around for very long, so that makes its use quite limited. In this case though, it's a very small overhead and I found it quite handy in order to be able to fix that bug.

jobserver/migrations/0003_remove_global_projectdeveloper.py
Jongmassey: Thanks - is there a reason CI didn't pick this up?

jobserver/authorization/roles.py
```diff
@@ -146,3 +145,21 @@ class SignOffRepoWithOutputs:
...
+
+class DeploymentAdministrator:
+    """
+    Run and cancel Jobs on any project, for development and maintenance purposes
+    including technical support for Approved Projects.
+    See Developer Permissions Log for the list of individuals who are approved for this role.
+    """
+
+    display_name = "Deployment Administrator"
+    description = "Run and cancel jobs"
```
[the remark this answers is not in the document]
Jongmassey: 🤦‍♂️ d'oh that's what I had meant to do. Thanks for the catch!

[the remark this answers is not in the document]
Jongmassey: image
 UI looks OK to me

jobserver/migrations/0003_remove_global_projectdeveloper.py
Jongmassey: Thanks, in future I will consider the reversibility of any migration I write!

lucyb (approved):

jobserver/migrations/0003_remove_global_projectdeveloper.py
iaindillingham: That's really useful, thanks @lucyb.

jobserver/migrations/0003_remove_global_projectdeveloper.py
iaindillingham: Archaeologists, see #4573 ⛏️

PR #4541, opened 2024-08-15
Only show repo sign off button to permissioned users

Jongmassey:
bug: workspaces with private repos will show the "admin section" to all users regardless of their permissions.

cause: the display logic is to check that the repo is private OR the user has one of two relevant permissions.

fix: this changes the display logic to check that the repo is private AND the user has one of two relevant permissions.

### DIFF
--- a/jobserver/views/workspaces.py
+++ b/jobserver/views/workspaces.py
@@ -305,9 +305,7 @@ class WorkspaceDetail(View):
         has_backends = request.user.is_authenticated and request.user.backends.exists()
 
         # should we show the admin section in the UI?
-        show_admin = (
-            can_archive_workspace or repo_is_private or can_toggle_notifications
-        )
+        show_admin = can_archive_workspace or can_toggle_notifications
 
         honeycomb_can_view_links = has_role(self.request.user, CoreDeveloper)
 
--- a/tests/unit/jobserver/views/test_workspaces.py
+++ b/tests/unit/jobserver/views/test_workspaces.py
@@ -488,6 +488,42 @@ def test_workspacedetail_authorized_private_repo_show_change_visibility_banner(
     assert response.context_data["show_publish_repo_warning"]
 
 
+def test_workspacedetail_authorized_private_repo_show_workspace_admin_panel(
+    rf, project_membership, role_factory
+):
+    project = ProjectFactory()
+
+    # a workspace with a "private" repo
+    workspace = WorkspaceFactory(
+        project=project, repo=RepoFactory(url="http://example.com/repo/private1")
+    )
+
+    user = UserFactory()
+    BackendMembershipFactory(user=user)
+    project_membership(
+        project=project,
+        user=user,
+        roles=[role_factory(permission=permissions.workspace_archive)],
+    )
+
+    # this is what defines "private"
+    class AnotherFakeGitHubAPI:
+        def get_repo_is_private(self, owner, name):
+            return name.startswith("private")
+
+    request = rf.get("/")
+    request.user = user
+
+    response = WorkspaceDetail.as_view(get_github_api=AnotherFakeGitHubAPI)(
+        request,
+        project_slug=project.slug,
+        workspace_slug=workspace.name,
+    )
+
+    assert response.status_code == 200
+    assert response.context_data["show_admin"]
+
+
 def test_workspacedetail_authorized_toggle_notifications(rf, role_factory):
     user = UserFactory(
         roles=[role_factory(permission=permissions.workspace_toggle_notifications)]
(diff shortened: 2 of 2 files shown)


Jongmassey:
before:

image


after:

image

jobserver/views/workspaces.py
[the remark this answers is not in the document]
Jongmassey: I've added positive and negative tests for this panel's presence

jobserver/views/workspaces.py
```diff
@@ -305,8 +305,8 @@ def get(self, request, *args, **kwargs):
         has_backends = request.user.is_authenticated and request.user.backends.exists()
 
         # should we show the admin section in the UI?
-        show_admin = (
-            can_archive_workspace or repo_is_private or can_toggle_notifications
+        show_admin = repo_is_private and (
```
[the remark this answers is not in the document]
Jongmassey: Ah, this is interesting. The admin panel has three buttons so the check should be for the permissions to do the things the buttons do, and the workspace/repository state to be valid for those buttons (private/public; archived or not; notifications settings etc.)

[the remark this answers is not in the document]
Jongmassey: So we need a slightly different boolean combination in the view

tests/unit/jobserver/views/test_workspaces.py
```diff
@@ -488,6 +484,37 @@ def get_repo_is_private(self, owner, name):
...
+    workspace = WorkspaceFactory(
+        project=project, repo=RepoFactory(url="http://example.com/repo/private1")
+    )
+
+    user = UserFactory()
+    BackendMembershipFactory(user=user)
+    project_membership(
+        project=project,
+        user=user,
+        roles=[role_factory(permission=permissions.workspace_archive)],
```
StevenMaude, line 502: nit: Another possible condition that will result in `show_admin` is `permissions.workspace_toggle_notifications` being enabled, which might also be worth testing.

tests/unit/jobserver/views/test_workspaces.py
```diff
@@ -48,6 +48,12 @@
 from ....utils import minutes_ago
 
 
+# this is what defines "private"
+class AnotherFakeGitHubAPI:
+    def get_repo_is_private(self, owner, name):
+        return name.startswith("private")
+
+
```
StevenMaude, line 56: :+1:

jobserver/views/workspaces.py
```diff
@@ -305,9 +305,7 @@ def get(self, request, *args, **kwargs):
         has_backends = request.user.is_authenticated and request.user.backends.exists()
 
         # should we show the admin section in the UI?
-        show_admin = (
-            can_archive_workspace or repo_is_private or can_toggle_notifications
-        )
+        show_admin = can_archive_workspace or can_toggle_notifications
```
StevenMaude, line 308: thought: I have a very limited understanding of roles and permissions. But, if I look at the current `roles.py`, it looks as if only `ProjectDeveloper` has either of `workspace_archive` or `workspace_toggle_notifications`. No other role does.

It does seem a little messy that we're checking for either of these two permissions instead of a single role. But I guess that type of duplication could be cleaned up in the forthcoming work on improving roles and permissions.

(It's fine to leave this for now; it's more a general observation on the current state of roles and permissions.)

tests/unit/jobserver/views/test_workspaces.py
```diff
@@ -488,6 +484,37 @@ def get_repo_is_private(self, owner, name):
...
 
+def test_workspacedetail_authorized_private_repo_show_workspace_admin_panel(
+    rf, project_membership, role_factory
+):
+    project = ProjectFactory()
+
+    # a workspace with a "private" repo
+    workspace = WorkspaceFactory(
+        project=project, repo=RepoFactory(url="http://example.com/repo/private1")
+    )
```
StevenMaude, line 495: thought: Without actually trying anything out, I *think* that regardless of the repository being private or public, it shouldn't make any difference to whether `show_admin` is enabled — which is what we're testing. That is, the outcome of the test is now primarily based on the user's permissions.

Where we *would* see a difference between a private and public repository is in the rendered template.

(I'm not too sure how much we favour testing content of rendered templates. Though, we do test a rendered template elsewhere in these tests though.)

StevenMaude (approved):
I think this is fine; I had a couple of other suggestions around the tests to consider.

jobserver/views/workspaces.py
Jongmassey: I had similar thoughts and backed out of any major changes on the basis we're taking a more holistic look at this and it's likely to change more majorly quite soon

PR #4543, opened 2024-08-19
Remove user github access_tokens from db

Jongmassey:
PR #4539 prevented future storage of users's GitHub access tokens when signing in, but left previously-stored tokens in the database.

Due to these tokens short expiry time these residual tokens in the db do no represent a security risk, but may confuse anyone looking at these data in the future.

This migration removes the residual tokens so that the database state is consistent with the current authentication pipeline.


Jongmassey:
This migration depends on the [most recent migration](https://github.com/python-social-auth/social-app-django/blob/master/social_django/migrations/0016_alter_usersocialauth_extra_data.py) from the [Django social auth app](https://pypi.org/project/social-auth-app-django/) such that this app's tables are created during the CI runs and thus the migration in this PR has something to act on.

jobserver/migrations/2025_remove_user_github_tokens.py
```diff
@@ -0,0 +1,20 @@
...
+
+from django.db import migrations
+from social_django.models import UserSocialAuth
+
+
+def remove_github_tokens(apps, schema_editor):
+    for user in UserSocialAuth.objects.all():
+        if "access_token" in user.extra_data:
+            del user.extra_data["access_token"]
+        user.save()
```
iaindillingham, line 11: We'd expect `"access_token" in user.extra_data` to be `True` for all users, given `jobserver/auth_pipeline.py`. Nevertheless, we could [lean on Django](https://docs.djangoproject.com/en/5.1/topics/db/queries/#has-key), which I hope leans on the database:

```suggestion
    for user in UserSocialAuth.objects.filter(extra_data__has_key="access_token"):
        del user.extra_data["access_token"]
        user.save()
```

jobserver/migrations/2025_remove_user_github_tokens.py
iaindillingham, line 1: What does the `2025_` prefix in the filename refer to?

iaindillingham (approved):
I've made a couple of inline comments; I also have an overall comment about consistency.

The migration calls the things we're removing "user_github_tokens", the function calls them "github_tokens", the commit calls them "github access-tokens", and I'm pretty sure I've read "pat" somewhere (although that may be a preceding PR). Each inconsistency introduces a small snag, which made this PR ever-so-slightly more time-consuming to review. Using the same name for the same thing would be really helpful.

I'm going to approve, but please consider my comments.

jobserver/migrations/2025_remove_user_github_tokens.py
Jongmassey: That's what `makemigrations` put on there - is there a way to override this/should I if so?

Jongmassey: 0027 was the last number pre-squashing

Jongmassey: I believe `makemigrations` got confused by the "2024" in the previous migration. I've renamed it to start with 0002 and the migration runs fine locally and in CI

iaindillingham: It's fine to rename unmigrated migrations. (It's not fine to rename migrated migrations!)

PR #4552, opened 2024-08-20
Remove `ensure_admins` command

### LINKED ISSUE #4035: Remove ensure_admins and related code paths
> [!NOTE]
>This should be done before https://github.com/opensafely-core/job-server/issues/4105

In an effort to move away from a boolean staff field, we ended up with the `CoreDeveloper` role as an unintentionally more complicated of `User.is_staff`.  That's a problem for a different time, but how we assign this role is also problematic.

#### Method
Via the dev&prod entrypoint scripts we run the `ensure_admins` management command.  This reads the list of strings defined in `ADMIN_USERS` to get a list of User instances by username (ignoring any that don't exist).  We then drop CoreDeveloper for all users and apply it to the list we built from the env var.

#### Pros
This was built to avoid the chicken and egg problem, both in dev and prod, of a running instance having no staff users, and thus no way in the UI for a user to set that, either for themselves or anyone else. Early on when we weren't planning to add all staff to the staff area this worked fairly well.  Adding users was done sparingly and very infrequently.

#### Cons
Over time we shifted our stance to giving any staff member access to the staff area.  After all, it exists to help staff do their jobs.  We now regularly find ourselves assigning CoreDeveloper to a user in the UI, confirming they can access the staff area, and then getting a message sometime later saying they can no longer access it.  This is because the person making the change has not updated the ADMIN_USERS env var so the change in the UI is effectively revoked. (IIRC I, the author of the design and implementation, am at ~100% failure rate with this for reference!)

This method also means the CoreDeveloper role acts differently to all the other roles, but it's not _clear_ that it's different.

#### The Fix
Remove:
* the `ensure_admins` management command
* the jobserver/authorization/admins.py module of code and its tests
* the call to `ensure_admins` in the entrypoints
* references to it in DEVELOPERS.md
* the reference to it in pyproject.toml
* any other places it's used or referenced in the codebase
* all references to ADMIN_USERS

Once deployed we can then drop ADMIN_USERS from the configuration in prod.

It would be ideal to have @bloodearnest go over what's done for this given he was involved in the feature and has been a part of various discussions to remove it over the years.

StevenMaude:
Fixes #4035.

This removes the `ADMIN_USERS` feature whereby users could be given the `CoreDeveloper` role based on environment configuration.

This will lead to less confusion on who has been assigned roles, and allow us to better audit assignment of roles in future.

This PR also adds a `make_user_core_developer` command for convenience when initially starting a *local* instance.


DEVELOPERS.md
```diff
@@ -244,9 +228,19 @@ In that situation you can follow the steps below to set up your local copy of th
...
+   > python3 manage.py shell_plus
+   ```
+
+   and then running in the `shell_plus` shell:
+
+   ```
+   >>> from jobserver.authorization import CoreDeveloper
+   >>> user = User.objects.get(username="<your_username>")
+   >>> user.roles.append(CoreDeveloper)
+   >>> user.save()
```
StevenMaude, line 242: thought: This could reduce to:

```suggestion
   ```sh
   > python3 manage.py make_user_core_developer <your_username>
```

if we go with a management command.

jobserver/authorization/admins.py
```diff
@@ -0,0 +1,16 @@
...
+from environs import Env
+
+from ..models import User
+from .roles import CoreDeveloper
+
+
+env = Env()
+
+
+@transaction.atomic
```
StevenMaude, line 11: question: I used the previously existing  module as a starting point for this module. 

Can this `transaction.atomic` decorator can be removed, since we now only have a single `.save()`?

iaindillingham (commented):
I'm going to comment, rather than approve, until we've checked off the items on the TODO list.

I'd suggest not adding the management command. It's useful only when working locally with a new DB, rather than from a dump. Your instructions for using the shell are clear, and using the shell is worthwhile.

My sense is that this change doesn't need an ADR. We're making the system less surprising, which is, I hope, evident from the issue and the PR.

StevenMaude:
@iaindillingham: thanks, I'll remove the admin command and leave it in the documentation only.

DEVELOPERS.md
StevenMaude: Leaving this as I removed the command.

pyproject.toml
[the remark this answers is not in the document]
StevenMaude: Done!

StevenMaude:
All that's left now is to remove the `ADMIN_USERS` environment variable from production after this PR is merged and deployed.

pyproject.toml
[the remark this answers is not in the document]
iaindillingham: Thank you 🙏🏻

iaindillingham (approved):

iaindillingham:
I've unset `ADMIN_USERS` on dokku4, @StevenMaude.

PR #4554, opened 2024-08-21
Replace `time-machine` with `pytest-freezer`

### LINKED ISSUE #4533: Replace time-machine with pytest-freezer
See [slack](https://bennettoxford.slack.com/archives/C01SE1WB6AW/p1723473271010089) 🧵 

We moved from pytest-freezegun to time-machine [here](https://github.com/opensafely-core/job-server/commit/fefeddda5858bfc9731a2cb8bfcc4c59ee573c88) when it appeared that the underlying freezegun project was unmaintained. Since then the project has been updated and the unmaintained pytest-freezegun plugin has been replaced by pytest-freezer which is now part of the larger pytest-dev project.

We've moved all our other projects over to using pytest-freezer. So for consistency, we could consider moving Job Server over to use it too as we're the only ones using time-machine.

StevenMaude:
Fixes #4533.


iaindillingham (approved):
It would be useful to know why `tick=False` isn't required by `freezer.move_to`. And if I were being _really_ nitpicky, I'd suggest merging 464985aa256f0b94bab9f474e527b1ef416e38d9 and c687531a11e464dc649fea1243146f354e6d0c41. Would we ever need to revert one and not the other?

StevenMaude:
Compared with `time-machine`, I'd add that `pytest-freezer` does feel like the kind of dependency we possibly want to avoid; it's currently a 25 line of code wrapper around `freezegun`, and that code has not been changed since October 2022.

But since we're using it elsewhere — [relevant GitHub search](https://github.com/search?q=org%3Aopensafely-core+pytest-freezer+path%3A**%2Frequirements.dev.in&type=code) — I think it's more useful to have consistency with our other projects, and debate its utility later.

PR #4555, opened 2024-08-21
Document use of `django-upgrade`

### LINKED ISSUE #4339: Consider removing django-upgrade
[django-upgrade](https://github.com/adamchainz/django-upgrade) is a tool for automatically upgrading Django project code. The tool seems to rewrite the code to match a target Django version.

Although `just django-upgrade` invokes the tool, I'm not aware we've actually used it. We should consider removing it; alternatively, we should consider documenting when we would use it.

StevenMaude:
Fixes #4339.

In #4339, it was proposed to remove this package, but it was decided that it was better to retain `django-upgrade` and provide some documentation for developers.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -20,6 +20,7 @@
 - [Deployment](#deployment)
 - [Testing](#testing)
   - [Slack Testing](#slack-testing)
+- [`django-upgrade`](#django-upgrade)
 - [Components](#components)
 - [Backends](#backends)
 - [Rotating the GitHub token](#rotating-the-github-token)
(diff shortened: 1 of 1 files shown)


DEVELOPERS.md
```diff
@@ -288,6 +289,23 @@ just test-ci
...
 More details on testing can be found in [TESTING.md](TESTING.md).
 
+## `django-upgrade`
+
+[`django-upgrade`](https://github.com/adamchainz/django-upgrade) is used
+to migrate Django code from older versions to the current version in use.
+
+`django-upgrade` is run via `just django-upgrade`.
+
+`django-upgrade` also gets run via `just check`
```
StevenMaude, line 299: question: One thing I noticed here is that `just check` states:

https://github.com/opensafely-core/job-server/blob/6cb9d56d579e04aac01dc53a66867779792c7384/justfile#L166

But `django-upgrade` always rewrites files in-place, making that comment incorrect. There's no `django-upgrade` option to check for changes only, in the same way that `black` and `ruff` have.

Should we move `django-upgrade` to only run with `just fix`? Or update the comment to reflect the behaviour?

iaindillingham: I'd think we should update the comment. Better still, we should do nothing and, in effect, pretend that `django-upgrade` doesn't rewrite files in-place -- the loss of error is worth the gain in clarity, IMHO. We run `just check` in CI, which is our last chance to check that we've upgraded Django correctly. We don't run `just fix` in CI.

iaindillingham (approved):
🚀

DEVELOPERS.md
StevenMaude: I'll leave it as is then.

Issue #4532, opened 2024-08-13
Remove `attrs`


StevenMaude:
Similar to #4491, we could consider removing `attrs`.

It's used in a few places. However, from a quick look, it doesn't seem like we often use much additional functionality over that which Python's dataclasses provide.

(This was prompted by reviewing #4531 and wondering what we used `attrs` for.)

StevenMaude:
This is actually dependent on interactive-templates; that depends on `attrs` too.

If we ever remove the use of interactive-templates, we can completely remove `attrs` as a dependency. In the meantime, we can clean up the uses of it in the job-server code directly.

StevenMaude:
#4456 is the first part of this.

iaindillingham:
As #4556 is merged, and as completely removing `attrs` is wrapped up in removing interactive-templates, I'm going to close this issue.

PR #4556, opened 2024-08-21
Reduce use of `attrs`

StevenMaude:
See #4532.

This goes a little bit towards removing `attrs` as a direct dependency. Apart from the two cases mentioned in the commit message, we're not using it in any exotic way.

It also makes the code clearer as using the standard Python dataclasses. It means there's less question on a code reader's part as to why `attrs` is being used.

### DIFF
--- a/jobserver/auditing/presenters/base.py
+++ b/jobserver/auditing/presenters/base.py
@@ -1,7 +1,7 @@
-from attrs import define
+from dataclasses import dataclass
 
 
-@define
+@dataclass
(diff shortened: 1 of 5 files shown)


Jongmassey (approved):

iaindillingham (approved):
It would be interesting to know why we required `attrs` in the first place, but sadly b4edbad76 isn't forthcoming 😞 Oh well, LGTM 🚀

PR #4559, opened 2024-08-22
Switch from `black` to `ruff format`

### LINKED ISSUE #4542: Switch from `black` to `ruff` for formatting
The RAP team are using `ruff format` for Airlock, and were positive about it.

Using Ruff is faster, removes a few development dependencies, and shouldn't really result in much divergence from Black's formatting anyway. (I ran the job-server code against it, and there were a handful of small, innocuous changes.)

StevenMaude:
Fixes #4542.

`ruff format` is considerably faster than `black`, while maintaining almost entirely the same formatting output. After deleting the cache for both and running against job-server commit 241415f7c70a1b6b8b1327ab047b5f5fd646f3b6:

```
$ time ruff format .
warning: The following rules may cause conflicts when used with the formatter: `ISC001`. To avoid unexpected behavior, we recommend disabling these rules, either by removing them from the `select` or `extend-select` configuration, or adding them to the `ignore` configuration.
9 files reformatted, 402 files left unchanged

real	0m0.073s
…
$ time black .
All done! ✨ 🍰 ✨
411 files left unchanged.

real	0m1.345s
…
```

(Note: the Ruff warning is fixed in this PR; it's because the configuration's not updated on the `main` branch where I ran the above.)

As a bonus, this also removes several additional development dependencies.

### DIFF
--- a/justfile
+++ b/justfile
@@ -151,20 +151,20 @@ test *args: assets
     $BIN/pytest -n auto -m "not verification and not slow_test" {{ args }}
 
 
-black *args=".": devenv
-    $BIN/black --check {{ args }}
+format *args=".": devenv
+    $BIN/ruff format --check {{ args }}
 
 
 django-upgrade *args="$(find applications interactive jobserver redirects services staff tests -name '*.py' -type f)": devenv
     $BIN/django-upgrade --target-version=5.0 {{ args }}
 
 
-ruff *args=".": devenv
+lint *args=".": devenv
     $BIN/ruff check --output-format=full {{ args }}
 
 
 # run the various dev checks but does not change any files
-check: black django-upgrade ruff
+check: format django-upgrade lint
 
 
 check-migrations: devenv
@@ -174,8 +174,8 @@ check-migrations: devenv
 
 # fix formatting and import sort ordering
 fix: devenv
-    $BIN/black .
     $BIN/ruff check --fix .
+    $BIN/ruff format .
 
 
 load-dev-data: devenv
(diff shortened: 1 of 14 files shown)


Jongmassey (approved):

justfile
```diff
@@ -174,8 +174,8 @@ check-migrations: devenv
 
 # fix formatting and import sort ordering
 fix: devenv
-    $BIN/black .
     $BIN/ruff check --fix .
+    $BIN/ruff format .
```
iaindillingham, line 178: What does `check --fix` do that `format` doesn't do?

pyproject.toml
```diff
@@ -121,6 +109,7 @@ exclude = [
   "docs",
   "htmlcov",
   "node_modules",
+  "outputs",
```
iaindillingham, line 112: And `static`?

iaindillingham (approved):
Couple of questions, but otherwise 🚀

justfile
StevenMaude: `ruff check` runs the linter and fixes any automatically fixable issues it finds (example: removing unused imports), but `ruff check` doesn't run the formatter.

It's planned for `ruff` to [eventually unify the two](https://github.com/astral-sh/ruff/issues/8232).

pyproject.toml
StevenMaude: Unless I've missed something, `static` is already included below.

`outputs` was the only entry for the Black excludes, that wasn't listed already in Ruff's excludes.

justfile
iaindillingham: I need to read up on the difference between linting and formatting!

PR #4561, opened 2024-08-22
Enable migrations when running tests in CI

### LINKED ISSUE #4558: Testing migrations
At present, we disable migrations when running the tests with pytest-django's [`--no-migrations`](https://pytest-django.readthedocs.io/en/latest/database.html#no-migrations-disable-django-migrations) flag. This means that the test database is created by inspecting the models, not by running `migrate`:

https://github.com/opensafely-core/job-server/blob/6cb9d56d579e04aac01dc53a66867779792c7384/pyproject.toml#L78

As a consequence, if we add a new migration, and don't test it by running `migrate`, then we won't know whether the new migration runs without error until it is deployed. (Thankfully, [migrations run inside transactions on PostgreSQL](https://docs.djangoproject.com/en/5.1/howto/writing-migrations/#non-atomic-migrations), so if a new migration fails, then the database won't be left in an inconsistent state.)

The pytest-django documentation suggests that disabling migrations might improve performance when there are "several" migrations. We recently squashed migrations, so this is not the case for job-server. Indeed, after running `just test` three times with and without the flag, `time` doesn't suggest any difference in performance.

## Considerations

At this stage, the solution-minded reader may ask, "Why not just remove the `--no-migrations` flag?" The answer: because if a new migration fails, then every test fails because every test has database access.

https://github.com/opensafely-core/job-server/blob/6cb9d56d579e04aac01dc53a66867779792c7384/tests/conftest.py#L57-L65

Furthermore, if a new migration fails, then the test database isn't dropped. The message "Got an error creating the test database: database "test_jobserver" already exists" is printed to stderr. If you miss the message, then you may wonder why the data you thought you added to the test database isn't the data that's in the test database.

Some considerations...

* Reverse the `--no-migrations` flag in CI by passing the `--migrations` flag. We're less concerned about the test database in CI, as it's used once. We're also less concerned about every test failing in CI, as they're not used for quick feedback. Indeed, we could pass the `--exitfirst` flag. However, we may find ourselves getting frustrated when the failures aren't associated with a new migration. (I'm looking at you, `coverage`.)

* Write a test that doesn't disable migrations, runs `migrate`, and catches the error. This could mean overriding pytest-django's [`django_db_setup`](https://pytest-django.readthedocs.io/en/latest/database.html#django-db-setup) fixture.

* We could investigate something like [django-test-migrations](https://github.com/wemake-services/django-test-migrations), but we don't need to test the logic of our migrations; just that our migrations run without error.

iaindillingham:
This PR attempted to address #4558, but was superseded by #4573. However, its actions demonstrate an interesting (and infuriating!) quirk of actions/checkout, so I'm going to keep it open and in draft until I've created an issue.


iaindillingham:
I talked about this during a recent tech team meeting. We don't need the PR (or the branch) any more.

PR #4573, opened 2024-08-27
Enable migrations everywhere

iaindillingham:
For the rationale, see `docs/adr/0023-enable-migrations-everywhere.md`. And for background, see #4561.


Jongmassey (approved):

Jongmassey:
>We will commit to running python manage.py migrate against a recent database dump when creating, and reviewing the creation of, a new migration. Doing so will follow many new code paths, and it will alert us to the less obvious errors, sooner.

I wonder if there is a way we could facilitate this by, for example, extending the `check-migrations` `just` command to not only check that there are any model changes without migrations but also that all migrations can run successfully? Obviously there are still the caveats around data migrations but I feel like an update to `check-migrations` to make it check migrations could help prevent future snafus

iaindillingham:
> I wonder if there is a way we could facilitate this by, for example, extending the `check-migrations` `just` command to not only check that there are any model changes without migrations but also that all migrations can run successfully? Obviously there are still the caveats around data migrations but I feel like an update to `check-migrations` to make it check migrations could help prevent future snafus

Thanks for your comment, @Jongmassey. `just check-migrations` runs `makemigrations --check`, which, as you say, exits non-zero if model changes are missing migrations. It doesn't load and run migrations, so wouldn't encounter any errors. There's also `migrate --check`, which exits non-zero if unapplied migrations exist. Similarly, it doesn't load and run migrations, so wouldn't encounter any errors.

Without third-party tooling, the only way I can think of to load and run migrations would be to create a test database and call `migrate`, but that's what the `django_db_setup` fixture does already. We _could_ populate the test database with [Django fixtures](https://docs.djangoproject.com/en/5.1/topics/db/fixtures/) (which are different to pytest fixtures!) but in my experience, they become hard to maintain. However, if they contained at least one instance of each model, then we would be alerted to the less obvious errors, sooner.

Creating a test database, populating it with Django fixtures, and calling `migrate` is similar to using django-test-migrations, albeit much less targeted.

Thanks again for your comment! It encouraged me to reflect more on the problem. I'll update the ADR to include the information in this comment.

PR #4574, opened 2024-08-28
Change repo-project from one-one to one-many

iaindillingham:
For much, but not all, of job-server's life, one repo could be associated with one project. However, there are now a small number of repos that should be associated with two projects: these repos have "legacy" and "transitional" projects. To accommodate these repos, we change the repo-project relationship from one-to-one to one-to-many.

Are you expecting to see changes to a model? And not expecting to see changes to a form? The repo-project relationship was actually one-to-many early in job-server's life, so the database already contains several repos that are associated with two projects. For this reason, the one-to-one repo-project relationship was enforced within a form and not within a model.

Are you concerned that we may no longer be supporting an important use-case? The original use-case for changing from repo-project relationship from one-to-many to one-one was (we think) to help a member of the public navigate from GitHub's website to job-server's website ("When was this code run?"). In practice, however, supporting this use-case isn't easy. We feel an alternative design is required.

### DIFF
--- a/jobserver/forms.py
+++ b/jobserver/forms.py
@@ -1,7 +1,7 @@
 from django import forms
 
 from .authorization.forms import RolesForm
-from .models import JobRequest, Project, Workspace
+from .models import JobRequest, Workspace
 
 
 class JobRequestCreateForm(forms.ModelForm):
@@ -135,7 +135,7 @@ class WorkspaceCreateForm(forms.Form):
         repo_url = cleaned_data.get("repo")
         branch = cleaned_data.get("branch")
 
-        if not (repo_url and branch):
+        if not (repo_url and branch):  # pragma: no cover
             return
 
         repo = next(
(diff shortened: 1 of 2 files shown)


iaindillingham:
Currently, seven repos are associated with more than one project (six with two projects; one with three projects).

@CLStables asked whether URLs from these repos' `README.md` files would point to meaningful pages on job-server. Alas, none of the seven repos has a suitable URL in their `README.md` file. However, if they had, then it would point to a list of projects with which the repo is associated. This list isn't explicitly sorted, but is implicitly sorted in ascending order of project creation date.

For example: <https://jobs.opensafely.org/repo/https%253A%252F%252Fgithub.com%252Fopensafely%252Fos-demo-research/>.

CLStables:
it would point to a list of projects with which the repo is associated

jobserver/forms.py
```diff
@@ -135,7 +135,7 @@ def clean(self):
         repo_url = cleaned_data.get("repo")
         branch = cleaned_data.get("branch")
 
-        if not (repo_url and branch):
+        if not (repo_url and branch):  # pragma: no cover
```
inglesp, line 138: I'd prefer to see this tested rather than excluded from coverage.

inglesp (approved):

jobserver/forms.py
iaindillingham: Thank you for your comment. I would too. I started writing a test, but I soon realised that I needed to learn more about forms in general as well as forms in job-server, and this issue is (to some degree) time-sensitive. I've made a note to read up, and to return to this line.

iaindillingham: See #4593. I was hoping to JDI, but haven't managed to. I created an issue before it dropped off my TODO list.

PR #4582, opened 2024-09-02
Update description of deployment administrator

iaindillingham:
The deployment administrator role is intended for general technical support, rather than for specific tech support (with a hyphen). To reduce confusion, we mirror the description in the Developer Permissions Log.

Thanks for clarifying, @lucyb and @CLStables.


iaindillingham:
I'm backing out of this PR.

PR #4591, opened 2024-09-06
Project page doesnt load when GitHub api is unreachable

### LINKED ISSUE #4534: Project page doesn't load when GitHub API is unreachable
We received this error on the project detail pages when the GitHub API was temporarily unavailable recently (see [slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1723554717290339) 🧵 ).
https://ebm-datalab.sentry.io/issues/5708489567/?alert_rule_id=10884789&alert_type=issue&notification_uuid=9b0e6bca-d903-4eaf-a0c7-1be130894177&project=5443358&referrer=slack 

This resulted in a 500 error back to users when they tried to view any project detail page (Dave tested it on the [internal project](https://jobs.opensafely.org/opensafely-internal/)). 

We can expect to receive these errors from time-to-time as GitHub's API may not always be available and should handle them more gracefully. 

Note: we received a [similar error](https://ebm-datalab.sentry.io/issues/5708512496/?alert_rule_id=10884789&alert_type=issue&notification_uuid=ecef0488-2232-4b76-a3a7-707898d01ab7&project=5443358&referrer=slack) from another endpoint at the same time

Jongmassey:
Fixes #4534 

I don't believe retrying in case of a `TimeoutError` here is necessary for this feature, but am happy to be convinced otherwise. For now, this just handles it and provides a "Github API unavailable" placeholder `Repo` for the relevant bit of the template

### DIFF
--- a/jobserver/views/projects.py
+++ b/jobserver/views/projects.py
@@ -173,7 +173,7 @@ class ProjectDetail(View):
                     is_private = self.get_github_api().get_repo_is_private(
                         repo.owner, repo.name
                     )
-                except requests.HTTPError:
+                except (requests.HTTPError, requests.Timeout, requests.ConnectionError):
                     is_private = None
                 span = trace.get_current_span()
                 span.set_attribute("repo_owner", repo.owner)
@@ -187,9 +187,12 @@ class ProjectDetail(View):
                 }
 
         # use the threadpool to parallelise the repo requests
-        yield from repo_thread_pool.map(
-            get_repo, repos, itertools.repeat(ctx), timeout=30
-        )
+        try:
+            yield from repo_thread_pool.map(
+                get_repo, repos, itertools.repeat(ctx), timeout=30
+            )
+        except TimeoutError:
+            yield {"name": "GitHub API Unavailable", "is_private": None, "url": ""}
 
 
 class ProjectEdit(UpdateView):
--- a/tests/unit/jobserver/views/test_projects.py
+++ b/tests/unit/jobserver/views/test_projects.py
@@ -1,5 +1,3 @@
-import time
-
 import pytest
 import requests
 from django.contrib.auth.models import AnonymousUser
@@ -214,13 +212,7 @@ def test_projectdetail_with_timed_out_github(rf):
 
     class BrokenGitHubAPI:
         def get_repo_is_private(self, *args):
-            response = requests.Response()
-            response.status_code = 504
-            time.sleep(31)
-            raise requests.HTTPError(
-                "504 Server Error: Gateway Timeout for url: https://api.github.com/user",
-                response=response,
-            )
+            raise TimeoutError
 
     response = ProjectDetail.as_view(get_github_api=BrokenGitHubAPI)(
         request, project_slug=project.slug
--- a/tests/unit/jobserver/views/test_projects.py
+++ b/tests/unit/jobserver/views/test_projects.py
@@ -1,3 +1,5 @@
+import time
+
 import pytest
 import requests
 from django.contrib.auth.models import AnonymousUser
@@ -186,6 +188,38 @@ def test_projectdetail_with_no_github(rf):
     class BrokenGitHubAPI:
         def get_repo_is_private(self, *args):
             raise requests.HTTPError
+
+    response = ProjectDetail.as_view(get_github_api=BrokenGitHubAPI)(
+        request, project_slug=project.slug
+    )
+
+    assert response.status_code == 200
+
+    # check there is no public/private badge when GitHub returns an
+    # unsuccessful response
+    assert not response.context_data["public_repos"]
+    assert response.context_data["private_repos"][0]["is_private"] is None
+    assert "Public" not in response.rendered_content
+
+
+def test_projectdetail_with_timed_out_github(rf):
+    project = ProjectFactory(org=OrgFactory())
+    WorkspaceFactory(
+        project=project, repo=RepoFactory(url="https://github.com/owner/repo")
+    )
+
+    request = rf.get("/")
+    request.user = UserFactory()
+
+    class BrokenGitHubAPI:
+        def get_repo_is_private(self, *args):
+            response = requests.Response()
+            response.status_code = 504
+            time.sleep(31)
+            raise requests.HTTPError(
+                "504 Server Error: Gateway Timeout for url: https://api.github.com/user",
+                response=response,
+            )
 
     response = ProjectDetail.as_view(get_github_api=BrokenGitHubAPI)(
         request, project_slug=project.slug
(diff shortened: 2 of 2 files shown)


tests/unit/jobserver/views/test_projects.py
```diff
@@ -199,6 +199,33 @@ def get_repo_is_private(self, *args):
     assert "Public" not in response.rendered_content
 
 
+def test_projectdetail_with_timed_out_github(rf):
+    project = ProjectFactory()
+    WorkspaceFactory(
+        project=project, repo=RepoFactory(url="https://github.com/owner/repo")
+    )
+    WorkspaceFactory(project=project, repo=RepoFactory(url="/path/on/disk/to/repo"))
```
lucyb, line 207: Minor nit, but I'm not sure that we need to create two workspaces here. I _think_ we're exercising the same code if we just have one, but I could be wrong.

lucyb (approved):
Looks good. Sorry this didn't get a review for ages.

tests/unit/jobserver/views/test_projects.py
Jongmassey: Test seems fine without the second one, I'll remove it then merge.

tests/unit/jobserver/views/test_projects.py
```diff
@@ -200,6 +200,32 @@ def get_repo_is_private(self, *args):
...
+        request, project_slug=project.slug
+    )
+
+    assert response.status_code == 200
+
+    # check there is no public/private badge when GitHub returns an
+    # unsuccessful response
+    assert not response.context_data["public_repos"]
+    assert response.context_data["private_repos"][0]["is_private"] is None
+    assert "Public" not in response.rendered_content
```
iaindillingham, line 226: `response.rendered_context` will call `response.render()`, which will render `templates/project/detail.html`. Alas, this template has a missing template variable. I'm going to guess that it's `Project.org`, so try the following...

```py
project = ProjectFactory(org=OrgFactory())
```

jobserver/views/projects.py
```diff
@@ -173,7 +173,7 @@ def get_repo(repo, ctx):
                     is_private = self.get_github_api().get_repo_is_private(
                         repo.owner, repo.name
                     )
-                except requests.HTTPError:
+                except (requests.HTTPError, TimeoutError):
```
iaindillingham, line 176: I'm curious as to why `TimeoutError` and not `requests.Timeout`.

tests/unit/jobserver/views/test_projects.py
Jongmassey: This only started failing once I rebased, I wonder what changed such that the tests passed previously. Either way, thanks for the hint - I'll give it a go!

iaindillingham: I reckon you branched and rebased either side of #4505.

jobserver/views/projects.py
Jongmassey: I wasn't aware of the latter, and typed the former out of habit! I think it'll be better with `requests.Timeout` having read the docs, thanks!

mikerkelly: I don't think this has been resolved in the latest commits, unless I'm missing something.

I think it's important to use requests.Timeout here not TimeoutError. I don't yet have access to sentry so I can't check the report -- but I assume it was requests.TimeoutError that would be raised if the GH API did indeed time-out. So I don't think the issue would be resolved without this change.

The [Python docs](https://docs.python.org/3/library/exceptions.html) say the builtin TimeoutError is something different, nothing to do with HTTP in general, I think:

> Raised when a system function timed out at the system level. Corresponds to errno [ETIMEDOUT](https://docs.python.org/3/library/errno.html#errno.ETIMEDOUT).

[requests.Timeout](https://github.com/psf/requests/blob/1ae6fc3137a11e11565ed22436aa1e77277ac98c/src/requests/exceptions.py#L71) does not sub-class TimeoutError.

jobserver/views/projects.py
```diff
@@ -187,9 +187,12 @@ def get_repo(repo, ctx):
...
 
         # use the threadpool to parallelise the repo requests
-        yield from repo_thread_pool.map(
-            get_repo, repos, itertools.repeat(ctx), timeout=30
-        )
+        try:
+            yield from repo_thread_pool.map(
+                get_repo, repos, itertools.repeat(ctx), timeout=30
+            )
+        except TimeoutError:
```
mikerkelly, line 194: Would this be cleaner and easier to follow if this `except` clause were part of the `try` block within `get_repo`? This localizes the try-except to the code it is protecting as much as possible. The `requests.HTTPError` and the timeout exception are similar in this context: different flavours of "the required resource was not available".

```
try:
	is_private = self.get_github_api().get_repo_is_private(
		repo.owner, repo.name
	)
except requests.HTTPError:
	is_private = None
except requests.Timeout:
	name = "GitHub API Unavailable"
	is_private = None
	url = ""
```
It's also not clear to me why the `HTTPError` block just sets `is_private = None`, whereas the new `TimeoutError` block changes the name and the URL. Could it just be one block instead as in your original commits? There may well be a good reason for that -- if so can it please be documented as it isn't obvious (to me)?

mikerkelly (dismissed):
Thanks, Jon.
* Test looks better as discussed in Slack.
* Re-opened one comment above from Iain about the exception used -- I think this has to be resolved before this can be merged to fulfil the requirements.
* One other comment about the structure and behaviour of the `Except` blocks -- possibly this is fine as is, unsure.

mikerkelly (commented):
I definitely violated the etiquette around PR workflow by clicking "Request changes". I do think the exception type issue needs resolution, but shouldn't have clicked the Big Red Button. Trying to rectify that!

jobserver/views/projects.py
Jongmassey: The reason is a `TimeoutError` and not one of the exceptions from the `requests` namespace is that it's not requests nor even the GitHub API that's giving the timeout. The code before this change would have gracefully handled the 504 error (gateway timeout) previously seen as it's an `HTTPError`. The problem is/was that it can take longer than the `ThreadPoolExecutor.map()` call's configured timeout for these to appear and so the ThreadPoolExecutor throws a 'TimeoutError'. Does that make sense?

Jongmassey: It's not possible to localise the handling of this timeout within the get repo function that's passed to map()

jobserver/views/projects.py
Jongmassey: >but I assume it was requests.TimeoutError

`requests` doesn't have `TimeoutError`, only `requests.Timeout`. The sentry issue is definitely the former

jobserver/views/projects.py
Jongmassey: >I think it's quick and useful to add the additional exceptions to the inner except: block. (requests.ConnectionError and requests.Timeout).

I agree, I'll add that to this PR. As for the other two I think raising tickets is a great idea

PR #4603, opened 2024-09-11
Remove most uses of `pragma: no branch`

### LINKED ISSUE #4562: Remove use of `pragma: no branch`
We should remove the `# pragma: no branch` introduced in #4503.

> I'm a bit concerned that we have `# pragma: no branch` in production code, especially when that's code that's just been changed. However, it might well be that it's a pragmatic solution and less risky than removing the `if f` (to use this code snippet as an example). I wouldn't want people to think it was okay these lines weren't tested in future though.
>
> I would recommend either removing the conditional (and manually testing that's okay) or adding in an extra unit test so that coverage doesn't complain. How and when to do that is up to you though.

_Originally posted by @lucyb in https://github.com/opensafely-core/job-server/pull/4503#discussion_r1725054562_

StevenMaude:
Fixes #4562.

This removes most uses of `pragma: no branch` from production code.

We decided to leave one exemption in the code, as it makes sense in context: we never expect that alternate execution branch to happen. See [the discussion](https://github.com/opensafely-core/job-server/pull/4603#discussion_r1757340223).

### DIFF
--- a/jobserver/forms.py
+++ b/jobserver/forms.py
@@ -113,13 +113,18 @@ class WorkspaceCreateForm(forms.Form):
 
         # has there been a repo selected already?
         if "data" in kwargs and "repo" in kwargs["data"]:
-            repo = next(  # pragma: no branch
-                (
-                    r
-                    for r in self.repos_with_branches
-                    if r["url"] == kwargs["data"]["repo"]
-                ),
-            )
+            try:
+                repo = next(
+                    (
+                        r
+                        for r in self.repos_with_branches
+                        if r["url"] == kwargs["data"]["repo"]
+                    ),
+                )
+            except StopIteration:
+                raise forms.ValidationError(
+                    "No matching repos found, please reload the page and try again"
+                )
         else:
             repo = self.repos_with_branches[0]
 
--- a/jobserver/middleware.py
+++ b/jobserver/middleware.py
@@ -38,9 +38,7 @@ class TemplateNameMiddleware:
 
         # flatten template_name to a string
         if isinstance(template_name, list):
-            template_name = next(  # pragma: no branch
-                (t for t in template_name if t), None
-            )
+            template_name = template_name[0]
 
         response.context_data["template_name"] = template_name
         return response
--- a/tests/unit/jobserver/test_forms.py
+++ b/tests/unit/jobserver/test_forms.py
@@ -200,3 +200,18 @@ def test_workspacecreateform_with_duplicate_name():
             'A workspace with the name "test" already exists, please choose a unique one.'
         ]
     }
+
+
+def test_workspacecreateform_with_no_repo_match():
+    project = ProjectFactory()
+    WorkspaceFactory(name="test")
+
+    data = {
+        "name": "test",
+        "repo": "different",
+        "branch": "test",
+        "purpose": "test",
+    }
+    repos_with_branches = [{"name": "test", "url": "test", "branches": ["test"]}]
+    with pytest.raises(ValidationError):
+        WorkspaceCreateForm(project, repos_with_branches, data=data)


tests/unit/jobserver/test_middleware.py
```diff
@@ -137,6 +137,21 @@ class DummyView(DetailView):
     assert response.context_data["template_name"] == "my_template"
 
 
+def test_template_name_middleware_with_template_name_as_a_list_without_a_name(rf):
```
StevenMaude, line 140: thought: This is a really clunky/confusing name, but it's in line with the test above.

jobserver/forms.py
```diff
@@ -113,13 +113,18 @@ def __init__(self, project, repos_with_branches, *args, **kwargs):
...
+                    (
+                        r
+                        for r in self.repos_with_branches
+                        if r["url"] == kwargs["data"]["repo"]
+                    ),
+                )
+            except StopIteration:
+                raise forms.ValidationError(
+                    "Unknown repo, please reload the page and try again"
+                )
```
StevenMaude, line 127: thought: This comes from `get_github_api().get_repos_with_branches(gh_org)` in `WorkspaceCreate`. That in turn pulls repositories from the OpenSAFELY researchers team.

We *always* expect that will contain repositories with branches.

I guess it's _possible_ that GitHub's API is working and gives a 200 response, but is broken and gives no repositories.

Also, I copied the message from elsewhere in the code, but maybe the error should be:

> No matching repos found, please reload the page and try again

StevenMaude: Also, I copied the message from elsewhere in the code, but maybe the error should be:

> No matching repos found, please reload the page and try again

tests/unit/redirects/test_models.py
```diff
@@ -44,6 +44,16 @@ def test_redirect_all_target_objects():
         RedirectFactory(**kwargs)
 
 
+def test_exception_when_no_field_name_found():
+    class FakeTarget:
+        name = ""
+
+    redirect = RedirectFactory(project=ProjectFactory(), old_url="/test/")
+    redirect.targets = lambda: [FakeTarget(), FakeTarget()]
```
StevenMaude, line 52: thought: The fact we have to resort to monkeypatching the object to test this is a smell.

Unless someone knows a way to create a valid `Redirect` which doesn't have a matching field name (which seems unlikely because these are baked into the class).

tests/unit/jobserver/test_middleware.py
```diff
@@ -137,6 +137,21 @@ class DummyView(DetailView):
...
+    class DummyView(DetailView):
+        model = Project
+        template_name = ""
+
+    request = rf.get("/")
+    response = DummyView.as_view()(request, pk=project.pk)
+
+    TemplateNameMiddleware(None).process_template_response(request, response)
+
+    assert response.context_data["template_name"] is None
```
StevenMaude, line 152: thought: This seems OK to capture the existing behaviour, but I'm not sure if this state can be reached, and if so, does it make sense to allow it.

That is: I don't know enough about how this `template_name` is created and how it might interact elsewhere. Would it be better to raise an exception here?

StevenMaude:
## What this PR is for

The goal of this PR to fix #4562, removing the `coverage` `no cover` in production code for the particular execution branches in question.

This PR does fix that issue.

However, I am not entirely convinced that this is all a good idea. Feedback on what we should do here is definitely welcome.

## More context

These exposed execution branches follow from the removal of `first()` in #4514 and #4503.

That change exposed that certain execution branches weren't covered. I think `coverage` wasn't able to determine these, but could once we swapped the external library code `first()` for Python's `next()` built-in.

These three cases all have the same pattern:

```
next((x for x in y if z))
```

There's something being iterated over with `next()`, looking for some kind of match.

We've not currently been testing the cases where the iterator is exhausted.

Note that:

* `next()`'s behaviour is that if there's no default value supplied — as in the `redirects/models.py` case here — and no matching value, a `StopIteration` exception is raised
* coverage doesn't complain about either side of the `if` branch not being reached; only about the case where we exhaust the iterator [^1]

## Where's the confusion?

For me, I don't really have enough understanding of what we can expect about the data that's being used as an input in each case.

* Are we adding convoluted tests to placate coverage?
  * The test of `Redirects` is particularly convoluted. The `CheckConstraints` on the `Redirects` class make it difficult to create a valid `Redirect` that falls into the non-matching branch we want to test. In the end, I had to fake out data to make this work.
* Is adding these tests OK to merely capture the current behaviour?
* Would we be better with something along the lines of:
  ```python
  try:
      next(x for x in y if z)
  except StopIteration:  # pragma: no branch
      assert False
```

where we never expect the failure case.

I've added specific comments for each part of the code changed.

[^1]: See https://github.com/nedbat/coveragepy/issues/475 for more on this.

tests/unit/redirects/test_models.py
lucyb: I agree, this seems like an unhelpful test. I actually think the `# pragma: no branch` is maybe more expressive in this instance (we're explicitly saying that we haven't anticipated this happening).

jobserver/forms.py
StevenMaude: I updated the error message.

tests/unit/redirects/test_models.py
StevenMaude: I removed this test and restored the `pragma` here.

tests/unit/jobserver/test_middleware.py
StevenMaude: @lucyb: Thanks for investigating! I wasn't entirely sure where to start with it.

If I have understood you correctly, I tried switching the test to use a ListView, with `get_queryset()` returning an empty list of objects.

This still resulted in a non-zero length list as `response.template_name`.

So I think your assertion here is correct, that a list is never empty. I'll amend the code accordingly.

tests/unit/jobserver/test_middleware.py
StevenMaude: This has gone now.

lucyb (approved):

PR #4606, opened 2024-09-13
Remove all information related to GitHub access tokens

Jongmassey:


jobserver/migrations/0005_remove_user_socialauth_expires.py
```diff
@@ -0,0 +1,20 @@
...
+
+from django.db import migrations
+from social_django.models import UserSocialAuth
+
+
+def remove_user_socialauth_expires(apps, schema_editor):
+    users = UserSocialAuth.objects.filter(extra_data__has_key="expires")
+    for user in users:
+        del user.extra_data["expires"]
+    UserSocialAuth.objects.bulk_update(users, ["extra_data"])
```
iaindillingham, line 11: Nice to see the `.bulk_update` method, here 👍🏻

jobserver/migrations/0005_remove_user_socialauth_expires.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("jobserver", "0004_remove_orgmembership_roles"),
+        ("social_django", "0016_alter_usersocialauth_extra_data"),
+    ]
+
+    operations = [migrations.RunPython(remove_user_socialauth_expires)]
```
iaindillingham, line 20: In the future, consider passing `RunPython(..., reverse_code=RunPython.noop())` to make the operation reversible. In this case, although we don't need to recreate the data, we may need to reverse the migration. I didn't know about `RunPython.noop()` but IIRC, @lucyb mentioned it in a previous PR.

tests/unit/jobserver/test_auth_pipeline.py
```diff
@@ -84,7 +84,11 @@ def test_pipeline_with_new_user(slack_messages, strategy):
 
     social = output["social"]
     assert social.user == user
+    assert social.extra_data["id"] == "1234"
+    assert social.extra_data["login"] == "dummy-user"
+    assert "auth_time" in social.extra_data
     assert "sekret" not in social.extra_data.values()
+    assert "token_type" not in social.extra_data
```
iaindillingham, line 91: Nitpick: I don't think we need to test for the absence of this value or this key. Neither are added to `social.extra_data` by `pipeline`; and `pipeline` doesn't simply pass through everything it receives from GitHub's response.

iaindillingham (approved):
Thanks for picking this up, @Jongmassey. I've made a couple of suggestions, but they're for next time.

jobserver/migrations/0005_remove_user_socialauth_expires.py
Jongmassey: Thanks, I've added that to this PR as it was a simple addition

PR #4615, opened 2024-09-19
Add explicit dependencies on building to Docker commands

benbc:


Jongmassey (approved):
Seems like a sensible change

PR #4633, opened 2024-09-30
Fix circular redirects caused by wildcards in paths

### LINKED ISSUE #4545: Pages with Redirect Error
Google Search console has highlighted the below pages as having redirect errors:

- https://jobs.opensafely.org/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/1688/v23wauxcsoibbf6c/
- https://jobs.opensafely.org/datalab/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/1718/
- https://jobs.opensafely.org/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/1688/oe2a5smsfcrvh6aa/
- https://jobs.opensafely.org/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/1603/
- https://jobs.opensafely.org/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/
- https://jobs.opensafely.org/datalab/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/1688/oe2a5smsfcrvh6aa/
- https://jobs.opensafely.org/datalab/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/1603/
- https://jobs.opensafely.org/datalab/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/1688/v23wauxcsoibbf6c/
- https://jobs.opensafely.org/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/1603/sqcfuzzbrcnhfkuu/
- https://jobs.opensafely.org/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/outputs/
- https://jobs.opensafely.org/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/1603/aemwzevstck23ipl/
- https://jobs.opensafely.org/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/1623/qzwhzgerbjfaoql4/

---

## Example request

```
❯ curl -I https://jobs.opensafely.org/factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/1688/v23wauxcsoibbf6c/
HTTP/2 301
server: nginx
date: Mon, 19 Aug 2024 09:53:16 GMT
content-type: text/html; charset=utf-8
location: /factors-associated-with-covid-19-vaccination/factors-associated-with-covid-19-vaccination-wip-branch/1688/v23wauxcsoibbf6c/
strict-transport-security: max-age=15724800; includeSubdomains
```

---

From Google docs ([source](https://support.google.com/webmasters/answer/7440203#multiple_redirects))

> ## Redirect error
> Google experienced one of the following redirect errors:
>
> - A redirect chain that was too long
> - A redirect loop
> - A redirect URL that eventually exceeded the max URL length
> - A bad or empty URL in the redirect chain

Jongmassey:
fixes #4545

### DIFF
--- a/redirects/middleware.py
+++ b/redirects/middleware.py
@@ -1,3 +1,4 @@
+from django.db.models import CharField, F, Value
 from django.db.models.functions import Length
 from django.shortcuts import redirect
 
@@ -27,11 +28,11 @@ class RedirectsMiddleware:
 
         # look for a direct or prefix match on the current URL
         redirection = (
-            Redirect.objects.extra(
-                where=["%s LIKE concat(old_url, '%%')"],
-                params=[request.path],
+            Redirect.objects.annotate(
+                url_length=Length("old_url"),
+                request_url=Value(request.path, output_field=CharField()),
             )
-            .annotate(url_length=Length("old_url"))
+            .filter(request_url__startswith=F("old_url"))
             .order_by("-url_length")
             .first()
         )
--- a/tests/unit/redirects/test_middleware.py
+++ b/tests/unit/redirects/test_middleware.py
@@ -39,3 +39,16 @@ def test_redirectsmiddleware_unknown_url(rf):
     response = RedirectsMiddleware(get_response)(request)
 
     assert response == "no match"
+
+
+def test_redirect_with_wildcard_in_path(rf):
+    w1 = WorkspaceFactory()
+    w2 = WorkspaceFactory()
+    RedirectFactory(old_url="/abc/123/test_workspace/", workspace=w1)
+    RedirectFactory(old_url="/abc/123/test-workspace/", workspace=w2)
+
+    request = rf.get("/abc/123/test-workspace/")
+
+    response = RedirectsMiddleware(get_response)(request)
+
+    assert response.url == w2.get_absolute_url()


redirects/middleware.py
```diff
@@ -28,7 +28,7 @@ def __call__(self, request):
         # look for a direct or prefix match on the current URL
         redirection = (
             Redirect.objects.extra(
-                where=["%s LIKE concat(old_url, '%%')"],
+                where=[r"%s LIKE concat(replace(old_url,'_','\_'), '%%')"],
```
StevenMaude, line 31: question: I don't know Django or its query API well enough, but might it be possible to rewrite this query, maybe with `filter(startswith=…`?

https://docs.djangoproject.com/en/5.1/ref/models/querysets/#startswith

One advantage of the abstracted lookups is that [they escape `%` and `_` for you](https://docs.djangoproject.com/en/5.1/topics/db/queries/#escaping-percent-signs-and-underscores-in-like-statements).

Django's documentation [actually advocates avoiding `extra`](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#extra), stating:

> This is an old API that we aim to deprecate at some point in the future. Use it only if you cannot express your query using other queryset methods.

Searching the codebase, it's the only use of `extra` we have, so if we can remove it, that's even better.

mikerkelly: We should do as Steve suggests. I think it's as simple as `Redirect.objects.filter(old_url__startswith=request.path)`. Definitely shouldn't require `.extra()`.

evansd: I think it's the reverse of that we need, which is why it isn't completely trivial. Would be do-able with [F-expressions](https://docs.djangoproject.com/en/5.1/ref/models/expressions/#f-expressions) I think, though you'd still need to the do the escaping to avoid the bug this PR is designed to solve.

Jongmassey: Yes, we need to check where the request path starts with `old_url` not the other way around, which is why `.extra()` was required. Sorry if I didn't phrase that clearly enough!

mikerkelly: Ah, I see, my bad! Can I suggest making the comment a bit more explicit about that? I did not read it the right way round.

mikerkelly:
```python
from django.db.models import F, Value, CharField

redirection = Redirect.objects.annotate(
    url_length=Length("old_url"),
    request_url=Value(request.path, output_field=CharField())
    ).filter(request_url__startswith=F('old_url')
    ).order_by("-url_length").first()
```
This passes existing and Jon's new unit test and I think it makes sense -- annotate in to each row the value we want to compare with, then we can use the `__startswith` lookup expression against that value, using an `F()` expression to get the `old_url` for that row.

No `.extra()` or custom SQL required and it handles the escaping automatically as it's now just standard querysets.
The generated SQL looks like (truncated listing of all the fields after the SELECT):

```
SELECT   ...,
         length("redirects_redirect"."old_url") AS "url_length",
         /abc/123/test-workspace/               AS "request_url"
FROM     "redirects_redirect"
WHERE    /abc/123/test-workspace/::text
LIKE replace(replace(replace(
        ("redirects_redirect"."old_url"), e'\\', e'\\\\'), e'%', e'\\%'), e'_', e'\\_') || '%'
ORDER BY 13 DESC
```

Does that work?

mikerkelly: Looking at the generated SQL again, the `request_url` line actually looks redundant, the WHERE just uses a string-literal. Maybe you can do it without the annotation, but I can't think how. I suppose it doesn't matter much, just wondering if the QS could be tighter.

Maybe that use of string literals in this comparison depends on the DB engine (`::` is postgres casting), and is perhaps not even possible in general, whereas comparing the `request_url` and the transformed `old_url` would work in any engine. So this is Django optimizing the WHERE line for postgres while leaving in the annotation we ask for.

Jongmassey: The generated SQL is much uglier but the Python is more tractable and that's what matters 👍

StevenMaude: Thank you to both Jon and Mike for looking at this!

StevenMaude (approved):

PR #4636, opened 2024-10-01
Rename CoreDeveloper Role

### LINKED ISSUE #4604: `CoreDeveloper` role might be mis-named

- [x] Update codebase

Jongmassey:
fixes #4604


iaindillingham (approved):
Thanks for picking this up, @Jongmassey. I'd like to think you honed your regex skill along the way 🙂 You may wish to consider fixing up the three commits into one, as if we ever needed to back out of this change, then we'd only have to perform a single revert.

Jongmassey:
Yes I'll squash and merge, I thought it would be slightly less intimidating to review as three!

PR #4639, opened 2024-10-02
Rename core developer (take 2)

### LINKED ISSUE #4604: `CoreDeveloper` role might be mis-named

Jongmassey:
fixes #4604 

~~In order for us to successfully migrate from one role to another, we need to be able to remove the old role from the codebase.~~

~~Prior to this PR, there was no graceful handling of the situation where a user record in the db had `roles` whose path was no longer present in the codebase.~~

~~This change handles this situation more gracefully by creating a skeleton role type with that path (but no model associations or permissions) in such a way that these "legacy" roles round trip from the database.~~

~~This change includes the change of name of role in the codebase, this more graceful role parsing, and a migration to change the existing role records in the database. The migration also includes local fake role classes and as such is not coupled to the old `CoreDeveloper` or new `StaffAreaAdministrator` roles in the codebase.~~

After discussion with @mikerkelly  and @iaindillingham we have decided that leaving a vestigial `CoreDeveloper` role class in the codebase for as long as there are dependent data migrations present is acceptable cruft.

Alternative approaches attempted in this PR were deemed to be too complex.

Following merging of this PR, an issue detailing the necessary removal of the vestigial class should be created.


Jongmassey:
I have tested locally this with a fresh dump of the prod database, running `manage.py migrate`  then `just run` - I can log in as myself and access all the staff area

mikerkelly:
I think the non-docs commits should be squished before merging, as they're atomic, I think? You don't want to roll back or deploy any one individually. Re-name and references work together, you need to migrate if your DB has been either side of these commits.

jobserver/authorization/roles.py
```diff
@@ -22,18 +22,22 @@
 
 class CoreDeveloper:
     """
-    Bennett staff member with administrator access to Job Server.
+    Temporary vestigial CoreDeveloper role to allow migration
```
mikerkelly, line 25: I'm not sure I understand why this exists? Is it because the references weren't changed in the same commit, failing CI? If this is merged with the references commit, you don't need this class anymore I think as nothing references it?

Jongmassey:
As the vestigial CoreDeveloper isn't assigned any models it doesn't appear in the Staff Area User page.
image

jobserver/authorization/roles.py
```diff
@@ -22,18 +22,22 @@
...
+    models = []
+    permissions = []
 
-    Note the name is misleading here as this does not imply what we generally mean by
-    "core developer". We plan to rename this role as part of a more general permissions
-    revamp.
+
+class StaffAreaAdministrator:
+    """
+    Bennett staff member with administrator access to Job Server.
```
mikerkelly, line 36: Worth summarising what that means? Access to all of the /staff app, can twiddle most models? Not obvious I think.

Could mention we might split this up in future?

jobserver/migrations/0008_coredeveloper_to_staffareaadministrator.py
```diff
@@ -0,0 +1,44 @@
+# Generated by Django 5.1.1 on 2024-10-02 11:00
```
mikerkelly, line 1: I think this is mostly hand-written and should have a docstring explaining why it exists, what the purpose and rationale are, and the approach taken. If anyone need to come back to where we are I think that will be very non-obvious.

jobserver/migrations/0008_coredeveloper_to_staffareaadministrator.py
```diff
@@ -0,0 +1,44 @@
...
+
+from django.db import migrations
+
+from jobserver.authorization.roles import StaffAreaAdministrator
+
+
+# We want to eventually remove the "real" CoreDeveloper so make a fake one here
+class CoreDeveloper:
```
mikerkelly, line 9: I think this idea works well, but I would like a slightly more in-depth explanation in comment.

jobserver/authorization/roles.py
Jongmassey: The proper description of these roles is decided and agreed on elsewhere, I'd rather not change them in the course of this PR as not to open that particular can of worms.

jobserver/authorization/roles.py
Jongmassey: I've come to realise that this is actually necessary to make the migration run when there is `jobserver.authorization.roles.CoreDeveloper` role records in the db (see the updated PR description).

jobserver/migrations/0008_coredeveloper_to_staffareaadministrator.py
Jongmassey: It turns out this isn't quite sufficient for this migration to run with the real `CoreDeveloper` deleted when there are `jobserver.authorization.roles.CoreDeveloper` role records in the db. If you can find a way to make is sufficient then that would make this PR much simpler and more robust!

jobserver/authorization/roles.py
mikerkelly: Reasonable, gotcha. Can it say "for info on what that means etc. refer to [XXX]" with a pointer to that elsewhere? Okay if not.

Jongmassey:
@mikerkelly  yes, will squash before merge - keeping them separate made it easier for me to iterate on reworking this.

jobserver/authorization/roles.py
Jongmassey: This is no longer the case after making `parse_role()` more resilient

jobserver/authorization/roles.py
Jongmassey: Thomas already has a PR in flight to remove these docstrings entirely so I'm going to resolve this because it will become irrelevant after a rebase!

tests/unit/jobserver/models/test_user.py
```diff
@@ -351,3 +353,20 @@ def test_create_user():
     assert user.email == "test@test.test"
     # Password is not stored in plaintext.
     assert user.password != password
+
+
+def test_user_legacy_roles_round_trip():
```
mikerkelly, line 358: Optionally would like a comment here explaining what we mean by legacy role and why this test exists as it's a little complex.

Jongmassey: I've added a docstring - do you think that covers it well enough?

mikerkelly:
Manual testing done on my local dev setup:

* I had a local DB with two User, one with `CoreDeveloper`.
* I applied migration 0008_coredeveloper_to_`StaffAreaAdministrator` and was able to use the staff site.
* I reversed the migration and the site was functional except I could not access the staff area (as I had `CoreDeveloper` not `StaffAreaAdministrator`). 403 not 500 error now when trying to bypass, as you'd hope. After reapplying the migration I could access the staff pages again.
* I used PDB to step into the migration and insert a spurious role `jobserver.authorization.roles.LegacyRole` on the User with `CoreDeveloper`. After stepping across the migration in each direction a few times, this spurious legacy role remained in the DB. So nothing is being dropped, even if unrecognized.
* If you put a class without `jobserver.authorization.roles` prefix in the table the site 500s whenever it tries to check roles. I think this is our expected behaviour given the current design of this package.

I think I realised what was making me uneasy: we have no tests of the migration code itself, which is custom code we've written. Such tests could be written by setting up the DB, calling `django.core.management.call_command`, then checking the DB state after (perhaps with `django.db.connection.cursor` to execute some SQL SELECT directly to avoid engaging the model code). I think ideally we would have such tests to validate and document the expected behaviour, but I would understand if you didn't feel that was necessary/cost-beneficial given how thoroughly we've tested it manually and that this has been pending a while.

I left a couple of comments with nits about docs. Optional. I suppose someone could always follow the git history to find this ticket and related ones.

I think it would be useful to have a note at the top of `jobserver/authorization/roles.py` that if any of these class names change, you need to make a custom migration, pointing to this one and/or this ticket. That might help someone that needs to do something similar in future. Optional.

Apart from that, good to merge. We discussed reducing the number of commits to make them more atomic.

mikerkelly (approved):

Jongmassey:
>If you put a class without jobserver.authorization.roles prefix in the table the site 500s whenever it tries to check roles. I think this is our expected behaviour given the current design of this package.

Yes, that's the current behaviour or `ensure_role_paths`:
https://github.com/opensafely-core/job-server/blob/b2cbb3bb0585bb0a3ead3fe1eb0f2515ddd54b20/jobserver/authorization/parsing.py#L4-L18


We _could_ alter this such that the error becomes softer  - e.g. we get notified that there's wonky roles in the db but the site remains functional. However, I think any circumstance that resulted in roles being put in the db that weren't in the `jobserver.authorization.roles` namespace would be a major problem and so failing hard is the right thing to do to.

Jongmassey:
From a quick search on the BI google drive that's all I find that I think needs updating. We will notify NOD and IG once this is done to ask them to update any of their documentation and processes of which we may be unaware

jobserver/migrations/0008_coredeveloper_to_staffareaadministrator.py
```diff
@@ -0,0 +1,35 @@
...
+
+from jobserver.authorization.roles import CoreDeveloper, StaffAreaAdministrator
+
+
+def change_user_role(apps, from_role, to_role):
+    User = apps.get_model("jobserver", "User")
+    users = User.objects.filter(roles__contains=from_role)
+    for user in users:
+        user.roles.remove(CoreDeveloper)
+        user.roles.append(StaffAreaAdministrator)
```
iaindillingham, line 13: I think you mean...

```suggestion
        user.roles.remove(from_role)
        user.roles.append(to_role)
```

because otherwise the reverse migration will fail with...

```
ValueError: list.remove(x): x not in list
```

iaindillingham (approved):
Thank you for reworking this, @Jongmassey. Although we end up with some cruft in `jobserver/authorization/roles.py`, we need not use class factories and fakes.

At present, the reverse migration is broken. I've made a suggestion that fixes this but approved the PR.

jobserver/migrations/0008_coredeveloper_to_staffareaadministrator.py
Jongmassey: I absolutely do mean that! Thanks for catching that

PR #4645, opened 2024-10-07
Fix `default_language_version`

iaindillingham:
pre-commit's `default_language_version` should have been bumped from 3.11 to 3.12 almost exactly a year ago (c1759e89), but it was missed. Those with a python3.11 binary won't have noticed, but those without, like @KatieB5, would. Thank you for spotting this, Katie!


KatieB5 (approved):
Thanks Iain!

Issue #4593, opened 2024-09-06
Replace `pragma: no cover` with a test


iaindillingham:
For expediency, #4574 added a `pragma: no cover`:

https://github.com/opensafely-core/job-server/blob/22f5abdc0bc2e4269fb9b214c037e22bcba7db83/jobserver/forms.py#L138

We should remove it and exercise the line with a test. It's worth taking some time to do so, as -- to me, at least -- it isn't clear how we might end up here, and what returning `None` from `WorkspaceCreateForm.clean` means.

mikerkelly:
I don't think this block of code achieves anything after #4725 and #4730 remove the rest of `clean()`. I suspect it existed to try and avoid that code accessing attributes of `repo` or `branch` when either was not provided in `data`. This would have caused `branch.lower()` to fail, for example, even though `branch` being `None` would independently have failed field validation. `clean()` is executed even if field-level validation fails, to allow for fuller error reporting.

So I think `clean()` can be removed altogether if #4730 is approved and merged. Nice.

PR #4654, opened 2024-10-08
Parametrize `test_workspacecreateform_success`

iaindillingham:
In this PR, we make a couple of small changes to two tests. These small changes help us see that the two tests are basically the same, save for two parameters. We then parameterise one test and remove the other, meaning we have the same tests for 14 fewer lines of code.

We use the [`pytest.mark.parametrize`](https://docs.pytest.org/en/latest/how-to/parametrize.html#pytest-mark-parametrize) decorator to parameterise the test, also using [`pytest.param`](https://docs.pytest.org/en/latest/reference/reference.html#pytest.param) to assign a more descriptive ID to each parameter set. Doing so also makes it easier to run the test with either parameter set:

```sh
pytest tests/unit/jobserver/test_forms.py::test_workspacecreateform_success[lower_case_name]

pytest tests/unit/jobserver/test_forms.py::test_workspacecreateform_success[mixed_case_name]
```

Or with both parameter sets:

```sh
pytest tests/unit/jobserver/test_forms.py::test_workspacecreateform_success
```

Although out of scope, I think it's worth noting that `test_workspacecreateform_success` would be a great candidate for [property-based testing](https://hypothesis.readthedocs.io/en/latest/). It's unlikely that every `name` will be either lower-case or mixed-case ASCII in real life, so verifying that `WorkspaceCreateForm.clean_name` works correctly with a wide range of `name`s is important.

I created this PR whilst working on #4593. It doesn't address that issue, but it does make that issue easier to address.

### DIFF
--- a/tests/unit/jobserver/test_forms.py
+++ b/tests/unit/jobserver/test_forms.py
@@ -116,7 +116,7 @@ def test_workspacecreateform_success():
     assert form.is_valid()
 
 
-def test_workspacecreateform_success_with_upper_case_names():
+def test_workspacecreateform_success_with_mixed_case_name():
     project = ProjectFactory()
     data = {
         "name": "TeSt",
--- a/tests/unit/jobserver/test_forms.py
+++ b/tests/unit/jobserver/test_forms.py
@@ -96,10 +96,17 @@ def test_jobrequestcreateform_with_bad_codelists(
         )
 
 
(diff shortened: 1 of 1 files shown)


KatieB5 (approved):
Thanks Iain! Feedback discussed on Friday.

PR #4658, opened 2024-10-09
Remove unused L4 file viewer code

### LINKED ISSUE #4588: Remove unused "Level 4 file viewer" code
Previously, we had a system for viewing unreleased output files which involved users logging in to Job Server from within Level 4 (using a single-use token) and then using a "hybrid" JavaScript app which combined a remote UI with file data from a local API. This allowed viewing unreleased files without that data every leaving the secure environment.

Airlock replaces this wholesale by running the entire application within the secure environment.

This PR updated the UI to replace references to the old system with references to Airlock:
 * https://github.com/opensafely-core/job-server/pull/4585

However it left in place of lot of now redundant code which should be cleaned up. A few examples are:

 * [ ] The [`Workspace.get_files_url`](https://github.com/opensafely-core/job-server/blob/a215c455cdb509634e4dce38c57d68db7739fc9b/jobserver/models/workspace.py#L162-L169) method and any tests which reference it.
 * [ ] The [`backend_files.html`](https://github.com/opensafely-core/job-server/blob/a215c455cdb509634e4dce38c57d68db7739fc9b/templates/workspace/backend_files.html) template and whatever view loads it, and whatever URL config points to that view.
 * [ ] The [single-use token](https://github.com/opensafely-core/job-server/blob/a215c455cdb509634e4dce38c57d68db7739fc9b/templates/login.html#L60-L80) login UI.
 * [ ] The [`signing.py`](https://github.com/opensafely-core/job-server/blob/c3dbea8127c5f8cc2c6073e748d0f53270f7fcbe/jobserver/signing.py) utility code used by the `WorkspaceBackendFiles` view (see #4602 for discussion)

Note that the single-use token **API endpoint** needs to remain as this is used by Airlock. As therefore does the single-use token **generator UI**. It's just logging in to Job Server itself using the token which is no longer supported.

Note also the the file viewer SPA needs to remain for now as this is used for viewing released output files. Longer term there's an ambition to replace this with something simpler.

Related Slack thread:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1725463761914569

inglesp:
References to the L4 file viewer were removed in #4585.  This cleans up the now-unreferenced code.

Fixes #4588.


rebkwok (approved):

mikerkelly:
Team REX will also review this before merge and deploy.

PR #4662, opened 2024-10-10
Unpin `urllib3`

StevenMaude:
v2.2.3 is the latest `urllib3` and we've had Dependabot pull requests update this version.

Example: https://github.com/opensafely-core/job-server/pull/4609

Therefore, this comment is no longer correct, and it implies the pinning is no longer required.

### DIFF
--- a/requirements.prod.in
+++ b/requirements.prod.in
@@ -35,8 +35,6 @@ slippers
 social-auth-app-django
 social-auth-core
 structlog
-# 2.1.0 appears to be causing some problems with exporting
-# https://github.com/opensafely-core/job-server/issues/3955
-urllib3<=2.2.3
+urllib3
 whitenoise[brotli]
 xkcdpass


StevenMaude:
It seems Dependabot ignored/ignores the pinning in the input requirements:

https://github.com/opensafely-core/job-server/commit/c8cc49db6bbabe87cbb773b1474691e1ab86ffa8

:man_shrugging:

StevenMaude:
As pointed out by @lucyb on Slack, it could be that the `urllib3` version relates to the cause of some OpenTelemetry errors. In that case, we should actually repin the version, and retain the comment (or some form of it).

StevenMaude:
I'm going to close this and open an issue for it instead, as it's unclear whether we should actually pin to an older version.

StevenMaude:
Closed in favour of a decision to be made in #4715.

PR #4663, opened 2024-10-10
Remove unneeded dependencies from L4 file viewer

### LINKED ISSUE #4661: Remove redundant dependencies after removing L4 file viewer code
As a result of #4588 we can now remove:
- Pydantic
- itsdangerous

Pydantic may still be a requirement of the current version of the pipeline library. However, it's just been removed from there, so I think updating to the latest version means we should be able to remove it from Job Server too.

StevenMaude:
This PR updates `opensafely-pipeline` which facilitates removing `pydantic` and `itsdangerous`.

Fixes #4661.


StevenMaude:
This is no longer needed with #4664 and #4666, courtesy of the RAP team.

PR #4665, opened 2024-10-11
Don't render large YAML in JobRequestDetail view.

### LINKED ISSUE #3704: Limit the size of project.yamls shown on the Job Request Detail page
We recently received a [Sentry error](https://ebm-datalab.sentry.io/issues/4553630533/?alert_rule_id=10884789&alert_type=issue&notification_uuid=a157c3c1-b5db-487b-9274-c78157fa8335&project=5443358&referrer=slack) because someone tried to view a 2.25MB (43,000 lines) project.yaml. This caused a timeout and when I tried to replicate it locally (which shouldn't have the timeout issue), it caused the browser tab to become unresponsive due to the large size.

We could optimise how we're rendering this file. However, it's too large a file for someone to reason about or usefully use, so we should put a limit on the size of the file we're willing to render. If a file is too large we should provide an informative error and maybe point to the file in GitHub if that's easy to do.

mikerkelly:
Fixes #3704.

The `job-request-detail` template doesn't display the rendered project YAML if the context flags `project_yaml` sub-key `is_too_large` (from #3870, which was also towards #3704). Therefore the view code doesn't need to render the definition in that case. Doing so was slowing down page rendering by 5 or more seconds in some cases.

The extant code arbitrarily defined "large" as longer than 10,000 characters, but the comments mentioned ~1MB , which doesn't match. I've updated the comment. I've not considered updating the arbitrary value used.

As well as a unit test for the "too large" case, I added one for the rendering actually taking effect, as it was missing. And added checks that the `project_yaml` `is_too_large` or `is_empty` are as they should be in the other "happy path" tests.

### DIFF
--- a/jobserver/views/job_requests.py
+++ b/jobserver/views/job_requests.py
@@ -281,14 +281,20 @@ class JobRequestDetail(View):
     def get_project_yaml(self, job_request):
         is_empty = job_request.project_definition == ""
 
-        # We know files that are 2.43MB take too long to render with pygments
-        # in production and crash browser tabs locally, so we're limiting the
-        # size.  The current 1MB is arbitrary until we can get some telemetry
-        # from production.
-        is_too_large = len(job_request.project_definition) > 10_000  # ~1MB
+        # Is this file too large to render? Files around 2.43MB crash tabs and
+        # render slowly. The 10K character limit (~10-40KB) is arbitrary. It
+        # could be refined, perhaps with telemetry. Length is an imperfect
+        # proxy for render cost, but better than size, which ignores encoding.
+        is_too_large = len(job_request.project_definition) > 10_000  # ~10-40KB
 
         if is_empty:
+            # Nothing to render, may as well not call render_definition.
             project_definition = ""
+        elif is_too_large:
+            # Skip rendering as the template shouldn't display the result.
+            # Return a human-readable string in case the template behavior
+            # changes and shows project_definition despite is_too_large.
+            project_definition = "This file is too large to render."
         else:
             project_definition = mark_safe(
                 render_definition(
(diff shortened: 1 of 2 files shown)


Jongmassey (approved):
Looks good to me. Tested locally with a fresh prod db dump and tested with The Very Large Job Request (/comparing-disparities-in-rsv-influenza-and-covid-19/disparities-comparison-rsv-flu-c19/23608/) and it is well-behaved, as it is with a regular-sized one.

jobserver/views/job_requests.py
```diff
@@ -283,12 +283,23 @@ def get_project_yaml(self, job_request):
 
         # We know files that are 2.43MB take too long to render with pygments
         # in production and crash browser tabs locally, so we're limiting the
-        # size.  The current 1MB is arbitrary until we can get some telemetry
+        # size.  The current ~10KB is arbitrary until we can get some telemetry
         # from production.
-        is_too_large = len(job_request.project_definition) > 10_000  # ~1MB
+        # Note: ~10KB assuming Latin-1 encoding, UCS-2 or UCS-4 would be ~2/4
+        # times as big. In any case, we're using length as only an approximate
+        # measure of compute required to render the definition.
```
iaindillingham, line 290: I had to read this comment a couple of times. Do you mean...

> We want to limit the project definition to 10kB. Assuming the Latin-1 encoding (1 byte/character), then 10kB corresponds to 10,000 characters. However, 10,000 characters correspond to 20kB in UCS-2 (2 bytes/character) or 40kB in UCS-4 (4 bytes/character).

iaindillingham (approved):

jobserver/views/job_requests.py
mikerkelly: Thanks, Iain. This is much clearer. On reflection, it's more important to highlight the approximate nature of the length measure than how encoding impacts size. I've rewritten both comments to be more focused -- let me know your thoughts. And thanks for pushing me to rethink how these comments could better serve their purpose.

PR #4666, opened 2024-10-11
Bump pipeline to v2024.10.11.170331

### LINKED ISSUE #4661: Remove redundant dependencies after removing L4 file viewer code

alarthast:
Fixes #4661.

- Part of https://github.com/opensafely-core/pipeline/issues/224
- Encompasses all the changes made in `pipeline` releases between March and now

- `pipeline` v2024.09.27.131013 removes the use of the pydantic, hence also
  - Use the new `Pipeline.build` syntax for tests (this PR)
  - #4664


evansd:
There's an additional change that we'll need to make here which is that the `opensafely-pipeline` dependency in `requirements.prod.in` needs to change to:
```
opensafely-pipeline[fastparser]@...
```

This will pull in the necessary dependencies to give us fast YAML parsing. More details in the pipeline PR:
 * https://github.com/opensafely-core/pipeline/pull/346

StevenMaude (approved):
Looks good; thanks!

PR #4668, opened 2024-10-14
Tidy CoreDeveloper vestigial indirect references.

### LINKED ISSUE #4604: `CoreDeveloper` role might be mis-named

mikerkelly:
Resolves #4604.

#4639 re-named the CoreDeveloper role to StaffAreaAdministrator. Direct references were updated. There are a couple of places with vestigial indirect references. In the management command `create_user` `--core-developer` flag and the `pytest`  fixture `core_developer` and therefore throughout `tests/`. Let's tidy these up as the vestigial references are potentially confusing and unclear. Better to reference the new role name.

### DIFF
--- a/jobserver/management/commands/create_user.py
+++ b/jobserver/management/commands/create_user.py
@@ -21,12 +21,14 @@ class Command(BaseCommand):
         parser.add_argument("--email", help="Defaults to username@example.com")
         parser.add_argument("--name", help="Defaults to username")
         parser.add_argument(
+            "-o",
             "--output-checker",
             action="store_true",
             help="Make user global OutputChecker",
         )
         parser.add_argument(
-            "--core-developer",
+            "-s",
+            "--staff-area-administrator",
             action="store_true",
             help="Make user global StaffAreaAdministrator",
         )
@@ -52,7 +54,7 @@ class Command(BaseCommand):
         roles = []
         if options["output_checker"]:
             roles.append(OutputChecker)
-        if options["core_developer"]:
+        if options["staff_area_administrator"]:
             roles.append(StaffAreaAdministrator)
 
         updated = False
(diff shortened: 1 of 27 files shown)


KatieB5 (approved):
Thanks for your work on this Mike! I'm happy to approve this PR. I'd also just like to say that this was a dream to review as a Junior Dev. You referenced the previous related issues, you gave clear context for your actions and you provided your sed command, all of which were very helpful.

jobserver/management/commands/create_user.py
```diff
@@ -21,12 +21,14 @@ def add_arguments(self, parser):
...
         parser.add_argument(
+            "-o",
             "--output-checker",
             action="store_true",
             help="Make user global OutputChecker",
         )
         parser.add_argument(
-            "--core-developer",
+            "-s",
+            "--staff-area-administrator",
```
iaindillingham, line 31: Looking back at the history of this management command...

```sh
git log --oneline -- jobserver/management/commands/create_user.py
```

...I see that Simon created it in 989c000f. It's not clear whether Simon intended to use it in scripts, but if he did, then this commit will break them. Could you drop into #team-rap and let them know?

iaindillingham (commented):
This PR was a joy to read, @mikerkelly, thank you for crafting it so expertly. I really appreciated the reference to the `find`/`sed` command in the commit message, and the note about `just fix`'s influence.

jobserver/management/commands/create_user.py
bloodearnest: Yes, we use to script set up for a local job-server for integration testing.

https://github.com/opensafely-core/airlock/blob/main/job-server/local-setup.sh#L64

However, this scripting is covered in CI, so this change should detect that and our new failing workflows dashboard should show it, so lets see if it all works as intended!

iaindillingham: I guess I saw this more as a breaking API change, and then wondered whether there were any non-human clients.

PR #4682, opened 2024-10-16
Don't reset invalid workspace creation form fields

### LINKED ISSUE #4618: Workspace creation: a validation error causes the github repo field to be reset
When creating a new workspace for a project, the GitHub repo field is reset when there is a validation error on form submission.

The other fields retain the values that have been entered, however the GitHub Repo field resets to the first item in the list. It's easy to miss that this has happened and not possible to change after the fact.

Steps to recreate:
1. Enter a workspace name with a space or a hypen
2. Enter a workspace purpose
3. Select a repo from the list
4. Click "Create workspace"

The form is refreshed with an error in the Workspace name, but all user entered values are the same, except for the GitHub repo.

Screenshot from 2024-09-19 17-43-21

Suggested fix:
1. Ensure the selected repo is preserved when the [clean_name](https://github.com/opensafely-core/job-server/blob/5baa27649eabd878fa63c53476d2bff23de84bce/jobserver/forms.py#L159) step fails in the form.
2. (Bonus) With a new form, default to selecting a blank value in the list, so that users always need to make a choice as selecting the first value by default is confusing.

mikerkelly:
Fixes #4618.

Also addresses some small issues with the tests and form.


tests/unit/jobserver/test_forms.py
```diff
@@ -104,27 +104,42 @@ def test_jobrequestcreateform_with_bad_codelists(
...
+        },
     ]
-    form = WorkspaceCreateForm(project, repos_with_branches, data)
+    data = {
+        "name": name,
+        "repo": "http://example.com/derp/test-repo2",
+        "branch": "test-branch2",
+        "purpose": "test purpose",
+    }
+    form = WorkspaceCreateForm(project, repos_with_branches, data=data)
```
iaindillingham: > These test cases were incorrectly passing `data` as a positional argument. A `Form` is initialised with POST data through the `data` _kwarg_.

This surprised me, Mike. Can you point to some documentation? It's not apparent from "[The Forms API](https://docs.djangoproject.com/en/5.1/ref/forms/api/)" page in the Django documentation that `MyForm(data)` is incorrect, but `MyForm(data=data)` is correct.

tests/unit/jobserver/test_forms.py
```diff
@@ -104,27 +104,42 @@ def test_jobrequestcreateform_with_bad_codelists(
     ],
 )
 def test_workspacecreateform_success(name, cleaned_name):
+    """
+    When the form is instantiated with valid data then:
+        * The form validates.
+        * The data ends up in the cleaned_data.
+        * The name field is cleaned by converting to lower case.
+    """
```
iaindillingham: I'm reluctant to add a comment (docstring) that's an English-language description of the implementation. If we thought of the implementation as a black-box, then what could we say about the intention of this function?

I think the intention of this function is to test that form and field validation works correctly (where validation encompasses cleaning). Perhaps renaming the function `test_workspacecreateform_validation` would help communicate the intention?

tests/unit/jobserver/test_forms.py
```diff
@@ -104,27 +104,42 @@ def test_jobrequestcreateform_with_bad_codelists(
...
 
     assert form.is_valid()
     assert form.cleaned_data["name"] == cleaned_name
+    assert form.cleaned_data["repo"] == "http://example.com/derp/test-repo2"
+    assert form.cleaned_data["branch"] == "test-branch2"
+    assert form.cleaned_data["purpose"] == "test purpose"
 
 
 def test_workspacecreateform_unknown_branch():
+    """When the form cleaned_data has an unknown branch, validation fails."""
```
iaindillingham, line 177: Here, and below, I feel we should incorporate the comments into the names of the functions. Doing so would make pytest's output clearer - if a test failed, then we wouldn't have to cross-reference with the function's comment to learn more about why the test failed.

jobserver/forms.py
```diff
@@ -98,10 +98,9 @@ class WorkspaceCreateForm(forms.Form):
     purpose = forms.CharField(help_text="Describe the purpose of this workspace.")
     branch = forms.CharField()
 
-    def __init__(self, project, repos_with_branches, *args, **kwargs):
```
iaindillingham, line 101: Great spot!

jobserver/forms.py
```diff
@@ -93,7 +93,11 @@ class WorkspaceArchiveToggleForm(forms.Form):
 
 class WorkspaceCreateForm(forms.Form):
     name = forms.SlugField(
-        help_text="Enter a descriptive name which makes this workspace easy to identify.  It will also be the name of the directory in which you will find results after jobs from this workspace are run."
+        help_text=(
+            "Enter a descriptive name for the workspace, which will also "
+            "be the directory name for results after job execution. Valid "
+            "characters: letters, numbers, underscores, and hyphens."
+        )
```
iaindillingham, line 100: Thanks for this. I've been meaning to create an issue/PR for ages.

iaindillingham (commented):
I'm going to submit my review, with the expectation that I'm going to return to it tomorrow. I really would like to know more about the `data` kwarg, though.

tests/unit/jobserver/test_forms.py
mikerkelly: Thanks, Iain, this is a good spot. I've indeed got that a bit wrong. The Django `Form` API allows either.

I could have said "our custom `WorkspaceCreateForm` only correctly handles POSTED `data` when supplied as a keyword-argument. This is because the class `__init__ ` code refers to `kwargs` directly." That would have been accurate. That's why I was finding my expanded success test was failing when `data` was passed by position.

But I'd gotten the implication of that the wrong way around -- the tests weren't broken, the form code design is (both before this PR, and with the changes as you initially reviewed them, as I continued the pattern of referring to `kwargs`.) It should access `self.data`, not subtly require the form to be instantiated with `data` passed by key-word to work correctly. It's not following the `Form` API properly. I'll fix.

This won't have been noticed so far because it's only used in one place in the prod code, and the test code was not robust enough to detect that the `data` it tried to supply was just being ignored in those tests.

tests/unit/jobserver/test_forms.py
mikerkelly: I'd be happy to add `_validation_fails` to the end of each. I think the docstrings are more legible when viewing the code and barely make the functions longer, so can we do both?

tests/unit/jobserver/test_forms.py
mikerkelly: Committed a rebased fix. The [Use form.data, not kwargs. Validate more in tests.](https://github.com/opensafely-core/job-server/pull/4682/commits/07cbf7ec293cf23d9ec62615869e28c2ff7a54dc) commit is now a bit overloaded / non-atomic but it would be a bit of effort to unpick it. Let me know if you think I should.

tests/unit/jobserver/test_forms.py
mikerkelly: I think one good alternative might be making the test more atomic -- this test now tests multiple things, I think. I haven't yet thought about the detail of how to do that.

tests/unit/jobserver/test_forms.py
mikerkelly: I need to rewrite the commit messages to reflect this. And the docstrings and comments can use some rework. But I think the form code and test code are reviewable.

I also think it's worth the tests explicitly checking the choice fields are constructed with the right options (would have caught this issue a different way if robust enough -- should have multiple repos and branches) and that `form.bound` is as we expect in each (which wouldn't have caught this issue but is worth sense-checking). What do you think?

iaindillingham (approved):
Thanks for talking this through earlier, @mikerkelly. I found the diff of all commits easier to follow than the diff of each commit. 

You can see the diff of all commits with the following:

```sh
git diff --merge-base main HEAD
```

Let's save the discussion about naming test functions and comments for another time. From memory, I think *[Unit Testing Principles, Practices, and Patterns](https://learning.oreilly.com/library/view/unit-testing-principles/9781617296277/)* has a lot of good advice.

PR #4687, opened 2024-10-17
Renumber ADRs sequentially

mikerkelly:
We have a few ADR with overlapping numerical parts of their name (2* 0018 and 2* 0023). This is probably just due to historical accident.  I re-named them to have unique sequential chronological numbers.

This may possibly break a few links to them in docs or Slack, but we prefer to make things consistent now. There are probably not many such links, and if we don't renumber it may also be confusing in future if a reference is made to "ADR 18" or "ADR 23" which is meant. And if we do it later when there are more ADRs it will be slightly more work.


iaindillingham (approved):
Thanks for this! Better to do sooner rather than later 🚀

PR #4701, opened 2024-10-22
Remove end-to-end test references

### LINKED ISSUE #4681: Remove references to e2e testing in TESTING.md
In [TESTING.md](https://github.com/opensafely-core/job-server/blob/main/TESTING.md), we currently describe Job Server's testing pyramid as a three layer structure that includes end-to-end testing. This is inaccurate: Job Server does not have any end-to-end tests. We should update the language to accurately reflect our testing pyramid hierarchy. We should also update the 'End-to-End' section of TESTING.md if we do not plan to add end-to-end tests anytime in the near future.

StevenMaude:
Fixes #4681. See also #4702.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -30,8 +30,6 @@ Job Server uses a testing pyramid, with the following layers from top to bottom:
 * Integration
 * Unit
 
-**At present, there are no end-to-end tests.**
-
 ### Unit Tests
 **Directory:** `tests/unit`
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -24,15 +24,13 @@ Tests should be fast where possible.
 
 
 ## Testing Pyramid
-Job Server uses a three layer testing pyramid, with the following layers from top to bottom:
-* End-to-End
+
+Job Server uses a testing pyramid, with the following layers from top to bottom:
+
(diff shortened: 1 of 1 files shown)


TESTING.md
```diff
@@ -24,15 +24,13 @@ Tests should be fast where possible.
...
-* Unit
 
-Below we describe how we draw the lines between each of those layers.
+Job Server uses a testing pyramid, with the following layers from top to bottom:
 
-Much like the developer tooling at Bennett Institute for Applied Data Science we define these layers as an ideal guide that should be adhered by default, but in the knowledge that things can deviate if there is a good reason.
+* Integration
+* Unit
 
+**At present, there are no end-to-end tests.**
```
iaindillingham, line 33: This line isn't necessary. We don't mention there aren't any other types of test, so why mention that there aren't any end-to-end tests?

iaindillingham (approved):
If we're removing them, then let's _really_ remove them 🙂

TESTING.md
StevenMaude: I left it in because it's part of the typical pyramid testing, but it's fine to take the line now, so let's do that.

PR #4707, opened 2024-10-28
Document and link to private repo policy

mikerkelly:
We display warnings to the user when they are viewing a workspace attached to a private repo nearing their public release date. Provide a link to the policy documentation in these messages.

Explain the related policy and process on the private repo dashboard, and add explanations in the relevant code sections with links to the policy document. Existing information did not make this clear.

Here is [the policy](https://docs.opensafely.org/repositories/#when-you-need-to-make-your-code-public) and a (Bennett internal) [Slack thread](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1730106471692929) discussing this and other dashboards. 

The template changes can be tested locally by visiting the [dashboard](http://localhost:8000/staff/dashboards/) and by editing the `WorkspaceDetail` view to set `show_publish_repo_warning = True` then visiting the page for a `Workspace` to which you have access (`/project-slug/workspace-slug/`).

### DIFF
--- a/jobserver/views/workspaces.py
+++ b/jobserver/views/workspaces.py
@@ -197,7 +197,7 @@ class WorkspaceDetail(View):
             name=self.kwargs["workspace_slug"],
         )
 
-        # get the first job for this workspace's repo
+        # Get the first job for this workspace's repo.
         first_job = (
             Job.objects.filter(job_request__workspace__project=workspace.project)
             .annotate(run_at=Least("started_at", "created_at"))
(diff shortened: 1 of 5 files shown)


templates/staff/dashboards/index.html
```diff
@@ -26,7 +26,7 @@
 
           {% url "staff:dashboards:repos" as staff_dashboards_repos_url %}
           {% #list_group_rich_item title="Private Repos" type="Dashboard" url=staff_dashboards_repos_url %}
-            <p>Private repos attached to one or more workspaces, where code was first run more than 11 months ago</p>
+            <p>Private research repositories linked to one or more workspaces, with initial code execution over 11 months ago, for IG monitoring</p>
```
Jongmassey, line 29: I'd leave off the "for IG monitoring" for now. I think we don't have a good handle on who uses these dashboards for what purposes and I'm not sure adding this gains us anything.

templates/staff/dashboards/repos.html
```diff
@@ -13,7 +13,7 @@
 {% endblock breadcrumbs %}
 
 {% block hero %}
-  {% staff_hero title="OpenSAFELY Private Repos" text="This table contains all private repos attached to one or more workspaces, where code was first run more than 11 months ago." %}
+  {% staff_hero title="OpenSAFELY Private Repos" text="This table contains all private research repositories linked to one or more workspaces, with initial code execution over 11 months ago. Per policy, these must be made public within 12 months. This dashboard is monitored by the Information Governance team for compliance." %}
```
Jongmassey, line 16: Ditto IG bit. I'm not sure we need to put the "why" of these dashboards here.

templates/workspace/detail.html
Jongmassey, line 1: 👍

jobserver/views/workspaces.py
Jongmassey, line 1: I appreciate the improvements to the comments here, but especially after adding a link to the policy in the template I think these comments are redundant and would be better removed! Not wedded to this, and am aware the whole topic of what role comments play in code is a controversial one!

Jongmassey (approved):

templates/staff/dashboards/index.html
mikerkelly: Agreed!

Jongmassey (approved):

PR #4725, opened 2024-11-07
Remove inaccessible `ValidationError`

iaindillingham:
`WorkspaceCreateForm.branch` is a `ChoiceField`, meaning that its `validate` method, which validates that the input is in its `choices` property, is called before its `clean` method. Consequently, unless its `clean` method is called directly, the `ValidationError` is inaccessible.

To put that another way, the `ValidationError` is inaccessible because it's raised by `Form.clean()`, but validation failed at `Field.validate()`. Django's [Form and field validation](https://docs.djangoproject.com/en/5.1/ref/forms/validation/) page would benefit from a diagram. Essentially:

1. `Field.clean()`
    1. `Field.to_python()`
    1. `Field.validate()`
    1. `Field.run_validators()`
1. `Form.clean_<fieldname>()`
2. `Form.clean()`

I'd suggest working backwards through this PRs commits.

### DIFF
--- a/jobserver/forms.py
+++ b/jobserver/forms.py
@@ -175,11 +175,6 @@ class WorkspaceCreateForm(forms.Form):
             msg = "Unknown repo, please reload the page and try again"
             raise forms.ValidationError(msg)
 
-        # normalise branch names so we can do a case insensitive match
-        branches = [b.lower() for b in repo["branches"]]
-        if branch.lower() not in branches:
-            raise forms.ValidationError(f'Unknown branch "{branch}"')
-
     def clean_name(self):
         name = self.cleaned_data["name"].lower()
 
--- a/tests/unit/jobserver/test_forms.py
+++ b/tests/unit/jobserver/test_forms.py
@@ -213,12 +213,12 @@ def test_workspacecreateform_unknown_branch_validation_fails():
             "branches": ["test-branch"],
         }
     ]
-    form = WorkspaceCreateForm(repos_with_branches)
-    form.cleaned_data = {
+    data = {
         "name": "test",
         "repo": "http://example.com/derp/test-repo",
         "branch": "unknown-branch",
     }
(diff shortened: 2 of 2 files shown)


lucyb (approved):
This is a nice change. It wasn't at all obvious that there was a ChoiceField being created and that the validation was also happening there, so thank you for your helpful explanation of what's happening. It's nice to see a much improved unit test too.

It looks like we may have slightly changed the error message being sent to users, but I think that's okay in this case. The new error returned from the form is "Select a valid choice. unknown-branch is not one of the available choices.", although I haven't checked how the template is handling this. Given that this error is highly unlikely to happen though, I don't think it's significant.

iaindillingham:
> It looks like we may have slightly changed the error message being sent to users

I don't think we have, although I understand why you reached this conclusion: form and field validation isn't straightforward 🙁

The error message from the removed `ValidationError` would have ended up in the special `__all__` field, because the removed `ValidationError` was raised by `WorkspaceCreateForm.clean()` but wasn't added to the `branch` field with [`WorkspaceCreateForm.add_error()`](https://docs.djangoproject.com/en/5.1/ref/forms/api/#django.forms.Form.add_error).

The `form_select` in `templates/workspace/create.html` (see below) didn't know that `__all__` had an error message that related to `branch`, so the `form_select` didn't display the error message. We might expect the error message to be displayed at the form-level, rather than at the field-level, but the `form` didn't include `{{ form.non_field_errors }}`.

Consequently, the error message was, and remains, that generated by `ChoiceField.validate()`:

> Select a valid choice. `<branch>` is not one of the available choices.

https://github.com/opensafely-core/job-server/blob/942f13139805c572fe1abc82c0594097a4ecfadb/templates/workspace/create.html#L22-L46

mikerkelly:
Great spot!

PR #4727, opened 2024-11-08
Fix `GitHubAPI` fakes verification tests

mikerkelly:
Fix some issues in verification tests due to #4717.

Verification tests aren't run on branch/merge CI as the ones that hit the remote APIs are costly. I must have overlooked some issues locally (coverage) and/or introduced the error manually between locally testing and committing in #4717.

### DIFF
--- a/tests/verification/test_utils.py
+++ b/tests/verification/test_utils.py
@@ -5,7 +5,7 @@ import pytest
 from .utils import assert_public_method_signature_equality
 
 
-pytestmark = [pytest.mark.disable_db]
+pytestmark = [pytest.mark.verification, pytest.mark.disable_db]
 
 
 class TestPublicMethodSignatureEquality:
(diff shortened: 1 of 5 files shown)


tests/verification/test_utils.py
iaindillingham, line 1: Would it be easier to add this file to `tool.coverage.run.omit` in `pyproject.toml`?

iaindillingham (approved):
Thanks for noticing that verification tests are run on a schedule, and that they were failing, @mikerkelly. Great to get this fixed quickly!

tests/verification/test_utils.py
mikerkelly: I think using `omit` would be equivalent to adding `#pragma: no cover` at module level. I think for a single module it would be clearer to mark that up inside that module and `omit` is most useful for patterns (`*/migrations/*` for example).

I think coverage for this file as a whole is useful. It's how I realised that its tests weren't being executed in CI. I hadn't marked the tests up as `verification`, mistakenly reading `-m verification` in the justfile as "module" or something rather than "marked". Coverage showed that. Perhaps any module with executable tests should have coverage enabled to help ensure the right set of tests are running.

I only want the dummy classes excluded from coverage, not the rest of the test code.

PR #4728, opened 2024-11-08
Filter by global roles

### LINKED ISSUE #4632: Filter by global role shows roles that aren't global
When I go to the Users page in the Staff Area and select the "Filter by global role" option, the dropdown box shows all roles rather than just the roles that can be globally assigned.

For example, ProjectDeveloper can only be assigned to a user on a project. It cannot be assigned globally, so should not be in this list.

Image

KatieB5:
Fixes #4632

Previously the Users page in the Staff Area had a "Filter by global roles" option which displayed all roles rather than just globally assignable roles. This commit adds a `global_roles` module so that the view does not need to know the details of the implementation.

### DIFF
--- a/jobserver/authorization/global_roles.py
+++ b/jobserver/authorization/global_roles.py
@@ -1,0 +1,10 @@
+from ..models import User
+from .utils import roles_for
+
+
+# A global role can be assigned to a User independently of ProjectMembership.
+# It either does not relate to Projects or grants permissions across ALL projects.
+
+GLOBAL_ROLES = roles_for(User)
+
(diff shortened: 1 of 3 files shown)


mikerkelly (approved):
Thanks Katie.

Katie and I paired on this ticket apart from writing the test.

I noticed the unit test includes an explicit list of expected role names. This made me wonder if the `GLOBAL_ROLES` definition could simply use the same list instead of being generated by `roles_for`. However, I think it's beneficial to have both approaches—one in the code and one in the test—for better verification. The minor downside is needing to update the test whenever `GLOBAL_ROLES` changes. That should be rare (maybe never) and easy.

PR #4729, opened 2024-11-08
Parametrize test

iaindillingham:
We parametrize the test to demonstrate that the filter operates on global *and* local roles. Ideally, the test wouldn't depend on us knowing that `OutputPublisher` is a global role and `ProjectDeveloper` is a local role, but making it so would add (arguably unnecessary) complexity. To clarify our intent, we assign descriptive IDs to each parameter set using `pytest.param`.

@KatieB5 and I parametrized the test when considering #4632. We didn't need to do so to address that issue, but running the test once for a global role and once for a local role has the benefit of documenting how we expect `UserList` to behave.


tests/unit/staff/views/test_users.py
```diff
@@ -588,11 +588,17 @@ def test_userlist_filter_by_invalid_org(rf, staff_area_administrator):
...
-def test_userlist_filter_by_role(rf, staff_area_administrator):
-    UserFactory(roles=[OutputPublisher])
-    UserFactory(roles=[ProjectCollaborator])
-
-    request = rf.get("/?role=OutputPublisher")
+@pytest.mark.parametrize(
+    "role",
+    [
+        pytest.param(OutputPublisher, id="global"),
+        pytest.param(ProjectDeveloper, id="local"),
```
mikerkelly, line 595: I can see we use 'local' as an antonym to 'global' for `ProjectMembership`-based roles vs. `User`-based roles in a few places. I find this wording a little tough to intuit the meaning of and wonder if "user-based" vs "project-based" might be simpler. Or "global" vs "non-global" or "project" vs "non-project". Hmm.

I'm not suggesting this needs to change here.

tests/unit/staff/views/test_users.py
```diff
@@ -588,11 +588,17 @@ def test_userlist_filter_by_invalid_org(rf, staff_area_administrator):
...
+    ],
+)
+def test_userlist_filter_by_role(rf, staff_area_administrator, role):
+    # first user is staff area administrator; second user is output publisher
```
mikerkelly, line 599: I think this comment is outdated now that the test is parameterised. 
```suggestion
    # Set up two users: one as the staff area administrator to request the view and
    # be filtered out, the other to be included in the response after filtering.
```

mikerkelly (approved):

tests/unit/staff/views/test_users.py
iaindillingham: Yup, we do. And yup, "global" vs "non-global" makes sense; as does "project" vs "non-project". (We also implemented, but don't actually use, `OrgMembership`-based roles, just to 🤯.)

tests/unit/staff/views/test_users.py
iaindillingham: Good spot!

iaindillingham:
> I imagine there may be other tests that want a `User` assigned an arbitrary global or non-global role?

I imagine so, too. Rather than updating `UserFactory`, we could create a new factory (e.g. `UserWithGlobalRoleFactory`) and, as you say, centralise the dependency. Thanks for the suggestion.

PR #4730, opened 2024-11-10
Remove unreachable `WorkspaceCreateForm` validation

mikerkelly:
Inspired by #4725, this PR removes more unreachable validation code in `WorkspaceCreateForm`.

The validation in question is already covered by identical checks in `__init__`, making it redundant and unreachable during the form’s normal lifecycle.

The previous test accessed it by creating an unbound form (without a `data` parameter) and then manually altering `cleaned_data` before calling `clean` directly. That's not a reasonable form lifecycle. An unbound form shouldn't invoke `clean`, nor should it be 'converted' to bound by adjusting variables post-instantiation. `cleaned_data` is public for reading purposes, not writing. Code that uses this form would not be able to reach these lines of code without being unreasonable.

### DIFF
--- a/jobserver/forms.py
+++ b/jobserver/forms.py
@@ -167,14 +167,6 @@ class WorkspaceCreateForm(forms.Form):
         if not (repo_url and branch):  # pragma: no cover
             return
 
-        repo = next(
-            (r for r in self.repos_with_branches if r["url"] == repo_url),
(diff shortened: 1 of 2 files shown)


iaindillingham (approved):
Less is more! Thanks for this 🙂

Issue #4715, opened 2024-10-30
Figure out what the version of `urllib3` should be


StevenMaude:
Following closing of #4662, we should decide to either:

* make the requirements file less misleading by removing the pinning and the associated comment, and use the latest v2 version of `urllib3`
* downgrade to the v1 of `urllib3`

The original reason `urllib3` was pinned relates to an OpenTelemetry problem in #3955. That issue may or may not have been subsequently resolved by updates in the OpenTelemetry packages.

We should check Sentry and see whether we are getting the same kind of errors as in #4662, and then decide what to do.

Note that while `urllib3` v1 still appears to be maintained (there was a release 2024-08-29), I'm not sure what the end-of-life date is for support. v2 has been out for around 18 months. Ideally, we would keep the `v2` version.

mikerkelly:
Bumping because [the issue is still occurring](https://ebm-datalab.sentry.io/issues/4807182834/?project=5443358&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=freq&statsPeriod=90d&stream_index=0), 310 times in last 24 hours, 72k times this year. 
#3955 suggests that perhaps this isn't leading to lost telemetry data but I'm not sure we're certain of that, and that would be problematic, particularly if it were systematically dropping some particular class(es) of data. 

This started Jan 4, seemingly after an update to urllib3 to 2.1.0, #3956 reverted to urllib3==1.26.18 and the issue was seemingly resolved. But now we're on [2.2.3](https://github.com/opensafely-core/job-server/blame/ea657c43cdd3a422119bb120b80d569ebbdcbb80/requirements.prod.txt#L924), I guess because the version was not pinned and Dependabot has bumped it.

Maybe we can just try reverting urllib again and pinning it. That would be quick to try. Or maybe there is some bug in our export of telemetry (to Honeycomb?) that the upgrade has revealed and we can resolve it.

StevenMaude:
All the relevant dependencies were updated in #4804.

This didn't break anything in Honeycomb, but hasn't resolved the error we get, since we still see it in Sentry.

StevenMaude:
As far as I can tell, the whole stack here is:

* nginx
* gunicorn
* Python3
* urllib3
* requests
* opentelemetry exporter/SDK

I'd add that we don't get any such problem for OpenCodelists, which has a similar stack.

StevenMaude:
[...]

PR #4804, opened 2025-01-14
Update OpenTelemetry dependency subtree

StevenMaude:
Related to #4715 — by updating the OpenTelemetry dependencies, it might be possible to have the latest urllib3 without error.

### DIFF
--- a/requirements.prod.in
+++ b/requirements.prod.in
@@ -21,11 +21,11 @@ nh3
 psycopg[binary]
 pygments
 python-ulid
-opentelemetry-exporter-otlp-proto-http==1.22.0
-opentelemetry-instrumentation-django==0.43b0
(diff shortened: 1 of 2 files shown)


StevenMaude:
See also #3892 — we have both any "old" and "new" HTTP and networking conventions configured in job-server for OpenTelemetry. I don't think anything should break with regard to that prior to a new major version of the OpenTelemetry SDK.

This PR is a minor version update.

StevenMaude:
It might be that we have to revert this, but I can review:

* the Honeycomb dashboards
* the Sentry error events

following merge to see if a revert is required.

We are using recent versions of OpenTelemetry packages in OpenCodelists, although I think there might be a bit more instrumentation in job-server, than OpenCodelists.

PR #4806, opened 2025-01-14
Revert urllib3 back to v1

StevenMaude:
This is the latest v1 release.

See #4715. Hopefully this stops the Sentry issue described there from appearing again. This is a short-to-medium term sticking plaster as at some point we'll have to switch to v2.

### DIFF
--- a/requirements.prod.in
+++ b/requirements.prod.in
@@ -35,7 +35,7 @@ social-auth-core
 structlog
 # 2.1.0 appears to be causing some problems with exporting
 # https://github.com/opensafely-core/job-server/issues/3955
-urllib3<=2.3.0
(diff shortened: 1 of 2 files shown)


StevenMaude:
To save a reader tracking back through previous PRs and issues, the context here is that we currently get "remote end closed connection" errors with the OpenTelemetry package, when exporting spans.

These apparently didn't happen previously with urllib v1 (#3956); since then, Dependabot managed to get us to update to v2, despite an attempt at pinning the urllib package version.

While still on urllib v2, we tried updating the OpenTelemetry packages in #4804, and the problem persists. So this PR will see if reverting to urllib v1 will stop the errors.

PR #4810, opened 2025-01-17
Allow release reuploads

rebkwok:
Airlock prevents users from adding an already released (via Airlock) file to a release request. This avoids unnecessary output-checking work re-reviewing files that have already been released. We may want to revisit that on Airlock later, but not just now.

However, currently Airlock only knows about files released via Airlock. If a file has been released pre-Airlock, Airlock will allow it to be added to the release request, and the job-server's check for already uploaded files prevents the release from being created.

This PR removes the check for already uploaded files, and also changes the `create_request` to a `get_or_create_request`, so we return an existing release with the same data if one matches.  It allows for the endpoint to be called by a different user (in the event that a release is created by one output-checker, and then a re-release attempt is made by another
output-checker). This can happen if not all files successfully upload from Airlock.

The create_release is wrapped in transaction.atomic, so if we found an existing matching release, we should be able to safely assume that all the request files have already been created.

### DIFF
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -302,12 +302,9 @@ class ReleaseWorkspaceAPI(APIView):
         metadata = serializer.validated_data["metadata"]
         release_id = metadata.pop("airlock_id", None)
 
-        try:
-            release = releases.create_release(
-                workspace, backend, user, files, metadata=metadata, id=release_id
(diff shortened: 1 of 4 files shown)


bloodearnest (approved):
I think this is simpler too, good change, thanks

PR #4840, opened 2025-02-06
Add copilot data in level4 api, and cleanups

### LINKED ISSUE #4837: Expose copiloting information in airlock API
This will allow us to provide better UX for copilots

bloodearnest:
This adds copiloted workspace data, so we can use that in airlock.

It also cleans up a few small issues, and completes the implementation of the staff field.

Fixes #4837

### DIFF
--- a/jobserver/api/releases.py
+++ b/jobserver/api/releases.py
@@ -569,28 +569,44 @@ class Level4AuthenticatedUser(serializers.Serializer):
     username = serializers.CharField()
     fullname = serializers.CharField(default=None, allow_blank=True)
     workspaces = serializers.DictField(default={})
+    copiloted_workspaces = serializers.DictField(default={})
     output_checker = serializers.BooleanField(default=False)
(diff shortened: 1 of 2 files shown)


Jongmassey (approved):
One q re:

>store user data in a persistant way in airlock,

What's the process for managing updates to this persisted data? I'm guessing that job server will continue to be the source of truth for user data?

bloodearnest:
> What's the process for managing updates to this persisted data? I'm guessing that job server will continue to be the source of truth for user data?

Yes, precisely.  On log in, and [every 15m whilst logged in](https://github.com/opensafely-core/airlock/blob/main/airlock/settings.py#L271), airlock will call this API, and update our user table with whatever it tells us. Our persistent storage is just a read-only cache, really.

The main motivation here is to be able to show the full name of other users in airlock,  rather than github username. See https://github.com/opensafely-core/airlock/issues/443 for more

PR #4859, opened 2025-02-12
Remove staff information from level 4 auth API

bloodearnest:
It is currently unclear how we are going to configure this, but it is definitely not related to StaffAreaAdministrator.

Removing it to avoid potential footguns.


Jongmassey (approved):
I trust that you are confident that this is not used in Airlock and on that basis this seems very sensible!

PR #4861, opened 2025-02-12
Send notifications for Airlock approved events

### LINKED ISSUE #4855: Send notifications for Airlock RequestApproved event
Airlock sends events to job-server when a release request changes status to Approved; currently nothing happens, because this state change didn't have any substantive effect for the user in Airlock.

However, now that the first part of the [release improvement](https://github.com/opensafely-core/airlock/issues/746) is done, files are released asynchronously. This means that a release request changes status to approved, and at this point the released files are pending upload. The request will not change status to Released until all files have completed uploading, so it does now make sense to add notifications for the change to approved.

Notifications should include:
- email author
- update github issue
- update slack

rebkwok:
Fixes #4855 

Airlock now releases files asynchronously, so an Approved request means that the release request has been approved and files will be uploaded in due course. Only when the last file is uploaded will the request change to Released. It therefore makes sense to send users notifications of the change of status to Approved.

This sends an email to the request author, adds a comment to the GitHub issue, and posts a notification to slack about the status change.

### DIFF
--- a/airlock/views.py
+++ b/airlock/views.py
@@ -11,6 +11,7 @@ from jobserver.models import User, Workspace
 
 from .config import ORG_OUTPUT_CHECKING_REPOS
 from .emails import (
+    send_request_approved_email,
     send_request_rejected_email,
     send_request_released_email,
     send_request_returned_email,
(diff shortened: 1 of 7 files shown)


airlock/views.py
```diff
@@ -158,12 +159,14 @@ def update_issue(airlock_event: AirlockEvent, github_api=None, notify_slack=Fals
         raise NotificationError(f"Error creating GitHub issue comment: {e}")
 
 
-def update_issue_and_slack(airlock_event, github_api=None, notify_slack=False):
+def update_issue_and_slack(airlock_event, github_api=None):
```
rebkwok, line 162: drive-by fix for unused parameter

templates/airlock/emails/request_released.txt
```diff
@@ -7,7 +7,8 @@ Dear {{ request_author }},
 Your files for workspace {{ workspace }} have been released.
 {% include "airlock/emails/includes/updates.txt" %}
 
-You can now view your <a href="{{ url }}">released files here.
+You can now view your released files here:
```
rebkwok, line 10: drive by fix for leftover html in txt template

inglesp (approved):

Issue #4852, opened 2025-02-11
The pipeline dependency update workflow is not updating the pinned `requirements.txt` file


StevenMaude:
It only updates the `.in` file.

Example: 68a45b78d7bf99df187cb59dc2558b9de4e3dbb7

(I noticed this on merging a subsequent update that then did update the `.txt` file.)

PR #4891, opened 2025-02-18
Fix automatic pipeline upgrade

bloodearnest:
Before, it just updated .in, now should also update .txt

Fixes: #4852

### DIFF
--- a/justfile
+++ b/justfile
@@ -110,7 +110,7 @@ upgrade env package="": virtualenv
     FORCE=true {{ just_executable() }} requirements-{{ env }} $opts
 
 # upgrade our internal pipeline library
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -110,7 +110,7 @@ upgrade env package="": virtualenv
     FORCE=true {{ just_executable() }} requirements-{{ env }} $opts
 
 # upgrade our internal pipeline library
-upgrade-pipeline:
+upgrade-pipeline: && requirements-prod
```
Jongmassey, line 113: TIL you could do that with just recipes.

Jongmassey (approved):
:+1:

PR #4896, opened 2025-02-19
Remove staff "create interactive project" option

### LINKED ISSUE #4867: Remove option to “Create an Interactive Project” in the Staff Area
And remove any associated code.

[Link to Projects: Staff Area](https://jobs.opensafely.org/staff/projects/)

StevenMaude:
Fixes #4867.

We're in the process of removing OpenSAFELY Interactive.

This is a self-contained commit that removes the option for staff to
create Interactive projects, along with the associated tests, while
still retaining the other remaining functionality.


templates/staff/project/list.html
```diff
@@ -15,12 +15,7 @@
 {% endblock breadcrumbs %}
 
 {% block hero %}
-{% url "staff:project-create" as staff_project_create_url %}
-{% #staff_hero title="Projects" %}
-  {% #button type="link" href=staff_project_create_url variant="primary" %}
-    Create an Interactive Project
-  {% /button %}
-{% /staff_hero %}
```
Jongmassey, line 23: Is this needed to close the `staff_hero` ?

Jongmassey (approved):
One minor Q could do with a look

templates/staff/project/list.html
StevenMaude: I thought that at first and had opening and closing template tags, but [it appears this is how slippers works](https://mitchel.me/slippers/docs/using-components/#block-and-inline-component-syntax).

Similar example in our code: https://github.com/opensafely-core/job-server/blob/3c25b965d6e6278333a3280d659cbee775b39eb4/templates/staff/redirect/list.html#L16-L18

Issue #4886, opened 2025-02-13
Remove `AnalysisRequestList`


StevenMaude:
## When will we know it's done?

Issue #4881, opened 2025-02-13
Remove ability to submit Interactive analysis requests


StevenMaude:
## Why are we doing this?

We're removing the OpenSAFELY Interactive functionality, and this is only used there.

## What are we doing?

* Remove `AnalysisRequestCreate` from `interactive/views.py`
* Remove `AnalysisRequestForm` from `interactive/forms.py`
* Remove `interactive/submit.py`
* Remove `interactive/management/osi_run.py`
* Remove `interactive/management/osi_run_and_release.py`
* Remove the URL path matching `AnalysisRequestCreate`
* Remove `notify_analysis_request_submitted` from `interactive/slacks.py`
* Remove `build_codelist()` and `from_codelist()`
* Remove `templates/interactive/analysis_request_create.html`
* Consider removing references in test JSON fixtures

## What needs to be done before this?

#4879.

## When will we know it's done?

It will not be possible to run or submit analysis requests.

Issue #4870, opened 2025-02-12
Remove references to Interactive analyses from Workspace templates


tomodwyer:
Remove any buttons or cards from these templates.

Blocked on #4886

StevenMaude:
May be blocked on #4888 and #4889.

PR #4898, opened 2025-02-19
Remove analysis_request_create permission

### LINKED ISSUE #4875: Remove the `analysis_request_create` permission
## Why are we doing this?

It may be easier to do this after removing other `AnalysisRequest` related code (see #4880).

## What are we doing?

* Remove the use of `analysis_request_create` from `dispatch()` in `interactive/views.py`, if `AnalysisRequestCreate` has not already been removed
* Remove the use of `analysis_request_create` from `WorkspaceAnalysisRequestList` in `jobserver/views/workspaces.py`, if that class has not already been removed
* Remove the use of `analysis_request_create` from `WorkspaceCreate` in `jobserver/views/workspaces.py`
* Remove the `analysis_request_create` permission from `jobserver/authorization/permissions.py

When `analysis_request_create` is no longer found in the job-server code.

Jongmassey:
Fixes #4875

### DIFF
--- a/interactive/forms.py
+++ b/interactive/forms.py
@@ -61,7 +61,7 @@ class AnalysisRequestForm(forms.Form):
             choices=codelist_type_choices
         )
 
-    def clean(self):
+    def clean(self):  # pragma: no cover
         cleaned_data = super().clean()
 
         time_ever = cleaned_data.get("time_ever")
--- a/interactive/views.py
+++ b/interactive/views.py
@@ -12,7 +12,7 @@ from interactive_templates.dates import END_DATE, START_DATE, WEEK_OF_LATEST_EXT
 from interactive_templates.schema import Codelist, v2
 
 from jobserver.authorization import has_permission, permissions
-from jobserver.models import Backend, Project
+from jobserver.models import Backend
 from jobserver.opencodelists import _get_opencodelists_api
 from jobserver.reports import process_html
 from jobserver.utils import build_spa_base_url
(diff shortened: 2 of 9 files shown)


mikerkelly (approved):
Thanks, Jon.

This has left some views in a strange state where they exist but aren't accessible. I can see that if we've removed the permissions check we don't want to allow the views to be exercised, but I think it would have been preferable to do the tickets that remove the views this touches first, then have this ticket remove any vestiges of the permission. I think that would have been clearer and less work. As things stand this is fine though, let's proceed and ideally do the tickets to remove those views next.

#4886 removes WorkspaceAnalysisRequestList so I think we should work on that soon as this PR removes the permission that makes it make sense.

@StevenMaude is currently working on #4881 which removes AnalysisRequestCreate. Steve, you might want to check for any conflicts when this is merged.

Small point, your commit changing the business code that doesn't pass CI is not really an atomic commit as it relies on other commits to fix its tests and coverage. If you roll back to the first commit it fails CI checks. I think they all belong together. Atomic commits are, in order of importance, self-contained, coherent, and small. Separately, this work could have been broken into smaller atomic commits based on which part of the business code was touched, but I don't think that was necessary.

lucyb:
I agree with the comments above. Since the PR was opened, the AnalysisRequestCreate view has been removed and I expect we're going to remove the Analysis Request List view very shortly, and others that this change touches. Given that Jon is on leave at the moment, I'm going to close the PR for now. If any parts are still relevant when he returns, he can always reopen it then.

Jongmassey:
This is still blocked by #4870 . Once this ticket is complete, a rebase should allow this to proceed.

Jongmassey:
This is still blocked by #4870 . Once this ticket is complete, a rebase should allow this to proceed.

PR #4899, opened 2025-02-20
Add `justfile` recipes to make and apply migrations

mikerkelly:
These tasks need to run fairly often -- whenever making model changes or pulling migrations.

### DIFF
--- a/justfile
+++ b/justfile
@@ -171,8 +171,18 @@ check: format django-upgrade lint
 
 check-migrations: devenv
     $BIN/python manage.py makemigrations --dry-run --check \
-    || echo "There is model state unaccounted for in the migrations, run python manage.py migrate to fix."
+    || echo "There is model state unaccounted for in the migrations, run `just migrate` to fix."
 
+# generate migrations for any model changes
+make-migrations: devenv
+    $BIN/python manage.py makemigrations
+
+# apply any unapplied migrations
+apply-migrations: devenv
+    $BIN/python manage.py migrate
+
+# generate migrations and apply unapplied ones
+migrate: make-migrations apply-migrations
 
 # fix formatting and import sort ordering
 fix: devenv


justfile
```diff
@@ -171,8 +171,19 @@ check: format django-upgrade lint
...
+
+make-migrations: devenv
+    $BIN/python manage.py makemigrations
+
+# This both checks model changes for migrations and runs migrate. That's what
+# devs want to do when changing models or when there are new migrations to
+# apply from a pull. It's convenient that it's a single command with a short name.
+# In the rare case you want to makemigrations without running them (for
+# example, to check or to edit them first), you can use make-migrations or run
+# manage.py commands manually.
```
StevenMaude, line 184: thought: It's maybe not immediately obvious, but when you run only `just` with a `justfile`, it helpfully shows the recipes and the comment line immediately preceding the recipe's contents.

Because of that, in that context, you lose the most helpful part of the comment which is

> … checks model changes for migrations and runs migrate

and instead see:

```
$ just
Available recipes:
…
migrate                       # manage.py commands manually.
…
```

The fragment shown is incomplete.

There are a few ways around this.

You could rework the way the comment is organised so that the one-line comment becomes something like:

```
Check model changes for migrations and then run migrate
```

This could be by either rewriting the comment so that the last line is:

> Checks model changes for migrations and runs migrate.

or by moving the more contextual details of the comment *inside* the recipe, leaving the summary as the preceding comment.

There is a cleaner option using the [`[doc]` option in `just`](https://github.com/casey/just/blob/7c63fb63b1517a95de950443d562cc71576ccc71/README.md#L1302-L1308). With `[doc]` you can mark that the preceding comment line doesn't apply, and then pass what actually should appear to `[doc()]`.

That does require everyone working on the project to be on v1.27 of `just` or higher; otherwise, `just` errors.

mikerkelly: > Everything looks fine; there's a little tidying up that can be done to make `just` show a helpful summary comment.

Thanks a lot, Steve. I'm not that familiar with justfiles and was unaware of this.

I've addressed your comment by adding "docstrings" and a separate recipe for making, applying, and both.

Issue #4889, opened 2025-02-17
Export all Analysis Reports from "real" projects


lucyb:
# Why are we doing this?

We have said that the details of all reports produced for the two research projects will be exported and stored as a spreadsheet in a secure section of our Google Drive. This allows us to recover them in case anybody ever needs that.

# How will we know when it's done?

See [this thread for details](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1739869019842079)

# What are we doing?

See [this thread for details](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1739869019842079)

mikerkelly:
I've put some thoughts on how to store this data in [this document](https://docs.google.com/document/d/1XlF-LF59OXRxBDC1o4mbW6_PPQ5CAu9LiP8wa0keiM4/edit?usp=sharing).

PR #4903, opened 2025-02-20
Remove ability to edit Interactive reports

### LINKED ISSUE #4883: Remove `ReportEdit` feature
## Why are we doing this?

## What are we doing?

* Remove `ReportEdit` from `interactive/views.py` 
* Remove `AnalysisRequestEditForm` from `interactive/forms.py` (and probably the entire `forms.py` module)
* Remove `templates/interactive/report_edit.html`
* Remove `report-edit` URL from `interactive/urls.py`
* Remove `get_edit_url()` from `AnalysisRequest` and from `templates/interactive/analysis_request_detail.html`
* …

Probably nothing, but #4884 may be sensible first.

When an Interactive user can no longer edit a report.

StevenMaude:
Fixes #4883.

We're removing the OpenSAFELY Interactive functionality from job-server.

This is a self-contained commit that removes the report editing feature from OpenSAFELY Interactive.


StevenMaude:
I think this should be OK to merge, but it might be better to wait for #4889 to be complete, in case anything needs editing on the existing reports.

Jongmassey (approved):

PR #4905, opened 2025-02-20
Remove ability to submit Interactive analysis requests

### LINKED ISSUE #4881: Remove ability to submit Interactive analysis requests

StevenMaude:
Fixes #4881.


jobserver/views/projects.py
```diff
@@ -82,10 +82,6 @@ def get(self, request, *args, **kwargs):
                 key=operator.itemgetter("name"),
             )
 
-        is_interactive_user = has_permission(
-            request.user, permissions.analysis_request_create, project=project
-        )
-
```
mikerkelly, line 88: I'm not sure I understand why the ProjectDetail view would change in this ticket. It doesn't seem relevant to the description. Did you mean to include it?

jobserver/views/workspaces.py
```diff
@@ -244,7 +244,6 @@ def get(self, request, *args, **kwargs):
         is_interactive_user = has_permission(
             request.user, permissions.analysis_request_create, project=workspace.project
         )
-        show_interactive_button = is_interactive_user and workspace.is_interactive
```
mikerkelly, line 247: I'm not sure I understand why this has changed. Can't there still be interactive users and projects in the DB at this point, we're just removing the views to create them?

mikerkelly (approved):
Thanks, Steve, mostly looking good. I just wonder if a couple of the commits were meant to be included here as I don't see the direct relevance?

jobserver/views/workspaces.py
StevenMaude: Yes, you're correct that we still have Interactive users and projects still.

As far as I can tell, the only use of `show_interactive_button` is in `templates/workspace/detail.html` to show a button to run an interactive analysis.

Here's the relevant part of the current `main` version of that template: https://github.com/opensafely-core/job-server/blob/8c177d0159cba9d76f5cac7484a987a791d7d665/templates/workspace/detail.html#L105-L117

If we don't remove this, and the associated part of the template, the button would still display, but be broken. (Whether in reality any projects meet the conditions to display a functioning button, I'm not entirely sure.)

Otherwise, we could leave this and come back to it in a final clean-up pass.

StevenMaude:
@mikerkelly: thanks for the thorough questioning! It made me go back and double-check.

Also:

> I also wonder if we should get rid of this now?
> 
> ```
> tests/unit/services/sentry_event.json:                "self": "<interactive.views.AnalysisRequestCreate object at 0x115ba19d0>",
> ```

I thought the same, but wasn't exactly sure what might be sensible to replace it with, if we want to ideally avoid having to update it again in future.

It should also be mentioned that there's actually a (albeit already public) user email in there too.

PR #4913, opened 2025-02-27
Remove use of interactive-templates functionality

### LINKED ISSUE #4856: Remove the inclusion of the `interactive-templates` package in job-server's repository
Remove the use of the `interactive-templates` package from job-server's repository

## Why are we doing this?

We're removing the OpenSAFELY Interactive functionality, and `interactive-templates` is only used there.

## What are we doing?

* Noting the amendment to ADR 0013
* Removing the `just update-interactive-templates` recipe
* Removing interactive-templates from the Python requirements
* Removing the `.github/workflows/update-interactive-templates.yml` workflow
* Editing `DEVELOPERS.md` to remove the reference to interactive-templates
* Archiving the interactive-templates repository when no longer included in job-server

Removing the import of `interactive_templates` from the following modules, or the modules entirely:

* `interactive/commands.py`
* `interactive/management/commands/osi_run.py`
* `interactive/submit.py`
* `interactive/views.py`

Removing the import of `interactive_templates` from the following *test* modules, or the modules entirely:

* `tests/integration/test_interactive.py`
* `tests/unit/interactive/conftest.py`
* `tests/unit/interactive/test_submit.py`
* `tests/unit/interactive/test_views.py`

It looks in all cases that this can be done cleanly; that is, all this code is only relevant to the OpenSAFELY Interactive code functionality.

*Edit: I think completing #4881 is now sufficient. When the associated PR (#4905) is merged, the uses of `interactive_templates` are removed.*

When:

* the only references in the source code to interactive-templates are in the `sentry_event` test fixtures (and arguably these could be removed or rebuilt to remove those references).
* the ADRs are updated
* the interactive-templates repository is archived

StevenMaude:
Fixes #4856.

We're removing the OpenSAFELY Interactive functionality, and this
imported package is no longer used.


docs/adr/0028-remove-the-use-of-interactive-templates-repo.md
StevenMaude, line 1: question: I'm not sure what the best thing to do is regarding the existing ADR that introduced the use of the interactive-templates repository:

* Is a new ADR necessary for this change?
* Do we also need to revoke the change the corresponding ADR (13) to have a status of "Revoked" or similar?

tomodwyer (approved):
Approved the code, waiting on that ADR change and the conflict.

docs/adr/0028-remove-the-use-of-interactive-templates-repo.md
StevenMaude: Done.

Issue #4860, opened 2025-02-12
Remove use of `attrs` entirely


StevenMaude:
## Why are we doing this?

It removes a third-party dependency entirely. We previously identified it for removal (#4532), but it was blocked by our transitive dependency on `attrs` through `interactive-templates`

## What needs to be done before this?

* Removing the use of the `interactive-templates` package (#4856).
* Removing the use of `pytest-subtests` (#4834).

## What are we doing?

* Removing the single use of `attrs` in `jobserver/runtime.py` ([version at time of writing](https://github.com/opensafely-core/job-server/blob/46eec14c69763b2fc2f7bd1e4401d4719e2c683c/jobserver/runtime.py#L1))
* Removing attrs from our Python `requirements.*.in` and `requirements.*.txt` files

## When will we know it's done?

When `attrs` is no longer included in our Python requirements.

StevenMaude:
Awaiting merge of #4915.

PR #4915, opened 2025-02-27
Remove use of `pytest-subtests`

### LINKED ISSUE #4834: Remove pytest-subtests
This package is used in [one test](https://github.com/opensafely-core/job-server/blob/b1b1d0944137ca4c19b627feaee7cc2be2530f05/tests/unit/jobserver/test_icons.py) currently.

Can we use `pytest.mark.parametrize` to drop the dependency?

StevenMaude:
Fixes #4834.

Refactor the single collection of tests that used `pytest-subtests` and use `parametrize()` instead.

If we used the package more, maybe we could justify leaving it be, but it seems a good candidate for dependency removal.

It also unlocks removing `attrs` (#4860).


tests/unit/jobserver/test_icons.py
```diff
@@ -2,29 +2,48 @@
...
 
+import pytest
 
-def test_all_icons(subtests):
-    """Ensure icon tag names defined in templates are valid tags"""
+
+# Logically, this feels like it could be a pytest.fixture.
+# However, it's not currently possible to use fixtures with parametrize:
+# https://github.com/pytest-dev/pytest/issues/349
+def icons():
```
mikerkelly, line 11: This code is broken as it doesn't return the right number of icons. This issue pre-exists this PR but we should fix it, and other issues with this function, now. Cross-check outside this code:

```
~/job-server$ grep -roh "{% icon_\w\+" templates/**/*.html | sed 's/{% //g' | sort | uniq | wc -l
39
```

There should be 39 icon matches, but only 14 are found for the tests (14 subtests). I think the main issue is using `search`, which as used here only returns the first result per-file. Probably it should be using `finditer` or `findall` instead. I've got several other issues with how this code is written and how readable that is that means I'm not sure that's the only issue affecting the accuracy, and I think we can just fix those other issues anyway, while also making it easier to see that the code is correct. Most of these issues pre-date this PR.

- The use of `itertools` seems totally unnecessary, and harder to understand than a simpler approach.
- The templates are all read twice, which is unnecessary. The regex matching is almost all of the run time, though. This won't be the case if we drop the first pass that filter down to `icon_templates` which is stated to be been done to simplify the later `itertools` code, but it doesn't seem so simple that it's easy to verify its accuracy.
- Some of the list comprehensions seem pointless.
- Some of the variable names aren't very descriptive.
- The function name could be more useful.
- Could use a docstring.

I can't make suggestions with the GitHub UI as the changes have removed some lines, but I think the following with `findall` and nested generator expressions is simple, clean, at least as time- and memory- efficient, and gets all the icons (I've checked the output against the grepped output):

```python

def find_icon_tag_names_used():
    """Find the name of template tags starting with `icon_` used in our templates."""
    icon_pattern = re.compile(r"{% (icon_[^\s]+)")
    template_paths = list(Path("templates").rglob("*.html"))
    return sorted(
        set(
            icon_tag_name
            for path in template_paths
            for icon_tag_name in icon_pattern.findall(path.read_text())
        )
    )
```

tests/unit/jobserver/test_icons.py
```diff
@@ -2,29 +2,48 @@
...
+    # Look for the start of an icon templatetag and capture the full name.
     icon_pat = re.compile(r"{% (icon_[^\s]+)")
 
-    # get a set of icon names in the Django templates
-    icons = set(
-        itertools.chain.from_iterable(
-            icon_pat.search(t.read_text()).groups() for t in icon_templates
+    # Get a sorted list of non-duplicated icon names in the Django templates.
+    # We use a sorted list instead of only a set
+    # to make this test work with parametrize and xdist:
```
mikerkelly, line 23: We can be more precise here and help the reader understand the point:
```suggestion
    # Use a sorted list so that the different xdist workers get a consistent ordering.
```

tests/unit/jobserver/test_icons.py
```diff
@@ -2,29 +2,48 @@
...
-    # it for each icon
-    config = Path("templates/components.yaml").read_text()
-    for icon in icons:
-        # use subtests to avoid failing the test at the first broken icon
-        with subtests.test(icon=icon):
-            assert icon in config
+    return icons
+
+
+@pytest.fixture(scope="module")
```
mikerkelly, line 36: I'm not sure using a fixture over a module-level constant gains us anything here and it seems slightly harder to explain? Not a strong opinion.

tests/unit/jobserver/test_icons.py
```diff
@@ -2,29 +2,48 @@
...
+def component_config():
+    # This is scoped to ideally to read the component config contents once,
+    # so we don't have to read it for each icon.
```
mikerkelly, line 39: I think "to ideally" is stray text?

Not sure this comment is needed at all. I think the intent is fairly clear.

mikerkelly (commented):
Thanks, Steve. I think this is a good approach and appreciate the push to remove unnecessary dependencies.

I think there's one issue that predates this PR that means a lot of the icons aren't actually tested. We should fix that. I've made several other mostly stylistic and documentation suggestions (again many addressing issues pre-dating this PR) that I think would be improvements but feel free to push back, they're subject to taste.

In general, I would suggest verifying tests against some outside source to make sure they're doing what you think.

I'm not sure I've seen the slower setup speed you mention in the commit comment. If you pass the `--durations=5` parameter you can see the setup time for just the first test doesn't change much compared to main, I think.

tests/unit/jobserver/test_icons.py
StevenMaude: Thanks; that's really helpful and more thorough than what I'd  done (ensuring the tests themselves worked as they were). I've incorporated these changes and added you as co-author of the commit.

tests/unit/jobserver/test_icons.py
StevenMaude: I've left this because it semantically feels like a fixture, and noting that perhaps makes its reason for inclusion more immediately obvious.

But I similarly don't have strong opinions, particularly as the test code here is not very long.

tests/unit/jobserver/test_icons.py
```diff
@@ -1,30 +1,38 @@
...
-        itertools.chain.from_iterable(
-            icon_pat.search(t.read_text()).groups() for t in icon_templates
+# Logically, this feels like it could be a pytest.fixture.
+# However, it's not currently possible to use fixtures with parametrize:
+# https://github.com/pytest-dev/pytest/issues/349
+def find_icon_tag_names_used():
+    """Find the name of template tags starting with `icon_` used in our templates."""
+    icon_pattern = re.compile(r"{% (icon_[^\s]+)")
+    template_paths = list(Path("templates").rglob("*.html"))
+    # Get a sorted list of non-duplicated icon names in the Django templates.
```
mikerkelly, line 14:
```suggestion
    # Get a sorted list of all of the icon names used in the Django templatetags.
```

"non-duplicated" could be interpreted as "occurring exactly once" which is not what we mean here.

tests/unit/jobserver/test_icons.py
```diff
@@ -1,30 +1,38 @@
...
+    # Note: this fixture does not get reused with pytest-xdist,
+    # so the file is read once per-worker, but this is very fast anyway.
+    # https://github.com/pytest-dev/pytest-xdist/issues/783
+    return Path("templates/components.yaml").read_text()
+
+
+@pytest.mark.parametrize("icon_tag_name", find_icon_tag_names_used())
+def test_icon_tag_names_config(icon_tag_name, component_config):
+    """Ensure icon tag names defined in templates appear in the icon component
+    configuration file"""
```
mikerkelly, line 37:
```suggestion
    configuration file."""
```
For punctuation consistency.

mikerkelly (approved):
Thanks, Steve. Looks good now. Two more really tiny and pedantic comments comments.

tests/unit/jobserver/test_icons.py
StevenMaude: Thanks; I think this makes it clearer. Have included this and the other suggestion in the single commit.

tests/unit/jobserver/test_icons.py
StevenMaude: Done.

StevenMaude:
@mikerkelly: thanks again for the thorough review and suggestions :eyes: — this part of the code is now considerably better for it.

PR #4917, opened 2025-02-27
Remove use of `attrs`

### LINKED ISSUE #4860: Remove use of `attrs` entirely

StevenMaude:
Fixes #4860.

### DIFF
--- a/jobserver/runtime.py
+++ b/jobserver/runtime.py
@@ -1,16 +1,11 @@
-from attrs import define, field
+from dataclasses import dataclass
 
 
-def less_than_60(instance, attribute, value):
-    if value > 59:
-        raise ValueError(f"{attribute} should be less than 60")
-
-
-@define
+@dataclass(frozen=True)
 class Runtime:
     hours: int = 0
-    minutes: int = field(default=0, validator=[less_than_60])
-    seconds: int = field(default=0, validator=[less_than_60])
+    minutes: int = 0
+    seconds: int = 0
     total_seconds: int = 0
 
     def __bool__(self):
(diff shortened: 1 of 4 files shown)


StevenMaude:
This needs #4913 and #4915 merging, then rebasing.

jobserver/runtime.py
```diff
@@ -24,3 +19,15 @@ def __str__(self):
             return "-"
 
         return f"{self.hours:02d}:{self.minutes:02d}:{self.seconds:02d}"
+
+    def _less_than_60(self, attribute):
+        if getattr(self, attribute) > 59:
+            raise ValueError(f"{attribute} should be less than 60")
```
StevenMaude, line 25: thought: We could consider validating that values aren't negative, but that would also apply to `hours`.

mikerkelly: I think checking that minutes and seconds are in the interval [0,59] and hours is greater than zero would be good things to do, with tests around the upper and lower bounds.

It's totally up to you whether you feel like doing that in this PR or not.

I also can't get too excited about it because I would much rather this class was representing time with library classes for representing times, rather than our own implementation. But that also doesn't need to happen in this PR.

mikerkelly (approved):
Looks good. Thanks for extirpating another dependency.

PR #4920, opened 2025-03-01
Use organisation slug for Airlock notification lookups

inglesp:
Fixes opensafely-core/airlock#854.

### DIFF
--- a/tests/unit/airlock/test_views.py
+++ b/tests/unit/airlock/test_views.py
@@ -2,10 +2,13 @@ from unittest.mock import patch
 
 import pytest
 
+from airlock.config import ORG_OUTPUT_CHECKING_REPOS
 from airlock.views import AirlockEvent, EventType, airlock_event_view
 from tests.factories import (
     BackendFactory,
     BackendMembershipFactory,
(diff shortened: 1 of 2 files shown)


tests/unit/airlock/test_views.py
```diff
@@ -136,12 +141,57 @@ def test_api_post_release_request_custom_org_and_repo(mock_create_issue, api_rf)
     ]
 
 
+@patch("airlock.views._get_github_api", FakeGitHubAPI)
+@patch("airlock.views.create_output_checking_issue")
+def test_api_post_release_request_non_default_org_and_repo(mock_create_issue, api_rf):
```
rebkwok, line 146: It might be worth using the `settings` fixture and defining a test-specific org in `ORG_OUTPUT_CHECKING_REPOS`. Otherwise this test is dependent on the actual settings for bristol and it'll break if they ever change their output checking repo

rebkwok (approved):

tests/unit/airlock/test_views.py
inglesp: Good point.

PR #4929, opened 2025-03-04
Create a default issue template

lucyb:
This is intended to make it easier for us to create issues that are inline with the delivery tasks guidance and to help ensure we have the information we need when picking up the issue.

### DIFF
--- a/.github/ISSUE_TEMPLATE/default.md
+++ b/.github/ISSUE_TEMPLATE/default.md
@@ -1,0 +1,23 @@
+---
+name: Default issue
+about: Create a standard delivery task
+title: ""
+labels: ""
+assignees: ""
+---
+
+# Why are we doing this?
+
+...
+
+# How will we know when it's done?
+
+...
+
+# What are we doing?
+
+...
+
+---
+
+[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)


.github/ISSUE_TEMPLATE/default.md
StevenMaude, line 1: thought: Apparently [Markdown issue templates with YAML frontmatter](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/manually-creating-a-single-issue-template-for-your-repository) are now "legacy", and [YAML issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) are the newer way of doing this.

I don't think that particularly matters too much here, right now, though, and having a working template is better than not having one.

StevenMaude (approved):
Thanks for this; it was on my list to do, and kept getting bumped down.

.github/ISSUE_TEMPLATE/default.md
lucyb: Good point. Although, I'm not sure about calling them legacy just yet. The YAML issue forms are still in public preview, although I expect they'll become the default over time. I didn't particularly want to use textareas and as far as I could see having a block of markdown that persisted when the issue was saved wasn't possible. Maybe textareas would have been okay? I will leave it as-is for now though.

Issue #4872, opened 2025-02-12
Remove ability to request publication of Interactive reports


StevenMaude:
## Why are we doing this?

We're removing the OpenSAFELY Interactive functionality, and this is part of that.

## What needs to be done before this?

Possibly #4889.

## What are we doing?

* Remove the URL pattern matching `/publish/` from `interactive/urls.py` 
* Remove `get_publish_url()` from `models.py`
* Remove `PublishRequestCreate` from `jobserver/views/reports.py`
* Remove the templates relating to publishing:
  * `templates/interactive/publish_request_create.html`
  * `templates/interactive/publish_request_create_locked.html`
* Remove `notify_copilots_of_publish_request` in `jobserver/slacks.py`
* Remove any associated tests
* TODO:
  * Can we remove `publish_request` from `interactive/models.py`?
* TODO: see if any more is needed in this section …
  
## When will we know it's done?

When an Interactive user can no longer request to publish an Interactive report.

PR #4934, opened 2025-03-05
Remove Interactive from Releases API

### LINKED ISSUE #4877: Remove Interactive from Releases API
## Why are we doing this?

## What are we doing?

* Removing `is_interactive_report()` from `jobserver/api/releases.py`, and the code that conditionally runs following a successful check
* Removing `send_report_uploaded_notification()` from `interactive/emails.py`
* Removing `notify_report_uploaded()` from `interactive/slacks.py`
* Removing `templates/emails/notify_report_uploaded.txt` and `templates/emails/notify_report_uploaded.html`
* Remove `interactive/management/commands/create_testing_analysis.py`
* Remove `create_report` from `interactive/commands.py`

TODO.

When there is no longer an import of `interactive` in `jobserver/api/releases.py` and the other associated code and templates are removed.

StevenMaude:
Fixes #4877.


StevenMaude:
Might be better waiting for #4872, perhaps? So that it's no longer possible for a user to create a publish request.

lucyb (approved):

Issue #4858, opened 2025-02-12
Remove the use of the OpenSAFELY Interactive GitHub token


StevenMaude:
## Why are we doing this?

We're removing the OpenSAFELY Interactive functionality, and this authentication is only used there.

## What needs to be done before this?

* Removing use of `interactive/submit.py` (probably #4881).

## What are we doing?

* Removing the use of `INTERACTIVE_GITHUB_TOKEN` everywhere, currently in:
  * `docker/docker-compose.yaml`
  * `.github/workflows/main.yml`
  * `dotenv-sample`
  * `jobserver/settings.py`
* Updating `tests/unit/services/test_sentry.py` to a different environment variable other than `INTERACTIVE_GITHUB_TOKEN` for testing
* Removing the `INTERACTIVE_GITHUB_TOKEN` from the Dokku deployment
* Revoking the corresponding token in GitHub
* Removing the "Rotating the OSI GitHub token" section from `DEVELOPERS.md` ([current version at time of writing](https://github.com/opensafely-core/job-server/blob/46eec14c69763b2fc2f7bd1e4401d4719e2c683c/DEVELOPERS.md#L352-L359))
* Removing the details of `opensafely-interactive-token` from the tech support playbook in the team manual
* Updating or revoking ADR 0014 ([version at time of writing](https://github.com/opensafely-core/job-server/blob/46eec14c69763b2fc2f7bd1e4401d4719e2c683c/docs/adr/0014-osi-token.md))
* Removing the entry for `INTERACTIVE_GITHUB_TOKEN` in `services/sentry.py`, **only once the token has been removed from all use and revoked**

## When will we know it's done?

When:

* there are no references to `INTERACTIVE_GITHUB_TOKEN` in the job-server repository, except any relevant ADRs
* ADR 0014 has been updated or revoked
* the job-server deployment no longer has this token in its environment
* the existing `INTERACTIVE_GITHUB_TOKEN` has been revoked
* the team manual no longer refers to this token

StevenMaude:
I've removed `INTERACTIVE_GITHUB_TOKEN` from the deployment.

[...]

PR #4937, opened 2025-03-06
Remove most references to Interactive GitHub token

StevenMaude:
Relates to #4858.

See the commit message for more explanation.

### DIFF
--- a/docs/adr/0014-osi-token.md
+++ b/docs/adr/0014-osi-token.md
@@ -2,7 +2,7 @@
 Date: 2023-03-24
 
 ## Status
-Accepted
+Deprecated
 
 ## Context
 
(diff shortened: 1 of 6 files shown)


docs/adr/0014-osi-token.md
StevenMaude, line 1: thought: If it's preferred, I can write a brief ADR to explain removal of the token and the deprecation of the existing ADR. Although I think it was mentioned in planning that maybe some of the changes might just be covered in an ADR that goes into more detail about removing OpenSAFELY Interactive.

tomodwyer: I think cover it in the main ADR, not worth it for such a small change!

tomodwyer (approved):
Sentry stuff makes sense in the commit message.

PR #4940, opened 2025-03-07
Remove interactive AnalysisRequestDetail view

### LINKED ISSUE #4885: Remove `AnalysisRequestDetail`
## Why are we doing this?

## What are we doing?

* Remove `AnalysisRequestDetail` from `staff/views/analysis_request.py`
* Remove the `templates/staff/analysis_request/detail.html` template
* Remove `AnalysisRequestDetail` from `interactive/views.py`
* Remove the `templates/interactive/analysis_request_detail.html` template
* Remove the corresponding URL for the views
* Remove `jobserver/reports.py` since `process_html()` should only be used for a detail page 

May be blocked on #4888 and #4889.

#4886 might be useful to do first.

When users and staff members cannot view details of analysis requests.

KatieB5:
Closes #4885 

Removed template, URLs, and get_absolute_url().


tomodwyer (approved):
Paired on this 🚀

PR #4945, opened 2025-03-10
Better airlock notification error handling

rebkwok:
This improves the airlock notifications so that when there are multiple notification actions (github/email/slack), an error in one action doesn't prevent others happening.

### DIFF
--- a/airlock/slack.py
+++ b/airlock/slack.py
@@ -3,9 +3,16 @@ from services import slack
 from .config import ORG_SLACK_CHANNELS
 
 
-def post_slack_update(org, update_url, issue_title, update_text):
+def post_slack_update(org, update_url, issue_title, update_text, github_error):
     channel = ORG_SLACK_CHANNELS.get(org, ORG_SLACK_CHANNELS["default"])
 
-    title_link = slack.link(update_url, f"{issue_title} has been updated")
-    message_text = f"{title_link}\n{update_text}"
+    if update_url is not None:
+        title = slack.link(update_url, f"{issue_title} has been updated")
+    else:
+        title = f"{issue_title} has been updated"
+    message_text = f"{title}\n{update_text}"
+
+    if github_error:
+        message_text += f"\nA GitHub error occurred: {github_error}"
+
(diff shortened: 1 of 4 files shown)


tests/unit/airlock/test_views.py
```diff
@@ -281,6 +281,53 @@ def test_api_airlock_event_error(api_rf, event_type, updates, error):
...
+    BackendMembershipFactory(backend=backend, user=user)
+
+    # A request approved event sends 3 notifications:
+    # - sends an email that includes the release URL,
+    # - updates the github issue
+    # - posts to slack
+    # This notification attempt will error on sending the email, because
+    # we haven't created a Release for it, and it'll error creating the
+    # GitHub issue update because we've mocked it that way. All events will
+    # be attempted, and the two errors will be reported in the response
```
inglesp, line 322: It feels slightly brittle to have this test depend on us not setting up the database in the way that we would expect to find it in production.  Is it feasible to mock both errors?

airlock/slack.py
```diff
@@ -3,9 +3,18 @@
...
-    message_text = f"{title_link}\n{update_text}"
+    if update_url is not None:
+        title = slack.link(update_url, f"{issue_title} has been updated")
+    else:
+        title = f"{issue_title} has been updated"
+    message_text = f"{title}\n{update_text}"
+
+    if github_error:
+        message_text += (
+            f"\nA GitHub error occurred, please contact tech-support: {github_error}"
```
inglesp, line 17: What can tech-support do here?

inglesp (approved):

tests/unit/airlock/test_views.py
rebkwok: yes

airlock/slack.py
rebkwok: Mostly just check what went wrong? I can remove this if we think it's not worth telling people to contact tech support. They probably will anyway, if they see an error reported

inglesp: OK.  No strong opinion from me.

PR #4950, opened 2025-03-11
Remove Interactive login form

### LINKED ISSUE #4866: Remove Interactive login form
Remove OpenSAFELY Interactive Login form and associated code, including any changes to the user model.

StevenMaude:
Fixes #4866.

See the commit message for details.

### DIFF
--- a/templates/login.html
+++ b/templates/login.html
@@ -23,7 +23,7 @@
       </p>
   </div>
 
-  <div class="mb-6 pb-12 border-b border-b-slate-300">
+  <div class="mb-6 pb-12">
       {% #card title="GitHub Login" container=True %}
       <div class="prose prose-sm mb-6">
           <p>If you do not already have a GitHub account, then you will need to create one before signing in.</p>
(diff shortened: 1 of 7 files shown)


StevenMaude:
I think this is roughly there, but could do with:

* tidying up the changes and commit message(s)
* checking the expected behaviour in browser (particularly as we don't have browser tests)
* reviewing if anything else can be removed following these changes

templates/login.html
```diff
@@ -23,7 +23,7 @@ <h1 class="text-3xl break-words font-bold text-slate-900 mt-2 md:mt-0 md:col-spa
       </p>
   </div>
 
-  <div class="mb-6 pb-12 border-b border-b-slate-300">
+  <div class="mb-6 pb-12">
```
Jongmassey, line 26: What's the rationale for removing `border-b border-b-slate-300` from this div's classes?

Jongmassey (approved):

templates/login.html
StevenMaude: To my eye, it looked a little bit unbalanced having the border for just one login entry. 

When we have one login entry, it looks like a divider, but it's not actually dividing anything. (It looks less so when we have two entries, even though we still have a border at the bottom of the final entry.)

I'm happy to defer to @tomodwyer though, whichever Thomas prefers. I'm not really tied to either :smiley:

## Page before these changes

old-login-options

## Page after these changes with the border *retained*

new-login-options-with-border

## Page after these changes with the border *removed*

new-login-options-without-border

Jongmassey: At first I think it seemed better with the border but looking at them side by side I agree with you that it's better without!

tomodwyer: Yup, I think remove the border 👍

PR #4959, opened 2025-03-13
Display hint strip in development

mikerkelly:
If running in development, display a soothing green strip at the top of each page to reassure dev users that they are not accidentally on the real live production site.

image

Don't display the red one on staff pages unless in production.  Staff users don't need to be warned about the risks of changing production data with the strip when unless they're actually in the production environment.


mikerkelly:
Tom has extended this in #4964 which I think is a straight improvement and we should merge that instead. This one can be closed.

PR #4962, opened 2025-03-13
Remove `AnalysisRequest` model

StevenMaude:
See #4880.

PR #4963, opened 2025-03-14
Remove references to Interactive analyses from Workspace templates

KatieB5:


Jongmassey:
Whilst the ticket does say "templates" there's also some parts of the underlying view that feed the bit of the template you have removed that can also be removed ([here](https://github.com/opensafely-core/job-server/blob/3da08055153603817bbe89f998af2e0ca4ca98a6/jobserver/views/workspaces.py#L215-L220) and [here](https://github.com/opensafely-core/job-server/blob/3da08055153603817bbe89f998af2e0ca4ca98a6/jobserver/views/workspaces.py#L225) ) as without this template code they are no longer executed. Which, in turn, would also lead to [this test](https://github.com/opensafely-core/job-server/blob/3da08055153603817bbe89f998af2e0ca4ca98a6/tests/unit/jobserver/views/test_workspaces.py#L490) being removed leaving things cleaner overall :)

Jongmassey (commented):
My reading of this ticket is that this should also remove the data that this bit of the template renders from the underlying view (see [this comment](https://github.com/opensafely-core/job-server/pull/4963#issuecomment-2725306627) ) / [this commit](https://github.com/opensafely-core/job-server/commit/7ffe9b04c9218a14002b19053561c77bee7a225a)

mikerkelly:
It looks like [Steve's draft PR](https://github.com/opensafely-core/job-server/pull/4962) for removing the `AnalysisRequest` model includes removing the view code and tests that Jon points out. So I think it would be fine to just merge this PR as is, re-base Steve's PR, and go from there.

Jongmassey (approved):
LGTM :rocket:

PR #4964, opened 2025-03-14
Add a fixed development banner

tomodwyer:
Building on the work in #4959 – this adds a development banner which is fixed to the top of the page.

This also improves the banner at the top of the staff area, and removes the need for an extra CSS file in the staff area.

https://github.com/user-attachments/assets/0c69ed7e-3eeb-494f-96cc-f097d66b927b


mikerkelly (approved):
Looks good to me, thanks. Looks better in the UI and the use of a partial template for the dev banner and just putting the staff banner in the staff pages base template rather than mucking around with `::before` both make a lot of sense.

PR #4969, opened 2025-03-17
Revert possibly problematic Interactive migrations

mikerkelly:


mikerkelly:
This is incomplete and was to address an incident that has now been resolved relating to #4962.

PR #4974, opened 2025-03-18
Remove `LoginWithToken`

### LINKED ISSUE #4939: Airlock makes LoginWithToken view redundant
# Why are we doing this?

Removing unused code is good.

The LoginWithToken view and its associated form are removed OR we decide that it's still needed.

# What are we doing?

The [LoginWithToken view](https://github.com/opensafely-core/job-server/blob/a7d4f0de2437f09843432a09d43e8b203c7fd21a/jobserver/views/users.py#L162) was used when logging into L4 to see the Outputs Viewer. This has been replaced with Airlock, which I believe uses the [Level4TokenAuthenticationAPI](https://github.com/opensafely-core/job-server/blob/1f5984653ac5949da6a36a678ebdf738d2bf474c/jobserver/api/releases.py#L635).

**We should check that this is the case before making any changes.**

---

StevenMaude:
Fixes #4939.

This removes the `LoginWithToken` view and `TokenLoginForm`. These are no longer needed since users now login to Airlock with tokens.

### DIFF
--- a/jobserver/forms.py
+++ b/jobserver/forms.py
@@ -61,11 +61,6 @@ class JobRequestCreateForm(forms.ModelForm):
                 )
 
 
-class TokenLoginForm(forms.Form):
-    user = forms.CharField()
-    token = forms.CharField()
-
-
 class RequireNameForm(forms.Form):
     name = forms.CharField()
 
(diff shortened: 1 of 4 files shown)


inglesp (approved):
Fine from RAP's perspective -- Airlock uses `jobserver.api.releases.Level4TokenAuthenticationAPI`, which is unaffected by this change.

lucyb (approved):
:broom:

PR #4979, opened 2025-03-19
Remove interactive reporter role

### LINKED ISSUE #4874: Remove the `InteractiveReporter` role
## Why are we doing this?

* We must remove the role from any existing users that have that role and ensure that all developers have done this locally before deploying any code changes.
* ~TODO: Do we even need to delete Interactive-only users entirely? (Or will they have broken pages, if we remove some of the other conditional parts of templates?)~ We should keep Interactive-only users, so we have an audit trail. We don't delete users in Job Server.
* Remove the ability to create users (`interactive/commands.py`) as `create_user()` uses `InteractiveReporter` (completed in #4902)
* #4875
* #4870
* #4866 

## What are we doing?

* **Possibly removing the role from existing users who have role on the site?**
* Remove `InteractiveReporter` from `jobserver/authorization/roles.py`
* Remove `InteractiveReporter` from `jobserver/authorization/__init__.py`
* Update the docstring for `all_roles()` in `jobserver/models/user.py` to refer to another role other than `InteractiveReporter`

When there are no longer references to `InteractiveReporter` in the job-server repository (except ADR 0011, which is a passing mention and probably doesn't need to be superseded).

Jongmassey:
Fixes #4874


jobserver/authorization/roles.py
```diff
@@ -48,15 +48,14 @@ class StaffAreaAdministrator:
...
-    ]
-    permissions = [
-        release_file_view,
-    ]
+    """
+    Vestigial role class to support data migration to remove assignments of this role to users.
+    Not to be used outside of this migration.
+    Must be removed once this data migration is eliminated.
+    """
+
```
lucyb, line 56: You will need to keep the display_name here, otherwise you will get an error when viewing some users in the staff area (such as my Interactive only user).

The error is occurring because you've not removed the roles from ProjectMembership, but there might be something else I've missed.

If you decide to remove roles from ProjectMemberships in code, you'll need to use the project_members.remove(..) "command".
https://github.com/opensafely-core/job-server/blob/04a50694712f8b20e6130d2f3df3f7f425719df9/jobserver/commands/project_members.py#L74

Jongmassey: That's a good point, I'd forgotten about project level roles. 

However, this raises another question - who should the "by" be in this data migration? Me? Do we want/need the audit log to be correct for this operation?

lucyb: Actually, slight correction, I think the migration needs to call `update_roles` and not `remove`, because the latter also removes the project membership, which I don't think we want.

I think we want the audit log to be correct, but the question is what correct means in this case. We don't need to provide a valid username, but perhaps it's easier if it's you. We'll want to inform the IG team once the migration has been applied too.

Jongmassey: OK I've modified the migration accordingly (with a bit of _fun_ along the way, see the comment in the code). Could you take another look?

jobserver/migrations/0013_remove_interactivereporter.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def remove_interactive_reporter_project_role(apps):
+    # The ProjectMembership model has the ImmutableModelMixin
+    # which provides an `override` parameter to the overridden model save()
+    # (which is absent in the base model save() method)
+    # and is provided when called within update_roles().
+    # The recommended way of accessing django model classes in migrations
+    # is using apps.get_model(), however this just returns a vanilla model
+    # class with no mixins and so the call to save(override=True) fails.
```
lucyb, line 18: I'm not sure that I understand this comment. The call to `project_members.update_roles` appears to work as expected, but you state the call to `save` failed?

lucyb (approved):
Thank you for considering the implications for the interactive-only users. Fortunately, they should be unable to login now and are internal users, so there shouldn't be any impact.

jobserver/authorization/roles.py
lucyb: Thanks. I've reapplied it locally and that seems to work fine. I've added a comment to clarify the comment. It does sound like it was a bit of a pain to write.

jobserver/migrations/0013_remove_interactivereporter.py
Jongmassey: Update roles calls save within itself, but it expects the save that's overridden in that mixin not the base save

PR #4980, opened 2025-03-19
ADR for `initiative:sunset-interactive`

### LINKED ISSUE #4863: Write an ADR to explain the removal of OpenSAFELY Interactive
## Why are we doing this?

We're removing the OpenSAFELY Interactive functionality. It may be useful to record that decision, and its consequences on the codebase.

## What are we doing?

Writing an appropriate ADR.

This doesn't have any prior dependency on other issues.

When there is an ADR documenting the decision.

mikerkelly:
Closes #4863.

Also deprecate Interactive-related ADRs.

### DIFF
--- a/docs/adr/0029-sunset-osi-interactive.md
+++ b/docs/adr/0029-sunset-osi-interactive.md
@@ -1,0 +1,42 @@
+# 29. Sunset OpenSAFELY Interactive
+
+Date: 2025-03
+
+## Status
+
+Accepted.
+
+## Context
+
+OpenSAFELY Interactive (OSI) was a Job Server app allowing users to request
+simple analyses.  We are removing OSI.
+
+Refer to the Team Manual for more information:
+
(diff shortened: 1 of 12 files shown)


docs/adr/0029-sunset-osi-interactive.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Context
+
+OpenSAFELY Interactive (OSI) was a Job Server app allowing users to request
+simple analyses.  We are removing OSI.
+
+Refer to the Team Manual for more information:
+
+https://github.com/ebmdatalab/team-manual/blob/main/docs/products/opensafely-interactive.md
+https://bennett.wiki/products/opensafely-interactive/
```
tomodwyer, line 17: I think it should either be only the link to the wiki, or a permalink of the current commit of that file, incase we move it or change to a CMS in the future.

tomodwyer (approved):
One comment, otherwise all good

PR #4990, opened 2025-03-20
Remove `interactive` reference in `redirects` migrations

mikerkelly:


StevenMaude:
I think Mike already would know/expect this behaviour but I'll note it:

* for an up-to-date local database, that is: up-to-date with migrations currently on `main`, this branch works fine:
  ```
  No migrations to apply.
  ```
* for a database in a state _prior_ to the recent `interactive` migration in #4962, this branch fails to run:
  ```
  django.db.migrations.exceptions.NodeNotFoundError: Migration interactive.0003_delete_analysisrequest depends on nonexistent node ('redirects', '0003_remove_redirect_redirects_redirect_only_one_target_model_set_and_more'). Django tried to replace migration redirects.0003_remove_redirect_redirects_redirect_only_one_target_model_set_and_more with any of [redirects.0001_initial_squashed_2025_03] but wasn't able to because some of the replaced migrations are already applied.
  ```
  * The issue here is that the squashed migration in this branch also contains a much older [redirects migration](https://github.com/opensafely-core/job-server/commit/d9ccca01fba06161d6db3d4ef541a75b80024075#diff-149f06c4c00834f965942d899ec7c76c088b06e58a5fea17dccc1161e90126df) that almost every working database copy will have applied a while ago.

I don't think that should matter, because we know that the production database *does* have the latest migrations applied.

For local development, individuals will need to apply those migrations first from a previous commit, or start afresh with a new database. (Mike already noted this in Slack, but it's worth mentioning here for visibility.)

StevenMaude (approved):

mikerkelly:
Thanks for the details and thinking that through, @StevenMaude. Yes, I agree.

Had thought about / briefly experimented with keeping a vestigial 0003 but I hope to remove interactive altogether very soon after this PR in #4991 so it didn't seem worth it.

Issue #4876, opened 2025-02-13
Review whether any database cleanup is required


StevenMaude:
There's at least one `interactive` app specific migration; do we still need the migration?

Is there any other database cleanup that can be done following removal of Interactive? Can any fields be removed? Can any other models be removed?

## What needs to be done before this?

All of the code removal, so #4873 should be complete first.

mikerkelly:
- https://github.com/opensafely-core/job-server/pull/5000
- https://github.com/opensafely-core/job-server/pull/5001
- https://github.com/opensafely-core/job-server/pull/5002

We don't anticipate a need to roll back the `applications` or `jobserver` apps, this is a bit tidier, and we believe squashing the migrations may make applying them from scratch (such as when running tests) slightly faster. These migrations can always be resurrected from git history if needed. This has also removed the 'data only' parts of migrations and cleaned up any vestigial references to those.

PR #4991, opened 2025-03-20
Remove `interactive` app

### LINKED ISSUE #4873: Check we have not missed removing anything following the removal of Interactive code
## Why are we doing this?

We're removing the OpenSAFELY Interactive functionality. We want to have no vestigial code or dependencies remaining. It's possible that we don't find everything that can be removed at the outset of specifying what's to be done for this initiative.

## What are we doing?

Search through the code again following all of the other code removal tasks to:

* ✅ review any remaining Interactive frontend and backend code to confirm it's still needed, and remove it if not
* ✅  check for other code in our codebase that was used by removed Interactive code, and is therefore no longer needed (tools like [vulture](https://github.com/jendrikseipp/vulture) can help here, although may miss things) -- see various comments downthread.
* check for unneeded dependencies (these should have already been identified, but it's possible we've missed some; there are several tools that can do this)
* ✅ reviewing ADRs to see if we need to update or supersede any more that we missed -- done in #4863.

All of the other code removal and clean-up tasks.

When the review is complete, and any other remaining unneeded code has been removed.

mikerkelly:
Partially fixes #4873. Related to #4876.

~Note that #4990 should be merged and this PR rebased before merging.~

We are sunsetting OpenSAFELY Interactive, refer to docs/adr/0029-sunset-osi-interactive.md. Remove the last references to the `interactive` app now it has no meaningful content. Along with references to the app in configuration and tooling.

The last remaining content in the app was the migrations. We don't need them anymore. Without the model `AnalysisRequest` the migrations are a no-op, adding the model then ultimately removing it. So having no migrations is equivalent, if starting from scratch. Production has these migrations applied. Databases still in a state from before those migrations will have old `interactive` data sitting in their DB but this isn't harmful. Devs can live with old interactive data in their dev DBs if they didn't apply the migrations yet, and if they don't want that, they can always do one of: start afresh with their dev DB, pull down prod, drop those tables manually, or get the migrations from git history.

Note that #4979 removes the InteractiveReporter role, but this doesn't depend on the `interactive` app; permissions apply to `User`, `Project`, and `ProjectMembership` instances, and views (and all related views have been removed already).


StevenMaude:
Ran this branch on a database that was up-to-date with the migrations from current `main`, and everything is OK:

```
Operations to perform:
  Apply all migrations: applications, auth, contenttypes, jobserver, redirects, sessions, social_django
Running migrations:
  No migrations to apply.
```

StevenMaude:
This looks good to me; thanks for this and the other removal-related PRs.

It might warrant a rebase just to remove the need for the [merge commit](https://github.com/opensafely-core/job-server/pull/4991/commits/1a75541d2bd62dd5b8f41539ffe74355357f463b).

It's also probably worth announcing to both REX and RAP developers that they may need to update local database setups when updating code checkouts to the latest `main`. That could avoid any confusion. (Particularly so if these migration changes are the final ones now.)

StevenMaude (approved):

Issue #4699, opened 2024-10-22
Require `User.fullname` to be populated


mikerkelly:
`User.fullname` is a `TextField` that defaults to an empty string. There are 27 `User` on the live server with `fullname == ''`. If we can manually fill in that field for each that would allow us to enforce `fullname` being populated in the model (non-null and non-empty?). That could allow us to be rid of the `User.name` property which uses `fullname` but falls back to `username` if `fullname` is empty. This would simplify the `User` model and populate the DB with better data for managing or viewing `User` instances.

PR #4995, opened 2025-03-21
Require User.fullname

Jongmassey:
Part fixes [#4699](https://github.com/opensafely-core/job-server/issues/4699)

#5045 backfills the currently-blank usernames must be merged before this PR.

~~TODO: rebase once 5045 is merged such that the migration graph is correct~~


Jongmassey:
Given recent issues with deploying PRs with multiple migrations we may wish to split this into two - backfill first, then the model change

Jongmassey:
This will also probably need communicating to NOD (?) that we've backfilled full names for 27 users who were either dormant or system accounts

tests/factories/user.py
```diff
@@ -9,3 +9,4 @@ class Meta:
 
     username = factory.Sequence(lambda n: f"user-{n}")
     email = factory.Sequence(lambda n: f"user-{n}@example.com")
+    fullname = factory.SelfAttribute("username")
```
mikerkelly, line 12: I think it's worth making the generated `username` and `fullname` distinct so there's no risk of a test mistakenly passing if one is mistakenly used where the other is expected. Maybe something like:

```suggestion
    fullname = factory.LazyAttribute(lambda obj: obj.username.replace("-", " ").title())
```

I think that "looks" more like a full name (`User 43` vs. `user-43` as the `username`).

mikerkelly (commented):
```python
fullname = models.TextField(blank=False)
```

This doesn't achieve the desired outcome. Django Field reference
https://docs.djangoproject.com/en/5.1/ref/models/fields/#blank

> Note that this is different than [`null`](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.Field.null "django.db.models.Field.null"). [`null`](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.Field.null "django.db.models.Field.null") is purely database-related, whereas [`blank`](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.Field.blank "django.db.models.Field.blank") is validation-related. If a field has `blank=True`, form validation will allow entry of an empty value. If a field has `blank=False`, the field will be required.

Our `User` get created by social auth so I don't think form validation is relevant. You can see it isn't working as intended by creating a `User` with blank `fullname` or seeing that there are still test cases that create users with blank `fullname` (`test_user_ordering`, `test_create_user`).

I think the model ought to be applying a constraint at database level that the field not be blank, otherwise we're relying on social auth providing a `fullname` without validating that.

```python
models.CheckConstraint(condition=~models.Q(fullname=""), name="fullname_non_blank"),
```

We should have a test for this.

I don't think we should make the (`blank=False`) change at all as it implies we rely on it when we don't. It is also in the wrong commit, I think ("Remove user.name[...]" instead of "Enforce fullname non-blank in User").

If we have the above constraint we don't need this part of the ordering constraint:

```python
# Empty 'fullname' comes last.
Case(
    When(fullname="", then=1),
    default=0,
    output_field=models.IntegerField(),
),

File: tests/factories/user.py (hunk 1 above)
Comment (Jongmassey) [reply tests/factories/user.py:12]: That's an excellent suggestion, I'll check that it still works with all the tests.

**Comment by Jongmassey** [comment]
>I think the model ought to be applying a constraint at database level that the field not be blank, otherwise we're relying on social auth providing a fullname without validating that.

I agree, I did encounter the lack of enforcement of the population of fullname when creating User objects in doing this and just thought it was a strange Djangoism. If a constraint makes it behave like I presumed it should behave then that's what I shall do!

File: tests/unit/jobserver/models/test_user.py [hunk 2]
```diff
@@ -298,19 +298,32 @@ def test_user_ordering():
...
     assert user.email == "test@test.test"
     # Password is not stored in plaintext.
     assert user.password != password
+
+
+def test_user_fullname_non_blank():
+    """Test the non-blank constraint on User"""
+    with pytest.raises(IntegrityError):
+        u = User(username="foo", email="foo@bar.com")
+        u.save()
```
mikerkelly, line 329: I think this effectively tests the expected behaviour, but we may as well make the use of `fullname` as explicit as possible in the docstring and `init` call.

```suggestion
    """Test the non-blank constraint on User.fullname."""
    with pytest.raises(IntegrityError):
        u = User(username="foo", email="foo@bar.com", fullname="")
        u.save()
```

mikerkelly (approved):
Thanks Jon, looking good!

tests/unit/jobserver/models/test_user.py
Jongmassey: Good idea, I'll do that

PR #4999, opened 2025-03-24
Update the developer documentation for fresh installs

tomodwyer:
This change updates the docs to list the current state of setting up Job Server from scratch for local installs. There are still issues that need to be resolved, but are outside the scope of this PR.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -222,29 +222,33 @@ Vite builds the assets and outputs them to the `assets/dist` folder.
 
 [Django Staticfiles app](https://docs.djangoproject.com/en/3.2/ref/contrib/staticfiles/) then collects the files and places them in the `staticfiles/assets` folder, with the manifest file located at `assets/dist/.vite/manifest.json`.
 
-### Data Setup
+### Setting up a fresh install
+
 Sometimes it's useful to have a fresh local installation or you may not have authorization to download a production backup.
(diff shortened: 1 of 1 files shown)


DEVELOPERS.md
```diff
@@ -9,29 +9,32 @@
...
+    - [Restoring Backups](#restoring-backups)
     - [Steps](#steps)
   - [Docker Compose](#docker-compose)
   - [Frontend development (CSS/JS)](#frontend-development-cssjs)
   - [Running the local asset server](#running-the-local-asset-server)
   - [Compiling assets](#compiling-assets)
-  - [Data Setup](#data-setup)
-  - [Upgrade OpenTelemetry dependencies](#pip-opentelemetry)
+  - [Setting up a fresh install](#setting-up-a-fresh-install)
+  - [pip-opentelemetry](#pip-opentelemetry)
```
StevenMaude, line 19: thought: I'd have probably kept the "Upgrade OpenTelemetry dependencies" contents entry here and amended the section title, rather than the other way around, because "pip-opentelemetry" is not a very good title. But I don't have too strong opinions about it.

StevenMaude (approved):

DEVELOPERS.md
tomodwyer: Oh I thought my plugin just re-ordered them, didn't realise it changed the titles! Let me fix that.

PR #5001, opened 2025-03-24
Remove `jobserver` data migrations

mikerkelly:
These data migrations don't affect schema and aren't needed anymore now that they've been applied to production. Removing them may be slightly faster and tidier, and will also allow us to tidy up any remaining references to them, and make squishing migrations slightly easier.

Original tickets: 0002 -- #4543, 0003 -- #4540 , 0005 -- #4606, 0006 -- #4606, 0008 -- #4639, 00013 -- #4979.

This also allows us to remove the vestigial role classes. These aren't needed now the data migrations that referenced them to remove them are gone.


Jongmassey (approved):
:tada:

PR #5029, opened 2025-03-26
Update psycopg[binary] to 3.2.6

StevenMaude:
This may require a little monitoring on deployment, since there's an open, although slightly vague, issue on psycopg performance degradation with 3.2:

psycopg/psycopg#888


mikerkelly (approved):
It seems to work fine locally.

@StevenMaude https://github.com/psycopg/psycopg/issues/888 does not show a lot of evidence as you note, but it's probably worth light monitoring on any such update, can you please pick a couple of headline stats and decide how we're going to monitor them? Maybe from the Job Server Honeycomb dashboard and comparing this week's stats with next week's for example?

StevenMaude:
From Honeycomb, it doesn't look like we get any database performance statistics directly, so we have to infer it from request time.

From the existing Honeycomb board, we could review the existing "Average & P95 duration of requests by route".

Looking at HTTP response status codes/durations could also be useful; with a degradation of database performance, we might expect:

* a greater number of error codes proportionally (due to timeouts)
* potentially different error codes appearing (maybe a HTTP 500 error code?)
* longer times for those errors to appear

Example:

```
Visualize
    HEATMAP(duration_ms)
    MAX(duration_ms)
    P95(duration_ms) 

Where
    http.status_code exists 

Group By
    http.status_code
```

We can review the next week following deployment (so a week today) and compare to the previous week.

StevenMaude:
The tests aren't noticeably slower, at any rate.

StevenMaude:
I compared request times and error code counts [on Slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1743681624933939). I don't think there's anything to worry about here.

PR #5030, opened 2025-03-26
Update social-auth-core dependency subtree

StevenMaude:


StevenMaude:
I checked locally that login and logout still work as expected.

mikerkelly (approved):

PR #5032, opened 2025-03-26
Update miscellaneous dependencies

StevenMaude:
These are mostly transitive dependencies that are shared by multiple packages.

### DIFF
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -438,9 +438,9 @@ tenacity==9.0.0 \
     --hash=sha256:807f37ca97d62aa361264d497b0e31e92b8027044942bfa756160d908320d73b \
     --hash=sha256:93de0c98785b27fcf659856aa9f54bfbd399e29969b0621bc7f762bd441b4539
     # via stamina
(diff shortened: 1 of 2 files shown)


StevenMaude:
I think this pretty much rounds off all the updates. If there are more that are found after merging all the other PRs, they could be added here. Equally, I'm happy to break this PR up into smaller PRs if it's deemed useful.

StevenMaude:
> I don't see what the benefit of splitting into multiple PRs would be.

It's fair to leave it as is; I didn't have a strong opinion.

My thinking was around "the updates in this PR interact with multiple areas of the system". That contrasts some of my previous recent PRs (such as #5030) where I've tried to contain to more limited code paths or subsystems.

(We do typically merge multiple unrelated Python dependency PRs in quick succession with Dependabot at the moment, and batch all Python updates into one regular dependency PR for the opencodelists repository.)

Except for whitenoise, these packages are transitive dependencies. Also, these updates are either fairly small bits of specialised functionality — tokenize-rt, python-ipware, django-ipware, or are long-lived, stable packages.

PR #5035, opened 2025-03-27
Switch from Dependabot to update-dependencies-action

### LINKED ISSUE #4412: Consider alternatives to Dependabot
Working with Dependabot can be time-consuming. Recently, for example, I asked why Dependabot hadn't bumped django-permissions-policy since v4.12.0,[^1] which was released on 2022-06-05. The latest version is 4.20.0, which was released on 2024-06-19. The reason? We'd instructed Dependabot to ignore django-permissions-policy with `@dependabot ignore`, and since then had been bumping it ourselves. The following table shows the timeline.

| Date       | PR    | Description                                  |
|------------|-------|----------------------------------------------|
| 2022-01-11 | #1516 | Dependabot bumped from 4.6.0 to 4.7.0        |
| 2022-01-17 | #1529 | we bumped from 4.7.0 to 4.8.0                |
| 2022-02-17 | #1613 | we instructed Dependabot to ignore           |
| 2022-02-18 | #1614 | we created issue asking for it to be removed |
| 2022-07-25 | #1946 | we bumped from 4.8.0 to 4.12.0               |
| 2022-08-30 | #1614 | we closed issue                              |
| 2024-07-03 | #1613 | we instructed Dependabot not to ignore       |
| 2024-07-03 | #4411 | Dependabot bumped from 4.12.0 to 4.20.0      |

In summary, django-permissions-policy was never removed. It was, in effect, pinned by the instruction to Dependabot between 2022-02-17 and 2024-07-03.

Creating a timeline is time-consuming. Whilst it's not necessary, it is useful to understand what happened when. Here, it seems as though we'd forgotten about the instruction to Dependabot and had been living with the consequences ever since. "The consequences" aren't limited to bumping a dependency ourselves; they include not using the latest version of a dependency in production.

I'd argue that `@dependabot ignore` is a bug, not a feature. Either way, it's certainly a headache: See dependabot/dependabot-core#2255. We've had several conversations about Dependabot's limitations. We should consider alternatives, such as Mend.io's [Renovate](https://www.mend.io/renovate/). @evansd linked to [a blog post by Jamie Tanna](https://www.jvt.me/posts/2024/04/12/use-renovate/) about it.[^2]

[^1]: https://bennettoxford.slack.com/archives/C069SADHP1Q/p1719946121312349
[^2]: https://bennettoxford.slack.com/archives/C01T2HACV3K/p1713174069344909

StevenMaude:
Fixes #4412.

This switches from Dependabot to opensafely-core/update-dependencies-action for updating Python dependencies.

It reuses the existing workflow that was updating the `pipeline` dependency, and switches to updating weekly.

While looking at how this works for Airlock, it also borrow an improvements from Airlock with respect to updating dependencies. Specifically:

* the two `requirements.*.txt` files are checked for consistency of package versions, where packages are listed in both requirements files

### DIFF
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -3,7 +3,7 @@ name: Update python dependencies
 on:
   workflow_dispatch:
   schedule:
-    - cron:  "0 23 * * *"
+    - cron:  "0 23 * * SUN"
 
 jobs:
   update-dependencies:
@@ -24,6 +24,6 @@ jobs:
     - uses: bennettoxford/update-dependencies-action@v1
       with:
         token: ${{ steps.generate-token.outputs.token }}
-        # Note: just update pipeline library for now, as job-server uses dependabot for python upgrades
         update_command: |
           just upgrade-pipeline
+          just update-dependencies
--- a/justfile
+++ b/justfile
@@ -109,6 +109,13 @@ upgrade env package="": virtualenv
     test -z "{{ package }}" || opts="--upgrade-package {{ package }}"
     FORCE=true {{ just_executable() }} requirements-{{ env }} $opts
 
+# Upgrade all dev and prod dependencies.
+# This is the default input command to update-dependencies action
+# https://github.com/bennettoxford/update-dependencies-action
+update-dependencies:
+    just upgrade prod
+    just upgrade dev
+
 # upgrade our internal pipeline library
 upgrade-pipeline: && requirements-prod
     ./scripts/upgrade-pipeline.sh requirements.prod.in
--- a/tests/test_requirements.py
+++ b/tests/test_requirements.py
@@ -1,0 +1,42 @@
+# Taken from Airlock:
+# https://github.com/opensafely-core/airlock/blob/ca23669bcf22bf719bd79103ecc53da0a7c53e2f/tests/test_requirements.py
+from pathlib import Path
+
+
+PROJECT_ROOT = Path(__file__).parents[1]
+
+
+def test_requirements_are_consistent():
+    # The production and development dependencies aren't intended to overlap (the dev
+    # dependencies are supposed to be _extra_ packages needed for development) but
+    # due to transitive dependencies the `requirements.dev.txt` file can end up
+    # including packages which are also included in `requirements.prod.txt`.
+    #
+    # We use a `--constraint` argument in `requirements.dev.in` to ensure that whenever
+    # we compile the development dependencies we stay consistent with the current
+    # production dependencies. However it's still possible, by doing things in a certain
+    # order, to accidentally end up with divergent versions. This can lead to confusion
+    # and makes the actual installed set of packages dependent on the order in which the
+    # files are read, which is bad.
+    #
+    # The below test ensures that wherever the two files specify the same dependency
+    # they agree on the version.
+    prod_packages = parse_requirements(PROJECT_ROOT / "requirements.prod.txt")
+    dev_packages = parse_requirements(PROJECT_ROOT / "requirements.dev.txt")
+    common_packages = prod_packages.keys() & dev_packages.keys()
+    for package in common_packages:
+        assert prod_packages[package] == dev_packages[package], (
+            f"Inconsistent versions of '{package}' in "
+            f"requirements.prod.txt and requirements.dev.txt"
+        )
+
+
+def parse_requirements(filename):
+    packages = {}
+    for line in filename.read_text().splitlines():
+        line = line.strip()
+        if not line or line.startswith(("#", "--")):
+            continue
+        package, _, version = line.partition(" ")[0].partition("==")
+        packages[package] = version
+    return packages
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,11 +1,6 @@
 version: 2
 
 updates:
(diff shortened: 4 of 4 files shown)


StevenMaude:
As things stand, as of 3b8c95bbcc76fad0548c1d4d76bd45a43919438a, we're still using `pip-compile` to update the dependencies automatically. This doesn't allow us to restrict to dependencies older than a week.

I tried using `uv` with `exclude-newer` which seems to work.

The differences are that:

* `--allow-unsafe` has no effect, so `pip` and other packages that `pip-compile` considers "unsafe in a requirements file" end up in the main list of requirements
* the command header line is different since we're using `uv`:
  ```
  # This file was autogenerated by uv via the following command:
  #    uv pip compile --allow-unsafe --generate-hashes --strip-extras --exclude-newer 2025-03-20 requirements.dev.in
  ```

The only downside of using `uv` to update the dependencies this is that everyone developing job-server should use `uv` when upgrading dependencies, otherwise we could get unnecessary changes when updating dependencies in future.

`uv` is installable by `pip` and cross-platform, so itself could be a developer requirement, instead of `pip-tools`. It's also considerably faster than `pip-compile`.

tests/test_requirements.py
StevenMaude, line 1: question: Should we note that this came from Airlock in a comment, or is the mention in the commit comment sufficient?

justfile
```diff
@@ -100,14 +100,16 @@ install-precommit:
     test -f $BASE_DIR/.git/hooks/pre-commit || $BIN/pre-commit install
 
 
-# upgrade dev or prod dependencies (specify package to upgrade single package, all by default)
-upgrade env package="": virtualenv
+# compile all requirements.in files (by default only adds/removes packages; `-U` upgrades all; `-P <package>` upgrades one)
+compile-reqs *ARGS: devenv
```
lucyb, line 104: (Warning: drive-by comment) It might be worth taking a look at how OpenCodelists does this:

https://github.com/opensafely-core/opencodelists/blob/f7ad55852b63dce37f390a2e15a7056144232aec/justfile#L118-L123

It would be good to be consistent, but if not, then it would be good to know why we're doing something different.

StevenMaude:
We decided not to switch to `uv` right now, and the behaviour will be more like Dependabot's default behaviour, that is, we'll get updates without any delay.

I've opened an issue to consider switching to `uv`: #5049.

justfile
StevenMaude: I think the Airlock approach is actually better, as alluded to in the associated commit message (a86761b0771a86a63301302055f9ab69b8ef6539), because it reduces the risk of production and development dependencies not being in sync.

That said, the test taken from Airlock also guards against this, so I've left this as in OpenCodelists for now, to reduce the changes in this PR. If we change one, we should change both.

.github/workflows/update-dependencies.yml
```diff
@@ -24,6 +24,16 @@ jobs:
     - uses: bennettoxford/update-dependencies-action@v1
       with:
         token: ${{ steps.generate-token.outputs.token }}
-        # Note: just update pipeline library for now, as job-server uses dependabot for python upgrades
         update_command: |
           just upgrade-pipeline
+
+    - name: Check weekday
+      run: |
+        echo "WEEKDAY=$(date +%A)" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
StevenMaude: To make things simpler, I'm proposing checking weekly for all updates: including `pipeline`. The workflow can be run manually if a `pipeline` or other update is needed between the scheduled runs.

tests/test_requirements.py
```diff
@@ -0,0 +1,42 @@
...
+def test_requirements_are_consistent():
+    # The production and development dependencies aren't intended to overlap (the dev
+    # dependencies are supposed to be _extra_ packages needed for development) but
+    # due to transitive dependencies the `requirements.dev.txt` file can end up
+    # including packages which are also included in `requirements.prod.txt`.
+    #
+    # We use a `--constraint` argument in `requirements.dev.in` to ensure that whenever
+    # we compile the development dependencies we stay consistent with the current
+    # production dependencies. However it's still possible, by doing things in a certain
+    # order, to accidentally end up with divergent versions. This can lead to confusion
```
mikerkelly, line 18: "However, it's still possible to change the production dependencies later and end up with divergent versions"?

tests/test_requirements.py
```diff
@@ -0,0 +1,42 @@
...
+    # and makes the actual installed set of packages dependent on the order in which the
+    # files are read, which is bad.
+    #
+    # The below test ensures that wherever the two files specify the same dependency
```
mikerkelly, line 22:
```suggestion
    # The below test verifies that wherever the two files specify the same dependency
```

I'm not sure if it makes sense to say a test that doesn't change anything "ensures" something. It can still be false after the test is run.

mikerkelly (approved):

tests/test_requirements.py
StevenMaude: I think your wording is more correct, but I'll leave this (and the other suggestion) because it's a duplication of the code taken from elsewhere.

mikerkelly: I'm not sure there's any motivation to keep this exactly like the original. But I don't feel strongly about it. It's easy enough to understand. It could be easier, that's all.

PR #5045, opened 2025-03-31
Backfill fullnames

Jongmassey:
Since f88e84280d78e6abbca166b18232116d9398509d (Apr 2023) the user sign up form has required a `fullname` but at the time this wasn't enforced in the model.

This left 27 (at the time of writing) vestigial nameless users.

In advance of enforcing the non-nullness of User.fullname in the model this commit backfills a full name for these users.

First priority is known system accounts for which we declare a name. Next is looking for applications from a given user which have a full name populated and taking that.
Next is to try and parse firstname.lastname@[domain] email addresses. Finally if these fail then populate with the username.


mikerkelly (approved):
We should remove this migration not-too-long after it has been applied in production as it will have no on-going relevance. Possibly flag to devs that they may want to apply the migration if they have stuff in a local DB they'd like to keep, as throwing it away will make the migration that actually changes the constraint not possible to apply. Please make sure we do that.

I have done a minimal review of the code as it's not going to persist and isn't used in any other code. Looks good to me.

Tested locally by restoring a backup of production and running `select username, email from jobserver_user where fullname = '';` to find the affected accounts, then `select username, email, fullname from jobserver_user WHERE username IN ({list of names});` to see the result of applying the migration.

I think each case where this uses something other than username the result looks reasonable. I think there are some cases where we can do something better manually. I'll discuss that in [this Slack thread](https://bennettoxford.slack.com/archives/C080XJJ0KDJ/p1743454243688709), taking appropriate care with any Personal Data.

Issue #4423, opened 2024-07-05
Deploys are non-idempotent


Jongmassey:
If a [deploy](https://github.com/opensafely-core/job-server/blob/3806de0d0213e6837bf9ffb25fc3772a8a10fb79/.github/workflows/main.yml#L148-L211) fails on a step after [Deploy Image](https://github.com/opensafely-core/job-server/blob/3806de0d0213e6837bf9ffb25fc3772a8a10fb79/.github/workflows/main.yml#L148-L211), the action cannot be re-run without this step erroring.

This step [calls](https://github.com/opensafely-core/job-server/blob/3806de0d0213e6837bf9ffb25fc3772a8a10fb79/.github/workflows/main.yml#L193) `git:from-image job-server $SHA`.
If this has previously run with the same SHA, it returns a non-zero exit code and the message 
>No changes detected, skipping git commit
 !     Call 'ps:rebuild' on app to rebuild the app from existing source
Error: Process completed with exit code 1.

It is therefore not possible to re-run the action to ensure the successful completion of any steps after this one which have previously failed ([example](https://github.com/opensafely-core/job-server/actions/runs/9806602093/job/27078821468)).

madwort:
another option might be to set [`continue-on-failure`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) on the relevant steps (i.e. just honeycomb marker & sentry release steps). As you say, re-deploying a version risks clashing with subsequent deploys, and if these markers aren't created at the time that the version is deployed they lose their usefulness a bit. I think the easiest thing is to just roll forward & deploy something new. 

Of course, if e.g. Sentry becomes generally unreliable we'd then have the problem of getting notified about that, but if it's just occasional intermittent failures like this then it would be ok?

StevenMaude:
@madwort: That's a good idea, and maybe that's the simplest thing to do.

I'll add my notes below.

I think there are some quirks with how this deployment process works, and it's worth making those quirks explicit, so we can decide if we're OK with it as is. (There's also a risk we overcomplicate it, which adds its own potential problems.)

StevenMaude:
# Summary of current quirks

If you don't want to read the rest of my notes (next post):

[...]

PR #5053, opened 2025-04-02
Finalize Sentry release separately from creation

### LINKED ISSUE #5046: Mark sentry release as unfinalized before deploy, then finalize after
# Why are we doing this?

#5041 re-ordered job server's [deploy workflow](https://github.com/opensafely-core/job-server/blob/main/.github/workflows/deploy.yml) such that a sentry release is created before it is deployed to dokku, such that in case of that deploy failing we know which release failed to deploy.

However, we now have the situation where the "currently live" release information is wrong in case of a failed deploy.

Sentry has the concept of [finalization](https://docs.sentry.io/cli/releases/#finalizing-releases) of releases, which we could use to differentiate between a release that has built and when it has been deployed.

When, during a deploy, the newly-created sentry release shows as unfinalized until the dokku deploy has been successful.

# What are we doing?

Altering the existing sentry deploy job to pass the finalized parameter as false.

Adding a new sentry deploy job to the deploy workflow after the dokku deploy job to set the finalize the release.

---

StevenMaude:
Fixes #5046.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -65,6 +65,7 @@ jobs:
           SENTRY_PROJECT: job-server
         with:
           environment: production
+          finalize: false
 
       - name: Deploy image
         run: |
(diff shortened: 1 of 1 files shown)


StevenMaude:
Also see #4423 as that's slightly related.

Just adding the extra deploy step in the workflow may be fine for the time being (and is what was proposed in #5046, and will address the currently slightly broken state).

mikerkelly (approved):

PR #5062, opened 2025-04-07
Replace @remotedevforce/tom-select with tom-select

### LINKED ISSUE #5028: Replace abandoned RemoteDevForce/tom-select dependency
[RemoteDevForce/tom-select](https://github.com/RemoteDevForce/tom-select) is now abandoned. We should switch back to the (now maintained version of) [orchidjs/tom-select](https://github.com/orchidjs/tom-select)

tomodwyer:
Closes #5028


Jongmassey (approved):
Tested on the staff area job request page as an example of something that uses multiselect and it seems good to me!

PR #5069, opened 2025-04-10
Pin django to v5.1.x and update

Jongmassey:
With this change to requirements.prod.in,
`pip-compile --allow-unsafe --generate-hashes --strip-extras --output-file=requirements.prod.txt requirements.prod.in --upgrade-package=django` updates django to 5.1.8 rather than 5.2


StevenMaude (approved):
thought: as I mentioned in the [OpenCodelists PR](https://github.com/opensafely-core/opencodelists/pull/2423); not essential, but maybe worth a requirements file comment or mentioning in the commit message the reason for pinning?

PR #5081, opened 2025-04-23
Document deployment of migrations approaches

mikerkelly:
We've discussed this in the team -- let's document it in the developer-facing docs for future reference.

Duplicative of https://github.com/opensafely-core/opencodelists/pull/2452 but we don't have a better place to put this for now, and it is relevant to both repos.


StevenMaude (approved):
The content's fine and was also reviewed in opensafely-core/opencodelists#2452.

Maybe both entries could be reduced to one Team Manual playbook entry, if this is the policy we're applying across all the Team REX repositories? And then a link from each `DEVELOPERS.md` to that entry?

But I don't have any strong objection to the duplication, as the advice is unlikely to change frequently, so it's fine to leave this as is.

mikerkelly:
@StevenMaude thanks, that approach broadly makes sense -- maybe at some point we'll have Team Rex ways of working docs in the team manual or elsewhere. I'm not quite sure if it fits as a playbook. Also I'm aware this repo may be moving away from Team REX in the next few months.

PR #5111, opened 2025-05-30
Fix ADR link in TESTING.md

rw251:


rw251:
Changed at the same time in #5112

PR #5133, opened 2025-06-24
Improve performance of JobRequests endpoint

evansd:
This endpoint is intended to return all "active" JobRequests, which means all JobRequests with active jobs and any JobRequests with no jobs at all (which means they have been recently created and not yet acknowledged by the RAP Controller).

Previously it did this using a JOIN between the JobRequest and Job tables and the a DISTINCT to squash the results back down to one row per JobRequest. However we can also do the same thing using subqueries and this turns out to be significantly faster. Running the equivalent queries through the management console shows a speed-up of somewhere between 25x and 30x.

### DIFF
--- a/jobserver/api/jobs.py
+++ b/jobserver/api/jobs.py
@@ -15,7 +15,7 @@ from rest_framework.views import APIView
 from jobserver.api.authentication import get_backend_from_token
 from jobserver.emails import send_finished_notification
 from jobserver.github import _get_github_api
-from jobserver.models import JobRequest, Stats, User, Workspace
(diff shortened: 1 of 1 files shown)


lucyb (approved):
This is a really nice spot, thank you.

I've not done any local performance testing, but I trust the checking you've done, and we can easily monitor the change using Honeycomb.

bloodearnest:
Nice work! ~35s down to ~<1s

https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/result/8t6AtbqVSoP?vs=hideCompare&tab=overview

PR #5152, opened 2025-07-10
Remove unused GitHub API from `JobAPIUpdate` view

mikerkelly:
The last reference to this in the view code was removed in commit ddd62303f82d38347c88ccbc295dbf6186b16d8d when removing OpenSafely Interactive notifications. So it's no longer required.


StevenMaude (approved):
Good spot :mag:; I'd missed that originally.

PR #5157, opened 2025-07-10
Enable `db` for unit/integration tests. Remove `disable_db`.

mikerkelly:
The `enable_db_access_for_all_tests` fixture in `tests/conftest.py` wasn't working when tests were run in isolation. Change the approach to instead put `autouse=True` `db` fixtures in `tests/unit/conftest.py` and `test/integration/conftest.py` where they are needed.

Remove `disable_db` marker that disallowed test scopes from accessing the DB. There were no tests other than the verification ones that were marked up. They can get the same effect more simply by not being affected by the new `conftest.py`. They won't spin up a test database when run separately from the unit/integration tests.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -62,12 +62,10 @@ We have a custom `@pytest.mark.slow_test` marker that should be used on test
 that take more than about 0.1s to run. These are excluded from `just test` to
 provide a quicker feedback loop.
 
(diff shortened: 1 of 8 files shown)


rebkwok (approved):
:+1: Definitely prefer this to the previous way (even had it not been broken for tests in isolation).

PR #5167, opened 2025-07-16
Use non-default Vite `port` to avoid clashes

mikerkelly:
Before this change, if you ran `run-assets` while another Vite server was running somewhere on your local development system on also on the default port, the new Vite server would automatically pick another free port, which would then not match the origin configured and assets wouldn't load.

This would happen if trying to run `job-server` alongside `opencodelists`, for example.

Using port `8001` as this one more than `8000` used by the `just run` webserver, so it's in the same "namespace".

Some templates and settings have the port hardcoded in dev mode. Change those and therefore use `strictPort` as it's required for those to work.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -213,7 +213,7 @@ To run the development server:
 1. Update the `.env` file to `ASSETS_DEV_MODE=True`
 2. Run `just assets-run`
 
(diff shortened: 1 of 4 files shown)


mikerkelly:
I wasn't sure this was actually working right locally, closing for now (should have been a draft PR, oops).

Issue #5068, opened 2025-04-09
Django 5.2 Upgrade


Jongmassey:
# Why are we doing this?

We are currently running Django 5.1, for which the end of extended support.
Django 5.2 was [released](https://docs.djangoproject.com/en/5.2/releases/5.2/) 2025-04-02.

Our dependency updating system is [starting to suggest](https://github.com/opensafely-core/job-server/pull/5058/files) upgrading to 5.2.

There are some major changes in 5.2 which might cause us problems, most notably that we currently [use](https://github.com/opensafely-core/job-server/blob/a72d03669d48398ffb858f4b46dbf2cc3448cec2/docker/docker-compose.yaml#L5) Postgres 13 and the 5.2 release notes state

>Django 5.2 supports PostgreSQL 14 and higher

# How will we know when it's done?

When Job Server runs successfully on Django 5.2 in production.

# What are we doing?

- [ ] Pin django to `>=5.1,<5.2` such that other dependency updates may proceed.
- [ ] Test the upgrade to 5.2
- [ ] Un-pin django to allow updates to proceed

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

StevenMaude:
This should be unblocked now.

lucyb:
Out of interest, did you try using django-upgrade to do this? It's currently a dev dependency and I'm wondering if it's actually needed?

Issue #4739, opened 2024-11-13
Update PostgreSQL to newer version


StevenMaude:
The PostgreSQL version in use is apparently 13, judging from the Docker configuration, as introduced here [back in 2021](https://github.com/opensafely-core/job-server/commit/096c8b60d359661e27c9d2b955b715311fb67795).

It's not been updated to a new major version since, as far as I can tell, although I can't check what Digital Ocean is actually running.

The [end of life for PostgreSQL 13](https://www.postgresql.org/support/versioning/) is a year today — 2025-12-23 — so we should update before then.

This will require work along the lines of:

* some local testing to see if there any issues with upgrading the database
* fixing any issues if there are
* upgrading the PostgreSQL version in `docker-compose.yml`
* backing up the database immediately prior to upgrade (this may happen automatically on upgrade)
* upgrading the database via Digital Ocean, which apparently can happen with no downtime
* confirming the upgrade has succeeded
* developers working on job-server upgrading their local setups, if they're not using Docker to run the database

One potential issue depending on how access to the current database is controlled; for password authentication, [newer versions of PostgreSQL enforce a newer password authentication scheme than older versions](https://www.percona.com/blog/postgresql-14-and-recent-scram-authentication-changes-should-i-migrate-to-scram/). It may be that any such database credentials need reconfiguring before or after the PostgreSQL upgrade.

StevenMaude:
To answer my own question on "what do Digital Ocean do when a PostgreSQL is end-of-life?", it looks likely that Digital Ocean immediately drops support and automatically upgrades existing databases in the next upgrade window.

That's what [they did with PostgreSQL 12](https://docs.digitalocean.com/release-notes/#2024-11-14).

lucyb:
There's some useful additional information [here](https://bennettoxford.slack.com/archives/C63UXGB8E/p1752075783192209?thread_ts=1752075774.840209&cid=C63UXGB8E), including confirmation that they'll automatically upgrade us to PG16. It's worth reading carefully as it contains some important tasks to do before upgrading (e.g upgrading the extensions first).

StevenMaude:
Taking my name off this for now as I'm not actively working on it, but I'll return to it, if no-one else picks it up.

StevenMaude:
I did a slightly shonky test of migrating the data with `pg_upgrade`.

This could be refined, but I'll add my notes for reference.

🔨 indicates parts of the process that could be improved.

# Steps to migrate a database locally for testing

[...]

PR #5169, opened 2025-07-17
Switch to base-docker 24.04

StevenMaude:
This upgrades to newer Ubuntu and gets us a newer version of the
PostgreSQL client which should be backwards compatible with our existing
database, while allowing us to upgrade to PostgreSQL 16 in future.

This avoids the additional step of having to use the PostgreSQL
repository with Ubuntu 22.04 to get a suitable PostgreSQL client.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -9,7 +9,7 @@ ARG NODE_VERSION
 # the latest base image, by design.
 #
 # hadolint ignore=DL3007
-FROM ghcr.io/opensafely-core/base-docker:22.04 as base-python
+FROM ghcr.io/opensafely-core/base-docker:24.04 as base-python
 
 # we are going to use an apt cache on the host, so disable the default debian
 # docker clean up that deletes that cache on every apt install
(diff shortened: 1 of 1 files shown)


StevenMaude:
This is one approach to getting a newer PostgreSQL client, which we'll want if we're to address #4739, which in turns allows us to upgrade to Django 5.2 (#5068).

If this change might be potentially disruptive to any other ongoing work, the other alternative is adding the PostgreSQL repositories and sticking with the existing base-docker:22.04 image, and we can always do this upgrade later.

StevenMaude:
From the [PostgreSQL documentation](https://www.postgresql.org/docs/16/app-psql.html):

> * psql works best with servers of the same or an older major version. Backslash commands are particularly likely to fail if the server is of a newer version than psql itself. However, backslash commands of the `\d` family should work with servers of versions back to 9.2, though not necessarily with servers newer than psql itself. The general functionality of running SQL commands and displaying query results should also work with servers of a newer major version, but this cannot be guaranteed in all cases.
>
>   If you want to use psql to connect to several servers of different major versions, it is recommended that you use the newest version of psql. Alternatively, you can keep around a copy of psql from each major version and be sure to use the version that matches the respective server. But in practice, this additional complication should not be necessary.

StevenMaude:
`psql` appears to work locally when running everything in Docker:

```
$ docker exec -it "$CONTAINER_ID" psql postgres://user:pass@db:5432/jobserver
psql (16.9 (Ubuntu 16.9-0ubuntu0.24.04.1), server 13.21 (Debian 13.21-1.pgdg120+1))
Type "help" for help.

jobserver=# 
```

I also tried some simple commands.

lucyb (approved):

PR #5176, opened 2025-07-23
Upgrade to JS linter Biome version 2

### LINKED ISSUE #5151: Upgrade to Biome v2
## Why are we doing this?

[Biome v2 has been released](https://biomejs.dev/blog/biome-v2/) and we should upgrade to using it, so that we receive the latest security updates and features.

## How will we know when it's done?

When v2 is used in the schema and for fixing our JS/TS files. When we moved from ESLint to Biome, we brough over a lot of legacy rules. Where possible we should try to simplify the ruleset and/or use the default rules.

## What are we doing?

- Upgrade to v2
- Update the schema to v2
- Remove as many unnecessary rules as possible

---

tomodwyer:
Closes #5151


package.json
```diff
@@ -13,9 +13,9 @@
     "test:single": "npx vitest -t"
   },
   "devDependencies": {
-    "@biomejs/biome": "^1.9.4",
     "@testing-library/dom": "^10.4.1",
     "@testing-library/jest-dom": "^6.6.4",
+    "@biomejs/biome": "^2.1.2",
```
rw251, line 19: and there was me thinking you were well into stuff like this being alphabetical :stuck_out_tongue_winking_eye:

PR #5184, opened 2025-07-30
Make notifications about github issues more robust (rebase fix)

Providence-o:
Copy of #5130 created to fix issues post-rebase.

Fixes https://github.com/opensafely-core/airlock/issues/952

Currently, if something goes wrong with creating/updating a github issue, we don't attempt to retry the process and we are not promptly alerted.


airlock/issues.py
```diff
@@ -36,6 +37,16 @@ def ensure_issue_status_labels(org, repo, github_api):
     return repo_labels
 
 
+def get_github_max_retry():
+    return settings.DEFAULT_MAX_GITHUB_RETRIES
```
rebkwok, line 41: Why do we need a function to just return the setting?  When it was used as `max_tries` in the `backoff` library, it could be a callable, but didn't have to be.  Now it's just being called in `stamina.retry`, so I don't think the function adds anything useful?

airlock/issues.py
```diff
@@ -204,3 +186,39 @@ def update_output_checking_issue(
...
+
+def update_output_checking_issue(
+    release_request_id,
+    workspace_name,
+    updates,
+    org,
+    repo,
+    github_api,
+    notify_slack,
+    request_author,
```
rebkwok, line 199: It would be nice not to have to pass around all these arguments (especially as they're positional args, so order matters).  

Probably out of scope for this ticket, but I think we could reuse the AirlockEvent object. It has most of the things that we pass to the output checking issues. And updates could be fetched in the issues functions instead of being called first in views.py.

rebkwok (approved):
A couple of minor comments (one of them not for actionning, as I think it's out of scope). 

Thanks for the nice modular commit re-work on this PR, made reviewing super easy!

PR #5189, opened 2025-08-01
Fix verification test failures introduced in #5184

Jongmassey:
This fixes a failing verification test  introduced in #5184 that expects a `IssueNotFound` to be raised when no issues matching the given title are found. Prior to this, the non-existence of the issue was not handled correctly and subsequent API calls would try to append a comment to a nonexistent issue, resulting in an unhandled HTTP 404 error.

### DIFF
--- a/jobserver/github.py
+++ b/jobserver/github.py
@@ -315,6 +315,8 @@ class GitHubAPI:
 
         if count > 0:
             return items[0]["number"]
+        else:
+            raise IssueNotFound()
 
     def create_issue_comment(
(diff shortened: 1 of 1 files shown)


Jongmassey:
My first attempt at this fix actually causes problems elsewhere. Going to put this down for now as I've got other more urgent things to be looking at.

rebkwok:
I had already fixed this in a PR that apparently didn't get merged to the problem branch; should be able to extract it

PR #5203, opened 2025-08-11
Add defaults for new RAP settings

rebkwok:
Add a default for new RAP settings. This is set to an empty string as we expect it always to be set from environment variables, both in dev and prod, and we don't need/want it to be a valid default. The default ensures that the Airlock dev integration continues to work and doesn't error due to missing env variables that it doesn't use.


mikerkelly (approved):
If the environment variable is not set then the management commands, crons, and sync loop(s) that use the API will fail, but a local developer will only need them to work if doing end-to-end testing with local RAP controller, in which case they will need to set the settings to point at it. So defaulting to "" seems fine.

PR #5220, opened 2025-08-20
Fix: correct cancel job request button text

BrenoNunes96:
This pull request fixes the text on the "Cancel this job request" button in the `detail.html` template. 

### Changes:
- Updated button label when the job can still be cancelled.
- Improves clarity for end users interacting with job requests.

No functional logic was changed — this is a UI text fix only.


lucyb:
Thank you for the suggested change. Unfortunately, we aren't currently set up to allow contributions from external developers, so we are unable to accept the PR at this time.

PR #5233, opened 2025-08-28
Add `jobserver/rap_api.py` and `cancel_jobs` management command.

### LINKED ISSUE #5206: Job Server to use cancel endpoint via a management command
# Why are we doing this

We have an endpoint for cancelling jobs. Lets make Job Server use it in stages:

- [ ] Management command that can trigger a cancel request to the RAP API.
- [ ] #5216
- [ ] Once that's tested in production, enable it for all backends. Should be fine anyway as endpoint is idempotent so doesn't matter if sync loop and endpoint cancel the same job/action.
- [ ] Remove the code from the controller's sync loop.

Previous ticket: https://github.com/opensafely-core/job-runner/pull/1131

# What are we doing

This Issue is for the first bullet. It should be possible to cancel a job running on the test backend via a management command, using the new Cancel endpoint. Test it in dev and production as well as automated tests.

Raising issue for the next/later bullet points if they don't already exist.

mikerkelly:
Fixes #5206.

Also adds `jobserver/rap_api.py` to centralize the code that actually makes the different API calls and their shared functionality.

### DIFF
--- a/jobserver/management/commands/cancel_jobs.py
+++ b/jobserver/management/commands/cancel_jobs.py
@@ -1,0 +1,37 @@
+"""Management command to cancel Jobs via the RAP API."""
+
+import structlog
+from django.core.management.base import BaseCommand
+
+from jobserver import rap_api
+
+
+class Command(BaseCommand):
+    """Management command to cancel Jobs via the RAP API."""
+
+    help = "Cancel RAP jobs via the RAP API."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "job_request_id",
+            type=str,
+            help="ID of the JobRequest to cancel.",
+        )
+        parser.add_argument(
+            "actions",
+            nargs="+",
+            type=str,
+            help="List of actions to cancel for the given JobRequest.",
+        )
+
+    def handle(self, *args, **options):
+        logger = structlog.get_logger(__name__)
+        try:
+            logger.info(
+                rap_api.cancel(
+                    options["job_request_id"],
+                    options["actions"],
+                )
+            )
+        except Exception as exc:
+            logger.error(exc)
--- a/jobserver/management/commands/cancel_jobs.py
+++ b/jobserver/management/commands/cancel_jobs.py
@@ -13,15 +13,15 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
-            "job_request_id",
+            "rap_id",
             type=str,
-            help="ID of the JobRequest to cancel.",
+            help="ID of the RAP to cancel. Equivalent to JobRequest.identifier.",
         )
         parser.add_argument(
             "actions",
             nargs="+",
             type=str,
-            help="List of actions to cancel for the given JobRequest.",
+            help="List of actions to cancel for the given RAP.",
         )
 
     def handle(self, *args, **options):
@@ -29,7 +29,7 @@ class Command(BaseCommand):
         try:
             logger.info(
                 rap_api.cancel(
-                    options["job_request_id"],
+                    options["rap_id"],
                     options["actions"],
                 )
             )
--- a/jobserver/rap_api.py
+++ b/jobserver/rap_api.py
@@ -1,0 +1,87 @@
+"""Methods to interact with the RAP Controller via its API.
+
+These allow control and query of Reproducible Analytics Pipelines executed in a
+Backend to service a Job Request.
+
+Reference for how the server defines the API endpoints:
+https://controller.opensafely.org/controller/v1/docs/
+https://github.com/opensafely-core/job-runner/tree/main/controller/webapp/api_spec
+
+The functions corresponding to each endpoint should not leak HTTP-layer
+information or the implementation detail that we use the requests library. So
+they should return subclasses of RapAPIError for relevant status codes or other
+errors, for example. They should probably not need to transform the response
+body defined in the spec. Clients of this module will interpret those further.
+"""
+
+from urllib.parse import urljoin
+
+import requests
+from django.conf import settings
+
+
+class RapAPIError(Exception):
+    """Base exception for this module. A problem contacting or using the RAP
+    API."""
+
+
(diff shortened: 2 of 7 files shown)


jobserver/rap_api.py
```diff
@@ -59,6 +63,7 @@ def _api_call(request_method, endpoint_path):
         response = request_method(
             urljoin(settings.RAP_API_BASE_URL, endpoint_path),
             headers={"Authorization": settings.RAP_API_TOKEN},
+            json=json,
```
madwort, line 78: As discussed before, I think it's not /really/ valid to do a GET request with a json, although I don't know off-hand what requests will do with it. If job-runner doesn't present any GET endpoints that are willing to receive it, I don't think it'll come up, it's just slightly weird to see this generic `request_method` and then have specific parameters passed to it. Is it overkill to throw a RapAPISettingsError if request_method is GET & json is not None?

tests/unit/jobserver/management/commands/test_cancel_jobs.py
```diff
@@ -0,0 +1,32 @@
...
+        "jobserver.rap_api.cancel",
+        lambda job_request_id, actions: test_response_body,
+    )
+
+    call_command("cancel_jobs", *_FAKE_ARGS)
+
+    assert log_output.entries[0] == {
+        "event": test_response_body,
+        "log_level": "info",
+    }
```
madwort, line 20: this test does so little that I'm /almost/ tempting to replace it with `pragma: nocover`.... we should keep it though.

jobserver/rap_api.py
```diff
@@ -0,0 +1,128 @@
...
+body defined in the spec. Clients of this module will interpret those further.
+"""
+
+from urllib.parse import urljoin
+
+import requests
+from django.conf import settings
+
+
+class RapAPIError(Exception):
```
madwort, line 23: only out of curiosity - why `*Error` rather than `*Exception`?

jobserver/rap_api.py
```diff
@@ -0,0 +1,128 @@
...
+    if not (settings.RAP_API_BASE_URL and settings.RAP_API_TOKEN):
+        raise RapAPISettingsError(
+            "A required environment variable RAP_API_BASE_URL and/or RAP_API_TOKEN is not set"
+        )
+
+    # Check the method is one covered in the tests of this module.
+    allowed_methods = {
+        requests.get,
+        requests.post,
+    }
```
madwort, line 65: I think this is a _little_ bit overkill, but I am pleased to see it here

jobserver/management/commands/cancel_jobs.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class Command(BaseCommand):
+    """Management command to cancel Jobs via the RAP API."""
+
+    help = "Cancel RAP jobs via the RAP API."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "job_request_id",
```
madwort, line 16: I know we're building in inconsistency, but
```suggestion
            "rap_id",
```

tests/unit/jobserver/management/commands/test_check_rap_api_status.py
```diff
@@ -4,62 +4,52 @@
...
-    settings.RAP_API_TOKEN = "bad_token"
-    settings.RAP_API_BASE_URL = TEST_BASE_URL
-
+def test_check_rap_api_status_bad_token(rap_api_token):
     with responses.RequestsMock() as rsps:
         rsps.add(
             responses.GET,
             TEST_STATUS_URL,
-            match=[responses.matchers.header_matcher({"Authorization": "bad_token"})],
+            match=[responses.matchers.header_matcher({"Authorization": rap_api_token})],
```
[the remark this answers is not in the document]
mikerkelly: Yep makes sense, can override the setting in this one.

jobserver/rap_api.py
mikerkelly: I think we may as well do this.

All the current `RapAPIError` are about the environment external to the process (environment variables for the settings, the network comms for the others).

I'm currently throwing `ValueError` for disallowed `request_method` parameters as I'm anticipating that being a programming error rather than something we see in the normal course of operation. I can make a module-specific one for that and use it here as well. They're both about bad parameters (in this case, a bad combination of parameters).

jobserver/rap_api.py
mikerkelly: Reasonable question, no particularly good reason. They do all indicate error conditions in this module I think it's accurate so far at least.

I tend to see about a 50-50 split in people calling application-defined subclasses of `Exception` `*Error` or `*Exception` and that seems to be the case in both `job-server` and `job-runner` also.

jobserver/management/commands/cancel_jobs.py
mikerkelly: Yes, let's mirror the remote API params. And the `help_text` can hint about the situation.

tests/unit/jobserver/management/commands/test_check_rap_api_status.py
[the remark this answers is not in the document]
mikerkelly: Oh, the test doesn't exist anymore. So nothing to do now I think.

madwort (approved):

tests/unit/jobserver/test_rap_api.py
```diff
@@ -119,6 +119,15 @@ def test_requests_get(self, rap_api_base_url, rap_api_token):
...
         assert response.json() == response_body
 
+    def test_request_get_and_json(self):
+        """Test that request_method requests.get disallowed with JSON parameter.
+
+        A payload within a GET request message has no defined semantics.
+        """
+        path = "some/path/"
+        with pytest.raises(ValueError, match="request_method not allowed"):
+            _api_call(requests.put, path)
```
madwort, line 129: `_api_call(requests.get, path)`

madwort (approved):

PR #5256, opened 2025-09-02
Fix `cancel` API wrapper method and improve tests

mikerkelly:
`cancel` needs to send a POST request to the RAP API.

Add tests that the endpoints actually use the right parameters, as well as how they handle the different status codes.

### DIFF
--- a/tests/unit/jobserver/test_rap_api.py
+++ b/tests/unit/jobserver/test_rap_api.py
@@ -171,14 +171,23 @@ def patch_api_call(monkeypatch):
 
     Usage:
         patch_api_call(fake_json, status_code=200)
+
+    Returns:
+        called: dict with params the tested function called api_call with
     """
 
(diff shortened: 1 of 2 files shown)


tests/unit/jobserver/test_rap_api.py
```diff
@@ -171,14 +171,23 @@ def patch_api_call(monkeypatch):
...
-            "jobserver.rap_api._api_call", lambda *args, **kwargs: fake_response
-        )
-        return fake_response
+        called = {}
+
+        def fake_api_call(request_method, endpoint_path, json=None):
+            called["request_method"] = request_method
+            called["endpoint_path"] = endpoint_path
+            called["json"] = json
+            return fake_response
```
madwort, line 187: I feel like this could be (waves hands) some kind of generic mock object?

PR #5257, opened 2025-09-03
Improve `rap_api.cancel` exception message

mikerkelly:
Including the `details` from the response body will make it slightly easier to debug what went wrong on the controller side.


madwort (commented):
Useful. I would probably add this to the test:

```
assert exc.match('gory')
```

madwort (approved):

Issue #5228, opened 2025-08-27
Create management command to call `/rap/status` endpoint


lucyb:
# Why are we doing this?

We are migrating in stages:

* Current status: sync loop runs in the rap-controller & sends requests to job-server
* This ticket: provide a manual command in job-server to call the RAP API on job-controller to update all jobs corresponding to a specified list of RAPs
* End goal: a new system which runs in job-server and calls the RAP API on the rap-controller

# How will we know when it's done?

* We can run a management command on job-server which updates job-server's knowledge of the state of the jobs as understood by the rap-controller.

# What are we doing?

* [x] Extend the rap_api.py in job-server as necessary to support the `rap/status` endpoint
* [x] Create a new management command to call the `rap/status` endpoint & dump the results to the terminal
* [ ] Extend the management command to update job-server's knowledge of the jobs state
* [ ] Deploy & verify on the test-backend

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

madwort:
tested the management command on the production server with some rap_ids corresponding to the test backend

PR #5260, opened 2025-09-04
Create rap/status management command

madwort:
* will fix https://github.com/opensafely-core/job-server/issues/5228

### DIFF
--- a/jobserver/management/commands/rap_status.py
+++ b/jobserver/management/commands/rap_status.py
@@ -1,7 +1,5 @@
 """Management command to update the status of a RAP's Jobs via the RAP API."""
 
-import json
-
 import structlog
 from django.core.management.base import BaseCommand
 
@@ -24,10 +22,8 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         logger = structlog.get_logger(__name__)
         try:
-            json_response = json.loads(
-                rap_api.status(
-                    options["rap_ids"],
-                )
+            json_response = rap_api.status(
+                options["rap_ids"],
             )
 
             for job in json_response["jobs"]:
--- a/jobserver/management/commands/rap_status.py
+++ b/jobserver/management/commands/rap_status.py
@@ -6,6 +6,9 @@ from django.core.management.base import BaseCommand
 from jobserver import rap_api
 
 
+logger = structlog.get_logger(__name__)
+
+
 class Command(BaseCommand):
     """Management command to update the status of a RAP's Jobs via the RAP API."""
 
@@ -20,7 +23,6 @@ class Command(BaseCommand):
         )
 
     def handle(self, *args, **options):
-        logger = structlog.get_logger(__name__)
         try:
             json_response = rap_api.status(
                 options["rap_ids"],
--- a/jobserver/management/commands/rap_status.py
+++ b/jobserver/management/commands/rap_status.py
@@ -1,0 +1,41 @@
+"""Management command to update the status of a RAP's Jobs via the RAP API."""
+
+import json
+
+import structlog
+from django.core.management.base import BaseCommand
+
+from jobserver import rap_api
+
+
+class Command(BaseCommand):
+    """Management command to update the status of a RAP's Jobs via the RAP API."""
+
+    help = "Update RAP's jobs status via the RAP API."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "rap_ids",
+            nargs="+",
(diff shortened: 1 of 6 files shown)


rebkwok (approved):
Have left a question, but it's more general than this PR

jobserver/management/commands/rap_status.py
```diff
@@ -0,0 +1,41 @@
...
+            "rap_ids",
+            nargs="+",
+            type=str,
+            help="IDs of the RAPs to update.",
+        )
+
+    def handle(self, *args, **options):
+        logger = structlog.get_logger(__name__)
+        try:
+            json_response = json.loads(
```
rebkwok, line 27: rap_api.status() is returning response.json(), which does the json decoding, so I think json.loads() will error here because it's expecting a string

tests/unit/jobserver/management/commands/test_rap_status.py
```diff
@@ -0,0 +1,47 @@
+from django.core.management import call_command
+
+
+def test_command(monkeypatch, log_output):
+    test_response_body = b'{"jobs": [{"identifier": "abc123", "rap_id": "abcdefgi12345678", "status": "succeeded"}], "unrecognised_rap_ids": []}'
+    test_log_output = "RAP: abcdefgi12345678 Job: abc123 Status: succeeded"
+
+    monkeypatch.setattr(
+        "jobserver.rap_api.status",
+        lambda rap_ids: test_response_body,
```
rebkwok, line 10: this needs to be a dict to mock what response.json() returns in rap_api.status

tests/unit/jobserver/test_rap_api.py
```diff
@@ -247,3 +248,37 @@ def test_bad_status_code(self, patch_api_call):
...
+        assert result == fake_json
+        mock_api_call.assert_called_once_with(
+            requests.post,
+            "rap/status/",
+            {
+                "rap_ids": self._fake_args,
+            },
+        )
+
+    def test_bad_status_code(self, patch_api_call):
```
rebkwok, line 277: Could test what happens if the response body isn't json? (Again, probably not for this PR)

jobserver/management/commands/rap_status.py
madwort: ah, you're right. I think this is a problem with the tests which I've unwittingly cribbed from the cancel endpoint - in the tests `response.json()` was returning a bytearray not a dict. In cancel it doesn't matter because it just dumps the value to logging, but I'm actually using it here. I will fix here & then fix the tests for cancel. 

For ref I've just run it outside of the tests, and it fails thusly:
```
$ just manage rap_status ht5xd5jeei4lnwnf d5uenk57whnd4rdc 1234567812345678
$BIN/python manage.py rap_status ht5xd5jeei4lnwnf d5uenk57whnd4rdc 1234567812345678
2025-09-05T09:52:02.555867Z [error    ] the JSON object must be str, bytes or bytearray, not dict
```

tests/unit/jobserver/test_rap_api.py
madwort: yes will do while I'm doing

tests/unit/jobserver/test_rap_api.py
```diff
@@ -157,12 +158,19 @@ def test_requests_post(self, rap_api_base_url, rap_api_token):
 class FakeResponse:
     """Lightweight fake response similar enough to requests.Response for these tests."""
 
-    def __init__(self, json_data, status_code=200):
-        self._json_data = json_data
+    def __init__(self, content=None, json_data=None, status_code=200):
+        if content and json_data:
+            raise ValueError("Tests should only specify content *or* json_data")
+
+        # Nb. in the real Response .content is a byte array, not a string
```
rebkwok, line 165: Should we make it a byte array in tests too then?

madwort: This would currently only really affect the error 500 codepaths, as all other tests interact with FakeResponse using only dicts - so it didn't seem to matter too much if it does dict->str->dict compared to dict->str->bytearray->str->dict for those tests.

With the current error 500 paths, I think it's currently resilient to whether it's a str or a bytearray (by which I mean it's sending it to a logger which I think will happily receive either), but maybe if it's worth doing it's worth doing accurately.

madwort: done

rebkwok (approved):

PR #5261, opened 2025-09-05
Test the `unpublish_snapshot` management command

iaindillingham:
46c67c7 added the `unpublish_snapshot` management command. For expediency, however, it did not add any tests. This commit adds them, and no longer omits `unpublish_snapshot` from coverage.

### DIFF
--- a/tests/unit/jobserver/management/commands/test_unpublish_snapshot.py
+++ b/tests/unit/jobserver/management/commands/test_unpublish_snapshot.py
@@ -1,0 +1,42 @@
+import pytest
+from django.core.management import call_command
+from django.core.management.base import CommandError
+
+from tests.factories import PublishRequestFactory, SnapshotFactory, UserFactory
+
+
+def test_command_no_snapshot():
+    with pytest.raises(CommandError, match="Snapshot '1' does not exist"):
+        call_command("unpublish_snapshot", 1, "a-username")
+
+
+def test_command_no_user():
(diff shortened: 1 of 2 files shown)


tests/unit/jobserver/management/commands/test_unpublish_snapshot.py
```diff
@@ -0,0 +1,42 @@
+import pytest
+from django.core.management import call_command
+from django.core.management.base import CommandError
+
+from .....factories import PublishRequestFactory, SnapshotFactory, UserFactory
```
lucyb, line 5: Minor thing, but in other tests we've done `from tests.factories import PublishRequestFactory` (etc) instead. This means our imports won't break if these tests are moved. It's also, I think, slightly clearer to read.

lucyb (approved):

tests/unit/jobserver/management/commands/test_unpublish_snapshot.py
iaindillingham: Ah, okay, I thought I was being consistent with other tests. For example:

https://github.com/opensafely-core/job-server/blob/f24c8833a9940435da6d8d0f83a4fb11c5698d2a/tests/integration/test_release_api_access.py#L4-L10

I agree with you! I'll change it. Switching from relative to absolute imports would be a useful, but boring, deck scrubbing task.

PR #5268, opened 2025-09-08
Get job-server to use the backend status endpoint implementation

### LINKED ISSUE #5207: Make Job Server use the backend status endpoint
# Why are we doing this?

We've added a backend status endpoint and a cron that contacts it and logs the results. Now we can change the Job Server behaviour to actually update its database based on the response.

Previous tickets:
- Adding the endpoint: https://github.com/opensafely-core/job-runner/issues/1107
- Adding a stub cron: https://github.com/opensafely-core/job-server/issues/5172
- Make the endpoint more useful: https://github.com/opensafely-core/job-runner/issues/1134
- Use the updated endpoint: https://github.com/opensafely-core/job-server/issues/5200

# What are we doing?

- Changing the management command / related code so that it (optionally?) actually updates Job Server's database.
- Checking this works as expected in a dev environment and production.
- Possibly removing the flags-related code from the controller's `sync` loop.
- Possibly moving the "contact the endpoint" code out into a separate module or package that can be partially shared between the Job Server clients of the RAP API.

#5190 cares about Flags -- make sure to implement in a way that doesn't clash and discuss with Katie if that PR isn't merged by the time this is done whether she or we want to update that new code to use the new-style flags from https://github.com/opensafely-core/job-runner/issues/1134

Providence-o:
Fixes #5207


jobserver/management/commands/check_rap_api_status.py
```diff
@@ -17,25 +17,23 @@ def handle(self, *args, **options):
...
     def update_backend_state(backend, request):
-        # Retrieve backend state sent up from job-runner. We might rename the header this is
-        # passed in at some point but for now this is good enough.
-        flags = rap_api.backend_status().get("backends", "")
+        # Retrieve the response gotten from the jobrunner endpoint
+        flags = rap_api.backend_status().get("backends", [])
 
-        # if not flags:  # this line should never be hit because we have a default dict compared with the old way
-        #     return
+        if not flags:
```
rebkwok, line 23: If we get no backends back from the RAP API, something's probably gone wrong, hasn't it? If you send an invalid client token (i.e. one that doesn't allow access to any backends in the controller), you get a 401, not an empty list. (It's possible we might change that, but even so, an empty list would suggest a problem). Maybe worth at least logging an error before returning?

tests/unit/jobserver/management/commands/test_check_rap_api_status.py
```diff
@@ -1,19 +1,62 @@
+import datetime
```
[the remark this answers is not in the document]
Providence-o: Adding a test like this returns a KeyError but `test_command_error` would also log the same error. The difference is in passing the fake nested function that raises an exception vs fake data with the wrong key. Would this be considered a repetition? 
```
def test_command_missing_backend_key(log_output, monkeypatch):
     test_response_body = {
        "other_backend": [
            {
                "name": "test",
                "last_seen": "2025-08-12T06:57:43.039078Z",
                "paused": {
                    "status": "off",
                    "since": "2025-08-12T14:33:57.413881Z",
                },
                "db_maintenance": {
                    "status": "off",
                    "since": None,
                    "type": None,
                },
            }
        ]
    }

    monkeypatch.setattr(
        "jobserver.rap_api.backend_status",
        lambda: test_response_body,
    )
    call_command("check_rap_api_status")
    assert "error" == log_output.entries[0]["log_level"]
```

tests/conftest.py
```diff
@@ -431,15 +431,15 @@ def site_alert():
 
 @pytest.fixture(autouse=True)
 def rap_api_token(settings):
-    """Django settting for the token for communicating with the RAP API."""
+    """Django setting for the token for communicating with the RAP API."""
```
lucyb, line 434: I really like this small "quality of life" improvement. It's easy to skip over mistakes like this, but making these small changes add up over time.

jobserver/management/commands/check_rap_api_status.py
```diff
@@ -1,14 +1,66 @@
+import datetime
```
lucyb, line 1: I don't think it really makes a difference here, but a common convention is to do `from datetime import datetime`, so that references in your code go from `datetime.datetime.fromisoformat` to `datetime.fromisoformat`.

tests/unit/jobserver/management/commands/test_check_rap_api_status.py
```diff
@@ -28,3 +71,133 @@ def fake_backend_status():
...
+        "jobserver.rap_api.backend_status",
+        lambda: test_response_body,
+    )
+
+    call_command("check_rap_api_status")
+
+    backend1.refresh_from_db()
+
+    assert backend1.rap_api_state == test_response_body["backends"][0]
+    assert backend1.last_seen_at == datetime.datetime.fromisoformat(
```
lucyb, line 190: It might be worth adding an extra assert to check that `last_seen_maintenance_mode` is also None, since you're now explicitly setting it.

lucyb (approved):
I think this looks good to go. There's really only one small addition to the tests that I've suggested. Once it's deployed, you'll also need to think about how to check that it's all behaving as you expect before moving on to the next stage.

Finally, it's often useful to say why you're making a particular change (not just what you're changing) in a commit message and in the overall PR description. There's a really useful blog post that explains this in more detail: https://cbea.ms/git-commit/.

Issue #4904, opened 2025-02-20
Remove unused fields from Backend model and associated forms


rebkwok:
> I think this means we can probably remove the parent_directory field from the Backend model?
> https://github.com/opensafely-core/job-server/blob/main/jobserver/models/backend.py#L25
> 
> And associated forms: https://github.com/search?q=repo%3Aopensafely-core%2Fjob-server%20parent_directory&type=code
> 
> Whether that's part of this ticket, or a ticket for REX to do, I am not sure.
> 
> 
>  

 _Originally posted by @bloodearnest in [#1720](https://github.com/opensafely/documentation/issues/1720#issuecomment-2656282820)_

lucyb:
Thank you for raising the ticket. I think you're right to put it onto our board, but I could do with some more information. What is the `parent_directory` field and why is it safe to remove?

bloodearnest:
`Backend.parent_directory` used to be used to give the user a hint where to look in the filesystem on the backend when we used filesync. That message was removed in 89b4d75eb215a84504907984c983cd597dbee2d4 when we pointed to airlock instead.  Thus the column and associated form handling are no longer needed.

Also `Backend.level_4_url` was used when launching the outputs-viewer from a Level 4 browser, to the point the SPA at the API to load the file data. Whilst we still use the output-viewer SPA to view released outputs, we no longer use it for level 4, after 6e69d97275ace9d55a245148a00cfb3e7869c5e3, so it and associated form handling can also go I think.

bloodearnest:
Damnit, every time.

PR #5282, opened 2025-09-16
Remove usage of Backend.parent_directory and associated code

nishtha-kalra:
Addresses issue #4904 
This PR is only removing `parent_directory` from `Backend` model.
Will create another PR to remove `level_4_url` as I didn't want to have two fileds removed in 1 PR in case of rollback.


StevenMaude (commented):
This looks almost done to me; the code changes in themselves look fine, just a thought about how best to deploy the changes :+1: I also like that you restricted this PR to the single field.

PR #5311, opened 2025-09-29
Update URLs in `TESTING.md`

### LINKED ISSUE #5280: Update URLs in TESTING.md
# Why are we doing this?

Some of the URLs in [TESTING.md](https://github.com/opensafely-core/job-server/blob/main/TESTING.md) are outdated. For example, the Django documentation links currently point to version 3.2.

- All links in `TESTING.md` resolve correctly.- 
- Links go to the latest or currently-used versions of the referenced documentation.

# What are we doing?

- [x] Update the URLs in `TESTING.md` so they point to relevant documentation and resources.

---

KatieB5:
Closes #5280.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -39,7 +39,7 @@ Test modules mirror their non-testing counterparts, so tests covering functions 
 
 Each function/class should have at least one test, ideally one for each exit condition.
 
-For views we use [`RequestFactory` instances](https://docs.djangoproject.com/en/3.2/topics/testing/advanced/#the-request-factory) as the default testing method, reserving Client instances for Integration tests (more below), testing the Python responses of views under test.
+For views we use [`RequestFactory` instances](https://docs.djangoproject.com/en/5.1/topics/testing/advanced/#the-request-factory) as the default testing method, reserving Client instances for Integration tests (more below), testing the Python responses of views under test.
 
 *Author's Note:* in an ideal world testing views would be bumped up to Integration tests since they typically test across at least one boundary, the database.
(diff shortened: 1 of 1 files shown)


TESTING.md
```diff
@@ -39,7 +39,7 @@ Test modules mirror their non-testing counterparts, so tests covering functions
 
 Each function/class should have at least one test, ideally one for each exit condition.
 
-For views we use [`RequestFactory` instances](https://docs.djangoproject.com/en/3.2/topics/testing/advanced/#the-request-factory) as the default testing method, reserving Client instances for Integration tests (more below), testing the Python responses of views under test.
+For views we use [`RequestFactory` instances](https://docs.djangoproject.com/en/5.1/topics/testing/advanced/#the-request-factory) as the default testing method, reserving Client instances for Integration tests (more below), testing the Python responses of views under test.
```
StevenMaude, line 42: thought: The alternative here would be to use the `dev` URL ([example](https://docs.djangoproject.com/en/dev/topics/testing/advanced/#the-request-factory)), which would mean that we shouldn't need to update the URLs again (provided Django doesn't move them).

That does risk it being out of sync with the actual behaviour in the code. So, it's also entirely reasonable as you've done to use the version corresponding to the Django version we're using. We then need to update these URLs when we update Django (provided we remember), but that's OK.

KatieB5: Thanks very much for highlighting this Steve! I hadn't considered using the `dev` URL for the Django docs - I'd just checked the version of Django that we were using, and linked to the corresponding version of the docs.

That's also a good point about updating the URLs when we come to update Django - I've left a comment on #5068  to double check and update any Django documentation URLs when we update the Django version.

PR #5321, opened 2025-10-02
Disable Sentry integration when running `shell`

mikerkelly:
This command is mainly used by devs to interrogate the database and entering code manually is error-prone. Unhandled exceptions caused by this activity are unlikely to be interesting, so let's not raise Sentry issues in this case.

For example, [this ](https://ebm-datalab.sentry.io/issues/6918084872/?alert_rule_id=10884789&alert_type=issue&project=5443358&referrer=slack) Sentry issue was raised due to an ad hoc use of the script.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -6,6 +6,13 @@ import sys
 
 
 def main():
+    # Disable Sentry when running the shell. This is mainly used by devs to
+    # interrogate the database and typing code manually is error-prone.
+    # Unhandled exceptions caused by this activity are unlikely to be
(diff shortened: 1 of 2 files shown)


StevenMaude (approved):
Thanks; this looks reasonable and I appreciate the comprehensiveness of the tests here.

mikerkelly:
Thanks, Steve.

If the env var name or how it is used changes this may silently stop doing anything. Hopefully we would think to check in that case, but if not then we are no worse off than we at now.

PR #5322, opened 2025-10-02
Remove level_4_url from Backend model

nishtha-kalra:
This PR is part of issue #4904 
### Description
It removes all references and usages from job-server which are listed as follows:
1. removed usage from fixture
2. removed level_4 from workspace as this is no longer used to show if we have files in level 4 for someone to access. This is discussed [here](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1759222227577909)
3. removed usage from staff view
4. removed from templates so no longer visible on create, detail and edit page
5. remove field from Backend model
6. remove url from create_backend command
7. updated tests accordingly

### Testing
Manually tested in the following ways:
1. Checked that the field is not available on the create and edit forms for Backend
2. Workspace edit and view works without any issues
3. The `create_backend` command throws error now if `--url ` argument is used
4. Tests are updated accordingly

### Note
Completely removing the field will require migration. For safe rollout and avoiding risky migration, I will create a follow up PR after merging this which will cover migration.


jobserver/views/workspaces.py
```diff
@@ -226,24 +226,11 @@ def get(self, request, *args, **kwargs):
...
-        can_view_files = is_privileged_user and has_backends
-
+    def get_output_permissions(self, workspace):
         # are there any releases to show for the workspace?
         can_view_releases = workspace.releases.exists()
 
         return {
-            "level_4": {
-                "disabled": not can_view_files,
-            },
```
StevenMaude: question: I believe this is safe, because of 438b49cbb44739a582e51756560a40706d65e267 which removed the use of `outputs.level_4.disabled` and there is no current use of this part of the context otherwise in the template. Is my reasoning correct?

(It seems worth double checking that this change is correct.)

StevenMaude:
@nishtha-kalra: Thanks for cleaning this up.

This looks fine, though I had one thing to double check. That's not because I think the change is incorrect, but because it's a significant functional change, and I do not know this code too well.

Finally, it also needs a rebase on `main` and the conflicts fixing, and it will be good to merge then.

PR #5343, opened 2025-10-08
Update text on Job Request page for test backends

### LINKED ISSUE #5255: Job Request page should display different text for `test` Backend
# Why are we doing this?

This [job request page](https://jobs.opensafely.org/opensafely-internal/disparities-comparison-test/24884/) claims I am an authorised researcher and have run this JobRequest against real patient data, neither of which are true. This was against the `test` backend.

Image

We should change this text when the Backend does not involve sensitive data. This should be fairly easy to do. If we don't we give the impression we are running internal tests with patient data, which would be inappropriate.

- [Offending template](https://github.com/opensafely-core/job-server/blob/d963b56a33b006eb1e208c6ba58c6f10b43b4191/templates/job_request/detail.html#L109)
```
            This page shows the technical details of what happened when
            the authorised researcher
            <strong>{{ job_request.created_by.fullname }}</strong>
            requested one or more actions to be run against real patient data
            in the
            {% spaceless %}
              {% link text="project" href=job_request.workspace.project.get_absolute_url %}
            {% endspaceless %}, within a secure environment.
```
- Possibly Backend should get a field identifying whether they are sensitive.

Possibly relevant issues as they touch the `Backend` model: 
- https://github.com/opensafely-core/job-server/issues/4904
- https://github.com/opensafely-core/job-server/issues/4760

Note: we should let people know that this text has been updated (most likely in the NOD slack channel)

# What are we doing?

...

---

tomodwyer:
Closes #5255

- Reformats the existing text for job requests run against secure backends
- Adds new text for job requests run against the test environment

### Before

Screenshot of secure environment job request on production

Screenshot of test backend job request on production

### After

Screenshot of secure environment job request with text changes

Screenshot of test backend job request with text changes

### DIFF
--- a/templates/job_request/detail.html
+++ b/templates/job_request/detail.html
@@ -104,34 +104,46 @@
     <div class="space-y-6 md:space-y-6 lg:col-span-2">
       {% #card container=True %}
         <div class="prose prose-oxford">
-          <p>
-            This page shows the technical details of what happened when
-            the authorised researcher
-            <strong>{{ job_request.created_by.fullname }}</strong>
-            requested one or more actions to be run against real patient data within a secure environment.
-          </p>
-          <p>
-            By cross-referencing the list of jobs with the pipeline section below, you can infer what
-            <a href="https://docs.opensafely.org/security-levels/">security level</a>
-            the outputs were written to.
-          </p>
-          <p>The output security levels are:</p>
-          <ul>
-            <li>
-              <strong>highly_sensitive</strong>
-              <ul>
-                <li>Researchers can never directly view these outputs</li>
-                <li>Researchers can only request code is run against them</li>
-              </ul>
-            </li>
-            <li>
-              <strong>moderately_sensitive</strong>
-              <ul>
-                <li>Can be viewed by an approved researcher by logging into a highly secure environment</li>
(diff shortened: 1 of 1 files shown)


templates/job_request/detail.html
```diff
@@ -103,33 +103,47 @@ <h1 class="text-3xl tracking-tight break-words font-bold text-slate-900 sm:text-
...
-            marked as
-            {% #code %}moderately_sensitive{% /code %}
-            can be viewed by an approved researcher by logging into a highly
-            secure environment. Only outputs marked as
-            {% #code %}moderately_sensitive{% /code %}
-            can be requested for release to the public, via a controlled
-            output review service.
-          </p>
+        <div class="prose prose-oxford">
+          {% if job_request.backend.name == "Test" %}
```
StevenMaude, line 107: thought: I'm kind of surprised there isn't an attribute on the Backend model that says whether the instance is test only, but there isn't, so we have to hardcode the name here.

That isn't a concern specifically about this PR though; it just highlights it.

StevenMaude (approved):
The structure of the changes looks fine to me, but as Thomas said on Slack, we'll need the text reviewing before merging this one.

PR #5383, opened 2025-10-17
RAP API integration tests

### LINKED ISSUE #5281: RAP API: integration tests
# Why are we doing this?

We currently have unit tests of [rap_api.py](https://github.com/opensafely-core/job-server/blob/main/tests/unit/jobserver/test_rap_api.py) and the related management commands ([1](https://github.com/opensafely-core/job-server/blob/main/tests/unit/jobserver/management/commands/test_create_backend.py) [2](https://github.com/opensafely-core/job-server/blob/main/tests/unit/jobserver/management/commands/test_cancel_jobs.py) [3](https://github.com/opensafely-core/job-server/blob/main/tests/unit/jobserver/management/commands/test_check_rap_api_status.py) [4](https://github.com/opensafely-core/job-server/blob/main/tests/unit/jobserver/management/commands/test_create_rap.py)). These are useful but because the API functions are patched in the tests of the management commands, we don't test the interface between them or any other parts of the systems. Integration tests would address this. 

Happy path tests exist for each backend.

# What are we doing?

These tests should only patch requests, not the whole `rap_api.py` functions and other imported code.

When we've done this we might change some of the existing unit tests like `test_rap_status_update_notifications_on_with_move_to_succeeded` to mock the `jobserver.emails` function rather than exercising it and checking `django.core.mail.outbox` to make it more of an isolated unit test, and have the integration test check that condition when integrating all the code units.

---

rebkwok:
Fixes #5281

Adds integration tests that mock requests to the RAP API so we can more fully test the rest of the system's interface to the RAP API itself (rather than mocking the rap_api module).

Mostly this just tests the happy path, on the assumption that error cases are tested by unit tests elsewhere.

### DIFF
--- a/tests/integration/test_rap_api.py
+++ b/tests/integration/test_rap_api.py
@@ -10,6 +10,7 @@ import pytest
 import responses
 from django.conf import settings
 from django.core import mail
+from django.core.management import call_command
 from django.urls import reverse
 
 from jobserver.authorization import permissions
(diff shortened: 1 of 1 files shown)


tests/integration/test_rap_api.py
```diff
@@ -0,0 +1,195 @@
+"""
+These tests use responses to patch the requests module where it is used to make calls
+to the RAP API. This allows us to test that the interface between the rap_api module
+and other parts of the system.
```
madwort: > This allows us to test that the interface between the rap_api module
and other parts of the system.

typo

PR #5387, opened 2025-10-20
Update docs on RAP API/jobrunner interface

### LINKED ISSUE #5251: Overarching documentation if not already sufficient
# Why are we doing this?

...

...

# What are we doing?

...

---

rebkwok:
Partially fixes #5251 (along with updated job-runner docs)

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -33,7 +33,7 @@
 - [Auditing events](#auditing-events)
   - [Presenters](#presenters)
 - [Interfaces](#interfaces)
-  - [Job Runner interface](#job-runner-interface)
+  - [RAP API (RAP Controller interface)](#rap-api-rap-controller-interface)
   - [Airlock interface](#airlock-interface)
 - [Risky migrations](#risky-migrations)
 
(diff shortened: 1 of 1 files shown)


madwort (approved):

DEVELOPERS.md
```diff
@@ -525,23 +525,84 @@ Job Server API endpoint (https://jobs.opensafely.org/api/v2 in production).
 Management commands might be used in other repos' tooling and CI. While not
 required, it's helpful to check for downstream impacts if you change their API.
 
-### Job Runner interface
+### RAP API (RAP Controller interface)
```
mikerkelly: Think that "Except where mentioned otherwise, URLs are relative to the root of the Job Server API endpoint (https://jobs.opensafely.org/api/v2 in production)." from the previous section now only applies to the Airlock interface section. Should it be moved there?

DEVELOPERS.md
```diff
@@ -525,23 +525,85 @@ Job Server API endpoint (https://jobs.opensafely.org/api/v2 in production).
...
+The [RAP API] is specified using the OpenAPI specification and is documented publicly.
 
+Refer to the [Job Runner documentation] for further information on the design principles
+and architecture of the RAP API, Controller and Agent.
+
+#### Authorization
+
+Job Server provides an Authorization token (the `RAP_API_TOKEN` environment variable)
+in the header of any request to the RAP API. This allows requests access to data about
+one or more backends in the RAP Controller. Refer to the [RAP API Auth]
```
mikerkelly: > This allows requests access to data about one or more backends in the RAP Controller.

I found this a bit hard to follow. Access sounds like a read operation, but it also includes create/cancel. And it sounded  like the backend is in the RAP controller not the data.

Maybe

> This determines which backends the request may be about.

DEVELOPERS.md
```diff
@@ -525,23 +525,85 @@ Job Server API endpoint (https://jobs.opensafely.org/api/v2 in production).
...
+RAP API (`GET /controller/v1/rap/status/`) in a loop (with a 60s - or other delay defined by the
+`RAP_API_POLL_INTERVAL` environment variable - pause between repeats),
+and updates Job Server's `Job`s' status based on the job data returned in the RAP API response.
+
+"Active" job requests are those that are pending or running, and also those
+in an unknown status. If an initial job request creation failed with an unknown
+error, this allows Job Server to request an update from the RAP Controller and
+update the unknown status accordingly.
+
+```mermaid
```
mikerkelly: Not rendering for me in the preview via "view file" - looks ok though so maybe just me?

image

mikerkelly (approved):

PR #5393, opened 2025-10-21
Add a test for database image version in CI

### LINKED ISSUE #5358: Ensure that the version of PostgreSQL in the local Docker developer setups and CI are consistent
# Why are we doing this?

When reviewing the database upgrade (#4739) and subsequent unintentional upgrade, I noticed that the PostgreSQL version in CI is configured there separately from Docker.

I opened a PR to synchronise these (#5357) for now, but we should look to ensure that these are consistent.

When we are confident that the two versions are kept synchronised, or we have some CI check that alerts us if the versions do not match. 

# What are we doing?

Probably one of:

* adding an automated check somewhere that ensures this
* using the Docker Compose database configuration in CI

StevenMaude:
Fixes #5358.

### DIFF
--- a/tests/test_repository.py
+++ b/tests/test_repository.py
@@ -1,0 +1,27 @@
+"""Tests of the literal repository content."""
+
+import pytest
+import yaml
+
+
+@pytest.mark.slow_test
+def test_development_db_version_is_consistent(request):
+    """This test is to ensure that the separately configured database service
+    in the GitHub Actions test workflow is consistent with the Docker Compose
+    configuration provided for local development.
+
+    When updating the database version, update it in both of those configurations,
+    and in this test."""
+    root_path = request.config.rootpath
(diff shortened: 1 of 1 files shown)


mikerkelly:
Thanks, Steve. Test function looks good but two small things

- it takes 50ms+ to run because of the file system access, so I would like to mark it as a slow test. 
- It is not a verification test as defined in TESTING.md, so I would like to move it to `tests/test_repository` or similar:

> We use a fake object, `FakeGitHubAPI`, to test our uses of GitHub's API. However, we want to [verify that fake](https://pythonspeed.com/articles/verified-fakes/), so we have a set of verification tests in `tests/verification/`.

Steve is away for a bit so I will make these minor changes and merge.

mikerkelly (approved):

PR #5404, opened 2025-10-27
Remove environs dependency

### LINKED ISSUE #4770: Remove environs
As [was suggested for OpenCodelists](https://github.com/opensafely-core/opencodelists/issues/2222), the reasoning being an excess of transitive dependencies for functionality that is not too difficult to achieve without.

tomodwyer:
This removes the `environs[django]` dependency, replacing it with simplified calls to `os.environ.get`.

Some notes:

- where possible, a named keyword argument of `default` has been added
- `dj_database_url` dependency has been retained for the database string
- `env.list` replaced by conditional list comprehension
- ~`DEBUG` check is different in development to testing~
    - ~in dev, `DEBUG=True` is expected~
    - ~in testing, `DEBUG=1` is expected~
    - ~fixing this issue is outside of the scope of this change~

Closes #4770

### DIFF
--- a/jobserver/settings.py
+++ b/jobserver/settings.py
@@ -68,7 +68,7 @@ if OLD_SECRET_KEY is not None:
     SECRET_KEY_FALLBACKS = [OLD_SECRET_KEY]
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.environ.get("DEBUG", default=False) == "True"
+DEBUG = os.environ.get("DEBUG", default="0") == "1"
 
 DEBUG_TOOLBAR = os.environ.get("DJANGO_DEBUG_TOOLBAR", default=False) == "True"
 
--- a/services/logging.py
+++ b/services/logging.py
@@ -6,7 +6,7 @@ import structlog
 
 
 # add logging before app has booted
-DEBUG = os.environ.get("DEBUG", default=False) == "1"
+DEBUG = os.environ.get("DEBUG", default="0") == "1"
 
 
 def timestamper(logger, log_method, event_dict):


jobserver/settings.py
```diff
@@ -33,22 +57,22 @@
...
 # after the time in SESSION_COOKIE_AGE elapses.
 # Refer to INSTALL.md for guidance.
-OLD_SECRET_KEY = env.str("OLD_SECRET_KEY", default=None)
+OLD_SECRET_KEY = os.environ.get("OLD_SECRET_KEY", default=None)
 if OLD_SECRET_KEY is not None:
     SECRET_KEY_FALLBACKS = [OLD_SECRET_KEY]
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = env.bool("DEBUG", default=False)
+DEBUG = os.environ.get("DEBUG", default=False) == "True"
```
mikerkelly: > - `DEBUG` check is different in development to testing
>     - in dev, `DEBUG=True` is expected
>     - in testing, `DEBUG=1` is expected
>     - fixing this issue is outside of the scope of this change

Agree that making `DEBUG` truthy value consistent with the other bool settings is out of scope here. Could you explain more what the distinction between dev and testing is here?

I think `DEBUG` as an environment variable should have one interpretation in both settings and logging. The value in `dotenv-sample` used is `1` and I think this  will be the value set in CI as well (not sure how best to check). So can we use that consistently in both places?

Tests pass locally for me with the following in both places, does this work?

> DEBUG = os.environ.get("DEBUG", default="0") == "1"

tomodwyer: That worked 👌

Issue #5352, opened 2025-10-09
Bug report: unapplied migration


madwort:
# Summary
Running local dev app throws a migration error

**NOTE: check whether a newer version has fixed this issue**

# What did you do and what did you expect to happen?

I created a fresh db on postgres 17 and ran migrations from scratch. Then when I ran the app I saw this error:

```
job-server$ just run
/usr/bin/just _compile requirements.prod.in requirements.prod.txt 
+ test requirements.prod.txt -nt .venv/.prod
+ exit 0
/usr/bin/just _compile requirements.dev.in requirements.dev.txt 
$BIN/python manage.py migrate
Operations to perform:
  Apply all migrations: applications, auth, contenttypes, jobserver, redirects, sessions, social_django
Running migrations:
  No migrations to apply.
DJANGO_DEBUG_TOOLBAR=1 $BIN/python manage.py runserver localhost:8000
2025-10-09T15:39:03.249794Z [info     ] Watching for file changes with StatReloader [django.utils.autoreload]
Performing system checks...

System check identified no issues (0 silenced).

You have 1 unapplied migration(s). Your project may not work properly until you apply the migrations for app(s): debug_toolbar.
Run 'python manage.py migrate' to apply them.
October 09, 2025 - 15:39:04
Django version 5.1.13, using settings 'jobserver.settings'
Starting development server at http://localhost:8000/
Quit the server with CONTROL-C.

```

# Impact of the bug

This just started happening today after updating to the latest `main`, as far as I know it doesn't affect me.

StevenMaude:
Hopefully, v6.1.0 (and up) of django-debug-toolbar fixes this. Otherwise, it's possible to [configure a setting](https://docs.djangoproject.com/en/5.1/ref/settings/#std-setting-MIGRATION_MODULES) as [mentioned in a django-debug-toolbar discussion](https://github.com/django-commons/django-debug-toolbar/discussions/2234#discussioncomment-14780069), and that avoids the need for the migration too:

```
MIGRATION_MODULES = {"debug_toolbar": None}
```

PR #5433, opened 2025-11-14
Disable migrations for the django-debug-toolbar

tomodwyer:
Linked to https://github.com/opensafely-core/job-server/issues/5352


StevenMaude:
Hopefully this is no longer necessary; I've recently done clean installs of job-server and not seen the migration warning appear.

tomodwyer:
@StevenMaude we discussed in stand up that applying this for now is probably ok, as we know we don't want any migrations from the debug toolbar, and we can revisit if they do a migration we want.

nishtha-kalra (approved):

Issue #5166, opened 2025-07-16
Github verification tests setup permissions mismatch with production


bloodearnest:
We recently added some code to the airlock app in job-server that [adds labels to the issues we create](https://github.com/opensafely-core/job-server/pull/5158) to track Airlock requests.  As part of this code, we created the labels if they didn't exist. We also added support for creating labels to job-servers github api, and matching [verification test](https://github.com/opensafely-core/job-server/pull/5158/files#diff-cc69b5d6d5da257144b7de11d170f97eadbb4a75a80430b133043b3b87ec891bR213).

This passed the Github verification tests in CI, as can be seen here: https://github.com/opensafely-testing/github-api-testing-private/issues. However, it failed in production, as the opensafely-readonly-bot only has read level access to the repo it's trying to create issues in, so it gets a 403.

Our test used the opensafely-testing/github-api-testing-private repo, as our production repos for this use case are also private. However, the openafely-testing-bot has *write* level permissions to that repo, so it can create labels, to the verification tests passed.  It has write level permissions explicitly because it is in the [researchers team](https://github.com/orgs/opensafely-testing/teams/researchers).  

Clearly, creating labels is not something we can currently do from job-server, which is good to know, we will change our code to not try create them.

However, our question is, how can we fix the verification test setup to fail if we try do this kind of thing again, so we don't hit it production?

Should the testing-bot have write access to this repo for other reasons? If it's meant to emulate the readonly-bot permissions, then it probably shouldn't have write access to anything?

Reviewing the current tests, there are none that look like they *require* write permissions on repos? Read is all you need to interact with issues and branches?

So could it be that the setup for opensafely-testing-bot in the opensafely-testing org is wrong, and it should just have read access to those repos? 

Whilst looking through the verification tests, there are few which no longer seem to be used in production code AFAICT, so maybe we can remove the tests for them?

- create_repo (I think we needed this for OSI?)
- add_repo_to_team (needs org admin I think?)
- set_repo_topics (needs org admin I think?)

AFAICT, all the others only need read permissions?

So possibly, simply removing opensafely-testing-bot from the researchers team in opensafely-testing org might fix this mismatch, assuming all the verification tests still pass?

StevenMaude:
I may not get round to this before someone else does. But some thoughts that might help whoever does tackle it.

## Changing what opensafely-testing-bot has access to

[...]

Issue #4862, opened 2025-02-12
Remove the use of `opensafely-interactive-bot`


StevenMaude:
## Why are we doing this?

We're removing the OpenSAFELY Interactive functionality, and this is only used there, to enable write access to repositories in the `opensafely` organization, via a Personal Access Token.

## What needs to be done before this?

Removing the use of the `INTERACTIVE_GITHUB_TOKEN` (#4858) that this account provides.

## What are we doing?

- [x] Confirming that the `opensafely-interactive-bot` account is only used for `INTERACTIVE_GITHUB_TOKEN` and nowhere else within OpenSAFELY. (For example: login to the account on GitHub and review the account's settings and installed apps…)
- [x] Removing the `opensafely-interactive-bot` from the [OpenSAFELY Interactive team](https://github.com/orgs/opensafely/teams/interactive/members). (Completed in #4865.)
- [x] ~Optionally: Deleting the `opensafely-interactive-bot` account.~ Not doing this.
- [x] ~Possibly: deleting shared credentials for `opensafely-interactive-bot` (*after confirming the account has been deleted*)~ Not doing this.
- [x] Removing/updating reference to `opensafely-interactive-bot` from the team manual (covered by ebmdatalab/team-manual#1030)
- [x] Update or revoke the ADR referring to `opensafely-interactive-bot` (covered by #4980)
- [x] Possibly: removing the associated mailbox. Not doing this.

## When will we know it's done?

Provided this account is used nowhere other than for OpenSAFELY Interactive:

* The `opensafely-interactive-bot` account has no permissions to interact with OpenSAFELY, and perhaps is even deleted
* Our documentation reflects the use of `opensafely-interactive-bot`

StevenMaude:
Does this need someone in IG to remove the bot from the `opensafely` organization (or at least notifiying that we are doing so)?

Or is it fine for us to remove access directly?

StevenMaude:
Since this involves the `opensafely` organization, I've asked IG to confirm this before we proceed. I'll pick up once that's OK, which I think it will be.

StevenMaude:
* Have removed `opensafely-interactive-bot` from the `opensafely-core` organization.

StevenMaude:
Closing as all tasks complete.

PR #5455, opened 2025-12-02
Remove vestigial GitHub API remnants

StevenMaude:
Relates to #5166.

This removes parts of the GitHub API and verification tests that were identified as no longer being in use. These relate to OpenSAFELY Interactive, and were never identified as candidates for removal then. Regardless, they are not used anywhere else in the code now.

(And, in some cases, the code may not work in production anyway, because it relied on the permissions of `opensafely-interactive-bot`. The `opensafely-interactive-bot` [account does not have access to write to repositories](https://github.com/opensafely-core/job-server/issues/4862) now.)

### DIFF
--- a/jobserver/github.py
+++ b/jobserver/github.py
@@ -50,14 +50,6 @@ class HTTPError(GitHubError):
     API."""
 
 
-class RepoAlreadyExists(HTTPError):
-    """Tried to create a repo that already existed."""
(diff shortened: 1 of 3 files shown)


tomodwyer (approved):
Coverage showing as 100% after these tests have been removed.

PR #5457, opened 2025-12-02
Move Staff Area access to a permission

### LINKED ISSUE #5445: Move staff area access to a permission rather than a role
# Why are we doing this?

In most areas of Job Server, we check someone's access against the permissions they have, which is generally considered best practice as it provides fine-grained control. However, for the staff area, we check against a role instead.

Moving the staff area access to a permission will allow us to assign it to other roles in future. At present, someone needs the StaffAreaAdmin role to view anything in the staff area. We expect that we'll soon be creating a new role, the ProjectAdmin (exact name TBC), who will also need access to the staff area.

This will remove `require_role` function and partially address https://github.com/opensafely-core/job-server/issues/4260.

All access to the staff area is controlled by a permissions check.

# What are we doing?

- [x] Creating a new permission, for staff area access
- [x] Replacing any StaffAreaAdmin role checks with checks for that permission
- [x] Removing any redundant code.

---

StevenMaude:
Fixes #5445.

This changes the code to require a specific permission for Staff Area access, instead of a specific role.

This allows us to grant more than one role that permission, in future.

However, for now, we'll keep this access to `StaffAreaAdministrator`, which results in no overall changes to access we've granted.

### DIFF
--- a/applications/views.py
+++ b/applications/views.py
@@ -9,9 +9,7 @@ from django.utils.decorators import method_decorator
 from django.views.generic import CreateView, RedirectView, UpdateView, View
 
 from jobserver.authorization import (
-    StaffAreaAdministrator,
     has_permission,
-    has_role,
     permissions,
 )
 from jobserver.hash_utils import unhash_or_404
@@ -192,7 +190,9 @@ def page(request, pk_hash, key):
     # check the user can access this application
     validate_application_access(request.user, application)
 
-    if application.approved_at and not has_role(request.user, StaffAreaAdministrator):
+    if application.approved_at and not has_permission(
+        request.user, permissions.staff_area_access
+    ):
         messages.warning(
             request, "This application has been approved and can no longer be edited"
         )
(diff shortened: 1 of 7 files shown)


applications/views.py
```diff
@@ -192,7 +190,9 @@ def page(request, pk_hash, key):
     # check the user can access this application
     validate_application_access(request.user, application)
 
-    if application.approved_at and not has_role(request.user, StaffAreaAdministrator):
+    if application.approved_at and not has_permission(
+        request.user, permissions.staff_area_access
```
StevenMaude: thought: I've tried to keep the diffs minimal, and easy to understand, because there are a large number of changes across multiple files.

So:

* sometimes we're using `permissions` as an import, because we already had that imported, and then accessing `staff_area_access` on that
* other times, we're importing `staff_area_access` directly and using that, which is an exchange for `StaffAreaAdministrator`

If it's thought preferable to make all the imports using the permissions consistent, perhaps to explicitly mark them as `permissions.staff_area_access`, that's also fine.

Though, it may be that the permissions get refactored into a class in #5459 subsequently, so I'm not sure it's worth worrying about too much now.

StevenMaude:
> You can possibly check this with your browser developer tools network tab; I now have.

I did this for some requests, but not for all. It's not impossible there's some unusual bug where we somehow render a 403 error page but do something else, but I assumed not.

PR #5472, opened 2025-12-08
Switch to `pyproject.toml` + `uv.lock`-based workflow for dependency management

KatieB5:
This work implements the final section of #5049 ; switching to a `pyproject.toml` + `uv.lock`-based workflow for dependency management.

**Summary of work**

1. Generate `pyproject.toml` + `uv.lock`
2. Remove old requirements files and the test for those files
3. Update justfile to switch requirements-based and pip-based commands to `uv sync` & `uv lock` commands.
4. Update Dockerfile and .dockerignore to use `uv sync`
5. Update GHA Workflows
6. Implement dependency cool down using `uv`

Note: the `uv.lock` in this PR was generated against the cutoff 2025-12-11T00:00:00Z. We may need to bump the cut off date, then update `uv.lock` again before merging.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -172,7 +172,7 @@ extend-ignore = [
 # See https://github.com/opensafely-core/repo-template/blob/main/DEVELOPERS.md for details.
 [tool.uv]
 required-version = ">=0.9"
-exclude-newer = "2025-12-11T00:00:00Z"
+exclude-newer = "2026-01-06T00:00:00Z"
 exclude-newer-package = {}
 
 [dependency-groups]
@@ -189,6 +189,7 @@ dev = [
     "pytest-freezer",
     "pytest-icdiff",
     "pytest-mock",
+    "pytest-playwright",
     "pytest-socket",
     "pytest-xdist[psutil]",
     "responses",
--- a/uv.lock
+++ b/uv.lock
@@ -7,7 +7,7 @@ resolution-markers = [
 ]
 
 [options]
-exclude-newer = "2025-12-11T00:00:00Z"
+exclude-newer = "2026-01-06T00:00:00Z"
 
 [[package]]
 name = "argon2-cffi"
@@ -101,11 +101,11 @@ wheels = [
 
 [[package]]
 name = "certifi"
-version = "2025.11.12"
+version = "2026.1.4"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/a2/8c/58f469717fa48465e4a50c014a0400602d3c437d7c0c468e17ada824da3a/certifi-2025.11.12.tar.gz", hash = "sha256:d8ab5478f2ecd78af242878415affce761ca6bc54a22a27e026d7c25357c3316", size = 160538, upload-time = "2025-11-12T02:54:51.517Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/e0/2d/a891ca51311197f6ad14a7ef42e2399f36cf2f9bd44752b3dc4eab60fdc5/certifi-2026.1.4.tar.gz", hash = "sha256:ac726dd470482006e014ad384921ed6438c457018f4b3d204aea4281258b2120", size = 154268, upload-time = "2026-01-04T02:42:41.825Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/70/7d/9bc192684cea499815ff478dfcdc13835ddf401365057044fb721ec6bddb/certifi-2025.11.12-py3-none-any.whl", hash = "sha256:97de8790030bbd5c2d96b7ec782fc2f7820ef8dba6db909ccf95449f2d062d4b", size = 159438, upload-time = "2025-11-12T02:54:49.735Z" },
+    { url = "https://files.pythonhosted.org/packages/e6/ad/3cc14f097111b4de0040c83a525973216457bbeeb63739ef1ed275c1c021/certifi-2026.1.4-py3-none-any.whl", hash = "sha256:9943707519e4add1115f44c2bc244f782c0249876bf51b6599fee1ffbedd685c", size = 152900, upload-time = "2026-01-04T02:42:40.15Z" },
 ]
 
 [[package]]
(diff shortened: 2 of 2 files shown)


KatieB5:
Thanks very much Steve :) 

I agree with you on the naming of `upgrade-all` so I'll update this, and I'll un-pin the versions in `pyproject.toml` to keep things cleaner. 

Changing `dynamic = ["version"]` to `version = "0.1.0"` in `pyproject.toml` didn't stop the `jobserver.egg-info` directory and files being produced when I tested locally, so I'll add to `.gitignore` and `.dockerignore` for now and make an issue for deck scrubbing, [as discussed in Slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1765993328801249?thread_ts=1763573364.089979&cid=C069SADHP1Q).

Re: removing the pipeline script, I think maybe we can remove it:

I updated the pinned version of the opensafely-pipeline to the [previous version](https://github.com/opensafely-core/pipeline/commit/0b7d3060819836bfd8f754d8f93e95ffb30a18f6) in `pyproject.toml`:
```
"opensafely-pipeline@https://github.com/opensafely-core/pipeline/archive/refs/tags/v2025.02.10.092559.zip"
```

Then ran `just clean` + `just devenv` to get a local venv with the older version of `opensafely-pipeline`. This updated `uv.lock` with the older version, and installed it into my local venv.

And then updated `pyproject.toml` following the [uv docs](https://docs.astral.sh/uv/concepts/projects/dependencies/#git) guidance:
```
dependencies = [
    ...
    "opensafely-pipeline",
    ...
]

[tool.uv.sources]
opensafely-pipeline = { git = "https://github.com/opensafely-core/pipeline/", branch = "main" }
```

I ran `just upgrade-package opensafely-pipeline` and got this:

```
uv lock --upgrade-package opensafely-pipeline
    Updated https://github.com/opensafely-core/pipeline/ (1efd568d4b2022e7313552be8807b7ccd38f0d0a)
Resolved 113 packages in 8.05s
Updated opensafely-pipeline v2025.2.10.92559 -> v2025.3.6.161237 (1efd568d)
Resolved 113 packages in 0.88ms
...
      Built opensafely-pipeline @ git+https://github.com/opensafely-core/pipeline/@1efd568d4b2022e7313552be8807b7ccd38f0d0a
Prepared 2 packages in 7.80s
Uninstalled 2 packages in 18ms
Installed 2 packages in 18ms
...
 - opensafely-pipeline==2025.2.10.92559 (from https://github.com/opensafely-core/pipeline/archive/refs/tags/v2025.02.10.092559.zip)
 + opensafely-pipeline==2025.3.6.161237 (from git+https://github.com/opensafely-core/pipeline/@1efd568d4b2022e7313552be8807b7ccd38f0d0a)
```
It seems `uv` has upgraded the package to the latest version.

StevenMaude:
@KatieB5: Maybe we still need the `upgrade-pipeline` script, I'm not entirely sure `uv` is giving us the existing behaviour we have with that.

When you updated, you got:

```toml
version = "2025.3.6.161237"
source = { git = "https://github.com/opensafely-core/pipeline?branch=main#1efd568d4b2022e7313552be8807b7ccd38f0d0a" }
```

Let's look at this.

## `version`

`version = "2025.3.6.161237"`

This is the latest version number; currently the latest [tag](https://github.com/opensafely-core/pipeline/tags) in the `pipeline` repository.[^1]

tags

This looks good, so far.

## `source`

But if we look at the next line, what do we see?

```toml
source = { git = "https://github.com/opensafely-core/pipeline?branch=main#1efd568d4b2022e7313552be8807b7ccd38f0d0a" }
```

There's a problem here.

[The commit](https://github.com/opensafely-core/pipeline/commit/1efd568d4b2022e7313552be8807b7ccd38f0d0a) that's linked in the updated `pyproject.toml` (`1efd568…`) doesn't match the actual [commit corresponding to the tag](https://github.com/opensafely-core/pipeline/commit/a6d1ac9c8259e24c43e729ac2cafd6d62d944f40) from March (`2025.3.6`): `a6d1ac9`.

If we look at commit `a6d1ac9` in the `pipeline` repository, we can see that GitHub labels the tag.[^2]
 
a6d1ac9-tag

Whereas `1efd568…` which `uv` has updated to in `pyproject.toml` is actually from 2025-12-17, not 2025-03-06.

1efd568-commit

## What's `uv` doing?

It looks like:

* `uv` is fetching the latest `pipeline` commit — currently `1efd568`
* mislabelling that commit as the desired version
  * in fact, the version number is the last known version number, so `uv` is using that in the absence of anything else

which we may not want.

We want the commit that corresponds to the specific version precisely.

## How do things work now?

If we want to preserve the existing behaviour, we should look how things are now.

From the existing `requirements.prod.txt` file, we see we're getting the source `.zip` download corresponding to the specific Git tag:

https://github.com/opensafely-core/job-server/blob/bb33758db5d46f90eacb79052b6ec4272539b5bd/requirements.prod.txt#L534

And `a6d1ac9` is, as we've seen, the commit corresponding to that tag.
 
## Conclusion

It might be that `uv` doesn't support our use case, and we have to update the `upgrade-pipeline.sh` script to edit `pyproject.toml` instead.

[^1]: In case you haven't used tags before, they're text markers that you can add to specific commits with `git tag`. They're stored in what might be a surprisingly [simple way](https://git-scm.com/book/en/v2/Git-Internals-Git-References): as a file in the repository.
[^2]: And we can interact with tags locally, again with `git tag`.

KatieB5:
Thanks very much Steve, this is really helpful context and detail. I'd noticed the different commits `uv` was using, and raised it for discussion in the [TeamREX Slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1766052806113849) as it felt like a non-trivial change to potentially be making. I think we're in agreement that we should continue specifying the tagged release and using the `upgrade-pipeline.sh` script to edit `pyproject.toml`.

docker/Dockerfile
```diff
@@ -108,16 +110,20 @@ ENV VIRTUAL_ENV=/opt/venv
...
+# --frozen for deterministic installs
+# (https://docs.astral.sh/uv/reference/cli/#uv-sync--frozen)
+# --no-dev so we only install prod dependencies
+#  (https://docs.astral.sh/uv/reference/cli/#uv-sync--no-dev)
+# --no-install-project so we only install the dependencies, not the
+# project (https://docs.astral.sh/uv/reference/cli/#uv-sync--no-install-project)
+# --directory /root lets `uv` detect the pyproject.toml, uv.lock etc.
 RUN --mount=type=cache,target=/root/.cache \
-    uv pip install --no-deps --require-hashes -r /tmp/requirements.prod.txt
+    --mount=type=bind,source=pyproject.toml,target=/root/pyproject.toml \
```
lucyb: Previously we were copying the `requirements.prod.txt` into the image and (I think) we could do the same with the pyproject.toml and uv.lock files. What are the advantages to using a bind mount instead? I'm not sure if I have a preference, but I am curious.

DEVELOPERS.md
```diff
@@ -271,17 +274,49 @@ In that situation you can follow the steps below to set up your local copy of th
 
 Workspace creation currently requires a link to a repo, so from this point onwards you will need to make changes directly in the database to create a workspace, and make job requests.
 
-### Compiling requirements
+## Dependency management
+We use [Dependabot](https://github.com/opensafely-core/job-server/blob/main/.github/dependabot.yml) and an [automated update workflow](https://github.com/opensafely-core/job-server/blob/main/.github/workflows/update-dependencies.yml) to keep our dependencies up to date.
```
lucyb: This is really nicely written and a good improvement on what was there before.

lucyb (commented):
Thanks again for the really good walkthrough earlier. It was very helpful for understanding the changes and the decisions behind them. That said, the commit messages are excellent and already do a good job of explaining the "why" behind the changes.

docker/Dockerfile
lucyb: Thanks, that's a really thorough explanation. I think either approach is fine and the bind mount approach has the advantage of acquiring the files exactly where they're used, so in some ways it's clearer. I would recommend sticking with it.

One thing though, I think both `pyproject.toml` and `uv.lock` will be in the final image. You're right that they won't be copied from the builder build stage; this is an intermediate image and we're only copying over `/opt/venv` into the final image. However, in the `job-server-prod` build stage we're doing `COPY . /app`, which will copy everything that's not explicitly excluded in the .dockerignore. I don't think this is a problem, but it's worth being aware of as it may cause issues in the future.

KatieB5: Thank you for confirming Lucy, in that case, I'll stick with the bind mount. 

Re: copying `pyproject.toml` and `uv.lock` into the final image when we do we do `COPY . /app`, thanks for pointing this out; it's good to be aware of for my own understanding now, and in future.

Issue #5474, opened 2025-12-08
Name and create new role - '`ServiceAdministrator`'


lucyb:
# Why are we doing this?

As part of the initiative to enable manual project set up in Job Server, we need to create a new role.

That role will eventually allow someone to create projects without an associated application, add researchers to the project and assign project-level roles to those researchers. 

# How will we know when it's done?

The new role exists, with a name that's been agreed by Mary and can be assigned to users.

# What are we doing?

- [X] Decide upon an appropriate role name (🧵 [thread](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1765789102408339)) -- 
  - Human name: "Service Administrator"
  - Class name: `ServiceAdministrator`
  - Description: Access the Create Project page. Create projects. Assign users to projects and project roles.
- [x] Creating a new role, without any permissions (for now) #5493 
- [x] Update other tickets relating to the new role with the new name and other initiative documentation like the DPV.
  - Searched for "ProjectAdmin" and done replacements with `ServiceAdministrator`, also in DPV.

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

PR #5493, opened 2025-12-18
Add `ServiceAdministrator` role and fixture.

mikerkelly:
This role will eventually allow someone to create projects without an associated application, add researchers to the project, and assign project-level roles to those researchers. Fixes https://github.com/opensafely-core/job-server/issues/5474, which I've taken the values from. The description is slightly inaccurate right now but that doesn't seem to matter; it will be accurate when we finish the work around this and before it's used for real.

This is part of [`initiative:manual-projects`](https://github.com/opensafely-core/job-server/issues?q=label%3A%22initiative%3Amanual-projects%22), and a step towards a general effort to split up `StaffAreaAdministrator`.

Add a fixture as tests may want to use this role once it has more than one permission. Often, a `UserFactory` for the specific relevant permissions will make the test more precise and less brittle, though.

### DIFF
--- a/jobserver/authorization/roles.py
+++ b/jobserver/authorization/roles.py
@@ -38,6 +38,18 @@ class StaffAreaAdministrator:
     ]
 
 
+class ServiceAdministrator:
+    display_name = "Service Administrator"
+    description = """DO NOT ASSIGN IN PRODUCTION.
+    Access the Create Project page.
+    Create projects.
+    Assign users to projects and project roles."""
+    models = [
+        "jobserver.models.user.User",
+    ]
+    permissions = []
+
+
 class OutputChecker:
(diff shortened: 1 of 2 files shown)


jobserver/authorization/roles.py
```diff
@@ -38,6 +38,17 @@ class StaffAreaAdministrator:
     ]
 
 
+class ServiceAdministrator:
+    display_name = "Service Administrator"
+    description = """Access the Create Project page.
```
lucyb: It would be worth adding something like "DO NOT USE", or similar, just in case someone is tempted to assign someone to this role. It shouldn't matter right now, because there are no permissions associated with it, but there will be shortly.

PR #5543, opened 2026-01-15
Remove deprecation warnings from Pytest filters

StevenMaude:
This PR removes more warnings that Pytest ignores.

It also documents why the few remaining `filterwarnings` entries exist.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -179,8 +179,15 @@ env = [
   "DEFAULT_OUTPUT_CHECKING_SLACK_CHANNEL=test-channel",
 ]
 filterwarnings = [
+    # Turn warnings into errors.
     "error",
+    # See issue #5540 for the test fixes required to remove
+    # PytestUnraisableExceptionWarning from filtering.
     "ignore::pytest.PytestUnraisableExceptionWarning:",
+    # The FORMS_URLFIELD_ASSUME_HTTPS and associated warning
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -179,11 +179,13 @@ env = [
...
+    # Turn warnings into errors.
     "error",
+    # See issue #5540 for the test fixes required to remove
+    # PytestUnraisableExceptionWarning from filtering.
     "ignore::pytest.PytestUnraisableExceptionWarning:",
-    "ignore:Call to deprecated method __init__:DeprecationWarning:",
-    "ignore:pkg_resources is deprecated as an API:DeprecationWarning:",
-    "ignore:Deprecated call to `pkg_resources.declare_namespace:DeprecationWarning:",
+    # The FORMS_URLFIELD_ASSUME_HTTPS warning can be removed
+    # when upgrading to Django >= 6.0.
```
mikerkelly: optional, further detail: Is it possible to give a link/reference or an explanation supporting this?
https://docs.djangoproject.com/en/6.0/releases/6.0/
>The FORMS_URLFIELD_ASSUME_HTTPS transitional setting is removed.

I think we are ignoring this warning because it is a deprecation warning, can we also say that? I think that is the main point. We could also say "along with the setting", presumably? I see that this is already documented in `settings.py`.

Issue #5553, opened 2026-01-21
Add dual format support(number and text) to `Project.number`


nishtha-kalra:
From the DPV [doc](https://docs.google.com/document/d/1qrx9EjSjv9mE2OpAo4nYudPQVqnujLq0U4zzSxDqX0U/edit?tab=t.0#heading=h.rbw2mysaauvy) and slack [thread](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1765884717676229), we have decided to handle both formats in the existing field name, but the type changed to a character-based field.
## What are we doing?
- [ ]  Reuse Project.number but treat it as text everywhere; add helpers for display/sorting/missing IDs. 
- [ ] Update backend validation to accept both numeric and new hyphenated values; form validation can be stricter for new entries while model validation still accepts legacy/null.
- [ ] Update frontend validation and templates, removing hard-coded # prefixes and using helpers so old/new/no-number cases display correctly.
- [ ] Introduce a default ordering on Project
- [ ] Clean up backend code that handled missing numbers ad hoc by switching to the new helpers.

Issue #5446, opened 2025-11-25
Use new project identifier format


lucyb:
# Why are we doing this?

The `project.number` or Approved Project ID is going to change for non-covid projects. It's currently an integer, it will become an alphanumeric, hyphenated, field. For example `POS-2025-2001`.

We should expect older project numbers to continue to be used, so should consider that in our validation checks.

A small number of older projects don't have a number assigned. These are due to be tidied up and assigned an application (or removed from Job Server) as part of another piece of work. We should expect them to continue to exist for some time though and take that into consideration for our validation checks.

This work does leave the `project.number` field misleadingly named. We should consider renaming this field in a separate ticket.

Note: the RAP API uses the `project.slug`, so this should not change for any existing project.

<img width="622" height="388" alt="Image" src="https://github.com/user-attachments/assets/8818f4e2-bd8b-4239-a261-26cab4a89baf" />

Image source: See [DPV](https://docs.google.com/document/d/1qrx9EjSjv9mE2OpAo4nYudPQVqnujLq0U4zzSxDqX0U/edit?usp=sharing) links (private link).

# How will we know when it's done?

- [ ] Project numbers can be provided in the new or old format.
- [ ] Project numbers are validated appropriately.
- [ ] Project numbers are displayed appropriately.
- [ ] The permissions files are updated to take the new format project number.
- [ ] Documentation should be updated appropriately (e.g team manual, OS Docs, OS website, Job Server docs).
- [ ] The Service Team are informed of the change.

# What are we doing?

- [ ] Deciding on an appropriate approach, either to update the existing `project.number` field or to add a new field (this latter approach may have wider implications that need considering though) [Slack thread](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1765884717676229) 
- [ ] Updating the database field to be a character field and adding validation, either allow letters, numbers and hyphens or match the pattern more closely if not too hard. We might want form field validation to only accept the new format, as new projects will always use the new one and that helps avoid some types of errors. That may apply to both form and model validation, if not too hard and accepting existing data.
- [ ] Updating the permissions files
- [ ] Updating the validation in the frontend
- [ ] Checking everywhere that the file is used in the frontend for any required updates - for example,  in a few places we add `#` before the project number, as it was a number `#{{ project.number }}`
- [ ] ...
[...]

Issue #5554, opened 2026-01-21
Change `project.number` field from integer to character


nishtha-kalra:
## What are we doing?
- [ ] Create and deploy a migration that alters Project.number to a CharField and backfills existing integers as strings by following risky migrations [guidelines](https://github.com/opensafely-core/job-server/blob/main/DEVELOPERS.md#risky-migrations).
- [ ] Verify backend/frontend changes still pass with the new type.

Issue #5510, opened 2026-01-06
Create project form - validation


lucyb:
# Why are we doing this?

We need to ensure there is appropriate client-side and server-side validation of fields in the create project form.
We also need to ensure the proper messages are returned to the user, so they know what's happened and what to do about it.

Dependent on #5473 and #5446.

# How will we know when it's done?

All fields in the create project form are validated appropriately and the user is clear what to do if invalid data is entered.

# What are we doing?

- [ ] Adding client-side validation to the create project form
- [ ] Adding server-side validation to the create project form

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

nishtha-kalra:
Some notes regarding this ticket that will help in implementation:
1. We could add a regex as part of the Project class attribute.

PR #5567, opened 2026-02-02
Treat project.number as a Char field in preparation for new project identifier format

nishtha-kalra:
## Description
Issue - https://github.com/opensafely-core/job-server/issues/5553
This sub issue is part of issue [5446](https://github.com/opensafely-core/job-server/issues/5446)
The [spike](https://github.com/opensafely-core/job-server/issues/5446#issuecomment-3813563272) covers the method we have taken to approach this change.

This PR focuses on solving the below requirements from the subticket:
- Reuse Project.number but treat it as text everywhere; add helpers for display/sorting/missing IDs.
- Update backend validation to accept both numeric and new hyphenated values; form validation can be stricter for new entries while model validation still accepts legacy/null.

## Testing
- manually tested the changes
- added new unit tests for validators and helper functions
- existing tests passes

## Next steps
- Follow up PR which will have migration file covering addition of validtor `validate_project_identifier`  on number field. This will not affect the field right now as only integers are supported. But wanted to cover all validation related things in 1 PR.
- Will be covering other AC's in next PR as wanted to keep PR smaller for easier review

### DIFF
--- a/jobserver/models/project.py
+++ b/jobserver/models/project.py
@@ -191,28 +191,42 @@ class Project(models.Model):
     def display_identifier(self):
         return self.identifier or ""
 
+    @staticmethod
+    def _coerce_numeric_identifiers(values):
+        """
+        Convert an iterable of raw identifier values into a list of integers.
+
+        Accepts ints (used as-is) and digit-only strings (trimmed and cast to int).
+        Any other values—POS-style identifiers, blank strings, None—are ignored.
+        """
+        numeric_values = []
+        for value in values:
+            if isinstance(value, int):
+                numeric_values.append(value)
+                continue
+
+            value_str = str(value).strip()
+            if value_str.isdigit():
+                numeric_values.append(int(value_str))
+
+        return numeric_values
+
     @classmethod
     def next_numeric_identifier(cls):
         """
-        Return the next sequential project number, ignoring any alphanumeric
-        identifiers or blanks. Returns None if no numeric IDs exist.
+        Return the next sequential project number.
+
+        Fetches all non-null identifiers from the database, coerces the numeric ones,
+        and returns max(value) + 1. Returns None if no numeric identifiers exist.
         """
-        project_numbers = []
-        for value in (
+        values = (
             cls.objects.exclude(number__isnull=True)
             .values_list("number", flat=True)
             .iterator()
-        ):
-            if isinstance(value, int):
-                project_numbers.append(value)
-                continue
+        )
 
-            # handle string representations of integers
-            value_str = str(value).strip()
-            if value_str.isdigit():
-                project_numbers.append(int(value_str))
-
-        if not project_numbers:
+        numeric_values = cls._coerce_numeric_identifiers(values)
+        if not numeric_values:
             return None
 
-        return max(project_numbers) + 1
+        return max(numeric_values) + 1
--- a/jobserver/project_validators.py
+++ b/jobserver/project_validators.py
@@ -35,6 +35,5 @@ def validate_project_identifier(value):
     if PROJECT_IDENTIFIER_PATTERN.match(normalized):
         return
 
-    raise ValidationError(
-        "Use either a numeric ID or the new POS-YYYY-#### identifier."
-    )
+    # TODO: update message when alphanumeric IDs are supported
+    raise ValidationError("Please use a numeric ID.")
--- a/tests/unit/jobserver/models/test_project.py
+++ b/tests/unit/jobserver/models/test_project.py
@@ -211,17 +211,31 @@ def test_project_display_identifier():
 
 
 def test_next_numeric_identifier():
-    ProjectFactory(number=1)
-    ProjectFactory(number="2")
+    ProjectFactory(number=100)
+    ProjectFactory(number=102)
 
-    assert Project.next_numeric_identifier() == 3
+    assert Project.next_numeric_identifier() == 103
+
+
+def test_next_numeric_identifier_handles_char():
+    ProjectFactory(number="101")
+
+    assert Project.next_numeric_identifier() == 102
 
 
 def test_next_numeric_identifier_handles_whitespace():
     ProjectFactory(number="  789  ")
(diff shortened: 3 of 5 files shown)


mikerkelly (commented):
Thanks Nishtha. It is good to see that the git commit messages are nice and clear about what has changed. However they don't give a rationale. It would be useful to have something in each so the log justifies the change in isolation. Even just a simple one-sentence "For forward compatibility when we migrate `Project.number` to a `CharField`".

I am not sure I understand how well this PR fits into your intended sequence, so I am finding it difficult to review. Sorry if I'm just being slow.

Have you tried making the change to charfield locally and seen that it broadly works with this approach? That would be a good check that nothing has been missed here for the forward-compatibility.

I am confused about your next steps. Are you saying that there will be more PRs to address the rest of #5553, then later a PR for the migration? Or will all the remaining changes be made along with the migration?

# Helpers

I think these would make more sense alongside the template changes where they will be used. (And any other places? I'm not sure exactly what "treat it as text everywhere" might include, have you done a survey?). I don't think it makes sense to review helpers in isolation from the functionality they are helping, as I don't know what they are going to achieve.

So I would prefer if you included these, if needed, in a PR with the changes that need them. Whether that's templates or something else.

# Validation

I am not sure that the validation changes make sense to do before the migration PR. They are only used in project creation, and the chances of that being run in the old container at same time as the migration is very remote. So I am not sure what we are defending against. And the very general validator you have written seems too general -- it accepts non-integer values on the `ApplicationApproveForm`.

https://docs.djangoproject.com/en/5.2/ref/validators/#how-validators-are-run

> Note that validators will not be run automatically when you save a model, but if you are using a ModelForm, it will run your validators on any fields that are included in your form.

This is a bit of a surprising thing about Django, the validators assigned to model fields are not run by default except for in `ModelForm`. We don't call project.clean or project.clean_fields anywhere that I can see, so I think that assigning this validator to the model `validators` is not doing anything at the moment.

The new form doesn't exist yet, so we don't need to worry about that. `ApplicationApproveForm` has a `IntegerField` validator that works until we change the field type. Your new validator will accept the new format immediately on that form after the migration, which seems wrong - that workflow should still only accept integers.

We will have two forms living side by side, the old `ApplicationApproveForm` (which is not a `ModelForm`) for integer-numbered projects and the new one for new-format ones. Each should only accept exactly its format.

The only place that `validate_project_identifier` is called is in the `ApplicationApproveForm`, where it is imported directly. But that from should only accept integer values for `project_number`, not nulls or the new format, so I think leaving it as an integer field and doing the type cast in the form clean code is more appropriate?  That is very simple so I don't see a problem adding it in the migration PR for https://github.com/opensafely-core/job-server/issues/5554, no other changes in this module required.

https://github.com/opensafely-core/job-server/issues/5510 will add validation for the new form, that can add a `RegexValidator` or other validator for the new format to the model if it wants.

If we really want to add something general as a fail-safe, I would suggest in `jobserver.actions.projects`, but it should just need to check for any of `None` or .isdigit() or the new format validator borrowed from the model. And I think this can wait until that exists, if needed at all. Given that we only create projects thru this in 1/2 places in the codebase, and that is unlikely to change.

What do you think? 

# next_numeric_identifier

It seems reasonable to make this handle both str and int ahead of the migration as it's not completely straightforward (but not necessary because the chances of a race are so remote). But we will want to remove the special-casing once we know the migration is good, for simplicity, so please take a note to do that somewhere in one of the issues.

I note that for this to work without further changes in the migration, we need to leave the field as an `IntegerField`, but I think that's fine as discussed above.

I think if there are no int values in the database we probably want to return 1, not null. Then the first instance with an int number gets 1, not `None`/`NULL`.

I think the approach in `_coerce_numeric_identifiers` is more complicated than it needs to be. If you have a variable and you know it is either an int or coerceable from str to int then `int(str(val))` works. So we construct the numeric values in a simple list comprehension as:

```
int_values = [
    int(str(val))
    for val in Project.objects.values_list('number', flat=True)
    if val and str(val).isdigit()]
```

nishtha-kalra:
Hi @mikerkelly, Thanks for the review. 
I had a chat with Lucy about this ticket last week, and we realised it might be better to break this change into even smaller PRs. The intent of this PR was to create a couple of PRs in preparation for the upcoming change to the `project.number` field from `int` to `char`. We realised that this PR may not be clear, so it would be better to break it into smaller parts.

Git commit messages: I will add a reason and my thought process for each commit going forward. I tried to make each commit clear and descriptive, but maybe I also need to explain why each one. Will do that going forward.

> I am not sure I understand how well this PR fits into your intended sequence, so I am finding it difficult to review. Sorry if I'm just being slow.

Yes, I realised that when I was discussing this PR with Lucy last week. We think that you thought the next step would be to do the model and migration change, but that's not what my intention was. In the PR description, I tried to mention the next steps as follow-up PRs regarding the same issue. There will be a couple more before I make the model change.

> Have you tried making the change to charfield locally and seen that it broadly works with this approach? That would be a good check that nothing has been missed here for the forward-compatibility.

That would break right now as that's not what I was trying to acheive with this PR. There are still a couple of more things to address before I can change the field.

>  I am confused about your next steps. Are you saying that there will be more PRs to address the rest of #5553, then later a PR for the migration? Or will all the remaining changes be made along with the migration?

Yes, there will be more PRs to address rest of #5553 and later a PR for migration. The idea of breaking this ticket into smaller pieces and tackling them one by one was to separate the code changes and the migration into different PRs for a better rollout. We had a discussion regarding this on [this](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1765884717676229) slack thread.

Regarding your comments on other parts:
I plan to address them as I create PRs covering those changes. Since I will separate out each change in different PRs, it will be easier to address those comments on the corresponding PRs. 
Please let me know if things are still confusing, and we can jump on a call.

nishtha-kalra:
Closing this PR and will open a couple of new ones, which will cover each change(template, validation, ordering, permission files, etc.) separately.

PR #5568, opened 2026-02-03
Project creation entry added to audit log

### LINKED ISSUE #5548: Show project created at and created by in the audit log
# Why are we doing this?

We've been asked to ensure that there's appropriate auditing in place for when we start to allow manual creation of projects.

Related to #5473.

It's possible to see who created a project and when in the audit log for a given project.

# What are we doing?

Enabling the project `created_at` and `created_by` fields to be displayed appropriately on the project audit log page.

Probably, per [this Slack thread](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1769514644635689), we should add a pseudo-auditable event to the start of the audit log, though there is no corresponding database entry. Possible approaches, either construct an in-memory AuditableEvent from the project details to pass to a new type of presenter function, or just directly construct a [PresentableAuditableEvent](https://github.com/opensafely-core/job-server/blob/a17a754797487f6556f9a8f743dec977a6b93f69/jobserver/auditing/presenters/base.py#L42) returned by such a function.

---

tomodwyer:
Closes #5548

This PR:

- Creates a new auditable event of type "project_created"
  - The event contains the project creation date, and who created it
- Triggers this event when a project is created
- Updates the audit log description text to confirm this
- Updates the functions for looking up users and projects for auditable events
- Adds tests for this auditable information

It also adds a migration to backfill creation dates in to the `jobserver_auditableevent` table for already created projects.


jobserver/migrations/0015_backfill_project_created_auditable_events.py
```diff
@@ -37,15 +35,11 @@ def add_project_created_events(apps, schema_editor):
...
-    return
-
-
 class Migration(migrations.Migration):
     dependencies = [
         ("jobserver", "0014_alter_user_backends_alter_user_orgs_and_more"),
     ]
 
     operations = [
-        migrations.RunPython(add_project_created_events, noop_reverse),
```
nishtha-kalra: What is the reason to do this?

jobserver/auditing/presenters/projects.py
```diff
@@ -0,0 +1,50 @@
...
+    except (Project.DoesNotExist, ValueError):
+        # Handle a ValueError here because Django will (helpfully!) try to cast
+        # the given pk to an int for BigAutoFields. However AuditableEvent's
+        # target_id and parent_id are TextFields which default to an empty
+        # string, we can't guarantee that's been populated, and int("") raises
+        # a ValueError.
+        return default
+
+
+def lookup_user(s: str) -> User | str:
```
mikerkelly: minor, naming: Could `s` be `username` instead, please, to be more descriptive?

jobserver/migrations/0016_alter_auditableevent_type.py
```diff
@@ -0,0 +1,86 @@
+from django.db import migrations, models
```
mikerkelly: nit, naming: this migration is about more than just changing the type, it also does the fairly complex backfill. So maybe it should mention that, 0016_alter_auditableevent_type_and_backfill?

mikerkelly (approved):
Thanks Thomas, this all looks really good now. Thanks for taking the effort to reshape the git history.

I rolled back and forth over the migration several times to test the "only apply once" logic. I think it's just about simple enough codewise that just testing it manually is a reasonable approach, and -- seems to work, with a production-ish mirror!

We could squash the migration and remove the backfill part once it's applied in production, but I don't think it matters that much.

Issue #5473, opened 2025-12-08
Create minimal create project form


lucyb:
# Why are we doing this?

_Prerequisites (now completed): #5444 (mock-ups of the form) and https://github.com/opensafely-core/job-server/issues/5474 (create new role)._

_Note: the old style project identifier will be used if we do this before https://github.com/opensafely-core/job-server/issues/5446_

We want a minimal form for creating projects, based on the mock-ups created in #5444 and the wording added by Mary in the comments below, but without any of the validation.

By creating a minimal form, we can confirm that what we've put into the mock-up can be done in Django and provide a higher fidelity demo for the Service Team. This also keeps the change as small as possible. We expect that only people with the `ServiceAdministrator` role will be able to access it and as that role hasn't been assigned to anyone yet, it shouldn't cause any issues on the production site.

# How will we know when it's done?

It's possible for someone with the `ServiceAdministrator` role to create a new project without an associated application in Job Server

# What are we doing?

- [ ] Add a new Django form
- [ ] Add a button on the staff area projects page for anyone with the  `ServiceAdministrator` role
- [ ] Add permission to view form to the `ServiceAdministrator` role
- [ ] Add a basic Playwright test
- [ ] Ensure that created at and created by shows up in the audit log

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

Issue #5444, opened 2025-11-25
Mockups for the create project form


lucyb:
# Why are we doing this?

We want to add a new route for the service team to create projects without needing an application held within Job Server. Creating some mockups will allow us to better understand the requirements for creating a new project, including the fields needed and the workflow ([Miro](https://miro.com/app/board/uXjVJOiLLTE=/?moveToWidget=3458764640151133610&cot=14)). The information learned will allow us to create further tickets to do the work to create the form itself.

# How will we know when it's done?

The mockups have been shared with Team REX and (if necessary) adjusted for any feedback. In particular, Mary will have confirmed that the workflow and the information gathered is correct.

# What are we doing?

- [ ] Creating a series of mockups showing how creating a new project would work.

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

mikerkelly:
[Slack thread with discussion of drafts](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1764770552848699).

PR #5578, opened 2026-02-06
Add minimal create project form

KatieB5:
Addresses the fourth step in #5473

### Current approach taken in this PR:
Our `ProjectCreateForm(ProjectEditForm)` sub-classes `ProjectEditForm(forms.ModelForm)`. This initially looked attractive because it allows reuse of shared logic (for example `clean_number()`) and fields, rather than duplicating these. 

However, this approach may be introducing additional complexity that we could potentially avoid:
- **Tight coupling between create and edit concerns**: `ProjectCreateForm` inherits behaviour intended for editing an existing instance. This means we need additional logic to override or undo assumptions and logic, based on the behaviour we've agreed for the `ProjectCreateForm` (for example: which fields are required, which fields should appear, auto-populated values, and values that are vs aren't user-editable on first render). 
- **Excluding fields increases cognitive overhead**: The create form currently excludes multiple fields inherited from `ProjectEditForm`. Readers must mentally diff the edit form against the create form (and the `Project` model) to understand what the create form actually contains.
- **Leaning less on Django’s built-in form behaviour than we could do**: the Slippers `input.html` component only supported the `hint_text="your string here"` placement by using the `custom_field=True` branch. This means  we bypass Django’s standard widget rendering for some fields and manually provide attributes in templates. This works, but means we are doing more work where we could do less.

Before taking this approach any further, I think it would be good to pause and consider alternative approaches. I have one in mind (described below) but I'm more than happy to talk through other approaches, or continuing with the current one (and modifying/improving it) if we agree it’s the best fit.

### Suggestion for an alternative approach:
It may be simpler, easier to reason about, and cleaner in terms of separating out logic, both for the current implementation and for future iterations (as we know we'll ultimately be improving and adjusting this form), to: 

- **Introduce a dedicated `ProjectCreateForm(forms.ModelForm)`**: Have a separate `ProjectCreateForm(forms.ModelForm)`, which subclasses `ModelForm`, explicitly lists only the fields users are expected to provide (`fields = [...]`), and relies on the `Project` model for defaults.
- **Share validation logic explicitly**: Extract `clean_number()` into a module-level helper (for example in `forms.py`) so it can be reused by both `ProjectEditForm` and `ProjectCreateForm`, and any future forms that may need it, without inheritance-based coupling. Note: we are updating our project number format in a separate piece of work, and this includes updating the validation (https://github.com/opensafely-core/job-server/issues/5446).
- **Use `help_text` instead of template-level custom fields**: Set `help_text` on the `ProjectCreateForm(forms.ModelForm)` fields where we want to display additional supporting copy in the UI. This would allow us to remove the `custom_field=True` attribute in our templates, and lean more on Django's built-in form/template display behaviour (e.g. `hint_text="Really long string with some super helpful info for users about the thing they should input"` could become `hint_text=field.field.help_text`.
- **Reuse existing mixins where appropriate**: There's a nice `SetUserMixin` in [`staff/views/site_alerts.py`](https://github.com/opensafely-core/job-server/blob/main/staff/views/site_alerts.py), which I think we might be able to make use of in the `ProjectCreateView(CreateView)`, instead of repeating similar logic.

Whichever approach we decide to take from here, we need:

- Unit tests (`tests/unit/staff/test_forms.py`)
- A 'success' page to direct users to after they've successfully created a new project (per the mockups: #5444); the current implementation redirects to the project detail page, which does not match the user journey we've agreed. Suggested approach to this from Thomas ([here in Slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1770372653148689))

**Helpful references:**
Django ModelForm documentation
https://docs.djangoproject.com/en/stable/topics/forms/modelforms/

### DIFF
--- a/staff/views/projects.py
+++ b/staff/views/projects.py
@@ -2,6 +2,7 @@ from django.contrib import messages
 from django.db import transaction
 from django.db.models.functions import Lower
 from django.shortcuts import get_object_or_404, redirect
+from django.utils import timezone
 from django.utils.decorators import method_decorator
 from django.views.generic import (
     CreateView,
@@ -25,6 +26,7 @@ from jobserver.models import AuditableEvent, Org, Project, ProjectMembership, Us
 
 from ..forms import (
     ProjectAddMemberForm,
+    ProjectCreateForm,
     ProjectEditForm,
     ProjectLinkApplicationForm,
     ProjectMembershipForm,
@@ -275,6 +277,22 @@ class ProjectMembershipRemove(View):
 
 @method_decorator(require_permission(Permission.PROJECT_CREATE), name="dispatch")
 class ProjectCreate(CreateView):
+    """Create a new Project instance."""
+
+    form_class = ProjectCreateForm
     model = Project
-    fields = []
     template_name = "staff/project/create.html"
+
+    @transaction.atomic
+    def form_valid(self, form):
+        project = form.save(commit=False)
+
+        project.created_by = self.request.user
+        project.updated_by = self.request.user
+
+        if not project.created_at:
+            project.created_at = timezone.now()
+
+        project.save()
+
+        return redirect(project.get_staff_url())


StevenMaude:
# Summary for the Django changes

I don't know this code or Django forms that well. Because of that, my opinions are flavoured by how I:

* personally prefer to work with Python more generally
* observe we do things in the codebase elsewhere

Here's a summary, followed by the rough stream of consciousness when I looked at this PR.

* An alternative to @KatieB5's proposals is a shared base class `ProjectBaseForm` with the shared functionality that both `ProjectCreateForm` and `ProjectEditForm`, with a `clean_number()` method, and any other shared functionality, if any.
* I agree that we could reuse `SetUserMixin` in the `ProjectCreate` view as it will make the code there simpler.
  * We may want to move `SetUserMixin` elsewhere, out of `site_alerts.py`, if it is being reused.

# The form itself

I'd forgotten that I needed to give myself the Service Administrator role to access the view.

After that, the form itself looks something like the mock up, which is good, and at least functionally on the right track :smiley:

# The Python changes

## Context in `forms.py`

Next I started looking at `forms.py`.

What I noted there is that the existing forms in `staff/forms.py` use:

* only inheritance of Django forms (the majority)
* inherit from a mixin and a Django form (a very small minority)
* inherit from a mixin and one of our form classes (one example only)

So the existing tendency there at least is to create new form classes.

There are a few mixins in job-server's code, but they're not that widely used.

## `ProjectCreateForm`

For the changes here, it seems like there are slightly awkward workarounds in both `ProjectEditForm` and `ProjectCreateForm` to circumvent behaviours that appear because of the inheritance of `ProjectEditForm`.

Ideally, a parent class shouldn't usually have to know about its subclasses. That suggests that inheritance might not be a good fit here.

(Of course, what's best also depends on how complicated any alternative is too.)

### Reuse of `clean_number`

* As @KatieB5 suggested, `clean_number()` looks like it's easy to move into a reusable function and takes the instance primary key and a project number as input on the form.
  * So, we could do that.
* _Another approach_ could be a shared superclass, like `ProjectBaseForm`. `ProjectBaseForm` could have only the minimum of shared functionality, including `clean_number()`, and nothing that needs overriding.
  * `ProjectEditForm` and `ProjectCreateForm` could both share what they need, without the more complex interdependency that is currently being introduced.
* `clean_number()` could also be a mixin, if it's required in other places other than the project forms. But if we don't need it elsewhere, then I think a `ProjectBaseForm` is fine.

### Managing the form fields

Again, with the inheritance, this is also harder to maintain and requires some care. Particularly so since `ProjectEditForm` includes fields, and then `ProjectCreateForm` excludes fields.

If we ever add a new field to `ProjectEditForm`, we have to then possibly remove it in `ProjectCreateForm`. So that too suggests against the existing implementation.

#### Field inclusion and exclusion

* A `ProjectBaseForm` would also remove the need for overriding the `required` state of `number`/`copilot` in `ProjectCreateForm`. This is also needed because of inheritance.
* While looking, I thought that `fields_to_disable_and_unrequire` is a bit awkward of a name.
  * Maybe something like `readonly_fields` which is a similar convention to that [used elsewhere in Django](https://docs.djangoproject.com/en/6.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.readonly_fields)? 

### Tidying up `form_valid`

* I agree that adding the existing `SetUserMixin` to the `ProjectCreate` class might be useful. Your code is duplicating the logic there.
  * It would be better to have one place where we do this. We might need to move that out of `staff/views/site_alerts.py` somewhere else.
* That could considerably simplify `ProjectCreate` to maybe something like:
  ```python
  @method_decorator(require_permission(Permission.PROJECT_CREATE), name="dispatch")
  class ProjectCreate(SetUserMixin, CreateView):
      """Create a new Project instance."""

      form_class = ProjectCreateForm
      model = Project
      template_name = "staff/project/create.html"

      def get_success_url(self):
          return self.object.get_staff_url()
  ```

PR #5601, opened 2026-02-11
Prep project number display and auto-numbering for new project identifier format changes

nishtha-kalra:
## Description

Issue #5553 - only 1 AC for this issue is covered in this PR

This PR is part of a set of PRs for forward compatibility as we migrate `Project.number` from `IntegerField` to `CharField`. To roll this change out safely, we will first update the code that uses `Project.number` to treat it as a `CharField`.
Before proceeding with these changes, I locally tested the migration and changed the `Project.number` from `IntegerField` to `CharField`. After doing so, I was able to view and edit Project number on Projects page but when I tried creating a new Project, there was an error as the logic that generates a new project number treats it as an integer.
This PR particularly only handles that part. There will also be some changes in views that will be affected when we change the Project.number to CharField.  In project page, we add # before the project number, as it was a number #{{ project.number }}. Since it won't be a number going forward, removing # before the id. 

I have also addressed the following related comments from my previous [PR](https://github.com/opensafely-core/job-server/pull/5567):

1. > Thanks Nishtha. It is good to see that the git commit messages are nice and clear about what has changed. However they don't give a rationale. It would be useful to have something in each so the log justifies the change in isolation. Even just a simple one-sentence "For forward compatibility when we migrate Project.number to a CharField".

I have now added a reason in my commit to make it clearer why I am making this change.

2. Regarding helpers - only added the helper that is being used.

3. Addresed following in `next_project_identifier` method:
> I think if there are no int values in the database we probably want to return 1, not null. Then the first instance with an int number gets 1, not None/NULL.

Returning 1 now when there is no digit in the database

>  I think the approach in _coerce_numeric_identifiers is more complicated than it needs to be

Removed that part and adjusted the method as suggested.

## Next Steps
I wanted to keep PRs small and simple so it's easier to review. Next steps that will be covered in another set of PRs before we migrate are:
1. Ordering on tables where Project ids are displayed - will start a discussion on Slack, asking what ordering looks best when we have None, Int, and alphanumeric ids.
2. Update permissions file
3. Update validation to accept both numeric and new hyphenated values - this could be done with migration as suggested in comments in my older PR linked above.

### DIFF
--- a/staff/views/applications.py
+++ b/staff/views/applications.py
@@ -76,8 +76,6 @@ class ApplicationApprove(FormView):
         }
 
     def get_initial(self):
-        # TODO: Project identifier should not be auto filled in the form when
-        # service team starts setting projects manually.
         project_number = Project.next_project_identifier()
 
         # set the value of project_name from the study_name field in the


tests/unit/jobserver/models/test_project.py
```diff
@@ -160,3 +160,17 @@ def test_project_org():
...
+    ProjectFactory(number="102")
+    ProjectFactory(number="103   ")
+
+    assert Project.next_project_identifier() == 104
+
+
+def test_next_project_identifier_returns_one_when_no_numeric_ids_exist():
+    ProjectFactory(number=None)
+
+    assert Project.next_project_identifier() == 1
```
mikerkelly: optional, test coverage: should probably include some examples of the new format in these tests? Could also add individual tests that negative and decimal (representing) values are ignored, and digit strings with leading and/or trailing whitespace.

But, I think as we are largely retiring this form due to the new applications process this is not worth spending much time on, and could be largely skipped.

PR #5602, opened 2026-02-12
Configure Dependabot for Python transitive dependencies

StevenMaude:
Specify that Python transitive dependencies should be regularly updated too.

This is how the previous update-dependencies action was configured.

Otherwise Dependabot only opens security PRs for transitive dependencies.


lucyb (approved):
In the commit message you mention Dependabot security PRs, but as I understand it, this will apply to all Dependabot PRs.

PR #5604, opened 2026-02-12
Update checkbox, input, and select form component

tomodwyer:
- Update the `form_checkbox` component
    - Alphabetise the order of the variables
    - Add label option which can be overwritten if a custom field
- Update the `form_input` component
    - Alphabetise the order of the variables
    - Add disabled and required options which can be overwritten if a custom field
- Update the `form_select` component
    - Allow setting or overwriting of variables based on state passed in
- Update all uses of `form_select` to match the updated component
- Update [UI Components page](https://jobs.opensafely.org/ui-components/) with new props, descriptions, and links

As part of this, I discovered the "Create new workspace" page would try to load a half-empty form if the GitHub API returned an error. This has been updated to match the failed response option, and return the `create_error.html` page.

### DIFF
--- a/templates/_components/form/checkbox.html
+++ b/templates/_components/form/checkbox.html
@@ -2,14 +2,12 @@
   {% var input_id=id %}
   {% var input_label=label %}
   {% var input_name=name %}
-  {% var input_required=required %}
   {% var input_value=value %}
   {% var label_for=id %}
 {% else %}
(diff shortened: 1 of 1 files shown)


tomodwyer:
Going to take another look at this in deck scrubbing. It needs more thought 😢

PR #5655, opened 2026-02-26
Codex/earlybird fp exact email rules

sebbacon:
Example when I'm trying to commit something with an email address in it:

image


mikerkelly:
I think that this was a proof-of-concept. Now there is https://github.com/bennettoxford/emailcheck/pull/2 being proposed with a similar approach, I am going to close this PR. Please re-open if required.

PR #5658, opened 2026-02-28
Apply custom project identifier ordering across staff views and management commands

nishtha-kalra:
## Description

Issue #5553 

As part of the `Project.number` transition to new identifier formats, simple ordering by number no longer gives the expected result. We now use the shared model helper` Project.apply_project_number_ordering()` so project lists are consistent across staff pages and management commands.

## Testing
Added unit tests wherever relevant.
Manually tested the changes. Please refer to [this](https://docs.google.com/document/d/1J22zyFTr04iW7QQA6sGKB1Fab1cUBuqrw1_TzePlXN8/edit?usp=sharing) doc to view screenshots.

### DIFF
--- a/jobserver/models/project.py
+++ b/jobserver/models/project.py
@@ -1,7 +1,7 @@
 import structlog
 from django.db import models
-from django.db.models import Case, IntegerField, Q, Value, When
-from django.db.models.functions import Cast, Lower, Substr
+from django.db.models import Case, CharField, F, IntegerField, Q, Value, When
+from django.db.models.functions import Cast, Lower
 from django.urls import reverse
 from django.utils import functional, timezone
 from django.utils.text import slugify
@@ -9,7 +9,34 @@ from furl import furl
 
 
 logger = structlog.get_logger(__name__)
-NUMERIC_IDENTIFIER_REGEX = r"^[0-9]+$"
+
+
+class ProjectQuerySet(models.QuerySet):
+    def order_by_project_identifier(self):
+        """
+        Return projects ordered by project identifier.
+        Ordering rules:
+        1. POS-format identifiers sort first, in reverse lexical order.
+        2. Numeric identifiers sort next, by numeric value descending.
+        3. Blank or null identifiers sort last.
+        4. Project name is used as a case-insensitive tie-breaker.
+        """
+        return self.annotate(
+            pos_format_lex=Case(
+                When(number__startswith="POS-", then=F("number")),
+                default=Value("", output_field=CharField()),
+                output_field=CharField(),
+            ),
+            numeric_value=Case(
+                When(number__regex=r"^[0-9]+$", then=Cast("number", IntegerField())),
+                default=Value(None, output_field=IntegerField()),
+                output_field=IntegerField(),
+            ),
+        ).order_by(
+            "-pos_format_lex",
+            F("numeric_value").desc(nulls_last=True),
+            Lower("name"),
+        )
 
 
 class Project(models.Model):
@@ -73,6 +100,8 @@ class Project(models.Model):
         on_delete=models.PROTECT,
         related_name="projects_updated",
     )
+
+    objects = ProjectQuerySet.as_manager()
 
     class Meta:
         constraints = [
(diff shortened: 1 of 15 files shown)


mikerkelly:
Hi Nishtha, thanks for this. I have some suggestions to make this more idiomatic and hopefully a bit simpler / more performant.

### Ordering

You have used a `@classmethod` for the ordering function. This works functionally but is not idiomatic. Methods that take and return QuerySet should be on [custom manager or custom queryset methods](https://docs.djangoproject.com/en/6.0/topics/db/managers/#calling-custom-queryset-methods-from-the-manager) so that they fit in with the QuerySet API and are composable with other QuerySet methods. For example, I expect to be able to filter a QuerySet then order it with this method as `Project.objects.filter(...).apply_project_number_ordering()`, but instead I have to do `Project.apply_project_number_ordering(Project.objects.filter(...))`. This would affect your changes to `UserDetailWithEmail.get_queryset`, for example. 

I don't think it is necessary to define the `number_type` in order to do the desired sorting, so long as we can sort rows with the default value of each annotation last, which I think we can.

I don't think it is necessary to analyse the year and sequence segments separately. Because the segments are always 4 digits long, we can just take reverse lexicographic over the entire number, when they start "POS-". This is as discussed in [the Slack thread ](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1771321697661689?thread_ts=1771320703.293479&cid=C069SADHP1Q)about the approach to take to this issue. If that works, it seems significantly simpler.

I agree with you and Steve that `startswith="POS-"` is sufficient and preferable to a more rigorous but slower regex, given our other guarantees.

You can make multiple annotations as separate parameters to a single `annotate` call and I feel that this is usually more readable.

I have tried the following locally and it seems to work on my local data that is based on production but with added "POS-" projects, but I have not tested it rigorously:

```python
class ProjectQuerySet(models.QuerySet):
    def standard_ordering(self):
        return (
            self.annotate(
                pos_format_lex=Case(
                    When(number__startswith="POS-", then=F("number")),
                    default=Value("", output_field=CharField()),
                    output_field=CharField(),
                ),
                numeric_value=Case(
                    When(number__regex=r"^[0-9]+$", then=Cast("number", IntegerField())),
                    default=Value(None, output_field=IntegerField()),
                    output_field=IntegerField(),
                ),
            )
            .order_by(
                "-pos_format_lex",
                F("numeric_value").desc(nulls_last=True),
                Lower("name"),
            )
        )

class Project(models.Model):
    ...
    
    objects = ProjectQuerySet.as_manager()
    
class ProjectList(ListView):
    queryset = Project.objects.all().standard_ordering()

```

Hopefully this helps but sorry if I missed something. I wanted to get you the feedback rapidly in case this was soon to merge.

### Testing

The view unit tests test the details of how the ordering works. This seems to repeat the tests of the function itself. A fixture as Steve suggests would improve the repetition somewhat, but I think it may be better just not to have unit testing for the view ordering at all? For the class-based views, all we are really testing is that our `queryset=` line has been applied, which seems to be testing the Django machinery, which is redundant. I think the fixture approach would be a better fit for integration testing of these views, if they have them.

nishtha-kalra:
> @nishtha-kalra: Thanks for working on this more; I know it's been open a while. I can look again at the current form of the PR.
> 
> I do really like Mike's approach (and, still not knowing Django well, I learned a lot from reading it and now being aware of the linked `Managers` documentation), as a cleaner interface into the ordering.
> 
> Whether we do that now, or subsequently, I think we should consider it. If it's simple enough to change it now, then it might be timely to do so now. Maybe it's worth spending 30 minutes on it, and seeing how far you get with it?
> 
> If this PR is seriously blocking other issues, and going to take a while to change, then we could merge as is, and refactor it later.
> 
> It feels isolated enough code that it's not too problematic for someone to change in the near future (swapping out the class method calls to an actual method, and/or changing querysets).

@StevenMaude Thanks for looking into the changes again. Yes, I like Mike's suggestions too. Since this PR is not blocking anything, I am happy to do these changes in this PR. I will let you know once I am done with them so you can review.

jobserver/models/project.py
```diff
@@ -1,15 +1,42 @@
...
+        4. Project name is used as a case-insensitive tie-breaker.
+        """
+        return self.annotate(
+            pos_format_lex=Case(
+                When(number__startswith="POS-", then=F("number")),
+                default=Value("", output_field=CharField()),
+                output_field=CharField(),
+            ),
+            numeric_value=Case(
+                When(number__regex=r"^[0-9]+$", then=Cast("number", IntegerField())),
```
StevenMaude: note: I think this regex has the same function as `DIGITS_PATTERN` in #5679. We could maybe reuse that here in future.

staff/views/users.py
```diff
@@ -123,21 +124,25 @@ def get_context_data(self, **kwargs):
...
-        )
-
         projects = [
             {
-                "name": m.project.title,
-                "roles": sorted(r.display_name for r in m.roles),
-                "staff_url": m.project.get_staff_url(),
+                "name": project.title,
+                "roles": sorted(
+                    role.display_name for role in project.user_memberships[0].roles
```
StevenMaude: thought: I think the existence of a user_membership is given by the filtering of projects with user memberships below, but accessing an element always has the risk of `IndexError` if it doesn't. If we're confident this is the case, then it's fine though.

StevenMaude (approved):
Thanks for your hard work on this, @nishtha-kalra. I think this looks a lot improved now.

(I based this subsequent review on the most recent commit only. I don't believe you'd modified previous commits, and that's helpful for reviewing a big set of changes.)

PR #5670, opened 2026-03-06
Add a test to check permissions project identifiers are entered correctly

### LINKED ISSUE #5657: Tests for permissions files
Add tests to detect manual-entry errors in [permission sets](https://github.com/opensafely-core/job-server/tree/main/jobserver/permissions).
Why? Because project identifier is now a `CharField` to support new alphanumeric format. We had to convert the project numbers added in the permissions allowlist to string. We want to add tests so that there are no errors when new numbers are manually added.
The numbers could either be a digit like string like`"123"`, a alphanuemric number like `POS-2025-2001` or `project.slug` if the project does not have a number.

StevenMaude:
Fixes #5657.

There was some discussion about the notion of project "number" getting decoupled from the project "identifier", so I've gone with "identifier" in this code for now, but mentioned "number". It should be a fairly straightforward `s/number/identifier` if it needs updating.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -35,6 +35,7 @@
     - [Both set or null](#both-set-or-null)
     - [updated\_at and updated\_by](#updated_at-and-updated_by)
 - [Actions](#actions)
+- [Permissions](#permissions)
 - [Auditing events](#auditing-events)
   - [Presenters](#presenters)
 - [Interfaces](#interfaces)
@@ -543,6 +544,23 @@ complex cross-model interactions or significant side effects.
 
 This is currently a convention, not an enforced rule. It may not be
 consistently applied.
+
+## Permissions
+
+The [jobserver.permissions](jobserver/permissions/) package is, by convention,
+the home for code that controls project access to restricted data sources or
+features.
+
+Care is needed when editing these files:
+
+* some of the modules are owned by IG (see [`CODEOWNERS`](.github/CODEOWNERS))
+* there are current existing conventions
+  * project permissions are usually defined by an iterable of strings
+    as a module-level value, along with a function that checks a value
+    for membership
+  * the name of that module-level value is typically `PROJECTS_WITH…PERMISSION`
+  * there are unit tests of those permissions values that validate the project
+    identifiers; if you create a new variable, add it to those tests
 
 ## Auditing events
 


jobserver/permissions/validator.py
```diff
@@ -0,0 +1,55 @@
+import re
+
+from staff.forms import PROJECT_IDENTIFIER_PATTERN
+
+
+# In #5658, there's a NUMERIC_IDENTIFIER_REGEX,
+# it might be preferable to use that for consistency once merged.
+NUMERIC_IDENTIFIER_REGEX = re.compile(r"^[0-9]+$")
```
StevenMaude: thought: This is a copy of what's used there already. I think we could probably avoid matching a leading zero though.

tests/unit/jobserver/permissions/test_permissions.py
```diff
@@ -0,0 +1,39 @@
...
+import pkgutil
+import re
+
+import pytest
+
+from jobserver import permissions
+from jobserver.permissions.validator import validate_project_identifier_in_permissions
+
+
+PERMISSIONS_VARIABLE_NAME_PATTERN = re.compile(r"^PROJECTS_WITH_[A-Z0-9_]+_PERMISSION$")
```
StevenMaude: thought: This is the current convention that the relevant names adhere to, but not sure it is codified anywhere.

We should probably document:

* the naming convention
* the specific type(s) that should be used for values of this kind
* the existence of these tests

in the `permissions/` directory.

jobserver/permissions/validator.py
```diff
@@ -0,0 +1,55 @@
...
+
+    It also does not validate values provided for a project identifier,
+    such as allowed ehrQL tables.
+
+    Project numbers in these configurations are strings and can be:
+
+    * project slugs
+    * entirely numeric identifiers
+    * identifiers of the format `POS-XXXX-YYYY` where X is a year, and Y
+      is a number with 4 digits or more.
```
StevenMaude: thought: Again, the "or more" specifies the current existing behaviour, and not necessarily what we want.

jobserver/permissions/validator.py
```diff
@@ -0,0 +1,55 @@
...
+        if project_identifier == "opensafely-internal":
+            continue
+
+        if NUMERIC_IDENTIFIER_REGEX.fullmatch(project_identifier):
+            continue
+
+        if PROJECT_IDENTIFIER_PATTERN.fullmatch(project_identifier):
+            continue
+
+        invalid_project_identifiers.add(project_identifier)
```
mikerkelly: Thought, style, optional: I would probably by habit write this as a comprehension as I find that easier to read, more compact, more "declarative style" (though just syntactic sugar), and sometimes easier for the compiler to optimise. Up to you though.

```python
invalid_project_identifiers = {
    project_identifier
    for project_identifier in project_identifiers
    if project_identifier != "opensafely-internal"
    and not NUMBER_REGEX.fullmatch(project_identifier)
}
```

tests/unit/jobserver/permissions/test_validator.py
```diff
@@ -0,0 +1,129 @@
+import pytest
+
+from jobserver.permissions.validator import validate_project_identifier_in_permissions
+
+
+# We use the identifiers also as names in the tests.
+# Therefore, they need to be in an ordered data structure,
+# to ensure they are consistent across xdist workers.
+project_identifier_test_cases = [
```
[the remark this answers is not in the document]
mikerkelly: Thought: could we decouple testing the regex from testing the `validate_project_identifier_in_permissions`? If we patch out the regex to return true/false in the function unit tests, we can separate the concerns.

This might be best left for https://github.com/opensafely-core/job-server/issues/5510, as probably all of the clients of the shared regex don't need to repeat similar testing of its behaviour. Each client test can patch it out, and the canonical regex tests can be comprehensive.

tests/unit/jobserver/permissions/test_validator.py
```diff
@@ -0,0 +1,129 @@
...
+def test_validate_project_identifier_in_permissions(
+    project_identifiers, expected_invalid
+):
+    identifiers = set(project_identifiers)
+
+    if not expected_invalid:
+        validate_project_identifier_in_permissions(identifiers)
+    else:
+        with pytest.raises(ValueError) as exc:
+            validate_project_identifier_in_permissions(identifiers)
```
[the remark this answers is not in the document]
StevenMaude: I think it's reasonable to leave this for now. There would be some duplication needed (assigning the test names, patching out the regex) to get the tests to work as they do.

StevenMaude:
I've added commits on here and made small changes, for convenience of review/diffing.

I think it's fine to rebase this into one or two commits when done.

mikerkelly (approved):
Thanks, Steve, think this looks really nice and clean now. A couple very minor comments on your latest, but happy to approve.

PR #5734, opened 2026-03-26
Make the service team responsible for creating new projects

### LINKED ISSUE #5503: Grant the Service Team access to create new projects
> [!WARNING]
> Please don't merge without confirming with Mary first. There's a slack [thread](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1774277768617739) with some more recent context.

# Why are we doing this?

We temporarily left `StaffAreaAdministrator` with permissions that will eventually belong to `ServiceAdministrator` only. Once that block of work is ready to hand over to the service team, we should remove these temporary permissions and ensure only the right people have the new role. 

https://github.com/opensafely-core/job-server/issues/5476 and https://github.com/opensafely-core/job-server/issues/5475 added the permissions.

 `ServiceAdministrator` is the only role with `user_edit_project_roles` permissions and any others that are just for the service admin role.

Note https://github.com/opensafely-core/job-server/issues/5637 may add `Permission.PROJECT_LINK_TO_APPLICATION`.

https://github.com/opensafely-core/job-server/pull/5669 requires `Permission.ORG_CREATE` be added to the role but not removed from `StaffAreaAdministrator` see [Slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1774278175853399?thread_ts=1774277768.617739&cid=C069SADHP1Q) also.

# What are we doing?

- Assigning the required members of the service team the `ServiceAdministrator` role.
- Removing `user_edit_project_roles` and any other permission identified above from the `StaffAreaAdministrator` role, and adding them to the `ServiceAdministrator` role if not already present.
- Updating the role description to remove the caveat about it not being used in production.
- Addressing any UI issues; we should not display functionality that the `StaffAreaAdministrator` cannot access (otherwise it will error when they try).
---

StevenMaude:
Fixes #5503.

Summary:
Screenshot from 2026-03-31 12-28-34

Testing:
Change in UI:
Not able to go to project details page if user doesn't have permissions to link application
Screenshot from 2026-03-30 18-10-33

Don't show edit/remove member buttons if user doesn't have permissions to edit roles
Screenshot from 2026-03-31 12-15-51

### DIFF
--- a/templates/staff/project/detail.html
+++ b/templates/staff/project/detail.html
@@ -87,17 +87,19 @@
           {% for membership in memberships %}
             {% #description_item title=membership.user.get_full_name|default_if_none:membership.user.username %}
               <div class="flex flex-col gap-1">
-                <div class="flex flex-row gap-2">
-                  {% #button type="link" variant="secondary-outline" href=membership.get_staff_edit_url|add:"?next="|add:request.path small=True %}
-                    Edit
-                  {% /button %}
-                  <form method="POST" action="{{ membership.get_staff_remove_url }}">
-                    {% csrf_token %}
-                    {% #button type="submit" variant="danger-outline" small=True %}
-                      Remove <span class="sr-only">from {{ project.title }}</span>
+                {% if user_can_edit_members %}
+                  <div class="flex flex-row gap-2">
+                    {% #button type="link" variant="secondary-outline" href=membership.get_staff_edit_url|add:"?next="|add:request.path small=True %}
+                      Edit
                     {% /button %}
-                  </form>
-                </div>
+                    <form method="POST" action="{{ membership.get_staff_remove_url }}">
+                      {% csrf_token %}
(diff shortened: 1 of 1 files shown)


mikerkelly (commented):
Thanks, Nishtha, Steve. This is close now.

I have a couple of minor comments on the text that Steve added.

Thanks for fixing up the tests, Nishtha. These look great now. I would marginally prefer that they now be squashed with the previous commit so that it does not have failing tests.

The UI changes look good for the Add members button. Thanks for fixing the additional Find and Link button.

I think that there is still some UI change to make. The edit and delete buttons should not be visible here without the project roles edit permission:

Image

This is on for example http://localhost:8000/staff/projects/foo/ I can see them with just StaffAreaAdministrator, but get a permission error when I click them. Would be better to hide the buttons please.

TODO:
- roles comments
- hide buttons on project detail page without permission

Issue #5699, opened 2026-03-17
Set a lead org for each project


tomodwyer:
# Why are we doing this?

- `is_lead` is not being set on the orgs when it is updated in the staff area
- `orgs` is a `ManyToManyField` ([source](https://github.com/opensafely-core/job-server/blob/53d7d1079a343840163b1d2f988ea1fdc626d0fb/jobserver/models/project.py#L44-L48)) - but we only ever show one in the UI, so set `is_lead` on the one we want to show ([source](https://github.com/opensafely-core/job-server/blob/53d7d1079a343840163b1d2f988ea1fdc626d0fb/jobserver/actions/projects.py#L28-L42))

We should make sure we either only set one org per project, or we always set a lead org when changing orgs on a project. A database check constraint cannot enforce "exactly one of these", so we will have to rely on the `actions.projects.edit` business logic to enforce this, for now.

Alternatively, now or later, make `lead_org` a foreign key, not picking from among the `is_lead` organisation from the `orgs` `ManyToManyField`. The primary organisation is semantically different enough to warrant its own foreign key. There are no projects in production that have more than one organisation right now. This keeps causing pain. Fixing the data model would make many parts of the application simpler and more consistent. It would be more work now, but may save it in the long run. This was added in #2827 and #3571 speculatively in case we wanted to represent more than one organisation collaborating on a project, without any use of it since then. If we do want to represent this in future, I think a better representation would be to use one foreign key for the main organisation and a many-to-many relation for collaborators. That reflects the reality that all projects have a main organisation, and the UI that wants to centre that organisation. We can just do the foreign key part now, leaving `orgs` in place.

# How will we know when it's done?

When each project only has one org, or only has one lead org.

# What are we doing?

- Set `is_lead` in `Projects.actions.edit` after call to `orgs.set`. https://github.com/opensafely-core/job-server/pull/5731
- fix `actions.project.edit`. It is currently mixing concerns by using form methods in a confusing way. Instead it should completely bypass the `form.save`, decoupling the action from the form. 
  - We recently changed this form and maybe that has silently broken this code in other ways, but it is hard to reason about. 
- make `ProjectEdit` view use the project edit action in the public site, so we consistently use the same code path (though for now as that only updates status and status description, there are no relevant side effects, that may not be true in future).

(Optional, new issue?) If we do the foreign key approach:
- PR to make `lead_org` a primary key
  - add `Project.lead_org` as a nullable `ForeignKey` (more easily searchable than `org`). 
[...]

PR #5745, opened 2026-03-31
Decouple `actions.projects.edit` from form

mikerkelly:
Fixes https://github.com/opensafely-core/job-server/issues/5699.

This was a TODO about doing this. The logic of the staff project edit view was moved to this function but relied on being passed in a model form. That was mixing concerns, which was confusing and hard to reason about. Instead, accept a data dict of fields to updated, constrained to be from an allowed set. That can be from a form, but need not be.

Use `actions.projects.edit` in main site `ProjectEdit` view. This ought to use the business-logic in the action to make the change for consistency and separation of concerns.

### DIFF
--- a/applications/views.py
+++ b/applications/views.py
@@ -124,9 +124,8 @@ class ResearcherCreate(CreateView):
         return super().dispatch(request, *args, **kwargs)
 
     def form_valid(self, form):
-        researcher = form.save(commit=False)
-        researcher.application = self.application
-        researcher.save()
+        form.instance.application = self.application
+        super().form_valid(form)
 
         return redirect(get_next_url(self.request.GET))
 
--- a/staff/views/orgs.py
+++ b/staff/views/orgs.py
@@ -48,11 +48,11 @@ class OrgCreate(CreateView):
     model = Org
 
     def form_valid(self, form):
-        org = form.save(commit=False)
-        org.created_by = self.request.user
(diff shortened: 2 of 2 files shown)


jobserver/views/projects.py
```diff
@@ -176,12 +176,9 @@ def form_valid(self, form):
     def get_object(self):
         project = get_object_or_404(Project, slug=self.kwargs["project_slug"])
 
-        user = self.request.user
-        user_has_project_manage = has_permission(
-            user, Permission.PROJECT_MANAGE, project=project
-        )
-
-        if not user_has_project_manage:
+        if not has_permission(
```
KatieB5: This is just a question out of curiosity. For the Staff Area `ProjectEdit` view, we use a method decorator to check that someone has the permission to view the project edit form:

`@method_decorator(require_permission(Permission.STAFF_AREA_ACCESS), name="dispatch")`

Is there a reason(s) we don't use that approach here, and opt for the conditional `has_permission` approach instead?

mikerkelly: I wondered the same thing then realised that it is because these are project-specific permissions and there is not a decorator version of the `require_permission` check that can check project-specific permissions. `has_permission` takes project as an optional parameter.

We could try and write one but it would need to be able to find the project object from the instance of the class that it is wrapping and I am not sure how challenging that would be.

PR #5752, opened 2026-04-01
Show users with `STAFF_AREA_ACCESS` as `copilot` choices.

### LINKED ISSUE #5750: Reduce list of users shown in the copilot field on the project create form
# Why are we doing this?

The list of copilots in the project create form is really long at the moment.

The list of copilots is shorter and easier for the copilots to use.

# What are we doing?

Only show users with either any role or the staff area admin role depending on the outcome of a conversation with copilots.

[Slack thread asking the question to the copilots.](https://bennettoxford.slack.com/archives/C028EJH752A/p1774954794484339) -- confirmed they will have this role until the copilot role is created. If the access is reassessed before that, we can always remove this and limit it to logged-in users, instead.

---

mikerkelly:
Fixes https://github.com/opensafely-core/job-server/issues/5750.

The list of all users displayed for the `copilot` field in the `ProjectCreateForm` and `ProjectEditForm` is very long in production -- anyone with a GitHub account can log in. Shortening it to a shorter valid set makes it slightly easier to navigate. This also makes manual errors less likely.

Ideally we would use a copilot-specific permission or role but that does not exist (yet?). So let's use `STAFF_AREA_ACCESS` as in practice all copilots do currently have that access for accessing the copilot dashboards (and formerly the application process).

Adjust the form field QuerySet to reduce the set of choices to those with roles that have the required permission. A function that returns User QuerySet is a User manager method so put it there, and this means that `with_permission` can be used more widely.

`potential_copilots` has the permission as a parameter to help with testing.

Add `ALL_ROLES` and `roles_with_permission` utility for making the `QuerySet` without hardcoding in the `StaffAreaAdministrator` role. Ideally all application behaviour should be driven by permissions, not roles, to decouple changes to the roles from the specific views et cetera.

Add `user_with_fake_role_factory` test fixture to aid with making `User` with fake permissions, to help with decoupling tests from specific roles and permissions. I updated existing tests using this where needed due to the copilot field change to show how this can be used, and also applied it to the rest of the integration tests and the `ProjectAddMember` unit tests for illustrative purposes.

The form tests use a fixture to patch the `QuerySet` returned by the `User.objects.potential_copilots` so that the other tests can be decoupled from the specific permission.

### DIFF
--- a/jobserver/models/user.py
+++ b/jobserver/models/user.py
@@ -31,8 +31,13 @@ class UserQuerySet(models.QuerySet):
 
     def potential_copilots(self, copilot_permission=Permission.STAFF_AREA_ACCESS):
         """
-        There is no copilot role, so we use staff area access as a proxy.
-        Copilots need access to view the copilot dashboard.
+        Return a QuerySet of User that could be a copilot.
+
+        There is no copilot role, so we use staff area access as a proxy. If we
+        add a more specific permission later we could use that instead.
+
+        Copilots need access to view the copilot dashboard, and may be assigned
+        to Project.copilot.
         """
         return self.with_permission(copilot_permission)
 
--- a/jobserver/models/user.py
+++ b/jobserver/models/user.py
@@ -15,11 +15,26 @@ from django.utils import timezone
 from django.utils.functional import cached_property
 from sentry_sdk import capture_message
 
+from jobserver.authorization.permissions import Permission
+from jobserver.authorization.utils import roles_with_permission
+
 from ..authorization.fields import RolesArrayField
 from ..hash_utils import hash_user_pat
 
 
 logger = structlog.get_logger(__name__)
+
+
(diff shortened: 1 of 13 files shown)


jobserver/models/user.py
```diff
@@ -15,13 +15,28 @@
...
 
 
 logger = structlog.get_logger(__name__)
 
 
+class UserQuerySet(models.QuerySet):
+    def with_permission(self, permission):
+        return self.filter(roles__overlap=roles_with_permission(permission))
+
+    def potential_copilots(self, copilot_permission=Permission.STAFF_AREA_ACCESS):
```
nishtha-kalra, line 32: Can we add a TODO or a note here that eventually we would want to replace this with a copilot specific permission?

nishtha-kalra (approved):
Thanks for clear and descriptive commit messages. 
Looks good, just a small suggestion/question.

PR #5770, opened 2026-04-08
Add get_display_name() method to User model

### LINKED ISSUE #5746: Display username with full name consistently in the staff area
# Why are we doing this?

The user full name (like "Mike Kelly") is a good way to identify users, but it may be ambiguous in some cases, as users may have more than one account, or different people may have similar or identical names. The username (tied to GitHub OAuth, like "mikerkelly") is unique and may be helpful to display wherever we list or display a user in the staff area.

https://github.com/opensafely-core/job-server/pull/5704 already did this for the project create form, and it has long been on the project edit from. But there are several other places in the staff UI that could benefit from this. 

Done on project edit form:

Image

Could be done on project detail page -- user links are good, but the text could be more explicit too:

Image

Image

All user displaying pages in the staff area have been updated, unless not appropriate.

# What are we doing?

Update the following Staff area pages to display a user's username alongside the user's full name.

---

KatieB5:
Closes #5746

The user full name (like "Mike Kelly") is a good way to identify users, but it may be ambiguous in some cases, as users may have more than one account, or different people may have similar or identical names. The username (tied to GitHub OAuth, like "mikerkelly") is unique and may be helpful to display wherever we list or display a user in the staff area.

#5704 already did this for the project create form, and it has long been on the project edit from. But there are several other places in the staff UI that could benefit from this. 

This PR updates the displayed name of a user from: **Full Name**, to: **Full Name (username)**, in several places throughout the staff area

### Questions for reviewer

1. Do we want to have the user's username displayed alongside the fullname in _all_ of the cases listed below?
2. I didn't add the username to the fullname displayed in breadcrumbs, as to me this seemed necessary; do you have an opinion?
3. To use `get_display_name()` in the Project and User Audit Log events table templates, I changed the presenters for projects, users and project members. Is there a better way to do this? I don't fully understand the presenters, `LinkableObjects` and `PresentableAuditableEvents` rationale/functionality, and I'm conscious a better approach may exist.

### Testing

- `just test`
- `just test-ci`
- `just run` (manually inspect pages/elements listed below)

### Updated staff area pages
I noted following pages where we display a user's full name, and could show the username alongside this.

- Applications list
- Applications detail 
- Application link
- Project detail 
- Project Audit log 
- User Audit log 
- Co-piloting dashboard 
- Project dashboard 
- Job request list 
- Job request detail 
- Org detail 
- Redirect detail 
- Repo detail 
- User list 
- User detail (with email and with Oauth)
- User edit global roles
- Workspace detail

### DIFF
--- a/jobserver/auditing/presenters/users.py
+++ b/jobserver/auditing/presenters/users.py
@@ -4,8 +4,10 @@ from .utils import lookup_user, roles_str_to_class_name_str
 
 
 def updated_roles(*, event: AuditableEvent) -> PresentableAuditableEvent:
+    actor_user_obj = lookup_user(event.created_by)
+
     actor = LinkableObject.build(
-        obj=lookup_user(event.created_by),
+        obj=actor_user_obj,
         link_func="get_staff_url",
     )
     user = LinkableObject.build(
@@ -15,6 +17,7 @@ def updated_roles(*, event: AuditableEvent) -> PresentableAuditableEvent:
     return PresentableAuditableEvent(
         context={
             "actor": actor,
+            "actor_user_obj": actor_user_obj,
             "created_at": event.created_at,
             "user": user,
             "before": roles_str_to_class_name_str(event.old),
--- a/jobserver/auditing/presenters/users.py
+++ b/jobserver/auditing/presenters/users.py
@@ -4,20 +4,20 @@ from .utils import lookup_user, roles_str_to_class_name_str
 
 
 def updated_roles(*, event: AuditableEvent) -> PresentableAuditableEvent:
-    actor_user_obj = lookup_user(event.created_by)
 
     actor = LinkableObject.build(
-        obj=actor_user_obj,
+        obj=lookup_user(event.created_by),
+        field="display_name",
         link_func="get_staff_url",
     )
     user = LinkableObject.build(
         obj=lookup_user(event.target_user),
+        field="display_name",
         link_func="get_staff_url",
     )
     return PresentableAuditableEvent(
         context={
             "actor": actor,
-            "actor_user_obj": actor_user_obj,
             "created_at": event.created_at,
             "user": user,
             "before": roles_str_to_class_name_str(event.old),
--- a/jobserver/models/user.py
+++ b/jobserver/models/user.py
@@ -210,6 +210,10 @@ class User(AbstractBaseUser):
         """Support Django's User contract"""
         return self.fullname
 
+    def get_display_name(self):
+        """Return a user's name in the format: fullname (username)"""
+        return f"{self.fullname} ({self.username})"
+
     def get_logs_url(self):
         return reverse("user-event-log", kwargs={"username": self.username})
 
--- a/jobserver/models/user.py
+++ b/jobserver/models/user.py
@@ -189,6 +189,11 @@ class User(AbstractBaseUser):
 
         return "".join(w[0].upper() for w in self.fullname.split(" "))
 
+    @property
+    def display_name(self):
+        """Return a user's name in the format: fullname (username)"""
+        return f"{self.fullname} ({self.username})"
+
     def get_absolute_url(self):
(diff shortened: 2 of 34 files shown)


jobserver/models/user.py
```diff
@@ -210,6 +210,10 @@ def get_full_name(self):
         """Support Django's User contract"""
         return self.fullname
 
+    def get_display_name(self):
```
mikerkelly: Driveby suggestion: could this be an [`@property`](https://docs.python.org/3/library/functions.html#property) instead, like initials?

> The @property decorator turns the voltage() method into a “getter” for a read-only attribute with the same name, and it sets the docstring for voltage to “Get the current voltage.”

This will allow it to be accessed with `user.display_name` rather than `user.get_display_name`. As this is statically derived directly from existing model fields, this feels like a good fit as it behaves like another read-only attribute.

templates/staff/user/detail_with_email.html
```diff
@@ -50,7 +50,7 @@
       {% endif %}
 
       {% #card title="User details" container=True container_class="flex flex-col items-start gap-4" %}
-        {% form_input field=form.fullname label="Full name" class="max-w-prose w-full" %}
+        {% form_input field=form.get_display_name label="Full name" class="max-w-prose w-full" %}
```
nishtha-kalra: I think this should still be fullname.

KatieB5: Thanks Nishtha, that's a good spot. I agree - I'll update it :)

jobserver/models/user.py
KatieB5: Thank you Mike, I think that's a nice suggestion and I'm happy to make that change.

jobserver/auditing/presenters/users.py
```diff
@@ -4,8 +4,10 @@
 
 
 def updated_roles(*, event: AuditableEvent) -> PresentableAuditableEvent:
```
nishtha-kalra, line 6:
```suggestion
def updated_roles(*, event: AuditableEvent) -> PresentableAuditableEvent:
    actor_user_obj = lookup_user(event.created_by)
    user_obj = lookup_user(event.target_user)

    actor = LinkableObject.build(
        obj=actor_user_obj,
        link_func="get_staff_url",
    )
    user = LinkableObject.build(
        obj=user_obj,
        link_func="get_staff_url",
    )
    return PresentableAuditableEvent(
        context={
            "actor": actor,
            "actor_user_obj": actor_user_obj,
            "created_at": event.created_at,
            "user": user,
            "user_obj": user_obj,
            "before": roles_str_to_class_name_str(event.old),
            "after": roles_str_to_class_name_str(event.new),
        },
        template_name="staff/auditable_events/user_updated_roles.html",
    )
```

nishtha-kalra (approved):
I have suggested the change needed to fix the issue where the name is not displayed when global roles are updated.
I'm approving the PR as everything looks good once this fix is added.
[slack thread](https://bennettoxford.slack.com/archives/C080XJJ0KDJ/p1777641510048499?thread_ts=1775744177.216019&cid=C080XJJ0KDJ)

jobserver/auditing/presenters/users.py
KatieB5: Thanks very much Nishtha! I appreciate you looking into solutions for this :) 

As discussed in [Slack](https://bennettoxford.slack.com/archives/C080XJJ0KDJ/p1778510155024909?thread_ts=1775744177.216019&cid=C080XJJ0KDJ), we've agreed to go with an alternative to the above approach, as this is slightly cleaner.

nishtha-kalra (approved):
Thanks, Katie, for updating the PR as discussed in the [Slack thread](https://bennettoxford.slack.com/archives/C080XJJ0KDJ/p1775744177216019)
Happy to approve it again, it is good to go now :)

PR #5785, opened 2026-04-14
Restrict visibility of workspace statuses to users with roles, to reduce potential public confusion

### LINKED ISSUE #5647: Restrict workspace status to users with particular roles
# Why are we doing this?

The workspace status can be "active" or "archived". This is useful for researchers when organising their work and is used in a few places across the system. However, it can be misleading to the public as it looks like it's related to the project status (i.e it seems like an ongoing project should have active workspaces and a closed project should have archived workspaces). In reality, the permissions system is unrelated to this. Longer term, we could tidy this up and make sure all workspaces are automatically archived for "closed" projects. However, for now we want to avoid presenting a confusing picture to anyone viewing a project.

The workspace status is only visible to researchers on a project or those with appropriate global roles.

# What are we doing?

Make sure the workspace status is only visible to researchers on a project or those with appropriate global roles on:
- [ ] The project detail view ([example](https://jobs.opensafely.org/analysis-of-the-pharmacy-first-element-in-the-plan-to-restore-access-to-primary-care-following-the-impact-of-covid-19/))
- [ ] ...

---

eli-miriam:
Closes #5647 

To find all pages where this info was used, I searched on `workspace.is_archived` across the repo, and then excluded all the pages (staff area, the 'yours' section, job requests & repo sign-off) where the necessary roles  to view workspace status would always be present by default.

The remaining pages which have now changed are:
- the project detail page; status pills for workspaces are only visible to logged-in role-having users. [http://localhost:8000/opensafely-internal/](http://localhost:8000/opensafely-internal/) demonstrates this
- the workspace detail page; the text at the top of the page denoting an archived workspace is only visible to logged-in role-having users. [http://localhost:8000/opensafely-internal/diamorphine-blank-dmd-check/](http://localhost:8000/opensafely-internal/diamorphine-blank-dmd-check/) demonstrates this
- the workspace logs page; the text at the top of the page denoting an archived workspace is only visible to logged-in role-having users. [http://localhost:8000/opensafely-internal/diamorphine-blank-dmd-check/logs/](http://localhost:8000/opensafely-internal/diamorphine-blank-dmd-check/logs/) demonstrates this.

I'm not 100% sold that the latter two changes are strictly necessary; in particular, the archived workspace note doesn't run the risk of someone mistakenly thinking that an 'active' workspace has a live link to sensitive data or so on. On the flip side, probably the only people who know and care about the distinction are strictly contained within the set of people to whom it's still visible anyway.

I don't believe there's documentation that needs updating; everywhere that the workspace status is referenced is already in sections that assume that the user has the roles that would get them visibility of it anyway.

The diff in the templates seems to be very inconsistent about whether or not it handles the extra `{% if user_can_view_workspace_statuses %}` checks nicely - all that's changed in the templates is surrounding the parts that deal with workspace status with the relevant context data check, even though some of the diffs look more involved.

### DIFF
--- a/jobserver/models/workspace.py
+++ b/jobserver/models/workspace.py
@@ -42,6 +42,9 @@ class Workspace(models.Model):
 
     name = models.TextField(unique=True)
     branch = models.TextField()
+
+    # We don't show this field publicly to anonymous users, or to users without roles,
+    # as it may cause confusion or be misleading to users without necessary context.
     is_archived = models.BooleanField(default=False)
     should_notify = models.BooleanField(default=False)
     purpose = models.TextField(default="")
--- a/jobserver/views/projects.py
+++ b/jobserver/views/projects.py
@@ -41,8 +41,9 @@ class ProjectDetail(View):
             request.user, Permission.PROJECT_MANAGE, project=project
         )
 
-        # An AnonymousUser doesn't have the has_any_roles property, so we use getattr
-        can_view_workspace_statuses = getattr(request.user, "has_any_roles", False)
+        can_view_workspace_statuses = (
+            request.user.is_authenticated and request.user.has_any_roles
+        )
 
         memberships = project.memberships.select_related("user").order_by(
             Lower("user__fullname"), "user__username"
--- a/jobserver/views/projects.py
+++ b/jobserver/views/projects.py
@@ -41,6 +41,9 @@ class ProjectDetail(View):
             request.user, Permission.PROJECT_MANAGE, project=project
         )
 
+        # An AnonymousUser doesn't have the has_any_roles property, so we use getattr
+        can_view_workspace_statuses = getattr(request.user, "has_any_roles", False)
+
         memberships = project.memberships.select_related("user").order_by(
             Lower("user__fullname"), "user__username"
         )
@@ -84,6 +87,7 @@ class ProjectDetail(View):
         context = {
             "can_create_workspaces": can_create_workspaces,
             "can_manage_project": can_manage_project,
+            "can_view_workspace_statuses": can_view_workspace_statuses,
             "first_job_ran_at": first_job_ran_at,
             "memberships": memberships,
             "outputs": self.get_outputs(workspaces),
--- a/templates/project/detail.html
+++ b/templates/project/detail.html
@@ -114,38 +114,40 @@
                     {{ workspace.name }}
                   </a>
                 </dd>
-                <dt class="sr-only">Status:</dt>
-                <dd class="ml-auto shrink-0">
-                  {% if workspace.is_archived %}
-                    <span class="relative group">
-                      {% pill variant="warning" text="Archived" %}
-                      {% tooltip position="-bottom-3" content="Creating new jobs is disabled" %}
-                    </span>
-                  {% else %}
-                    {% pill variant="primary" text="Active" %}
-                  {% endif %}
-                </dd>
-              </div>
-              <div class="flex flex-col gap-2 text-sm text-slate-600 sm:flex-row sm:gap-x-4">
-                <dt class="sr-only">GitHub repository:</dt>
-                <dd class="flex flex-row items-start overflow-hidden sm:max-w-[50%]">
-                  {% icon_github_outline class="mr-1.5 h-5 w-5 shrink-0 text-slate-400" %}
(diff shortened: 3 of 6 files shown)


templates/project/detail.html
```diff
@@ -114,38 +114,40 @@
...
+                    {% if workspace.is_archived %}
+                      <span class="relative group">
+                        {% pill variant="warning" text="Archived" %}
+                        {% tooltip position="-bottom-3" content="Creating new jobs is disabled" %}
+                      </span>
+                    {% else %}
+                      {% pill variant="primary" text="Active" %}
+                    {% endif %}
+                {% endif %}
+              </dd>
```
mikerkelly: I think this `</dd>` is closing a block opened inside the `if` tags so belongs before the `endif`?

jobserver/models/workspace.py
```diff
@@ -42,6 +42,9 @@ class Workspace(models.Model):
 
     name = models.TextField(unique=True)
     branch = models.TextField()
+
+    # We don't show this field publicly to anonymous users, or to users without roles,
+    # as it may cause confusion or be misleading to users without necessary context.
     is_archived = models.BooleanField(default=False)
```
mikerkelly, line 48: Thanks for thinking to add this!

jobserver/views/projects.py
```diff
@@ -41,6 +41,9 @@ def get(self, request, *args, **kwargs):
             request.user, Permission.PROJECT_MANAGE, project=project
         )
 
+        # An AnonymousUser doesn't have the has_any_roles property, so we use getattr
+        can_view_workspace_statuses = getattr(request.user, "has_any_roles", False)
```
mikerkelly: Suggestion, optional: I think it is slightly nicer to check that the user is authenticated before calling the method, rather than using `hasattr`. Django `User` has a built in property method for that. That may also make the intent clearer from the code and mean that the comment is not needed.

```python
 can_view_workspace_statuses = request.user.is_authenticated and request.user.has_any_roles
```

templates/project/detail.html
```diff
@@ -114,38 +114,40 @@
...
+            </div>
```
mikerkelly: I think all of the following lines have been indented by 2 characters. I am not sure they need to change at all. Is this intentional?

jobserver/views/projects.py
mikerkelly: Suggestion, optional: could mention here as well that users without roles probably don't have the context to usefully understand the workspace status. Or, `User.can_view_workspace_statuses` could be a model method to centralise the logic and explanation in one place rather than independently doing the same logic in 3 separate views.

mikerkelly (approved):
Thanks Eli, this works well. Just a couple of possible issues on one of the templates and optional suggestions on how/where to do the user checking logic. Happy to approve subject to any changes in response to these, happy to discuss anything if that helps.

I agree with the three templates changed. On the project detail page, the meaning of the pills is not so clear and not helpful to public viewers. On the workspace detail/log pages, the banners are written as instructions to the researcher, so best to omit for the public. And just "any user with roles" is fine for granularity, attempting to guess who else may benefit from the banner gets complicated. All other templates mentioning `is_archived` are staff pages or only usable by service users anyway:

```
~/job-server$ grep -ilIR is_archived templates/
templates/project/detail.html
templates/workspace/detail.html
templates/workspace/event_log.html


templates/yours/workspace_list.html
templates/repo/sign_off.html
templates/job_request/create.html
templates/staff/workspace/detail.html
templates/staff/workspace/edit.html
templates/staff/repo/detail.html
```

jobserver/views/projects.py
eli-miriam: I don't think a new model method gets us much clarity beyond your first suggestion, since in the view code we'd have to check authentication either way (as `AnonymousUser` doesn't get any of those methods by default and I saw a comment elsewhere indicating that it's non-trivial to add them). I'll definitely take your first suggestion though; that's much much nicer!

templates/project/detail.html
eli-miriam: definitely not intentional! Will fix, thankyou :)

templates/project/detail.html
eli-miriam: whoops thankyou! will fix

PR #5786, opened 2026-04-14
Only require `copilot` field for `POS-` projects.

### LINKED ISSUE #5779: Project creation: Only require copilot field for POS- projects
# Why are we doing this?

When creating a project, the form validates that a co-pilot is selected, and it is not possible to create a project without one.

This is fine for regular projects, but data development projects do not really have a co-pilot. *Indeed, the existing OpenSAFELY Internal project does not have a co-pilot.*

We can work around this, for example, by setting someone like Seb as the co-pilot (and later optionally removing them), but that requires extra steps and is not representative of the project's real state.

When we have made the co-pilot selection optional, possibly only for data development projects.

# What are we doing?

* Changing the project create form to make the co-pilot selection optional, only for data development projects, based on the project identifier being entered and starting with POS-.
* Only send a message to the copilots channel if the project identifier exists and starts with POS-

mikerkelly:
Fixes https://github.com/opensafely-core/job-server/issues/5779. In the ProjectCreateForm we want to not require copilot for non-POS projects such as data development projects that do not have copilots, but still require one for POS- projects that do.

In the form `clean` method, do this check, using the regex for POS- projects. If we implement the project category attribute, we could use that instead, maybe.

In `actions.projects.add`, only send the copilots a Slack message if a copilot was set, as there is nothing for them to do for other project types.

Adapt existing tests with additional parameters or checks to validate the new behaviour.

http://localhost:8000/staff/projects/create/

image

### DIFF
--- a/staff/forms.py
+++ b/staff/forms.py
@@ -7,7 +7,7 @@ from applications.models import Application, ResearcherRegistration
 from jobserver.authorization.forms import RolesForm
 from jobserver.backends import backends_to_choices
 from jobserver.models import Backend, Org, Project, SiteAlert, User, Workspace
-from jobserver.models.project import NUMBER_REGEX
+from jobserver.models.project import NUMBER_REGEX, POS_FORMAT_REGEX
 
 
 def user_label_from_instance(obj):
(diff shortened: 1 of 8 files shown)


staff/forms.py
```diff
@@ -143,12 +143,28 @@ def __init__(self, *args, **kwargs):
...
         _validate_slug(self.cleaned_data["name"])
         return self.cleaned_data["name"]
 
+    def clean(self):
+        # Only require copilot be set for projects with a POS-format project number.
+        # TODO: When we implement project category, the second part of the
+        # condition could use that.
+        if (
+            "copilot" not in self.cleaned_data or not self.cleaned_data["copilot"]
+        ) and POS_FORMAT_REGEX.fullmatch(self.cleaned_data["number"]):
```
StevenMaude: issue: I _think_ I found a bug here:

1. Create a `POS` project, like `POS-2026-2001`, which requires a co-pilot to be created.
1. Verify for yourself in the list of projects that the project is successfully created.
1. Go back to the "Create project" form and enter the same project ID again, but  *without* a co-pilot.
1. You should see a `KeyError` on `self.cleaned_data["number"]`.

This works fine *if* the co-pilot is also set the second time. The validation then correctly complains about the duplicate project ID.

(Please let me know if you can't reproduce it though, and I'll see if it's a quirk with my local setup.)

StevenMaude (approved):
Happy to approve as no issues with the code and how it works, but one bug to maybe check and fix if necessary.

staff/forms.py
mikerkelly: Thanks for catching this, Steve. Indeed if `clean_number` has an error, the key is not present so we need to be defensive to that possibility.

```python
number = self.cleaned_data.get("number")
copilot = self.cleaned_data.get("copilot")
if not copilot and number and POS_FORMAT_REGEX.fullmatch(number):
```
That is also a bit more readable.

Added `test_projectcreateform_duplicate_number` which seemed to be missing before. That failed before the above fix.

PR #5796, opened 2026-04-21
Mark more slow tests

mikerkelly:
TESTING.md says to mark slow tests that take more than about 0.1s to run. These
are excluded from `just test` to keep the feedback loop quick but are still run
in `just test-ci`.

There are a number of tests that consistently took longer than this. Most were
added in the last year and many use features like the test client, database
transactions, or are just long, so it makes sense that they are slow and not
obvious how to speed them up. Also, permutations of parametrised tests may add
up to a large amount if the individual tests are reasonably slow.

I used `hyperfine` for better profiling in main and this branch and got:

```
~/job-server$ hyperfine --warmup 1 'just test'
Benchmark 1: just test
  Time (mean ± σ):     27.267 s ±  3.372 s    [User: 95.866 s, System: 4.686 s]
  Range (min … max):   22.577 s … 32.413 s    10 runs
```

vs.

```
Benchmark 1: just test
  Time (mean ± σ):     19.746 s ±  2.771 s    [User: 81.794 s, System: 3.724 s]
  Range (min … max):   16.882 s … 23.891 s    10 runs

```

That shows about 8s improvement and lower variance.

Also added some explanation of work done to TESTING.md for future repeats.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -81,6 +81,10 @@ few times and see if there is a consistent improvement. Remember that the very
 first seems to get punished by including the worker preparation time in its
 setup and call time, so that may appear much worse than it actually is.
 
+To benchmark the suite as a whole (or any shell command) you can use a tool
+like `hyperfine`: `hyperfine --warmup 1 'just test' --export-markdown
+hyperfine.md`.
+
 #### Database access
 
 All unit and integration tests have access to the database by default. You can
--- a/TESTING.md
+++ b/TESTING.md
@@ -70,7 +70,16 @@ This allows testing functionality of the entire running system, as a user would 
 
 We have a custom `@pytest.mark.slow_test` marker that should be used on test
 that take more than about 0.1s to run. These are excluded from `just test` to
-provide a quicker feedback loop.
+provide a quicker feedback loop. They can still be run with `just test-ci`.
+
+To check this against specific tests or as a suite-wide check, run `just test
+-n0 --durations=20 --durations-min=0.1`. We use `-n0` so there is only one
+worker to reduce some sources of variance. Run several times and observe if any
+tests consistently take longer than this. You can specify a module to reduce
+the overall runtime.  Also, try running with and without suspects marked up a
+few times and see if there is a consistent improvement. Remember that the very
+first seems to get punished by including the worker preparation time in its
+setup and call time, so that may appear much worse than it actually is.
 
 #### Database access
 
--- a/tests/integration/staff/views/test_projects.py
+++ b/tests/integration/staff/views/test_projects.py
@@ -71,6 +71,7 @@ class TestProjectCreation:
         assert selected_orgs == bennett_org.pk
 
     @pytest.mark.django_db(transaction=True)
+    @pytest.mark.slow_test
     def test_projectcreate_post_success(
         self, client, request, slack_messages, user_with_fake_role_factory
     ):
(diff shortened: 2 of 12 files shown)


TESTING.md
```diff
@@ -70,7 +70,20 @@ This allows testing functionality of the entire running system, as a user would
 
 We have a custom `@pytest.mark.slow_test` marker that should be used on test
 that take more than about 0.1s to run. These are excluded from `just test` to
-provide a quicker feedback loop.
+provide a quicker feedback loop. They can still be run with `just test-ci`.
+
+To check this against specific tests or as a suite-wide check, run `just test
+-n0 --durations=20 --durations-min=0.1`. We use `-n0` so there is only one
```
StevenMaude, line 76: question: Is it worth having this command as a `justfile` recipe like `just test-find-slow`?

TESTING.md
StevenMaude, line 1: thought: With you having done this now across the code, I'm wondering how we can maintain it more as a process:

* Could we perhaps run this regularly (weekly?) in CI, maybe with a relaxation of the allowed duration, to alert us to particularly slow tests?
* Should we expect developers to be checking they're not introducing unmarked slow tests on changes they've made?

(I don't think this necessarily has to be answered now, or stops this PR though.)

TESTING.md
```diff
@@ -81,6 +81,10 @@ few times and see if there is a consistent improvement. Remember that the very
 first seems to get punished by including the worker preparation time in its
 setup and call time, so that may appear much worse than it actually is.
```
StevenMaude, line 82: suggestion:

Maybe worth clarifying the command line options?

Either by adding a link to the [relevant pytest documentation](https://docs.pytest.org/en/9.0.x/how-to/usage.html#profiling-test-execution-duration), or amending the text to explain:

```suggestion
To check this against specific tests or as a suite-wide check, run `just test
-n0 --durations=20 --durations-min=0.1`.

This finds the 20 slowest tests that take at least 0.1 seconds.
We use `-n0` so there is only one worker to reduce some sources of variance.

Run several times and observe if any tests consistently take longer than this. 
You can specify a module to reduce the overall runtime. Also, try running with
and without suspects marked up a few times and see if there is a consistent
improvement. Remember that the very first seems to get punished by including the
worker preparation time in its setup and call time, so that may appear much worse
than it actually is.
```

StevenMaude (approved):
> That shows about 8s improvement and lower variance.

Great! Thanks for this 😃

TESTING.md
mikerkelly: Good point. I think we do this very rarely and may want to tweak parameters so not sure it is worth it. It is not a regular workflow.

TESTING.md
mikerkelly: > Could we perhaps run this regularly (weekly?) in CI, maybe with a relaxation of the allowed duration, to alert us to particularly slow tests?

I think this is a nice idea but may be too expensive to develop when it is quite easy to do this manual process periodically instead.

> Should we expect developers to be checking they're not introducing unmarked slow tests on changes they've made?

I think that is what the existing TESTING.md document implies:

> We have a custom @pytest.mark.slow_test marker that should be used on test that take more than about 0.1s to run. These are excluded from just test to provide a quicker feedback loop.

I am not actually checking that manually unless I have a reason to worry about my tests, though. I think this is what most people are doing. And sometimes they miss that they have made a slow test.

Maybe one for discussion in the meeting about testing approach.

TESTING.md
mikerkelly: I marginally think it is clear enough as is. The option names are pretty intuitive. And breaking out to define interrupts the flow.

Issue #5782, opened 2026-04-13
Find out why we have one job request that's not being processed by RAP API


lucyb:
# Why are we doing this?

On the Job Server [status page](https://jobs.opensafely.org/status/) there's one job request that hasn't yet been processed by the RAP API. It's unclear how long it's been there for.

# How will we know when it's done?

...

# What are we doing?

* Investigating why the job request hasn't been processed
* Coming up with some suggestions for what to do next

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

evansd:
It's this one here:
https://jobs.opensafely.org/impact-and-inequalities-of-winter-pressures-in-primary-care-providing-the-evidence-base-for-mitigation-strategies/wp-work-package-1/25706/

I think it should probably have a `_status` of `unknown_error_creating_jobs` but I don't know why that didn't get assigned. Do shout if you think it could be a RAP Controller error.

eli-miriam:
We (Lucy & I) think the culprit is [here](https://github.com/opensafely-core/job-server/blob/main/jobserver/actions/rap.py#L109) - if the job request _did_ get the correct status assigned but then later was in `unrecognised_rap_ids`, it'd have its status switched over the `FAILED` (which is what we now see on the job request that Dave linked). That would cause it to show up in the [query on the status page](https://github.com/opensafely-core/job-server/blob/main/jobserver/views/status.py#L59), which is what triggered this investigation.

Next steps are:
- deciding if we want to change the status page, given that the RAP API makes some of the ways we describe the job queue less accurate
- assessing the status flow of job requests to decide whether the behaviour we've (probably) seen here is OK

mikerkelly:
https://github.com/opensafely-core/job-server/issues/5525 added annotation exclusions for special statuses:

> The [JobRequest.status](https://github.com/opensafely-core/job-server/blob/dd400b8ae5b65e7e8851f4b23cb8ce6ba8ca3057/jobserver/models/job_request.py#L66) field has a couple of values (NOTHING_TO_DO, UNKNOWN_ERROR_CREATING_JOBS) that can legitimately have no jobs associated with them, as the final state of the request, since the RAP API work removed the error signal jobs.

But I think this is not quite right and was probably the wrong fix (see below). My bad. So should we fix up the definition of unacked jobs and what should the definition be?

> deciding if we want to change the status page, given that the RAP API makes some of the ways we describe the job queue less accurate

[...]

PR #5798, opened 2026-04-21
Remove confusing and extraneous detail from the /status page

eli-miriam:
Per #5782 we've learned that [the status page](https://jobs.opensafely.org/status) currently shows counts of jobs that are either now meaningless due to RAP API changes (unacked jobs) or are simply not useful and constitute noise (total jobs ever run on a given backend). This PR removes that information from view, and from the page queries.

Here's the page as it currently stands:
image

Here's how it looks (running locally and therefore not in maintenance mode) after these changes:
image

### DIFF
--- a/jobserver/views/status.py
+++ b/jobserver/views/status.py
@@ -7,7 +7,7 @@ from django.utils import timezone
 from django.views.generic import View
 
 from ..backends import show_warning
-from ..models import Backend, Job, JobRequestStatus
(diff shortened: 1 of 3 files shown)


mikerkelly (approved):
Thanks Eli, this is a definite improvement on the clarity of the page. Thanks also for the incidental fixing up and clarifying in the tests.

PR #5811, opened 2026-04-29
Set uv exclude-newer

lucyb:
We're now setting the equivalent of exclude-newer in Dependabot's config, but Dependabot isn't updating correctly. Setting it in the pyproject.toml might help, shouldn't do any harm and will make it easier to do manual updates.

See [docs](https://docs.astral.sh/uv/guides/integration/dependabot/#dependency-cooldown).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -248,6 +248,7 @@ extend-ignore = [
 # See https://github.com/opensafely-core/repo-template/blob/main/DEVELOPERS.md for details.
 [tool.uv]
 required-version = ">=0.9"
+exclude-newer = "7 days"
 
 [dependency-groups]
 dev = [
--- a/uv.lock
+++ b/uv.lock
@@ -6,6 +6,10 @@ resolution-markers = [
     "python_full_version < '3.14'",
 ]
 
+[options]
+exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
+exclude-newer-span = "P7D"
+
 [[package]]
 name = "argon2-cffi"
 version = "25.1.0"


mikerkelly:
> We're now setting the equivalent of exclude-newer in Dependabot's config, but Dependabot isn't updating correctly.

See [Slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1777538493388099).

> there were no PRs for python updates for Job Server when I was working my through the dependarota yesterday. So, I had a look at the logs for the last dependabot run. It ran successfully, but didn't find any updates and hasn't been for several weeks at least.
> https://github.com/opensafely-core/job-server/actions/runs/24982770759/job/73148649093
> This line is suspicious, as there are currently packages eligible for upgrade
>> updater | 2026/04/27 07:55:39 INFO <job_1338367182> Found no dependencies to update after filtering allowed updates in /

pyproject.toml
```diff
@@ -248,6 +248,7 @@ extend-ignore = [
 # See https://github.com/opensafely-core/repo-template/blob/main/DEVELOPERS.md for details.
 [tool.uv]
 required-version = ">=0.9"
+exclude-newer = "7 days"
```
mikerkelly, line 251: Noted in passing: The comments on the previous lines come from the [repo-template](https://github.com/opensafely-core/repo-template/blob/main/pyproject.toml). I am not sure that they are that relevant any more if we are not using absolute timestamps either in dependabot configuration or here. So should we remove those comments?

I also find them a bit hard to interpret and the linked docs did not really help. I was not sure if they were talking about due to automated updates, manual, or both.

mikerkelly, line 251: Documentation suggestion: I don't think it is obvious that we intend that this is unrelated to the automatic update mechanism. So it might be worth briefly commenting on that and why this is here.

mikerkelly (approved):
This does not seem necessary for Dependabot which has its own cooldown configuration, but is consistent with the Astral documentation so should not hurt technically. I am pessimistic that adding additional description of the same filtering will make Dependabot filter less than it currently is, but we may as well try as we don't understand the problem Dependabot is having. Happy to approve.

pyproject.toml
lucyb: Good point. I'll leave that until after we know it doesn't affect how Dependabot functions though.

lucyb: It was a bit easier to reason about when we had `exclude-newer-package = {}` listed in this section. I can imagine that we will still want to use absolute timestamps for some packages. Although I agree about the comment not helping and will update it later.

PR #5819, opened 2026-04-30
Remove accidental `Project` `name='kasdlkfj'`

mikerkelly:
A Project was created in production by accident. Let's remove it.
See https://bennettoxford.slack.com/archives/C069SADHP1Q/p1777546132838079

We should also remove associated ProjectCollaboration, ProjectMembership, and
AuditableEvent instances.

Let's do this in a migration in a new maintenance app, so that the fix is
auditable, reviewable, repeatable, testable, avoids the possibility of manual
error, and fixes up development environment databases as well. This also
provides an example if we need to do this or other maintenance again. And it
avoids cluttering the feature apps and getting squashed.

See commit message for how this was testing locally.

### DIFF
--- a/jobserver/settings.py
+++ b/jobserver/settings.py
@@ -85,6 +85,7 @@ INSTALLED_APPS = [
     "airlock",
     "applications",
     "jobserver",
+    "maintenance",
     "redirects",
     "staff",
     "anymail",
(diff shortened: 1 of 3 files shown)


lucyb (approved):
Thanks for the helpful code comments. I like that we're following the same pattern as opencodelists with the maintenance app. We should probably copy the README you've added over to ocl as well.

PR #5823, opened 2026-05-06
Add AIRLOCK_READONLY_ACCESS permission for Level4AuthenticatedUser

### LINKED ISSUE #1176: Add a class of user who can view all workspaces/release requests but not approve requests
At present only output checkers have permission to view all workspaces, but there's a need for some developers who are not output checkers to be able to view all workspaces and release requests in order to help with debugging.

There's an open question as to which, if any, of our existing roles this maps to. The "Tech Support" role is the obvious candidate here, though only a subset of those on Tech Support will have access to Airlock.

Once the role is decided we'll need to thread it through from the `authorise` API in job-server:
https://github.com/opensafely-core/job-server/blob/ff9b9bede1080199e025a5f9aff863ee0d72695e/jobserver/api/releases.py#L577-L589

And then represent it in the User model in Airlock:
https://github.com/opensafely-core/airlock/blob/3d9f617ce7b4891831c59666107b3971e2410aed/users/models.py#L49-L55

And then extend the logic in the permissions check:
https://github.com/opensafely-core/airlock/blob/3d9f617ce7b4891831c59666107b3971e2410aed/airlock/permissions.py#L44-L57

Related Slack thread:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1777644928543169?thread_ts=1777635193.501999&cid=C069YDR4NCA

rebkwok:
This permission is passed to Airlock during authentication, and will allow certain users to view all workspaces (with readonly access) even if they do not have a role on the project.

The permission is currently attached to the DeploymentAdministrator role and will allow developers with L4 access to view files and logs for any workspace, which is useful when responding to tech support questions.

Airlock has already [been updated](https://github.com/opensafely-core/airlock/pull/1181) to accept and handle the new `readonly_access` field.

[Slack discussion](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1778061872689109)
Closes https://github.com/opensafely-core/airlock/issues/1176

### DIFF
--- a/jobserver/authorization/permissions.py
+++ b/jobserver/authorization/permissions.py
@@ -11,6 +11,8 @@ class Permission(StrEnum):
         member names via `auto()`.
     """
 
+    AIRLOCK_READONLY_ACCESS = auto()
+
     APPLICATION_MANAGE = auto()
 
     BACKEND_MANAGE = auto()
--- a/jobserver/authorization/roles.py
+++ b/jobserver/authorization/roles.py
@@ -133,7 +133,9 @@ class SignOffRepoWithOutputs:
 
 class DeploymentAdministrator:
     display_name = "Deployment Administrator"
-    description = """Run and cancel jobs on any project, for development and maintenance purposes, including technical support for Approved Projects.
+    description = """
+    Run and cancel jobs on any project, for development and maintenance purposes, including technical support for Approved Projects.
+    View files, logs and release requests for any Airlock workspace, for those with Level 4 access.
     See Developer Permissions Log for the list of individuals who are approved for this role.
     """
     models = [
--- a/jobserver/authorization/roles.py
+++ b/jobserver/authorization/roles.py
@@ -142,4 +142,5 @@ class DeploymentAdministrator:
     permissions = [
         Permission.JOB_CANCEL,
         Permission.JOB_RUN,
+        Permission.AIRLOCK_READONLY_ACCESS,
     ]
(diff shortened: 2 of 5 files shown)


evansd (approved):

jobserver/authorization/roles.py
KatieB5: Suggestion: update the `DeploymentAdministrator` description to reflect this newly-added permission, so the role description stays aligned with its current capabilities.

For example, something along the lines of:
"Run and cancel jobs on any project, for development and maintenance purposes, including technical support for Approved Projects. View files and logs for any Airlock workspace, for those with Level 4 access."

I appreciate this permission may move to a different role later based on the Slack discussion, but I think it's still worth keeping the description accurate for the current behaviour.

jobserver/authorization/permissions.py
```diff
@@ -11,6 +11,8 @@ class Permission(StrEnum):
         member names via `auto()`.
     """
 
+    AIRLOCK_READONLY_ACCESS = auto()
```
KatieB5, line 14: Suggestion: `AIRLOCK_WORKSPACE_READONLY` might be a clearer name here, as it scopes the permission more specifically to workspace access within Airlock? 

I also wondered whether we need 'ACCESS' in the permission name, since Permissions represent access/capabilities already?

I don't have strong feelings, so if you'd prefer to leave things how they are that's totally fine.

KatieB5 (approved):

jobserver/authorization/roles.py
rebkwok: Ah, yes, I did wonder if this description should be updated (I asked in the slack thread but didn't get a response to that particular question). Will update it.

jobserver/authorization/permissions.py
rebkwok: Workspace access also allows read-only access to all release requests, so I think the current name is accurate

rebkwok: I can update the comment in `build_level4_user` to state that, though

PR #5831, opened 2026-05-11
Fix bug in remove files from publish requests command

### LINKED ISSUE #5810: Fix `MultipleObjectsReturned` error when removing files from a draft snapshot
I attempted to a file from [this draft snapshot](https://jobs.opensafely.org/opensafely-internal/tpp-database-schema/outputs/139/) but got the below error:
```
jobserver.models.release_file.ReleaseFile.MultipleObjectsReturned: get() returned more than one ReleaseFile -- it returned 2!
```

**Sequence of events that led to the error:** 
1. Ran command to create the snapshot: 

Image

2. Snap shot created: 

Image

3. Attempt to remove `output/files_for_release/vaccination_reference.csv` from the draft (following the instructions in the [playbook](https://bennett.wiki/tech-group/tech-support/playbook/#removing-files-from-a-draft-publication)) using the same file path used in creating the draft which resulted in the `MultipleObjectsReturned` error:

Image

This process worked when [attempted locally](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1775837450197839?thread_ts=1775831638.957339&cid=C069SADHP1Q) 

4. Second attempt to remove the file using the file path in the snapshot (which is only the file name):

Image

nishtha-kalra:
## Description
Fixes #5810 
The remove_files_from_publish_request command looked up files by workspace and path before removing them from a snapshot. ReleaseFile names are not unique within a workspace, so older releases or other files with the same path could make the lookup return multiple rows and raised the error `ReleaseFile.MultipleObjectsReturned`.

Look up files through `snapshot.files` instead, so the command only matches files that are actually part of the snapshot being edited.

Update the management command tests to cover the new lookup behavior and included test for case where another ReleaseFile in the same workspace has the same name but it is not attached to the snapshot.

## Testing
Tested it locally using the following command and it worked:
`python manage.py remove_files_from_publish_request <snapshot_id> <file>`

### DIFF
--- a/jobserver/management/commands/remove_files_from_publish_request.py
+++ b/jobserver/management/commands/remove_files_from_publish_request.py
@@ -32,9 +32,7 @@ class Command(BaseCommand):
         release_files = []
         for path in paths:
             try:
-                release_file = ReleaseFile.objects.filter(
-                    workspace=snapshot.workspace
-                ).get(name=path)
+                release_file = snapshot.files.get(name=path)
                 release_files.append(release_file)
             except ReleaseFile.DoesNotExist:
                 raise CommandError(f"Unknown file path: {path}")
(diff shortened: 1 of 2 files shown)


tomodwyer (approved):
Tested locally and it all works as expected.

As a side note you can use `just manage` instead of `python manage.py` in your commands when running locally 👌

PR #5832, opened 2026-05-12
Update GitHub Actions with security and workflow improvements

tomodwyer:
The majority of these changes have been spotted by [Zizmor](https://github.com/zizmorcore/zizmor).

- Pin GitHub Actions versions to specific hashes
    - [`unpinned-uses`](https://docs.zizmor.sh/audits/#unpinned-uses)
- Disable persist credentials in checkout action
    - [`artipacked`](https://docs.zizmor.sh/audits/#artipacked)
- Add `name` and `needs` to each step
    - [`anonymous-definition`](https://docs.zizmor.sh/audits/#anonymous-definition)
- Pin Postgres image using sha256
    - [`unpinned-images`](https://docs.zizmor.sh/audits/#unpinned-images)
    - And update the associated test
- Cancel running jobs when workflow is re-triggered
    - [`concurrency-limits`](https://docs.zizmor.sh/audits/#concurrency-limits)
- Update Ubuntu to version 24.04
    - As that's the version we're using on the server, and is the LTS
- Fix template injection for GitHub Actor
    - [`template-injection`](https://docs.zizmor.sh/audits/#template-injection)
- Fix excessive permissions
    - [`excessive-permissions`](https://docs.zizmor.sh/audits/#excessive-permissions)

## Before

Screenshot of the GitHub Actions workflow before

## After

Screenshot of the GitHub Actions workflow after

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -2,7 +2,11 @@
 # configuring the production build
 services:
   db:
-    image: "postgres@sha256:b994732fcf33f73776c65d3a5bf1f80c00120ba5007e8ab90307b1a743c1fc16"
+    # When this updates, you need to also update:
+    # - .github/workflows/main.yml:47
+    # - tests/test_repository.py:30
+    # to match this SHA
+    image: postgres@sha256:0027bef26712baaee437a4ea48fdf3d2d2e2bc5f0d81615374408ca320f3c7e3 # v17
     environment:
       POSTGRES_USER: user
       POSTGRES_PASSWORD: pass
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -2,7 +2,7 @@
 # configuring the production build
 services:
   db:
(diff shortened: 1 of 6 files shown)


docker/docker-compose.yaml
```diff
@@ -2,7 +2,7 @@
 # configuring the production build
 services:
   db:
-    image: "postgres:17"
+    image: "postgres@sha256:b994732fcf33f73776c65d3a5bf1f80c00120ba5007e8ab90307b1a743c1fc16"
```
StevenMaude: suggestion: Could we get Dependabot to update this with a cooldown (and maybe that in `main.yml`, and if not, a reminder to update the version in CI/`test_repository.py` too?)

StevenMaude (approved):

docker/docker-compose.yaml
tomodwyer: I've done both, if you want to take another quick look before I merge.

PR #5834, opened 2026-05-13
POC for scrubbing data via the ORM.

mikerkelly:
We want to scrub some Django-managed model fields because they have personal data or secrets/tokens et cetera.

We can define in a model which fields require scrubbing and what they should be replaced with. Then a management command can do a bulk update on these fields via the ORM. Django can connect to a non-primary database in the cluster that has been loaded with a dump of the production database to run this command to scrub the data. Then it can be dumped out again without the sensitive data.

This POC just scrubs one table, User.

There are other tables not managed through Django models that contain data we might want to truncate as well, that is best done through raw SQL statements. That can also be done in a Django management command but is not shown here. Such tables include: `django_admin_log`, `django_session`, `auth_*`, and `social_auth_*`.

We could add some test that all models have both "scrub_fields" and "scrub_allowed" and that all fields on the model appear in exactly one of these.

Example invocation:

```
~/job-server$ just manage scrub_data
$BIN/python manage.py scrub_data
Scrubbed 408 User records from fields:dict_keys(['email', 'password', 'login_token', 'login_token_expires_at', 'pat_expires_at', 'pat_token'])
```

Then in a database shell:

```
jobserver=# select username, password, email, login_token from jobserver_user where username='mikerkelly';
  username  | password |         email         | login_token
------------+----------+-----------------------+-------------
 mikerkelly |          | whiteadam@example.net |
(1 row)
```

I can still use the site normally after logging back in to social auth.

image

### DIFF
--- a/jobserver/models/user.py
+++ b/jobserver/models/user.py
@@ -13,6 +13,7 @@ from django.db.models.functions import Lower
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.functional import cached_property
+from faker import Faker
 from sentry_sdk import capture_message
 
 from jobserver.authorization.permissions import Permission
@@ -23,6 +24,8 @@ from ..hash_utils import hash_user_pat
 
 
 logger = structlog.get_logger(__name__)
+
+fake = Faker()
 
 
 class UserQuerySet(models.QuerySet):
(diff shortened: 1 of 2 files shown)


jobserver/models/user.py
```diff
@@ -143,6 +146,26 @@ class User(AbstractBaseUser):
...
 
+    scrub_fields = {
+        "email": fake.email,
+        "password": "",
+        "login_token": None,
+        "login_token_expires_at": None,
+        "pat_expires_at": None,
+        "pat_token": None,
+    }
+    scrub_allowed = frozenset(
```
nishtha-kalra, line 157: Could this be `allowed_fields` instead of `scrub_allowed`?

PR #5845, opened 2026-05-19
Simplify `previous(filter_succeeded=True)`.

mikerkelly:
This branch of the `previous` function wants to filter a QuerySet of Job Requests to only include those where all related Jobs have status "succeeded". The previous implementation examines the max and min status and needs a comment the explain why that works. The new implementation simply counts the total and succeeded jobs and asserts equality. This is an easier-to-understand way to achieve the same goal.

This has no measureable impact on performance as both checks require iterating through all related jobs. Checked by viewing in Django debug toolbar a few times before and after.

Preexisting test `test_jobrequest_previous_succeeded` checks that *all* jobs must be succeeded, not at least one, with 3 test job requests, which seems sufficient.

### DIFF
--- a/jobserver/models/job_request.py
+++ b/jobserver/models/job_request.py
@@ -5,7 +5,7 @@ import sentry_sdk
 import structlog
 from django.contrib.postgres.fields import ArrayField
 from django.db import models
-from django.db.models import Max, Min, Q, prefetch_related_objects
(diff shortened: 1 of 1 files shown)


StevenMaude (approved):
Thanks; this is a clearer rewrite.

PR #5846, opened 2026-05-19
t1oo permission file changes in preparation for updates to permissions handling

rebkwok:
Note: this PR involves changes to the `t1oo.py` permissions file, which required IG team approval. It does not change any of the permitted projects, and it does not change the way permissions are currently handled.

The t1oo.py file predates the categorisation of permissions into population permissions, dataset permissions and (current unused) component access permissions which are passed as `analysis_scope` to the RAP API. In prepration for updating the handling of t1oo permissions to bring them in line with the way other permissions are handled, we move the file into the population_permission subdir, along with other population permissions files (for ndoo and gp-activations).

Note that no changes have been made to the project listed in the `t1oo.py` file; it has just been moved (in the first commit of this PR) and code changes have been made that will be required for the updated permissions handling (adding an ANALYSIS_SCOPE variable, renaming a function, and adding the t1oo permission to the `analysis_scope` object sent in the job request create request to the RAP API).

Documentation updates (with link to the moved t1oo.py file) in [draft PR](https://github.com/opensafely/documentation/pull/1980), to be merged after this one. 

Part 1 of https://github.com/opensafely-core/ehrql/issues/2801


jobserver/permissions/population_permissions/t1oo.py
```diff
@@ -8,8 +8,9 @@
 #
 # Note also that this file is linked to in the documentation. If you move or restructure
 # this file you should ensure the documentation is updated appropriately.
-# https://github.com/opensafely/documentation/blob/af88baf1/docs/type-one-opt-outs.md
+# https://github.com/opensafely/documentation/blob/1394ac771298611d40483e36a76b11ebb507ae7a/docs/type-one-opt-outs.md
```
rebkwok, line 11: Note: we're referencing the specific documentation commit here; usually this is fine, but in this case I also need to update the docs because the file location has moved. This commit ref should be fine as long as I don't have to rebase the draft PR; or I could change it to reference `main` which is obviously a changing ref, but would avoid this issue in future

jobserver/models/job_request.py
```diff
@@ -486,7 +484,7 @@ def update_status(
 
     @property
     def database_name(self):
-        if project_is_permitted_to_use_t1oo_data(self.workspace.project):
+        if project_has_permission(self.workspace.project):
```
evansd: If I was being super-persnickety I would have recommended importing the `t1oo` module and then this line would be `if t1oo.project_has_permission`, which would be slightly clearer.

But given that this is all going away in an upcoming PR I don't think it makes any difference.

jobserver/rap_api.py
```diff
@@ -210,6 +210,7 @@ def create(job_request):
     """
     analysis_scope = {}
     for population_module in [
+        population_permissions.t1oo,
```
evansd, line 213: I don't think this will work unless you add the `t1oo` module to `population_permissions/__init__.py`. But I think the monkeypatching in the tests is masking this fact.

evansd (approved):
I don't think this works as is, but have approved on the basis that it's an easy fix that doesn't need a re-review.

jobserver/rap_api.py
rebkwok: Hmm. You're right - I'll fix that and see if I can add a test that ensures it doesn't get missed in future

rebkwok: Actually, I think this does work. Without adding the t1oo to the __inti__.py, in a job-server django shell:

```
>>> from jobserver.permissions import population_permissions
>>> project = Project.objects.get(number="155")
>>> for population_module in [population_permissions.t1oo, population_permissions.ndoo]:
...     population_module.analysis_scope_for_project(project)
... 
'include_t1oo'
```

But, I think it's worth adding the test so we check including all the modules in the __init__ as we should anyway

jobserver/models/job_request.py
rebkwok: fixed it since I was making changes anyway

evansd (approved):

JulietUnderdown1 (approved):
Discussed with Becky. T1OO file is approved. No change to the projects listed.

PR #5858, opened 2026-05-21
Pin pre-commit hooks to commits

StevenMaude:
Not tags.


StevenMaude:
There's possibly an argument for adding pre-commit to Dependabot, so the versions don't go stale, but I've left that for now. We don't do that in the repo-template right now.

tomodwyer (approved):

PR #5868, opened 2026-05-28
Fix incorrect status showing on Job Request create template

tomodwyer:
Running and pending states were not accurately reflected on the page.

This PR also simplifies the code for updating the pill class and text content.

### DIFF
--- a/assets/src/scripts/job_request_create.js
+++ b/assets/src/scripts/job_request_create.js
@@ -28,34 +28,47 @@ async function getStatuses() {
 }
 
 /**
+ * @typedef {"succeeded"|"failed"|"none"|"running"|"pending"|"loading"} StatusState
+ */
+
+/** @type {Record<StatusState, {className: string, label: string}>} */
+const STATUS_STATES = {
+  succeeded: {
+    className: "pill--success",
+    label: "Succeeded",
+  },
+  failed: {
+    className: "pill--failed",
+    label: "Failed",
+  },
+  none: {
+    className: "pill--info",
+    label: "No status",
+  },
+  loading: {
+    className: "pill--info",
+    label: "Loading",
+  },
+};
+
+/**
  * Set the visible icon based on the status returned from the API
  * @param {Element} action - label element for an action
- * @param {("succeeded"|"failed"|"loading")} status - status returned from the API
+ * @param {StatusState|string} status - status returned from the API
  * @returns {void}
  */
 function setIcon(action, status) {
   const parentEl = action.closest(`[data-action]`);
   const pill = parentEl.querySelector("[data-action-status]");
+  const state = STATUS_STATES[status] || STATUS_STATES.loading;
 
-  pill.classList.remove("pill--success", "pill--failed", "pill--info");
-
-  if (status === "succeeded") {
-    pill.classList.add("pill--success");
-    return (pill.textContent = "Succeeded");
-  }
-
-  if (status === "failed") {
-    pill.classList.add("pill--failed");
-    return (pill.textContent = "Failed");
-  }
-
-  if (status === "none") {
-    pill.classList.add("pill--info");
-    return (pill.textContent = "No status");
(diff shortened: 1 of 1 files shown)


assets/src/scripts/job_request_create.js
```diff
@@ -27,35 +27,56 @@ async function getStatuses() {
   return statuses;
 }
 
+/**
+ * @typedef {"succeeded"|"failed"|"none"|"running"|"pending"|"loading"} StatusState
```
KatieB5: Do we know that these are the only statuses that an action can have? And if so, how do we know that? I tried looking in our codebase, and the job-runner codebase, but I couldn't find this distinct set of statuses for jobs.

assets/src/scripts/job_request_create.js
```diff
@@ -27,35 +27,56 @@ async function getStatuses() {
...
  * Set the visible icon based on the status returned from the API
  * @param {Element} action - label element for an action
- * @param {("succeeded"|"failed"|"loading")} status - status returned from the API
+ * @param {StatusState|string} status - status returned from the API
  * @returns {void}
  */
 function setIcon(action, status) {
   const parentEl = action.closest(`[data-action]`);
   const pill = parentEl.querySelector("[data-action-status]");
+  const state = STATUS_STATES[status] || STATUS_STATES.loading;
```
KatieB5: I think this means that if we get a status that's not in the list, we fallback to using "Loading". I

This might be OK. But I'm wondering if this is helpful to the user or to us i.e. is there potential for the action to not be in a loading state, but for the UI to tell us that it is?

assets/src/scripts/job_request_create.js
```diff
@@ -27,35 +27,56 @@ async function getStatuses() {
...
+};
+
 /**
  * Set the visible icon based on the status returned from the API
  * @param {Element} action - label element for an action
- * @param {("succeeded"|"failed"|"loading")} status - status returned from the API
+ * @param {StatusState|string} status - status returned from the API
  * @returns {void}
  */
 function setIcon(action, status) {
```
KatieB5: Suggestion: as we're no longer using icons to represent the status, we could rename this e.g. from `setIcon` to `setStatusPill`, or somethng nicer?

assets/src/scripts/job_request_create.js
tomodwyer: That's a good point - that was the previous fallback, but we should go to None I think. I'll make that change.

assets/src/scripts/job_request_create.js
tomodwyer: https://github.com/opensafely-core/job-runner/blob/7371c93046ffa2129dbaa08fafb6c527dd60640a/controller/models.py#L42

Job Runner contains this `StatusCode` class where the values map to PENDING, RUNNING, SUCCEEDED, and FAILED - then we add "loading" and "none" as the UI needs to adapt to those two states.

I will add a comment, and sort the order to match that class.

assets/src/scripts/job_request_create.js
tomodwyer: Good spot, yes we can clarify the language now it's not an icon.

assets/src/scripts/job_request_create.js
```diff
@@ -28,46 +28,69 @@ async function getStatuses() {
 }
 
 /**
- * Set the visible icon based on the status returned from the API
+ * @typedef {"pending"|"running"|"succeeded"|"failed"|"loading|"none""} StatusState
```
KatieB5:
```suggestion
 * @typedef {"pending"|"running"|"succeeded"|"failed"|"loading|"none"} StatusState
```

KatieB5 (approved):
Looks good to me. Just one small suggestion. The comments about where we get statuses are helpful, and there's some nice refactoring in setActionStatus - thank you!

PR #5869, opened 2026-05-28
Scrub sensitive fields in a management command

mikerkelly:

There are other tables not managed through our Django models that contain data we might want to scrub as well. Method to be determined, maybe raw SQL `TRUNCATE`.  Such tables include: `django_admin_log`, `django_session`, `auth_*`, and `social_auth_*`.

To make this useful in our production data dump process, we can extend this to connect to a non-primary database in the cluster that has been loaded with a dump of the production database to run this command to scrub the data. Then it can be dumped out again without the sensitive data.

Example invocation:

```
Scrubbed 926 ResearcherRegistration records from fields:dict_keys(['name', 'email', 'job_title', 'telephone', 'phone_type', 'github_username', 'daa', 'training_with_org', 'training_passed_at'])
Scrubbed 415 User records from fields:dict_keys(['email', 'password', 'login_token', 'login_token_expires_at', 'pat_expires_at', 'pat_token'])
Scrubbed 366 StudyPurposePage records from fields:dict_keys(['author_name', 'author_email'])
Scrubbed 415 ContactDetailsPage records from fields:dict_keys(['full_name', 'email', 'telephone', 'job_title', 'team_name', 'organisation'])
Scrubbed 130 SponsorDetailsPage records from fields:dict_keys(['sponsor_name', 'sponsor_email', 'sponsor_job_role', 'institutional_rec_reference'])
Scrubbed 10 Backend records from fields:dict_keys(['auth_token'])
```

Then in a database shell:

```
jobserver=# select username, password, email, login_token from jobserver_user where username='mikerkelly';
  username  | password |         email         | login_token
------------+----------+-----------------------+-------------
 mikerkelly |          | whiteadam@example.net |
(1 row)
```

While developing I tested manually by examining each table before and after adding its configuration and running the command. Not including the "before" here or in commit messages as they contain personal data.

I can still use the site normally after logging back in to social auth. Applications and Backend info in Staff pages look okay.

For `applications`, these are the fields that we identified in the previous iteration of data scrubing for `applications`. From inspecting the database I think there are
many more freetext fields that may contain personal data so we should consider
each carefully in a future PR before this is ready for production. Added a note to https://github.com/opensafely-core/security/issues/55.

### DIFF
--- a/applications/models.py
+++ b/applications/models.py
@@ -4,6 +4,7 @@ from django.db.models import Q
 from django.urls import reverse
 from django.utils import timezone
 
+from data_scrubbing.management.commands.scrub_data import fake
 from jobserver import hash_utils
 
 
@@ -219,6 +220,16 @@ class ContactDetailsPage(AbstractPage):
     # TODO: how will we tie this to an existing org, especially with typos?
     organisation = models.TextField(blank=True)
 
+    class DataScrubbing:
+        fields_to_scrub = {
+            "full_name": fake.name,
+            "email": fake.email,
+            "telephone": "",
+            "job_title": "",
+            "team_name": "",
+            "organisation": "",
+        }
+
 
 class CommercialInvolvementPage(AbstractPage):
     details = models.TextField(blank=True)
@@ -243,6 +254,12 @@ class StudyPurposePage(AbstractPage):
     is_covid_vaccine_eligibility_or_coverage = models.BooleanField(default=False)
     is_covid_vaccine_effectiveness_or_safety = models.BooleanField(default=False)
     is_other_impacts_of_covid = models.BooleanField(default=False)
+
+    class DataScrubbing:
+        fields_to_scrub = {
+            "author_name": fake.name,
+            "author_email": fake.email,
+        }
 
 
 class StudyDataPage(AbstractPage):
@@ -281,6 +298,14 @@ class SponsorDetailsPage(AbstractPage):
     sponsor_job_role = models.TextField(blank=True)
     institutional_rec_reference = models.TextField(blank=True)
     is_member_of_bennett_or_lshtm = models.BooleanField(default=False)
+
+    class DataScrubbing:
+        fields_to_scrub = {
+            "sponsor_name": fake.name,
+            "sponsor_email": fake.email,
+            "sponsor_job_role": "",
+            "institutional_rec_reference": "",
+        }
 
 
 class ShortDataReportPage(AbstractPage):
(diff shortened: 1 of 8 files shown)


applications/models.py
```diff
@@ -371,6 +396,19 @@ class PhoneTypes(models.TextChoices):
 
     created_at = models.DateTimeField(default=timezone.now)
 
+    class DataScrubbing:
+        fields_to_scrub = {
+            "name": fake.name,
+            "email": fake.email,
+            "job_title": "",
+            "telephone": "",
+            "phone_type": "iphone",
```
nishtha-kalra, line 405: reivew if iphone can be replaced.

nishtha-kalra: Ah, its an enum!

jobserver/models/user.py
```diff
@@ -143,6 +144,16 @@ class User(AbstractBaseUser):
     EMAIL_FIELD = "email"
     USERNAME_FIELD = "email"
 
+    class DataScrubbing:
+        fields_to_scrub = {
+            "email": fake.email,
```
nishtha-kalra, line 149: `User.email` is unique and faker does not guarantee uniqueness. We can either use `fake.unique` or a deterministic per-row value based on `pk`. I didn't face this issue though when testing so will create a follow up PR to solve this.

data_scrubbing/management/commands/scrub_data.py
```diff
@@ -0,0 +1,32 @@
...
+            objs = list(model.objects.all())
+            for obj in objs:
+                for attr_name, fake_value in scrub_fields.items():
+                    value = fake_value() if callable(fake_value) else fake_value
+                    setattr(obj, attr_name, value)
+
+            field_names = scrub_fields.keys()
+            model.objects.bulk_update(objs, field_names)
+            self.stdout.write(
+                f"Scrubbed {len(objs)} {model.__name__} records from fields:{field_names}"
```
nishtha-kalra, line 31: Suggested for follow up PR:
`f"Scrubbed {len(objs)} {model.__name__} records from fields: {', '.join(field_names)}"`

data_scrubbing/management/commands/scrub_data.py
```diff
@@ -0,0 +1,32 @@
+from django.apps import apps
+from django.core.management.base import BaseCommand
+from faker import Faker
+
+
+fake = Faker()
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs):
```
nishtha-kalra, line 10: Follow-up PR: add `help`

nishtha-kalra (approved):
Thanks Mike! The PR description and commit messages were very helpful.
There are some minor suggestions I plan to address in a follow-up PR, as they are not important. 
Also, since those changes will be minor, it will be easier for others to review them quickly.

PR #5878, opened 2026-06-03
Basic job that acts as a skeleton for dump scrubbed db job

nishtha-kalra:
Addresses https://github.com/opensafely-core/security/issues/52

- Instead of a daily job, created a yearly job so that it is easier to test it on dokku and it won't be triggered daily. After the job is created completely with sanitisation logic, we can shift it to a daily job.
- Created env variables for readonly jobserver database and jobserver_data_scrubbing database.
- Set the yearly job in app.json.

#### Note:
Need to create variables `JOBSERVER_READONLY_DATABASE_URL` and `JOBSERVER_SCRUBBED_DATABASE_URL` in dokku before merging this PR. I haven't created them now because I wanted a review of the names I have used.

#### Testing:
To test this locally:
1. Recreate `.env` from the updated `dotenv-sample` 
2. run the job `just manage runjob dump_scrubbed_db`

### DIFF
--- a/app.json
+++ b/app.json
@@ -19,6 +19,10 @@
     {
       "command": "python manage.py runjobs monthly",
       "schedule": "@monthly"
+    },
+    {
+      "command": "python manage.py runjobs yearly",
+      "schedule": "0 15 10 6 *"
     }
   ],
   "healthchecks": {
--- a/app.json
+++ b/app.json
@@ -22,7 +22,7 @@
     },
     {
       "command": "python manage.py runjobs yearly",
-      "schedule": "0 15 10 6 *"
+      "schedule": "0 17 10 6 *"
     }
   ],
   "healthchecks": {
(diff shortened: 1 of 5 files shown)


app.json
```diff
@@ -19,6 +19,10 @@
     {
       "command": "python manage.py runjobs monthly",
       "schedule": "@monthly"
+    },
+    {
+      "command": "python manage.py runjobs yearly",
+      "schedule": "0 14 4 6 *"
```
nishtha-kalra: Need to update it, depending on when this change is merged.

KatieB5 (approved):
Thank you Nishtha! This looks good to me - I've tested it locally as described in the commit message and can confirm I get the output logs ([outputs shown here in Slack message](https://bennettoxford.slack.com/archives/C080XJJ0KDJ/p1781090447459179?thread_ts=1780560863.729909&cid=C080XJJ0KDJ)).

PR #5880, opened 2026-06-04
Add just command and documentation for scrubbing local Job Server data

nishtha-kalra:
https://github.com/opensafely-core/security/issues/55

Related to the 3rd point in the ticket:

> Ask/automatically developers to run the latest version of that management command whenever they download the production dump. Maybe we can modify the just command so that it does this? Ask developers to delete after loading locally and delete that and any old copies. (Later scrubbing won't need to happen locally as it will be applied at source, but we can reduce the data lying outside production earlier this way, and we may still want to auto-delete the downloaded dump.)

Adding just command and doc before asking developers to use this command

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -177,9 +177,10 @@ After restoring a production database dump locally, run the data scrubbing comma
 just scrub-data
 ```
 
-This command removes or replaces sensitive data.
+This command removes or replaces sensitive data in your local database.
 
-Once the dump has been restored and scrubbed, the original downloaded dump file should be deleted, if present. This helps minimise the amount of raw production data stored outside production systems.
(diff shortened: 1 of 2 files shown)


DEVELOPERS.md
```diff
@@ -168,6 +169,18 @@ The important line to check for (typically at the very end) is `errors ignored o
...
+
+After restoring a production database dump locally, run the data scrubbing command before using the database for development or debugging:
+
+```bash
+just scrub-data
+```
+
+This command removes or replaces sensitive data.
+
+Once the dump has been restored and scrubbed, the original downloaded dump file should be deleted, if present. This helps minimise the amount of raw production data stored outside production systems.
```
tomodwyer: Seeing as the command deletes the dump, do we want to tell devs that's what it is doing?

nishtha-kalra: makes sense. will update it

tomodwyer (approved):

PR #5883, opened 2026-06-04
Add error handling for `scrub_data`

nishtha-kalra:
Convert database `OperationalError` failures into a `CommandError` that explains PostgreSQL must be running before using the command, and points Docker users to `just docker/db`.

Users were facing this issue as reported in [slack](https://bennettoxford.slack.com/archives/C63UXGB8E/p1780572225125979) message.

### DIFF
--- a/data_scrubbing/management/commands/scrub_data.py
+++ b/data_scrubbing/management/commands/scrub_data.py
@@ -1,5 +1,6 @@
 from django.apps import apps
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandError
+from django.db.utils import OperationalError
 from faker import Faker
 
 
@@ -12,6 +13,16 @@ class Command(BaseCommand):
     help = "Scrub sensitive data from selected fields"
 
     def handle(self, *args, **kwargs):
+        try:
+            self.scrub_data()
+        except OperationalError as error:
+            raise CommandError(
+                "Could not connect to the database. "
+                "Before running this command, make sure PostgreSQL is running. "
+                "If using Docker, start it with `just docker/db`."
+            ) from error
+
+    def scrub_data(self):
         models = {model for model in apps.get_app_config("jobserver").get_models()}
         models |= {model for model in apps.get_app_config("applications").get_models()}
 


mikerkelly:
Thanks for considering things that could go wrong. However, I don't think that we should add this error handling to this management command. The error message reported in Slack:

```
django.db.utils.OperationalError: connection failed: connection to server at "127.0.0.1", port 6543 failed: Connection refused
        Is the server running on that host and accepting TCP/IP connections?
error: Recipe `scrub-data` failed on line 99 with exit code 1
```

... is the same error message that will display when **any** management command fails due to the database not being up, including `runserver`/`just run`. Most commands use the database. Anyone developing Job Server is likely to be familiar with this message. It is fairly easy to interpret and the developer that pointed it out did interpret it correctly. There is nothing special about this management command in this regard. If we wanted such error handling, I think that we should put it in `manage.py` instead so that it applies to all commands. But I don't think this is necessary. I don't think we need to "handle" errors unless we can do something about them. Reraising a new error is only useful if the original is really unclear and will trip people up regularly, or if we want to hide implementation detail from an outer context.

I think it was worth adding a note to the Slack message asking developers to use this, and that is sufficient.

What do you think?

nishtha-kalra:
> Thanks for considering things that could go wrong. However, I don't think that we should add this error handling to this management command. The error message reported in Slack:
> 
> ```
> django.db.utils.OperationalError: connection failed: connection to server at "127.0.0.1", port 6543 failed: Connection refused
>         Is the server running on that host and accepting TCP/IP connections?
> error: Recipe `scrub-data` failed on line 99 with exit code 1
> ```
> 
> ... is the same error message that will display when **any** management command fails due to the database not being up, including `runserver`/`just run`. Most commands use the database. Anyone developing Job Server is likely to be familiar with this message. It is fairly easy to interpret and the developer that pointed it out did interpret it correctly. There is nothing special about this management command in this regard. If we wanted such error handling, I think that we should put it in `manage.py` instead so that it applies to all commands. But I don't think this is necessary. I don't think we need to "handle" errors unless we can do something about them. Reraising a new error is only useful if the original is really unclear and will trip people up regularly, or if we want to hide implementation detail from an outer context.
> 
> I think it was worth adding a note to the Slack message asking developers to use this, and that is sufficient.
> 
> What do you think?

Yes I agree, will close this out.

PR #5894, opened 2026-06-09
Make Run Jobs visible for DeploymentAdministrator

### LINKED ISSUE #5892: Bug report: "Run Jobs" button no longer displays for Deployment Administrators
# Summary
The "Run Jobs" button is no longer visible to Deployment Administrators on the workspace detail page.

This [commit](https://github.com/opensafely-core/job-server/commit/42da240e5073109454511d7a88ea1b94f8dfedb5) added an `is_member` conditional around some of the buttons on the workspace details page. However users with the "Deployment Administrator" role can run jobs on any workspace without being members of that workspace. So while the button should show for them it doesn't.

# Impact of the bug

Now the initial confusion is over the impact on me individually is fairly minimal: I can just add `/run-jobs/` to end of the workspace URL to navigate there. But it's probably worth fixing at some point to save future confusion for others.

tomodwyer:
As they might not be a member of a project.

This includes a refactor for the nasty if/else statement moving that logic to the view, and adding tests around the existing behaviour.

Closes https://github.com/opensafely-core/job-server/issues/5892

### DIFF
--- a/jobserver/models/workspace.py
+++ b/jobserver/models/workspace.py
@@ -6,6 +6,9 @@ from django.urls import reverse
 from django.utils import timezone
 from furl import furl
 
+from ..authorization import has_permission
+from ..authorization.permissions import Permission
+
 
 logger = structlog.get_logger(__name__)
 
@@ -145,6 +148,18 @@ class Workspace(models.Model):
 
     def __str__(self):
         return self.name
+
+    def get_run_jobs_disabled_reason(self, user):
+        if self.is_archived:
+            return "Jobs cannot be run on an archived workspace"
+
+        if not has_permission(user, Permission.JOB_RUN, project=self.project):
+            return "You do not have permission to run jobs on this workspace"
+
+        if not user.backends.exists():
+            return "You do not have permission to run jobs on any OpenSAFELY backends"
+
+        return None
 
     def get_absolute_url(self):
         return reverse(
--- a/jobserver/views/workspaces.py
+++ b/jobserver/views/workspaces.py
@@ -208,6 +208,8 @@ class WorkspaceDetail(View):
 
         outputs = self.get_output_permissions(workspace)
 
+        run_jobs_disabled_reason = workspace.get_run_jobs_disabled_reason(request.user)
+
         context = {
             "first_job": first_job,
             "honeycomb_can_view_links": honeycomb_can_view_links,
@@ -215,8 +217,10 @@ class WorkspaceDetail(View):
             "is_member": is_member,
             "outputs": outputs,
             "repo_is_private": repo_is_private,
+            "run_jobs_disabled_reason": run_jobs_disabled_reason,
             "show_admin": show_admin,
             "show_publish_repo_warning": show_publish_repo_warning,
+            "show_run_jobs_button": is_member or can_run_jobs,
             "user_can_archive_workspace": can_archive_workspace,
             "user_can_run_jobs": can_run_jobs,
             "user_can_toggle_notifications": can_toggle_notifications,
--- a/templates/workspace/detail.html
+++ b/templates/workspace/detail.html
@@ -81,32 +81,30 @@
     </div>
 
     <ul class="flex flex-row gap-2">
-      {% if is_member %}
-        {% if workspace.is_archived %}
-          <li>
-            {% #button type="button" variant="info" disabled=True tooltip="Jobs cannot be run on an archived workspace" %}
-              Run jobs
-            {% /button %}
-          </li>
-        {% elif not user_can_run_jobs %}
-          <li>
-            {% #button type="button" variant="info" disabled=True tooltip="You do not have permission to run jobs on this workspace" %}
-              Run jobs
-            {% /button %}
-          </li>
-        {% elif not user_has_backends %}
-          <li>
-            {% #button type="button" variant="info" disabled=True tooltip="You do not have permission to run jobs on any OpenSAFELY backends" %}
-              Run jobs
-            {% /button %}
-          </li>
-        {% else %}
-          <li>
-            {% #button class="plausible-event-name=Run+jobs+click" href=workspace.get_jobs_url type="link" variant="success" %}
-              Run jobs
-            {% /button %}
-          </li>
-        {% endif %}
+      {% if user_can_run_jobs and workspace.is_archived %}
+        <li>
+          {% #button type="button" variant="info" disabled=True tooltip="Jobs cannot be run on an archived workspace" %}
+            Run jobs
+          {% /button %}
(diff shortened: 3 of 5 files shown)


templates/workspace/detail.html
```diff
@@ -81,7 +81,7 @@ <h1 class="text-3xl tracking-tight wrap-break-word font-bold text-slate-900 sm:t
     </div>
 
     <ul class="flex flex-row gap-2">
-      {% if is_member %}
+      {% if is_member or is_deployment_administrator %}
```
evansd: You already have the `user_can_run_jobs` variable defined. It might make sense to use that here instead of `is_deployment_administrator` as it avoids encoding more specific facts about our current permissions structure into the template.

tomodwyer: Ha, as soon as I pushed this up and re-read the code I thought - there's got to be a better way. I've tweaked it now, bit of a nasty if/else statement, but it does what we want until I can improve the way we load disabled buttons.

evansd: Ha! Sorry for seagulling in on your PR. I just got a notification and thought I'd take a look.

jobserver/views/workspaces.py
```diff
@@ -208,15 +208,30 @@ def get(self, request, *args, **kwargs):
...
+            run_jobs_disabled_reason = "Jobs cannot be run on an archived workspace"
+        elif not can_run_jobs:
+            run_jobs_disabled_reason = (
+                "You do not have permission to run jobs on this workspace"
+            )
+        elif not has_backends:
+            run_jobs_disabled_reason = (
+                "You do not have permission to run jobs on any OpenSAFELY backends"
+            )
+
```
mikerkelly: Separation of concerns: I think that moving this logic into the view code instead of the template code is a clear  improvement but this is still not the right place for it. This is business logic about when jobs may be run in a workspace. A view is concerned with the request response cycle, and generally we prefer in this codebase "thin views" that  delegate business logic elsewhere. What do you think about making this chunk of logic a function on the `Workspace` model with signature `def get_reason_cannot_run_jobs(self, user)` returning the string reason or `None`?

As well as being a more logical organisation, that would also avoid the issue of the tests for this view now being tightly coupled to that business logic. We can test the business logic where it is defined and mock it out in the view tests to just get the positive/negative cases.

I can see that this "fat view" has other logic embedded in it but let's leave that alone for now.

tests/unit/jobserver/views/test_workspaces.py
```diff
@@ -432,6 +433,41 @@ def test_workspacedetail_authorized_run_jobs(rf, project_membership, role_factor
...
+    response.render()
+    assert b"Run jobs" in response.content
+    assert workspace.get_jobs_url().encode() in response.content
+
+
+def test_workspacedetail_deployment_administrator_run_jobs(
+    rf, project_membership, role_factory
+):
+    workspace = WorkspaceFactory()
+    user = UserFactory(roles=[DeploymentAdministrator])
```
mikerkelly: Testing for `DeploymentAdministrator` specifically tightly couples this view test to that specific role's permissions. We prefer creating testing based on the permission rather than the role to avoid this. This is still a valid distinct test where the user has `Permission.JOB_RUN` at the user/global level, not at the project level.

There are helpers in conftest `role_factory` and `user_with_fake_role_factory` that can be useful here.

tests/unit/jobserver/views/test_workspaces.py
```diff
@@ -432,6 +433,41 @@ def test_workspacedetail_authorized_run_jobs(rf, project_membership, role_factor
     assert response.status_code == 200
     assert response.context_data["user_can_run_jobs"]
     assert response.context_data["user_has_backends"]
+    assert response.context_data["show_run_jobs_button"]
+    assert response.context_data["run_jobs_disabled_reason"] is None
+
+    response.render()
+    assert b"Run jobs" in response.content
+    assert workspace.get_jobs_url().encode() in response.content
```
mikerkelly: Suggestion: These 7 asserts and the rendering are repeated in 5 tests with the only difference being the `run_jobs_disabled_reason`. Maybe a helper function can reduce repetition.

tests/unit/jobserver/views/test_workspaces.py
```diff
@@ -432,6 +433,41 @@ def test_workspacedetail_authorized_run_jobs(rf, project_membership, role_factor
...
+    assert response.context_data["user_has_backends"]
+    assert response.context_data["show_run_jobs_button"]
+    assert response.context_data["run_jobs_disabled_reason"] is None
+
+    response.render()
+    assert b"Run jobs" in response.content
+    assert workspace.get_jobs_url().encode() in response.content
 
 
 def test_workspacedetail_authorized_run_jobs_no_backends(
```
mikerkelly, line 438: Separation of concerns: as in view code above. If we move the business logic in the view code to a business logic function in a model we can mock the result and don't need to tightly couple this or have separate tests for the different cases, we only need one test for the "visible but disabled" case.

tests/unit/jobserver/views/test_workspaces.py
```diff
@@ -502,8 +620,12 @@ def test_workspacedetail_logged_out(rf):
     assert response.status_code == 200
 
     assert not response.context_data["user_can_run_jobs"]
+    assert not response.context_data["show_run_jobs_button"]
```
mikerkelly: I am not sure the additions to this function are germane to the behaviour under test. Maybe we can do without them to keep the test focus.

mikerkelly (commented):
Thanks, Thomas. I tried each different branch in a development environment and this works well. I have some comments about where the business logic should live and about how to structure the testing.

jobserver/views/workspaces.py
tomodwyer: That makes sense, thanks for the guidance - I will update that 👌

tomodwyer:
@mikerkelly thank you for that, I've reworked it and will get someone to review based on your comments, then rebase it in to a couple of concise commits.

jobserver/models/workspace.py
```diff
@@ -146,6 +149,18 @@ class Meta:
     def __str__(self):
         return self.name
 
+    def get_reason_cannot_run_jobs(self, user):
```
nishtha-kalra: If I understand this correctly, this function checks if user can run job and returns reason if it can't. Right? Would it be better if we use `get_run_jobs_disabled_reason`?

nishtha-kalra (approved):

jobserver/models/workspace.py
tomodwyer: Added that in as part of the rebase.

PR #5904, opened 2026-06-15
Add dump and restore steps for scrubbed database job

nishtha-kalra:
Fixes https://github.com/opensafely-core/security/issues/52
Update the `dump_scrubbed_db` job so it now creates a raw dump from the configured readonly JobServer database and restores that dump into the configured scrubbed database. The job stores the intermediate dump under `/tmp` and removes it in a finally block so the raw dump is cleaned up after success or failure.

Add Django database aliases for the readonly and scrubbed database URLs. This lets code use parsed Django database settings rather than passing raw environment variable URLs.

Use password-free PostgreSQL connection URLs for `pg_dump` and `pg_restore`, passing the database password via `PGPASSWORD` instead. This matches the existing `dump_db` job pattern and avoids exposing credentials in command arguments.

Update `dotenv-sample` so the scrubbed database points at a separate local database name, making the source and restore targets distinct in local setup.

### Testing
To test it locally:
1. Create a db where dump will be restored (this won't be required for dokku as db alread exists there)
`createdb -h localhost -p 6543 -U user jobserver_scrubbing`
2. Copy updated `dotenv-sample` to `.env`
3. Load these changes
`set -a; source .env; set +a`
4. Run the job
`just manage runjob dump_scrubbed_db`
The database will be cleared after restoring, as we don't want another copy of sensitive data on dokku.


jobserver/settings.py
```diff
@@ -278,12 +278,10 @@ def immutable_file_test(path, url):
...
-JOBSERVER_READONLY_DATABASE_URL = os.environ.get(
-    "JOBSERVER_READONLY_DATABASE_URL", default=""
-)
-JOBSERVER_SCRUBBED_DATABASE_URL = os.environ.get(
-    "JOBSERVER_SCRUBBED_DATABASE_URL", default=""
-)
+if readonly_database_url := os.environ.get("JOBSERVER_READONLY_DATABASE_URL"):
+    DATABASES["readonly"] = dj_database_url.parse(readonly_database_url)
+if scrubbed_database_url := os.environ.get("JOBSERVER_SCRUBBED_DATABASE_URL"):
+    DATABASES["scrubbed"] = dj_database_url.parse(scrubbed_database_url)
```
mikerkelly: Suggestions to make these clearer:

- Move these to just after where the `DATABASES` variable is defined, so all the database-related lines are together. At the moment they are in the `#Auth` section of settings, this does not seem right.
- Add some brief comments explaining what these are and pointing to other documentation. It is not clear at first glance what these are used for.
- I am not sure about "SCRUBBED_DATABASE" / "scrubbed" as names. That name sounds like the database has already or always been scrubbed, when this is not the case. Most of the time it will be empty, apart from when we are doing data scrubbing. Maybe it can be called the "DATA_SCRUBBING_DATABASE" and "data_scrubbing"?

dotenv-sample
```diff
@@ -2,7 +2,7 @@
 # point another postgres
 DATABASE_URL=postgres://user:pass@localhost:6543/jobserver
 JOBSERVER_READONLY_DATABASE_URL=postgres://user:pass@localhost:6543/jobserver
-JOBSERVER_SCRUBBED_DATABASE_URL=postgres://user:pass@localhost:6543/jobserver
+JOBSERVER_SCRUBBED_DATABASE_URL=postgres://user:pass@localhost:6543/jobserver_scrubbed
```
mikerkelly: Can we add comments here.

Maybe they should be commented out in this file? Most developers don't need to set these.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -12,14 +13,62 @@ class Job(YearlyJob):
...
-        logger.info("Scrubbed database dump job")
-        logger.info(
-            "Database path for readonly JobServer database",
-            database_path=readonly_db_url.path,
-        )
-        logger.info(
-            "Database path for scrubbed JobServer database",
-            database_path=scrubbed_db_url.path,
-        )
+        raw_dump = pathlib.Path("/tmp/jobserver_raw.dump")
```
mikerkelly: Please use the standard library module [tempfile](https://docs.python.org/3/library/tempfile.html) to manage temporary files. This ensures proper handling. Probably using [NamedTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) as a context manager is best here.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -12,14 +13,62 @@ class Job(YearlyJob):
...
+        readonly_db = settings.DATABASES.get("readonly")
+        scrubbed_db = settings.DATABASES.get("scrubbed")
+
+        if readonly_db is None:
+            raise RuntimeError("JOBSERVER_READONLY_DATABASE_URL is not found")
```
mikerkelly: "Not found" is not quite clear. I think "Not set" would be better.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -12,14 +13,62 @@ class Job(YearlyJob):
...
-            "Database path for readonly JobServer database",
-            database_path=readonly_db_url.path,
-        )
-        logger.info(
-            "Database path for scrubbed JobServer database",
-            database_path=scrubbed_db_url.path,
-        )
+        raw_dump = pathlib.Path("/tmp/jobserver_raw.dump")
+        readonly_db = settings.DATABASES.get("readonly")
+        scrubbed_db = settings.DATABASES.get("scrubbed")
```
mikerkelly: I am not sure about `scrubbed_db` as a variable name. That name sounds like the database has already or always been scrubbed, when this is not the case. Most of the time it will be empty, apart from when we are doing data scrubbing. Maybe it can be called the and `data_scrubbing_database` instead?

mikerkelly: We may as well spell out `database` throughout this file? Generally we try to avoid 1-3 character abbreviation names as they are harder to read.

jobserver/settings.py
mikerkelly: Question: can we both add these to `DATABASES` and keep the URL as a setting? Then `database_url` function is not needed, it can be use the URL setting directly.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -12,14 +13,62 @@ class Job(YearlyJob):
...
+    subprocess.check_call(
+        [
+            "pg_dump",
+            "--format=c",
+            "--no-acl",
+            "--no-owner",
+            f"--file={output}",
+            database_url(db),
+        ],
+        env={"PGPASSWORD": db["PASSWORD"]},
```
mikerkelly: This did not work for me. `pg_dump` is not found. Suggest:

```python
env = {**os.environ, "PGPASSWORD": db["PASSWORD"]}
```

... because otherwise you wipe out the existing environment, including PATH, which can make `pg_dump` fail to be found. And the same below.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -12,14 +13,62 @@ class Job(YearlyJob):
...
+            restore_database(scrubbed_db, raw_dump)
+
+            logger.info("Finished restoring dump into database")
+
+        finally:
+            logger.info("Deleting temporary source dump")
+            raw_dump.unlink(missing_ok=True)
+
+
+def database_url(db):
```
mikerkelly: Below I suggest just recording the URL in the settings and using that, so this function would not need to exist.

If not, suggest passing host/user/dbname flags directly, which is usually safer than building a URL (for example you have not considered quoting here).

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -12,14 +13,62 @@ class Job(YearlyJob):
...
+    return f"postgresql://{db['USER']}@{db['HOST']}:{db['PORT']}/{db['NAME']}"
+
+
+def dump_database(db, output):
+    if not output.parent.exists():
+        raise RuntimeError(f"Unknown output directory: {output.parent}")
```
mikerkelly: Is this necessary? I expect `pg_dump` will fail with a clear message in this case anyway.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -12,14 +13,62 @@ class Job(YearlyJob):
...
+
+    subprocess.check_call(
```
mikerkelly: Optional suggestion: `subprocess.run(..., check=True)` is more modern than `check_call`. You can do everything you need with that one function with different parameters. No big deal though.

mikerkelly (commented):
Thanks, Nishtha. I have some review comments inline.

Question: I also think that we agreed that the script should clear the scrubbing database after running, to avoid the data sitting at rest outside the other databases. Should we add that now, before we test that this works in production? Otherwise, we have another database with the sensitive data in it, until we do the rest of the implementation.

jobserver/jobs/yearly/dump_scrubbed_db.py
mikerkelly: [JobError](https://github.com/django-extensions/django-extensions/blob/32ddfb0499bec7f39f9fbb44568d2781cecd1f32/django_extensions/management/jobs.py#L14) could be a slightly better fit here, unsure if that would help the extension manage the exception like `CommandError` for base Django commands.

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: I have passed the host, port, username, and database name as separate arguments. Did not pass the URL directly to avoid exposing the password. (Idea taken from https://github.com/opensafely-core/job-server/commit/62d20c3475e00aa17c7d8c1ab77706452e2753a4)

jobserver/settings.py
nishtha-kalra: I have renamed all related variables to "scrubbing" instead of "scrubbed", except for `JOBSERVER_SCRUBBED_DATABASE_URL`, because it is already configured in Dokku and renaming it does not seem worth the operational change. The database alias is still `data_scrubbing`, which keeps the code intent clear.

nishtha-kalra:
> Thanks, Nishtha. I have some review comments inline.
> 
> Question: I also think that we agreed that the script should clear the scrubbing database after running, to avoid the data sitting at rest outside the other databases. Should we add that now, before we test that this works in production? Otherwise, we have another database with the sensitive data in it, until we do the rest of the implementation.

Thanks for the review Mike. 
Yes, I wanted to add that logic later, but I added it now since there will be another database with sensitive data.
I have also addressed all other comments and suggestions.

dotenv-sample
mikerkelly: As well as commenting out the lines, can you please also add a comment or two as documentation?

jobserver/settings.py
mikerkelly: As this process is not being used yet I think it is minimal effort to change that environment variable name in the code and in dokku, so can we please do that so that we use consistent naming? I think it worth doing that now while it is easy to avoid confusion and mistakes in future.

I think that means SSHing and running one dokku command, and changing the following occurrences in Job Server:

```
~/job-server$ rg SCRUBBED
INSTALL.md
32:dokku config:set job-server JOBSERVER_SCRUBBED_DATABASE_URL='xxxx'

jobserver/settings.py
172:DATA_SCRUBBING_DATABASE_URL = os.environ.get("JOBSERVER_SCRUBBED_DATABASE_URL")

jobserver/jobs/yearly/dump_scrubbed_db.py
24:            raise JobError("JOBSERVER_SCRUBBED_DATABASE_URL is not set")

dotenv-sample
5:# JOBSERVER_SCRUBBED_DATABASE_URL=postgres://user:pass@localhost:6543/jobserver_scrubbing

```

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -12,14 +14,79 @@ class Job(YearlyJob):
...
+        data_scrubbing_database = settings.DATABASES.get("data_scrubbing")
+
+        if readonly_database is None:
+            raise JobError("JOBSERVER_READONLY_DATABASE_URL is not set")
+
+        if data_scrubbing_database is None:
+            raise JobError("JOBSERVER_SCRUBBED_DATABASE_URL is not set")
+
+        with tempfile.NamedTemporaryFile(suffix=".dump") as raw_dump:
+            logger.info("Creating raw dump of readonly jobserver database")
```
mikerkelly: Suggestion: may as well include the temporary file name in the log message.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -12,14 +14,79 @@ class Job(YearlyJob):
...
+            dump_database(readonly_database, raw_dump.name)
+            try:
+                logger.info("Restoring dump into data scrubbing database")
+                restore_database(data_scrubbing_database, raw_dump.name)
+                logger.info("Finished restoring dump into database")
+            finally:
+                logger.info("Clearing data scrubbing database")
+                clear_database(data_scrubbing_database)
+
```
mikerkelly, line 43: Suggest: a log message when clearing the database is done would be useful to make it very explicit.

jobserver/settings.py
nishtha-kalra: renamed it now

mikerkelly (approved):
Thanks, Nishtha, looks good.

I think we should test this in production before trying to extend it.

PR #5914, opened 2026-06-17
Add `database` option to data scrubbing management command

mikerkelly:
Partly fixes https://github.com/opensafely-core/security/issues/55. Adds a required `database` option to the database scrubbing management command to enable this to be useful in production where we intend to run it against a non-default database.

Also uses an iterator to reduce memory usage, adds transaction wrapping, adds a `data_scrubbing` database alias (overlapping with https://github.com/opensafely-core/job-server/pull/5904/), and a guard clause to reduce the risk of accidentally running this against the default database in production.

To test, suggest getting a separate database with data from the default database loaded into it following https://github.com/opensafely-core/job-server/pull/5904/, then running `just manage scrub_data data_scrubbing` and using `psql` to confirm that the target database changed and the default one did not (for example by checking a particular username and email in the default database before and after).

### DIFF
--- a/data_scrubbing/management/commands/scrub_data.py
+++ b/data_scrubbing/management/commands/scrub_data.py
@@ -12,10 +12,18 @@ class Command(BaseCommand):
 
     help = "Scrub sensitive data from selected fields"
 
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "database_alias",
+            help="Alias of database to scrub, should be in Django setting DATABASES",
+        )
+
     def handle(self, *args, **kwargs):
+        database_alias = kwargs["database_alias"]
+
         models = {model for model in apps.get_app_config("jobserver").get_models()}
         models |= {model for model in apps.get_app_config("applications").get_models()}
-        with transaction.atomic():
+        with transaction.atomic(using=database_alias):
             for model in models:
                 data_scrubbing = getattr(model, "DataScrubbing", None)
                 if data_scrubbing is None:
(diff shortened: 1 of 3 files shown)


data_scrubbing/management/commands/scrub_data.py
```diff
@@ -11,26 +13,46 @@ class Command(BaseCommand):
...
-            )
+        with transaction.atomic(using=database_alias):
+            for model in models:
+                data_scrubbing = getattr(model, "DataScrubbing", None)
+                if data_scrubbing is None:
+                    continue
+                scrub_fields = getattr(data_scrubbing, "fields_to_scrub", None)
+                if not scrub_fields:
+                    continue
+                field_names = scrub_fields.keys()
```
nishtha-kalra, line 44: Optional - We could also write this as:
`field_names = list(scrub_fields)`

data_scrubbing/management/commands/scrub_data.py
```diff
@@ -11,26 +13,46 @@ class Command(BaseCommand):
 
     help = "Scrub sensitive data from selected fields"
 
+    def add_arguments(self, parser):
+        options = list(settings.DATABASES.keys())
+        parser.add_argument(
+            "database_alias",
+            help=f"Alias of database to scrub, options: {options}",
+        )
+        parser.add_argument(
```
nishtha-kalra, line 22: Really liked the `--i-am-sure` guard here, thanks!

nishtha-kalra (approved):
Looks good to me. A small suggestion, but otherwise, it looks good, and I was able to test it as recommended.

data_scrubbing/management/commands/scrub_data.py
mikerkelly: We could do. I think that this provides slightly more hint that this is coming from a dict, which helps to see what is going on. This is all tightly coupled so it seems okay. So I think that I will leave it.

PR #5928, opened 2026-06-23
Update 403 and 404 page messages for users

### LINKED ISSUE #5909: Add hints when a user is logged out
# Why are we doing this?

It is not always clear to a user when they are logged out. There are many public pages they can still access, even for project resources. They may then be confused to receive a 404 when accessing a permission-gated resource or action (not 403 so as not to confirm the existence of the resource). And some pages may be missing enhancements they would see if they were logged in (for example extra buttons). Sometimes this leads to tech support requests.

We can try to make it clearer when this has happened.

Current 404 page:

Image

Current 403 page:

Image

[Original Slack discussion ](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1781513927375419)

...

# What are we doing?

- Improving the message on the 403 and 404 pages so that the possible causes are clearer.
- Add some extra hints to the 403 and 404 pages if not `user.is_authenticated` to remind users that they may have been logged out and linking to the login page in a button.
- Add a `django.contrib.messages` message or other indication when the user visits a page that is enhanced by login and they are not logged in, with link. Could limit this to once per session to reduce spam when not relevant.
- Thomas mentioned the GitHub 404 has an overlay for logging in if the user is not logged in. Possibly we could do something like that.

Image

- Anything else?

The links can maybe include a `?next=` to redirect back to the resource.

Maybe limiting to once per session looks like:

```python
if not request.session.get('seen_login_hint'):
    messages.info(...)
    request.session['seen_login_hint'] = True
```

Maybe that can be a mixin for views.

---

KatieB5:
Part fixes #5909

Summary of work in this PR:
- [x] Improve the messages on the 403 and 404 pages so that the possible causes for these are clearer to users.
- [x] Add some extra hints to the 403 and 404 pages if not `user.is_authenticated` to remind users that they may have been logged out and linking to the login page in a button.

Testing:
- Updating the unit tests for the permission_denied and page_not_found_views
- Tested manually with a local dev site (`just run` and debug set to false, so I could see the error pages)
  - As a logged-out user, I navigated to http://localhost:8000/test/ (404), and http://localhost:8000/staff/ (403)
  - As a logged-in user, I navigated to http://localhost:8000/test/ (404), and http://localhost:8000/staff/projects/create/ (403, because I don't have `ServiceAdministrator` role and associated permissions required to access the `ProjectCreate` view)

<details>
  <summary>Screenshots from running a local site</summary>
  
**Error pages for a logged in user**
Screenshot from 2026-06-24 12-42-10
Screenshot from 2026-06-24 12-48-37

  **Error pages for a user who isn't logged in**
Screenshot from 2026-06-24 12-37-08
Screenshot from 2026-06-24 12-42-26

</details>

### DIFF
--- a/jobserver/views/errors.py
+++ b/jobserver/views/errors.py
@@ -41,6 +41,12 @@ def csrf_failure(request, reason=""):
 
 
 def page_not_found(request, exception=None):
+
+    if not request.user.is_authenticated:
+        error_message = "We could not find the page you are looking for. The link may be incorrect, the page may have moved, or you may need to log in to access it. Please check the URL, try logging in, or contact us if you need further help."
+    else:
+        error_message = "We could not find the page you are looking for. The link may be incorrect or the page may have moved. Please check the URL or contact us if you need further help."
+
     return TemplateResponse(
         request,
         "error.html",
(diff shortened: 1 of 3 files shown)


tomodwyer (approved):
Going to write a ticket to better style these pages, but the guidance for the user is now much better.

PR #5929, opened 2026-06-23
Data scrubbing: Truncate `django_session` and `social_auth` tables

mikerkelly:
Part addresses https://github.com/opensafely-core/security/issues/55.

These Django-defined and extension-defined tables may contain sensitive personal data or tokens. Let's just truncate these tables in full, as none of them seem to be necessary for the development site to work.

Use `sql_flush` so Django can write a backend-agnostic query list for us and
handle escaping et cetera. Use `allow_cascade=False` to force listing all
affected tables explicitly.

Tested manually by restoring a production copy then running the command and checking the database state before and after.

Some very old Django auth and admin tables are not present in the test database because we don't use them at all. So the integration test fails if you try to unconditionally truncate them. Probably the best option is to make a maintenance migration to drop these very old unused tables in production, so that it aligns with the tables that we actually use in the code, then we don't have to worry about scrubbing them or reintroduce them in the tests.

```
    # Default Django User tables. Different to jobserver_user.
    # Guard against future use or old production data that may have personal data.
    "auth_user",
    "auth_user_groups",
    "auth_user_user_permissions",
    # Can contain references to many other models including personal data.
    # Guard against future use or old production data that may have personal data.
    "django_admin_log",
```

### DIFF
--- a/applications/models.py
+++ b/applications/models.py
@@ -239,7 +239,6 @@ class ContactDetailsPage(AbstractPage):
     telephone = models.TextField(blank=True)
     job_title = models.TextField(blank=True)
     team_name = models.TextField(blank=True)
-    # TODO: how will we tie this to an existing org, especially with typos?
     organisation = models.TextField(blank=True)
 
     class DataScrubbing:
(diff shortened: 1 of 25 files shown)


mikerkelly:
Close in favour of https://github.com/opensafely-core/job-server/pull/5936

PR #5930, opened 2026-06-23
Fix error in scrubbing job encountered while testing in production

nishtha-kalra:
While testing the job in production after merging [this](https://github.com/opensafely-core/job-server/pull/5904) PR, I encountered an error and related [Sentry alert](https://ebm-datalab.sentry.io/issues/7570238993/?alert_rule_id=15682564&alert_type=issue&notification_uuid=d237a5c8-86ca-45ac-bad2-deb2f41fd179&project=5443358).

This PR fixes that error and add related error handling:
- Give the `jobserver_data_scrubbing` user `SELECT` permissions on sequences in the jobserver database, in addition to table `SELECT` permissions.
- Strip scrubbing database URLs from Sentry events. Add the `JOBSERVER_SCRUBBING_DATABASE_URL` and
`JOBSERVER_READONLY_DATABASE_URL` environment variables to the Sentry token scrub list so their full values are replaced before events are sent to Sentry.
- Handle error when database commands fail. Run `pg_dump`, `pg_restore` and `psql` through a helper that captures stderr output and raises `JobError` when the command fails. This makes failures visible in Sentry and helps to debug any issues.

## Testing
- The job ran successfully in production once when correct permissions were granted.
Related [slack](https://bennettoxford.slack.com/archives/C03D1G814BA/p1782222341533529) thread.
- Tested manually that JobError was thrown when the command failed.

### DIFF
--- a/jobserver/jobs/yearly/dump_scrubbed_db.py
+++ b/jobserver/jobs/yearly/dump_scrubbed_db.py
@@ -61,7 +61,7 @@ def run_database_command(command, database_config):
     try:
         subprocess.run(
             command,
-            env={**os.environ, "PGPASSWORD": database_config["PASSWORD"]},
+            env={"PATH": os.environ["PATH"], "PGPASSWORD": database_config["PASSWORD"]},
             capture_output=True,
             text=True,
             check=True,
--- a/jobserver/jobs/yearly/dump_scrubbed_db.py
+++ b/jobserver/jobs/yearly/dump_scrubbed_db.py
@@ -56,15 +56,23 @@ def database_connection_args(database_config):
 
 
 def run_database_command(command, database_config):
-    result = subprocess.run(
-        command,
-        env={**os.environ, "PGPASSWORD": database_config["PASSWORD"]},
-        capture_output=True,
-        text=True,
-    )
+    logger.info("Running database command", command=command)
 
-    if result.returncode != 0:
-        raise JobError(f"{command[0]} failed: stderr={result.stderr}")
+    try:
+        subprocess.run(
+            command,
+            env={**os.environ, "PGPASSWORD": database_config["PASSWORD"]},
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+    except subprocess.CalledProcessError as error:
+        raise JobError(
+            f"{command[0]} failed\n"
+            f"returncode={error.returncode}\n"
+            f"stdout={error.stdout}\n"
+            f"stderr={error.stderr}"
+        ) from error
 
 
 def dump_database(database_config, output):
--- a/jobserver/jobs/yearly/dump_scrubbed_db.py
+++ b/jobserver/jobs/yearly/dump_scrubbed_db.py
@@ -55,8 +55,20 @@ def database_connection_args(database_config):
     ]
 
 
+def run_database_command(command, database_config):
+    result = subprocess.run(
+        command,
+        env={**os.environ, "PGPASSWORD": database_config["PASSWORD"]},
+        capture_output=True,
+        text=True,
+    )
+
+    if result.returncode != 0:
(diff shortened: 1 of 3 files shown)


jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -55,8 +55,20 @@ def database_connection_args(database_config):
     ]
 
 
+def run_database_command(command, database_config):
```
mikerkelly, line 58: Suggestion: can we log `command` before trying to run it? I think that that will make the logs more legible. As does not include the password part I think that this is okay.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -55,8 +55,20 @@ def database_connection_args(database_config):
...
+    result = subprocess.run(
+        command,
+        env={**os.environ, "PGPASSWORD": database_config["PASSWORD"]},
+        capture_output=True,
+        text=True,
+    )
+
+    if result.returncode != 0:
+        raise JobError(f"{command[0]} failed: stderr={result.stderr}")
```
mikerkelly: Suggestion: also include the return code and stdout as these may have useful information.

I think it will be more Pythonic to pass `check=True` and then catch the [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) and use its attributes for this to raise the `JobError`. EAFP.

services/sentry.py
```diff
@@ -89,6 +89,8 @@ def parse(data):
         "SENTRY_DSN",
         "SOCIAL_AUTH_GITHUB_KEY",
         "SOCIAL_AUTH_GITHUB_SECRET",
+        "JOBSERVER_SCRUBBING_DATABASE_URL",
+        "JOBSERVER_READONLY_DATABASE_URL",
```
mikerkelly, line 93: These are good to scrub, but we should also scrub `PASSWORD` and probably `PGPASSWORD` for prudence sake. That is the issue that was directly shown in https://bennettoxford.slack.com/archives/C03D1G814BA/p1782222341533529

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: added log

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: Makes sense. done

services/sentry.py
nishtha-kalra: added these.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -55,8 +55,28 @@ def database_connection_args(database_config):
     ]
 
 
+def run_database_command(command, database_config):
+    logger.info("Running database command", command=command)
+
+    try:
+        subprocess.run(
+            command,
+            env={**os.environ, "PGPASSWORD": database_config["PASSWORD"]},
```
mikerkelly: Hi Nishtha, looking at this again I realise that passing the whole environment down into the subprocess is not needed, only PATH and PGPASSWORD. And the other environment variables contain more secrets. So probably it is more secure to only pass what is needed. If I make the following change, I can still run this:

```python
env={"PATH": os.environ["PATH"], "PGPASSWORD": database_config["PASSWORD"]},
```

Sorry for the flaw in the original suggestion.

Would you mind testing that change and merging?

mikerkelly (approved):

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: That's a good point, added this change now

PR #5931, opened 2026-06-24
Update uv tooling config in Dockerfile

### LINKED ISSUE #5854: Update uv dependency management
# Why are we doing this?

Whilst working through #2456, files relating to jobserver's uv tooling for dependency management were referenced. There were a few things we could potentially consider updating.

# What are we doing?

Potentially do some or all of the following:
- [x] update jobserver Dockerfile:
   - [x] tidy up comments on [lines 37 and 38](https://github.com/opensafely-core/job-server/blob/f77b5e5f8320119ecf3259ec3358d4a41b7ed9ea/docker/Dockerfile#L37), as we're no longer using deadsnakes ppa (#5824)
   - [x] update `UV_VERSION` [on line 5](https://github.com/opensafely-core/job-server/blob/f77b5e5f8320119ecf3259ec3358d4a41b7ed9ea/docker/Dockerfile#L5) to 0.11 (from 0.9)
- [x] update jobserver [GHA workflows](https://github.com/opensafely-core/job-server/tree/main/.github/workflows)
  -  [ ] ~~update to ubuntu 24.04 for Verification tests workflow (all other workflows used 24.04 already, except Deploy which uses 22.04)~~ (Decided not to action this per [discussion with the team in Slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1782381572807649?thread_ts=1782373419.715179&cid=C069SADHP1Q))
  - [x] check if we can remove `uv` from the host runner set up (`install: uv` and `cache: uv`)

---

KatieB5:
Closes #5854 

Work done:
- [x] update jobserver Dockerfile:
   - [x] update old comments, as we're not using deadsnakes ppa to install python (#5824)
   - [x] update `UV_VERSION` to 0.11 (from 0.9), to match what we use in other services e.g. [opencodelists](https://github.com/opensafely-core/opencodelists/blob/69005f45b8903bd263df0591057c23e61aaf4048/docker/Dockerfile#L5)
- [x] update jobserver [GHA workflows](https://github.com/opensafely-core/job-server/tree/main/.github/workflows)
  - [x] check if we can remove `uv` from the host runner set up (`install: uv` and `cache: uv`)
    - the workflow steps that install `uv` on the host runner do need `uv`, as they execute `just` recipes that call `uv` commands, so the current config remains unchanged.

Testing:

- Ran `just docker-build` locally; ran successfully to completion
- Ran `just docker-serve` with the newly-build image; successfully ran a local dev site
- For the verification workflow, I went to our [Verification tests workflow page](https://github.com/opensafely-core/job-server/actions/workflows/verification-tests.yml) and ran the workflow on my branch; it succeeded.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,7 +2,7 @@
 
 ## These build args are used in FROM lines, so need to be declared globally and have defaults set.
 ARG NODE_VERSION
-ARG UV_VERSION=0.9
+ARG UV_VERSION=0.11
 
 #################################################
 #
@@ -34,8 +34,7 @@ RUN --mount=type=cache,target=/var/cache/apt \
 ENV UV_LINK_MODE=copy
 # compile at installation time, not import time
 ENV UV_COMPILE_BYTECODE=1
-# we are providing python via deadsnakes ppa
-# (as we require dynamic linking of openssl for security reasons)
+# we are using the ubuntu 24.04 system default python (3.12)
 ENV UV_PYTHON_DOWNLOADS=never
 # set the directory for the venv
 ENV UV_PROJECT_ENVIRONMENT="/opt/venv"


docker/Dockerfile
```diff
@@ -2,7 +2,7 @@
 
 ## These build args are used in FROM lines, so need to be declared globally and have defaults set.
 ARG NODE_VERSION
-ARG UV_VERSION=0.9
+ARG UV_VERSION=0.11
```
StevenMaude, line 5: suggestion: If we're setting a minimum here, it might be worthwhile to consider whether we need to bump the minimum in `pyproject.toml`.

StevenMaude (approved):

docker/Dockerfile
KatieB5: I think that makes sense from a clarity and consistency perspective, and it would help prevent potential drift between local development and production. 

But I'm unsure if we need to, or should, make this change. I think it would mean everyone working on jobserver locally would probably need to update their `uv` version, and it tightens the supported toolchain for no functional gain (at least, we don't depend on any features of 0.11 that I know of). Unhelpfully, I'm not actually sure why we pinned to 0.9 across our services originally; I suspect maybe this was the latest major version at the time of switching to `uv`. There's also the question of what we're using our `pyproject.toml` for: is meant to describe our minimum supported tool versions, or just what we happen to use in our project.

Looking at [opencodelists](https://github.com/opensafely-core/opencodelists/blob/05c106e36ccd1784bfe51966cf6434ea812aa91f/pyproject.toml#L133), [job-runner](https://github.com/opensafely-core/job-runner/blob/a8faebff99069b73b7ea2e5fbd8f5337830c77d7/pyproject.toml#L98) and [airlock](https://github.com/opensafely-core/airlock/blob/d5e7eb54b310f0509e3033fcdeb5e2cb1de05167/pyproject.toml#L134), they have 0.11 in their `Dockerfile`, and >=0.9 in their `pyproject.toml`.

StevenMaude: @KatieB5: That's OK 😃 I think it's fine to leave this as is, for now.

It may be that we review what we do here again in future.

It's also possible that [future improvements in `uv` make this simpler to manage too](https://github.com/astral-sh/uv/issues/11065). It would be preferable to not rely on the version someone has installed locally.

PR #5932, opened 2026-06-24
Address pytest parametrize deprecation notice

lucyb:
In the latest version of pytest (v9.1.0) a deprecation warning is added for the use of non-Collection iterables in `@pytest.mark.parametrize`, as this can cause unexpected issues when the iterables are exhausted after the first iteration. This change replaces all uses with a list instead.

This will allow us to apply the latest dependency updates from #5922.

### DIFF
--- a/tests/paired_fields_utils.py
+++ b/tests/paired_fields_utils.py
@@ -54,16 +54,26 @@ def get_all_paired_fields(lsuffix, rsuffix, *, exclude=None):
         else:
             models.pop(model)
 
+    paired_fields = []
     for model, field_names in models.items():
         factory = get_factory(model)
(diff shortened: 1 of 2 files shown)


mikerkelly (approved):
https://github.com/pytest-dev/pytest/releases
https://github.com/pytest-dev/pytest/issues/13409

> Using non-~collections.abc.Collection iterables (such as generators, iterators, or custom iterable objects) for the argvalues parameter in @pytest.mark.parametrize <pytest.mark.parametrize ref> and metafunc.parametrize <pytest.Metafunc.parametrize> is now deprecated.

> These iterables get exhausted after the first iteration,
leading to tests getting unexpectedly skipped in cases such as running pytest.main() multiple times,
using class-level parametrize decorators,
or collecting tests multiple times.

PR #5933, opened 2026-06-24
Add more sensitive tokens to Sentry scrubbing

### LINKED ISSUE #4219: Review how we're scrubbing data before sending it to Sentry
We're currently stripping sensitive data before sending it to Sentry here:
https://github.com/opensafely-core/job-server/blob/44e871a0153e2e61aa8cf1f696ee3a130ab26b76/services/sentry.py#L33

However, this isn't perfect and Sentry recommend this [approach](https://docs.sentry.io/platforms/python/data-management/sensitive-data/#scrubbing-data). Their approach requires turning off the option to send pii to Sentry and then strips sensitive session information as well as anything that looks like credentials. We should consider whether we want to do this. If not, we should consider unit testing our custom code.

We should also extend our `tokens` list with any newer ENV VARs and think about how we'll keep it up to date.

mikerkelly:
Part fixes https://github.com/opensafely-core/job-server/issues/4219.

These contain or could contain secrets so it seems better security to never send them to Sentry at all rather than rely on Sentry-side scrubbing to avoid these leaking.

Found by looking at the environment variables in Job Server production and looking for things that might be sensitive. Except PASSWORD and PGPASSWORD which are likely risks in either subprocesses or as part of the DATABASES setting.

We should do a more thorough check later.

https://github.com/opensafely-core/job-server/issues/4219 also suggests changing our approach to the one Sentry reccomends. We should do that later. https://docs.sentry.io/platforms/python/data-management/sensitive-data/#scrubbing-data

### DIFF
--- a/services/sentry.py
+++ b/services/sentry.py
@@ -82,17 +82,22 @@ def parse(data):
 
     # tokens with values we want to strip
     tokens = [
+        "DATABASES",
+        "DATABASE_URL",
+        "CSP_REPORT_URI",
         "GITHUB_TOKEN_TESTING",
         "JOBSERVER_GITHUB_TOKEN",
+        "JOBSERVER_SCRUBBING_DATABASE_URL",
+        "JOBSERVER_READONLY_DATABASE_URL",
+        "MAILGUN_API_KEYPASSWORD",
+        "PASSWORD",
+        "PGPASSWORD",
         "RAP_API_TOKEN",
         "SECRET_KEY",
         "SENTRY_DSN",
+        "SLACK_BOT_TOKEN",
         "SOCIAL_AUTH_GITHUB_KEY",
         "SOCIAL_AUTH_GITHUB_SECRET",
-        "JOBSERVER_SCRUBBING_DATABASE_URL",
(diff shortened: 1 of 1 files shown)


services/sentry.py
```diff
@@ -82,11 +82,18 @@ def parse(data):
 
     # tokens with values we want to strip
     tokens = [
+        "DATABASES",
+        "DATABASE_URL",
+        "CSP_REPORT_URI",
         "GITHUB_TOKEN_TESTING",
         "JOBSERVER_GITHUB_TOKEN",
+        "MAILGUN_API_KEYPASSWORD",
+        "PASSWORD",
```
nishtha-kalra, line 93: `PASSWORD` and `PGPASSWORD` covered in [this](https://github.com/opensafely-core/job-server/commit/c584aead7a112f3b639477495152397fa0ebfb39) commit.

mikerkelly: Reordered to be back in alphabetical order. That is consistent with what was there before that commit. It may be easier to understand and group alike things based on prefix.

nishtha-kalra (approved):

PR #5934, opened 2026-06-25
Extend dump_scrubbed_db job to call scrubbing command and create the scrubbed dump

nishtha-kalra:
https://github.com/opensafely-core/security/issues/67

For testing:
1. Create jobserver_scrubbing db
`createdb -h localhost -p 6543 -U user jobserver_scrubbing`

2. Run the job
`just manage runjob dump_scrubbed_db`

### DIFF
--- a/dotenv-sample
+++ b/dotenv-sample
@@ -9,6 +9,8 @@ DATABASE_URL=postgres://user:pass@localhost:6543/jobserver
 # JOBSERVER_READONLY_DATABASE_URL=postgres://user:pass@localhost:6543/jobserver
 # Database where data will be restored and scrubbed.
 # JOBSERVER_SCRUBBING_DATABASE_URL=postgres://user:pass@localhost:6543/jobserver_scrubbing
+# Path where the scrubbed database dump will be written.
+# JOBSERVER_SCRUBBED_DUMP_PATH=/tmp/jobserver_scrubbed.dump
 
 # Turn on debug
 DEBUG=1
--- a/dotenv-sample
+++ b/dotenv-sample
@@ -10,7 +10,7 @@ DATABASE_URL=postgres://user:pass@localhost:6543/jobserver
 # Database where data will be restored and scrubbed.
 # JOBSERVER_SCRUBBING_DATABASE_URL=postgres://user:pass@localhost:6543/jobserver_scrubbing
 # Path where the scrubbed database dump will be written.
-# JOBSERVER_SCRUBBED_DUMP_PATH=/tmp/jobserver_scrubbed.dump
+# JOBSERVER_SCRUBBED_DUMP_PATH=jobserver_scrubbed.dump
 
 # Turn on debug
 DEBUG=1
--- a/jobserver/jobs/yearly/dump_scrubbed_db.py
+++ b/jobserver/jobs/yearly/dump_scrubbed_db.py
@@ -16,7 +16,7 @@ class Job(YearlyJob):
     help = "Scrubbed database dump job (in progress)"
 
     def execute(self):
-        scrubbed_dump_path = pathlib.Path("/storage/jobserver_scrubbed.dump")
+        scrubbed_dump_path = pathlib.Path(settings.SCRUBBED_DATABASE_DUMP_PATH)
         temp_scrubbed_dump_path = scrubbed_dump_path.with_suffix(".dump.tmp")
         readonly_database = settings.DATABASES.get("readonly")
         data_scrubbing_database = settings.DATABASES.get("data_scrubbing")
--- a/jobserver/jobs/yearly/dump_scrubbed_db.py
+++ b/jobserver/jobs/yearly/dump_scrubbed_db.py
@@ -17,6 +17,7 @@ class Job(YearlyJob):
 
     def execute(self):
         scrubbed_dump_path = pathlib.Path("/storage/jobserver_scrubbed.dump")
+        temp_scrubbed_dump_path = scrubbed_dump_path.with_suffix(".dump.tmp")
         readonly_database = settings.DATABASES.get("readonly")
         data_scrubbing_database = settings.DATABASES.get("data_scrubbing")
 
@@ -43,12 +44,19 @@ class Job(YearlyJob):
                 logger.info("Scrubbing data scrubbing database")
                 call_command("scrub_data", "data_scrubbing")
                 logger.info("Finished scrubbing database")
-
-                logger.info("Creating scrubbed database dump", path=scrubbed_dump_path)
-                dump_database(data_scrubbing_database, scrubbed_dump_path)
-                logger.info(
-                    "Finished creating scrubbed database dump", path=scrubbed_dump_path
-                )
+                try:
+                    logger.info(
+                        "Creating scrubbed database dump",
+                        path=temp_scrubbed_dump_path.name,
+                    )
+                    dump_database(data_scrubbing_database, temp_scrubbed_dump_path)
+                    temp_scrubbed_dump_path.replace(scrubbed_dump_path)
+                    logger.info(
+                        "Finished creating scrubbed database dump",
+                        path=scrubbed_dump_path.name,
+                    )
+                finally:
+                    temp_scrubbed_dump_path.unlink(missing_ok=True)
             finally:
                 logger.info("Clearing data scrubbing database")
                 clear_database(data_scrubbing_database)
--- a/jobserver/jobs/yearly/dump_scrubbed_db.py
+++ b/jobserver/jobs/yearly/dump_scrubbed_db.py
@@ -37,6 +37,7 @@ class Job(YearlyJob):
                 restore_database(data_scrubbing_database, raw_dump_path)
                 scrub_database(settings.DATA_SCRUBBING_DATABASE_ALIAS)
                 dump_database(data_scrubbing_database, temp_scrubbed_dump_path)
+                # Note: `Pathlib.replace` requires source and destination to be on the same filesystem.
                 temp_scrubbed_dump_path.replace(scrubbed_dump_path)
             finally:
                 clear_database(data_scrubbing_database)
--- a/jobserver/jobs/yearly/dump_scrubbed_db.py
+++ b/jobserver/jobs/yearly/dump_scrubbed_db.py
@@ -17,8 +17,10 @@ class Job(YearlyJob):
 
     def execute(self):
         scrubbed_dump_path = settings.SCRUBBED_DATABASE_DUMP_PATH
-        readonly_database = settings.DATABASES.get("readonly")
-        data_scrubbing_database = settings.DATABASES.get("data_scrubbing")
+        readonly_database = settings.DATABASES.get(settings.READONLY_DATABASE_ALIAS)
+        data_scrubbing_database = settings.DATABASES.get(
+            settings.DATA_SCRUBBING_DATABASE_ALIAS
+        )
 
         if readonly_database is None:
             raise JobError("JOBSERVER_READONLY_DATABASE_URL is not set")
@@ -28,14 +30,14 @@ class Job(YearlyJob):
 
         with tempfile.TemporaryDirectory(
             prefix="dump-scrubbed-db-", dir=scrubbed_dump_path.parent
-        ) as temp_dir:
-            temp_dir = pathlib.Path(temp_dir)
-            raw_dump_path = temp_dir / "raw.dump"
-            temp_scrubbed_dump_path = temp_dir / scrubbed_dump_path.name
+        ) as temp_dir_path:
+            temp_dir_path = pathlib.Path(temp_dir_path)
+            raw_dump_path = temp_dir_path / "raw.dump"
+            temp_scrubbed_dump_path = temp_dir_path / scrubbed_dump_path.name
             dump_database(readonly_database, raw_dump_path)
             try:
                 restore_database(data_scrubbing_database, raw_dump_path)
-                scrub_database("data_scrubbing")
+                scrub_database(settings.DATA_SCRUBBING_DATABASE_ALIAS)
                 dump_database(data_scrubbing_database, temp_scrubbed_dump_path)
                 temp_scrubbed_dump_path.replace(scrubbed_dump_path)
             finally:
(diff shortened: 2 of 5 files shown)


jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -43,12 +44,19 @@ def execute(self):
                 logger.info("Scrubbing data scrubbing database")
                 call_command("scrub_data", "data_scrubbing")
                 logger.info("Finished scrubbing database")
-
-                logger.info("Creating scrubbed database dump", path=scrubbed_dump_path)
-                dump_database(data_scrubbing_database, scrubbed_dump_path)
-                logger.info(
-                    "Finished creating scrubbed database dump", path=scrubbed_dump_path
-                )
+                try:
```
mikerkelly: Suggestion: I think that it will be simpler to use one [TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) context manager about the whole block for all the different commands and files. Then you can just point the subprocesses at subpaths of that and the standard library will handle cleanup. Rather than one `NamedTemporaryFile` and one case where you are manually handling cleanup. It will also be more secure and get other benefits from the standard library.

That also reduces the level of nesting which should make it more readable.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -36,6 +40,23 @@ def execute(self):
                 logger.info("Restoring dump into data scrubbing database")
                 restore_database(data_scrubbing_database, raw_dump.name)
                 logger.info("Finished restoring dump into database")
+
+                logger.info("Scrubbing data scrubbing database")
```
mikerkelly: Minor suggestion, taste: there are quite a few log statements now. Maybe it would make it a bit easier to read the control flow of the main function if the log statements were in the subfunctions?

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -14,6 +16,8 @@ class Job(YearlyJob):
     help = "Scrubbed database dump job (in progress)"
 
     def execute(self):
+        scrubbed_dump_path = pathlib.Path("/storage/jobserver_scrubbed.dump")
```
mikerkelly: Configuration: We should not hardcode our production locations into scripts. This may break if production layout changes and will make it harder to run locally (would need to create that directory). Instead we should use our standard mechanism of an environment variable and a Django setting.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -36,6 +40,23 @@ def execute(self):
...
                 restore_database(data_scrubbing_database, raw_dump.name)
                 logger.info("Finished restoring dump into database")
+
+                logger.info("Scrubbing data scrubbing database")
+                call_command("scrub_data", "data_scrubbing")
+                logger.info("Finished scrubbing database")
+                try:
+                    logger.info(
+                        "Creating scrubbed database dump",
+                        path=temp_scrubbed_dump_path.name,
```
mikerkelly: Logging detail: `temp_scrubbed_dump_path` is a `Path`, so `name` represents the final component, not the whole path. I think wherever we use `Paths` we just want to pass the whole variable, or maybe `str(var)` as kwargs to the logger, please check.

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: done

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: done

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: passing the whole variable now.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -23,23 +26,20 @@ def execute(self):
...
-            )
-            dump_database(readonly_database, raw_dump.name)
-            logger.info(
-                "Finished creating raw dump of readonly jobserver database",
-                path=raw_dump.name,
-            )
+        with tempfile.TemporaryDirectory(
+            prefix="dump-scrubbed-db-", dir=scrubbed_dump_path.parent
+        ) as temp_dir:
+            temp_dir = pathlib.Path(temp_dir)
```
mikerkelly: Naming suggestion: `temp_dir_path` to avoid shadowing the variable from the context manager, and for consistency with `raw_dump_path`, `scrubbed_dump_path`, `temp_scrubbed_dump_path`.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -23,23 +26,20 @@ def execute(self):
...
+            raw_dump_path = temp_dir / "raw.dump"
+            temp_scrubbed_dump_path = temp_dir / scrubbed_dump_path.name
+            dump_database(readonly_database, raw_dump_path)
             try:
-                logger.info("Restoring dump into data scrubbing database")
-                restore_database(data_scrubbing_database, raw_dump.name)
-                logger.info("Finished restoring dump into database")
+                restore_database(data_scrubbing_database, raw_dump_path)
+                scrub_database("data_scrubbing")
```
mikerkelly: Suggestion: there are now two instances of "data_scrubbing" in this file, it might be better to use a variable to avoid risk of inconsistency. And both this and "readonly" are hardcoded references to strings in `settings.py`, it might be better to define them as variables in that file both for use in `DATABASES` and for import here.

This is a bit more significant than it would be if we had automated testing of this module. As it is, an error or change could lead to this breaking on the scheduled run, which is not a disaster, but we may as well guard against it.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -87,9 +88,13 @@ def dump_database(database_config, output):
         ],
         database_config,
     )
+    logger.info("Finished creating database dump", path=output)
 
 
 def restore_database(database_config, input_dump):
+    logger.info(
+        "Restoring database dump", path=input_dump, database=database_config["NAME"]
```
mikerkelly: Naming suggestion: maybe `input_dump` should be `dump_file_path` or similar for consistency with other file path variables in this module?

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -76,6 +76,7 @@ def run_database_command(command, database_config):
 
 
 def dump_database(database_config, output):
+    logger.info("Creating database dump", path=output)
```
mikerkelly: Naming suggestion: maybe `output` should be `dump_file_path` or similar for consistency with other file path variables in this module?

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -23,23 +26,20 @@ def execute(self):
...
-            logger.info(
-                "Creating raw dump of readonly jobserver database", path=raw_dump.name
-            )
-            dump_database(readonly_database, raw_dump.name)
-            logger.info(
-                "Finished creating raw dump of readonly jobserver database",
-                path=raw_dump.name,
-            )
+        with tempfile.TemporaryDirectory(
+            prefix="dump-scrubbed-db-", dir=scrubbed_dump_path.parent
```
mikerkelly: I don't think that we need to specify a directory here, we won't do anything with the file while it exists and it should stop existing after the script has ended, even in the event of error. Leaving it as the default may also help the OS to clean it up if something did go wrong and it persisted.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -102,9 +107,21 @@ def restore_database(database_config, input_dump):
...
+    logger.info(
+        "Finished restoring dump into database",
+        path=input_dump,
+        database=database_config["NAME"],
+    )
+
+
+def scrub_database(database_alias):
+    logger.info("Scrubbing database", database=database_alias)
+    call_command("scrub_data", database_alias)
```
mikerkelly: For consistency with `run_database_command` with could catch `CommandError` and reraise as `JobError`. Not sure how much this helps with the management of the job.

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: Updated

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: done

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: updated

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: updated

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: The `scrub_data` command already provides a useful failure message, but I agree we can raise it as a `JobError` to be consistent with `run_database_command`.

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: Makes sense. Removed the dir here.

jobserver/settings.py
```diff
@@ -160,6 +160,8 @@ def get_env_var(name):
 ]
 
 WSGI_APPLICATION = "jobserver.wsgi.application"
+READONLY_DATABASE_ALIAS = "readonly"
+DATA_SCRUBBING_DATABASE_ALIAS = "data_scrubbing"
```
StevenMaude, line 164: thought: These could possibly be a [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum) of allowed database aliases instead, rather than plain strings.

But I think it's also OK as is for the limited uses we have now.

jobserver/jobs/yearly/dump_scrubbed_db.py
```diff
@@ -14,32 +16,30 @@ class Job(YearlyJob):
...
+            temp_scrubbed_dump_path = temp_dir_path / scrubbed_dump_path.name
+            dump_database(readonly_database, raw_dump_path)
             try:
-                logger.info("Restoring dump into data scrubbing database")
-                restore_database(data_scrubbing_database, raw_dump.name)
-                logger.info("Finished restoring dump into database")
+                restore_database(data_scrubbing_database, raw_dump_path)
+                scrub_database(settings.DATA_SCRUBBING_DATABASE_ALIAS)
+                dump_database(data_scrubbing_database, temp_scrubbed_dump_path)
+                temp_scrubbed_dump_path.replace(scrubbed_dump_path)
```
StevenMaude, line 41: question: This failed for me running in job-server in a VS Code dev container.

This is because [`os.replace()`](https://docs.python.org/3/library/os.html#os.replace) is used in the implementation of `Pathlib.replace`, and that:

> …may fail if *src* and *dst* are on different filesystems.

For dev containers, this is true: `/tmp` is *inside* the Docker container, and the repository is bind mounted from the host.

My setup not working for this with the code as written, isn't a big problem. I probably don't need to use it outside of this testing, and can workaround it if I do need it to work.

My question is:

**Are we sure that the two files on the server will always be on the same filesystem?**

If so, that's fine.

***

Note that it's possible to at least avoid the failure in Python 3.14 by using [`Pathlib.move()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.move):

> If both paths are on the same filesystem, the move is performed with [`os.replace()`](https://docs.python.org/3/library/os.html#os.replace). Otherwise, this path is copied (preserving metadata and symlinks) and then deleted.

though the `.move()` operation is not guaranteed to have the same atomicity property as `.replace()`.

(We're still on Python 3.12 here, so that's not something we can even debate right now anyway.)

dotenv-sample
```diff
@@ -9,6 +9,8 @@ DATABASE_URL=postgres://user:pass@localhost:6543/jobserver
 # JOBSERVER_READONLY_DATABASE_URL=postgres://user:pass@localhost:6543/jobserver
 # Database where data will be restored and scrubbed.
 # JOBSERVER_SCRUBBING_DATABASE_URL=postgres://user:pass@localhost:6543/jobserver_scrubbing
+# Path where the scrubbed database dump will be written.
+# JOBSERVER_SCRUBBED_DUMP_PATH=jobserver_scrubbed.dump
```
StevenMaude, line 13: question: Do we need to update some of the other references to `jobserver.dump`? I can see them in:

* `DEVELOPERS.md`
* `docker/justfile`
* `justfile`

Or would we wait until the job becomes daily/hourly as opposed to yearly?

StevenMaude (approved):
I think this is about done now. I had a couple of questions, but I don't think much should be needed to added to this PR.

jobserver/jobs/yearly/dump_scrubbed_db.py
StevenMaude: It may be worth noting this caveat in a comment:

```
# Note: `Pathlib.replace` requires source and destination to be on the same filesystem.
```

nishtha-kalra: That's a good use case; thanks for testing this so thoroughly. I initially specified the directory like:
```
with tempfile.TemporaryDirectory(
            prefix="dump-scrubbed-db-", dir=scrubbed_dump_path.parent
        ) as temp_dir_path:
```
But then removed it after a [suggestion](https://github.com/opensafely-core/job-server/pull/5934#discussion_r3485693786).
I think bringing that back might help?

jobserver/settings.py
nishtha-kalra: That's a good point. Maybe we can turn it into enum in case there are more aliases in future?

dotenv-sample
nishtha-kalra: My plan is to first test this on Dokku after merging this PR. Once it works, I will switch this to a daily job. Then will retire the job that creates jobserver.dump. Will update all the references at that time.

jobserver/jobs/yearly/dump_scrubbed_db.py
nishtha-kalra: added a comment now 😃

mikerkelly: Agree that a comment is sufficient here. Keeping atomicity lets us avoid thinking about other types of error. It was worth thinking about this possibility, but it is pretty remote and it is okay if this job fails temporarily, nothing on the main disk will change and the temporary directory should be cleaned up automatically. And we should have a monitor right this so if this does somehow change in future in production and fail, someone should notice and be able to fix.

PR #5953, opened 2026-07-06
ADR for new jobserver db scrubbing process

nishtha-kalra:
- https://github.com/opensafely-core/security/issues/57
- Added ADR for the new process
- Superseded the old ADR


lucyb (approved):
This looks good, thank you. It's more comprehensive than the first version, which I think reflects your/our understanding of the problem. I particularly like where you've said _why_ we've made particular decisions and what we compared them against. For example the decision to do an daily dump rather than hourly and why we chose not to scrub certain values.

PR #5954, opened 2026-07-06
Add management command for raw database dump

nishtha-kalra:
## Description

https://github.com/opensafely-core/security/issues/53

Add a `dump_raw_data` management command to create raw JobServer database dumps manually.

The command uses the same safer subprocess pattern as the scrubbed dump job: it passes database connection details as `pg_dump` arguments, provides the password through `PGPASSWORD`, and raises `CommandError` with `stdout/stderr` when `pg_dump` fails.

Add a configurable raw dump output path so the production `/storage` location is set through environment configuration rather than hardcoded in the command.

Tip for reviewer - This script is doing the same thing as  [hourly dump job](https://github.com/opensafely-core/job-server/blob/main/jobserver/jobs/hourly/dump_db.py). Instead of a job, it is a management command with some improvements.

## Next steps

This is a step towards replacing the scheduled hourly `dump_db` job. The raw dump command should be used only when there is an approved need for a raw production copy.

Once this PR is merged and scrubbed dump is available on dokku4, I will remove the [hourly dump job](https://github.com/opensafely-core/job-server/blob/main/jobserver/jobs/hourly/dump_db.py) and will ask developers to use this management command whenever a raw dump is required.

### DIFF
--- a/tests/unit/jobserver/management/commands/test_dump_raw_data.py
+++ b/tests/unit/jobserver/management/commands/test_dump_raw_data.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+from types import SimpleNamespace
 
 import pytest
 from django.core.management import call_command
@@ -14,22 +15,29 @@ TEST_DBNAME = "test_dbname"
 
 
 @pytest.fixture
-def default_database(settings):
-    database_config = {
-        "HOST": TEST_HOST,
-        "PORT": TEST_PORT,
-        "USER": TEST_USER,
-        "PASSWORD": TEST_PASSWORD,
-        "NAME": TEST_DBNAME,
-    }
-    settings.DATABASES["default"] = database_config
-    return database_config
+def fake_database_settings(monkeypatch, tmp_path):
+    fake_settings = SimpleNamespace(
+        DATABASES={
+            "default": {
+                "HOST": TEST_HOST,
+                "PORT": TEST_PORT,
+                "USER": TEST_USER,
+                "PASSWORD": TEST_PASSWORD,
+                "NAME": TEST_DBNAME,
+            }
+        },
+        RAW_DATABASE_DUMP_PATH=tmp_path / "jobserver.dump",
+    )
+
+    monkeypatch.setattr(
+        "jobserver.management.commands.dump_raw_data.settings", fake_settings
+    )
+
+    return fake_settings
 
 
-def test_dump_raw_data_command(
-    settings, default_database, monkeypatch, tmp_path, capsys
-):
-    settings.RAW_DATABASE_DUMP_PATH = tmp_path / "jobserver.dump"
+@pytest.fixture
+def fake_subprocess_calls(monkeypatch):
     calls = []
 
     def fake_run(command, env, capture_output, text, check):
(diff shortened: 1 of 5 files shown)


tests/unit/jobserver/management/commands/test_dump_raw_data.py
```diff
@@ -0,0 +1,129 @@
...
+            "default": {
+                "HOST": TEST_HOST,
+                "PORT": TEST_PORT,
+                "USER": TEST_USER,
+                "PASSWORD": TEST_PASSWORD,
+                "NAME": TEST_DBNAME,
+            }
+        },
+        RAW_DATABASE_DUMP_PATH=tmp_path / "jobserver.dump",
+    )
```
StevenMaude: thought: This is OK, but it's a little bit unusual. There is also `override_settings` in Django, which we do use in other tests.

tests/unit/jobserver/management/commands/test_dump_raw_data.py
StevenMaude, line 1: question: All the tests are just checking the commands are given correctly. Is it worthwhile to add a simple end-to-end check that the dump management command works, perhaps?

This isn't necessarily what it should exactly look like, but with some back-and-forth with ChatGPT:

```python
@pytest.mark.slow_test
@pytest.mark.django_db(transaction=True)
def test_command_can_dump_postgres_test_database(tmp_path):
    user = UserFactory()
    user.username = "TestPostgresDumpCommand"
    user.save()

    output_path = tmp_path / "test.dump"

    call_command(
        "dump_raw_data",
        output=str(output_path),
    )

    assert output_path.exists()
    assert output_path.stat().st_size > 0

    result = subprocess.run(
        [
            "pg_restore",
            "--data-only",
            "--table",
            "jobserver_user",
            "--file=-",
            str(output_path),
        ],
        capture_output=True,
        text=True,
        check=True,
    )

    assert "TestPostgresDumpCommand" in result.stdout
```

I'm not sure how much validation is required; at this point, we're testing the behaviour of `pg_dump`, which isn't really our concern.

StevenMaude:
I'll approve as I don't really have any issue with the code itself, just the tests, which seem less critical.

StevenMaude (approved):

tests/unit/jobserver/management/commands/test_dump_raw_data.py
nishtha-kalra: Thanks, indeed, `override_settings` is a better fit for the output path, so I have updated that. It cannot override individual values nested under `DATABASES["default"]`, and overriding the entire `DATABASES` setting produces a Django warning and risks removing required database defaults (I encountered a [failure](https://github.com/opensafely-core/job-server/actions/runs/28932941235/job/85841200633) earlier because of that). I have therefore used `patch.dict` for the database connection values, which preserves the rest of the configuration and restores the original values after the test.

tests/unit/jobserver/management/commands/test_dump_raw_data.py
nishtha-kalra: I was a bit confused about the tests too and was not sure if it was worth adding an integration test. But I think it is good to add it, and your concern is valid. Added an integration test now, similar to the one you suggested.

nishtha-kalra: I added the suggested integration test, but it exposed a PostgreSQL version mismatch in the CI environment.
The CI database service runs PostgreSQL 17, while the `pg_dump` executable on the runner is version 16. 
Since addressing this requires changes beyond the tests, and this PR has already been approved, I have removed the integration-test commit from this branch to keep the PR focused. I will merge this PR and then open a separate PR containing the integration test and the CI fix.

mikerkelly:
I tested this again in production on 25 August 2026. See  https://bennettoxford.slack.com/archives/C069SADHP1Q/p1787649799832629

PR #5955, opened 2026-07-07
Show status message for pending jobs

### LINKED ISSUE #5911: Show job status code/message for pending jobs on Job Request Detail Page
# Why are we doing this?

Users have reported wanting more insight into job state beyond `status` for pending jobs. The job detail page does display the `status_message` but it is hard to get an overall picture when viewing the list of jobs in the job request detail page. The job request detail page [does display the status code/message](https://github.com/opensafely-core/job-server/blob/ef9fbf8efef8aeb9c62f2f5df66ea3f4150ebebd/templates/job_request/detail.html#L179) when the job is failed but there are interesting sub statuses when the job is pending, too (waiting for DB worker, waiting on dependencies...).

Here is [what it looks like](https://jobs.opensafely.org/opensafely-internal/disparities-comparison-test/25110/) for error jobs currently:

Image

Maybe display the same information also for pending jobs in the job request detail page.

[One user report](https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1780491247850409?thread_ts=1779286911.577239&cid=C01D7H9LYKB):

> When jobs are pending it would be nice to have more info on WHY - blocked by dependency, maintenance mode, or simply queued

Jobs have `status` (coarse), `status_code` (fine-grained), `status_message` (human-readable, more-or-less corresponding to `status_message` but sometimes with extra information).

# What are we doing?

Adding an extra conditional block to display status code/message for pending jobs.

---

tomodwyer:
Closes #5911

List of jobs, with pending jobs now showing a status message

### DIFF
--- a/templates/job_request/detail.html
+++ b/templates/job_request/detail.html
@@ -176,12 +176,14 @@
                       {{ job.identifier }}
                     </code>
                   </dd>
-                  {% if job.status == "failed" and job.status_message %}
-                    <dt class="sr-only">Error:</dt>
-                    <dd class="mt-1 bg-red-50 text-red-900 border-red-200 wrap-break-word font-mono text-sm pt-2 pb-1.5 px-3 rounded border">
-                      {% if job.status_code %}<strong>{{ job.status_code }}:</strong>{% endif %}
-                      <span>{{ job.status_message }}</span>
-                    </dd>
+                  {% if job.status == "failed" or job.status == "pending" %}
+                    {% if job.status_message %}
+                      <dt class="sr-only">Status message:</dt>
+                      <dd class="{% if job.status == "failed" %}bg-red-50 text-red-900 border-red-200 {% endif %}{% if job.status == "pending" %}bg-orange-50 text-orange-900 border-orange-200 {% endif %}mt-1 wrap-break-word font-mono text-sm pt-2 pb-1.5 px-3 rounded border">
+                        {% if job.status_code %}<strong>{{ job.status_code }}:</strong>{% endif %}
+                        <span>{{ job.status_message }}</span>
+                      </dd>
+                    {% endif %}
                   {% endif %}
                 </dl>
               </li>


templates/job_request/detail.html
```diff
@@ -176,12 +176,14 @@ <h1 class="text-3xl tracking-tight break-words font-bold text-slate-900 sm:text-
                       {{ job.identifier }}
                     </code>
                   </dd>
-                  {% if job.status == "failed" and job.status_message %}
-                    <dt class="sr-only">Error:</dt>
-                    <dd class="mt-1 bg-red-50 text-red-900 border-red-200 wrap-break-word font-mono text-sm pt-2 pb-1.5 px-3 rounded border">
-                      {% if job.status_code %}<strong>{{ job.status_code }}:</strong>{% endif %}
-                      <span>{{ job.status_message }}</span>
-                    </dd>
+                  {% if job.status == "failed" or job.status == "pending" %}
```
nishtha-kalra, line 179: Maybe we could write line 179-180 as:
`{% if job.status_message and job.status in "failed,pending" %}`

nishtha-kalra (approved):
Looks good, just a small suggestion.
Also, not sure if it is worth adding a test for this.

templates/job_request/detail.html
tomodwyer: TIL I did not know we could write that in Django templates! My only concern is that it checks substrings, but I guess we will always have valid full strings as `job.status` - so maybe not worth worrying?

nishtha-kalra: That's a good point. For clarity and removing this confusion I think what you did is the best way. I had another solution in mind like:
`job.status_message and (job.status == "failed" or job.status == "pending")`
but checked that templates don't support that.

tomodwyer: Yeah that was the first thing I tried ha, I just assumed brackets would work.

I'll leave it as a double if statement for now, if we add in more, I think we should probably do the check in the view and just pass it to the template.

PR #5960, opened 2026-07-08
Unpin pytest-socket

### LINKED ISSUE #5888: Update and unpin `pytest-socket`
# Why are we doing this?

We pinned `pytest-socket` in #5884 as tests were failing with the upgrade to v0.8.0. This is a new upgrade, out after a long period from the previous release that we're using.

This is a development-only dependency.

When the package is unpinned in `pyproject.toml`.

# What are we doing?

* Reviewing whether `--disable-socket --allow-hosts=localhost` is a solution.
* Reviewing whether we want to do that (is that in line with how we expect the disabling of sockets to work?)
* Making that change if so, or finding an alternative solution if not.

StevenMaude:
Fixes #5888.

`pytest-socket` was [introduced as a replacement](https://github.com/opensafely-core/job-server/pull/5063) for the unmaintained `pytest-network`.

There's no specific commit message explaining why `pytest-network` was added in 1b8542a3f9dd154de796933fb3e3541cd8af0ca4 (as part of #408).

The only context I can find for the use of `pytest-network` (which we replaced with `pytest-socket` in #5063) is from George in response to being asked [what is being mocked in job-server tests](https://bennettoxford.slack.com/archives/C63UXGB8E/p1619794117190700?thread_ts=1619793981.189600&cid=C63UXGB8E):

> anything with external network access, there are a few views
> where we have to hit the GitHub API for data
> I've even gone as far as adding pytest-network to throw an error
> if there's any network access in the tests

Whether we still want to enforce this is another question, but, in this context, allowing localhost seems OK here, since that will still restrict external access.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -170,7 +170,7 @@ skip_covered = true
 show_contexts = true
 
 [tool.pytest.ini_options]
-addopts = "--disable-socket --tb=native --maxprocesses=6 --strict-markers"
+addopts = "--disable-socket --allow-hosts=localhost --tb=native --maxprocesses=6 --strict-markers"
 DJANGO_SETTINGS_MODULE = "jobserver.settings"
 env = [
   "JOBSERVER_GITHUB_TOKEN=empty",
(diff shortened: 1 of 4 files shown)


StevenMaude:
It's probably worth just confirming that `just test` works locally for someone else.

nishtha-kalra (approved):

PR #5974, opened 2026-07-15
Avoid race condition in `db` service

iaindillingham:
Three small quality of life improvements to the operation of the `db` service. Essentially, these avoid a race condition that humans are unlikely to experience (you're too slow, human!), but computers aren't, if you'll excuse the double negative.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -20,6 +20,8 @@ services:
       interval: 10s
       timeout: 10s
       retries: 5
+      start_period: 40s
+      start_interval: 5s
 
   # base service, exists to hold common config, but is not actually used directly
   base:
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -16,7 +16,7 @@ services:
     volumes:
       - postgres_data:/var/lib/postgresql/data/
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready"]
+      test: ["CMD-SHELL", "pg_isready; exit $(( $? != 0 ))"]
       interval: 10s
       timeout: 10s
       retries: 5
--- a/docker/justfile
+++ b/docker/justfile
@@ -58,7 +58,7 @@ smoke-test host="http://localhost:8000": _dotenv
 
 # manually run the docker db. Useful if you want to use it but with venv based dev env
 db: build
-    docker compose up -d db
+    docker compose up -d --wait db
 
 # if the database container exists and is running, stop it from running
 stop-db:


docker/docker-compose.yaml
```diff
@@ -20,6 +20,8 @@ services:
       interval: 10s
       timeout: 10s
       retries: 5
+      start_period: 40s
+      start_interval: 5s
```
mikerkelly, line 24: Taste: I think on balance I agree explicit is better than implicit in this case, as we don't manage many of these, and probably most of us working on them are not Docker experts so being as explicit as possible about all involved parameters is probably helpful.

Correctness: [reference/compose-file/services/#healthcheck] shows a configuration with these values chosen but those are not the default values. [dockerfile/#healthcheck] states the defaults: 

```
--start-period=DURATION (default: 0s)
--start-interval=DURATION (default: 5s)
```

This commit would change the start period from 0s to 40s, adding delay that is not needed. I don't think that we should do that without reason. We should avoid changing the behaviour instead.

Instead can we specify 0s/5s to be explicit without changing the behaviour?

Optional, documentation: It could also be helpful to add a comment above the healthcheck section linking to [dockerfile/#healthcheck] where the interpretation of those values is defined, and a separate one noting that the last two values are the default. This foregrounds useful information about the configuration without being overly visually busy, and avoids some need for git archaeology. Presumably the other values were set non-default for some good reason but let's not worry about those for now I guess.

[reference/compose-file/services/#healthcheck]: https://docs.docker.com/reference/compose-file/services/#healthcheck
[reference/dockerfile/#healthcheck]: https://docs.docker.com/reference/dockerfile/#healthcheck
[dockerfile/#healthcheck]: https://docs.docker.com/reference/dockerfile/#healthcheck

docker/justfile
```diff
@@ -58,7 +58,7 @@ smoke-test host="http://localhost:8000": _dotenv
 
 # manually run the docker db. Useful if you want to use it but with venv based dev env
 db: build
-    docker compose up -d db
+    docker compose up -d --wait db
```
mikerkelly, line 61: Thought, ignorable: `--wait` implies `-d` so `-d` is now technically redundant. But I think on balance we may as well leave it in for clarity and to avoid future errors if the `--wait` is removed again for some reason without readding the `-d`.

[reference/cli/docker/compose/up/](https://docs.docker.com/reference/cli/docker/compose/up/) says:

> --wait		Wait for services to be running|healthy. Implies detached mode.

mikerkelly: Suggestion, optional: We could add a timeout to potentially avoid this hanging forever if `db` does not return. Maybe this is not that important if it is just running in a developer's terminal. Maybe it will sometimes / someday run in CI or production though?

[reference/cli/docker/compose/up/](https://docs.docker.com/reference/cli/docker/compose/up/) says:

> --wait		Wait for services to be running|healthy. Implies detached mode.
> --wait-timeout		Maximum duration in seconds to wait for the project to be running|healthy

Maybe 30s?

mikerkelly (commented):
Oof, good spot, thanks. Nice PR. Appreciate the detailed commit messages including references. Very easy to follow. Couple of queries/suggestions in comments.

If we are feeling superabundantly helpful we could notify the maintainer of https://github.com/peter-evans/docker-compose-healthcheck that https://github.com/opensafely-core/job-server/commit/4e81b33b2aeee49dd94be2f99c7028f8a02bc71e cites as the source for our approach and let them know about the issue and our proposed fix, maybe by raising an issue on that repository. That also higher chance of others seeing it if the maintainer does not fix it.

docker/docker-compose.yaml
iaindillingham: Thanks for highlighting that these values aren't the default values. The default value of `start_period` (`0s`) means that, in effect, the start period doesn't exist. This means that `start_interval` is redundant, and any failures will decrement `retries`.

That's probably why these values were implicit.

Making these values implicit was confusing. Making these values explicit is differently confusing.

Help me help you Mike. I've used my daily allocation of brain tokens. Would you like me to use default values or non-default values?

mikerkelly: I think we should leave values as they were, at the defaults, unless there is some reason to change them. Adding them here was supposed to be an explicitness-enhancement, not a material change in what this configuration actually does. The question is do we make those default values explicit or not.

I think this discussion has convinced me that not listing these default values is the wisest way to make this readable and understandable. It is not as explicit as possible, but sparse is better than dense, and readability counts. If we list them we probably need a comment to explain that they are not doing anything in practice, which probably hinders more than helps when someone wants to understand what is distinctive and special in this configuration. I think if there were 20 default values unchanged we'd probably leave them out unless they were relevant, as they would just get in the way. I think the same logic applies here, but was harder to reason about than in a more extreme case.

So let's drop this commit, if you don't mind.

docker/justfile
mikerkelly: :bump:

PR #5975, opened 2026-07-15
Add more test factories 

mikerkelly:

I am all about that `factoryboy`.

This test was using a mixture of factories and direct models for instantiation. Now the reader does not have to wonder why.

It also makes the test read more simply by not showing the arbitrary test values chosen -- all that matters for the test is that we have instantiated the model, not what values its attributes have.

This will make any future tests that need those models simpler, more consistent and reliable in future. 
    
It also removes some unnecessary imports from the test.

Started organizing factories for third party packages in subdirectories of `factories`. That will distinguish them clearly from the ones that we define, avoiding some possible confusion, clutter, and clashes.

Also some very quick refactors on the test to make it more consistent.


tests/factories/django/contrib/sessions/__init__.py
```diff
@@ -0,0 +1,2 @@
+"""Contains factories from django.contrib.sessions.models so that they are
+eparated from the factories for models we define."""
```
tomodwyer:
```suggestion
separated from the factories for models we define."""
```

tomodwyer (approved):
One typo otherwise all good. Going commit by commit this was super easy to follow.

> I am all about that factoryboy.

😆

PR #5976, opened 2026-07-16
Remove use of `faker` dependency for data scrubbing

mikerkelly:

Using a generator that we define for unique values and hardcoded values otherwise is simpler and has better uniqueness and is probably faster (not tested). So we don't need this at all in data scrubbing. Factoryboy still depends on it, so we cannot remove it altogether.

We don't actually need "realistic" fake values in the scrubbed data for
applications. The e-mail addresses used here are not required to be unique.
This application is also not under active development, it is just not that
important if all the values are the same, and we can use a generator to make
them if that turns out to be important.

Example fake e-mail generated: `260716073644_1@example.com`.

The e-mail generator embeds the start time of the generator and an loop counter so should be unique within-process and between consecutive runs triggered by humans, but is not globally unique. This should be sufficient to avoid hitting uniqueness constraints in the data scrubbing process.

Before this change, if you ran data scrubbing repeatedly on the same database,
as in manual testing, sometimes it would generate the same "unique" addresses
between runs, failing the uniqueness constraint. This can happen more often
than one might expect due to the ["birthday problem"](https://en.wikipedia.org/wiki/Birthday_problem).


nishtha-kalra (approved):
Looks good, thanks!

PR #5977, opened 2026-07-16
Document `DataScrubbing` and how to use it when changing models and fields

mikerkelly:

Documentation will help those changing this module and when deciding how to scrub new models and fields.

### DIFF
--- a/data_scrubbing/management/commands/scrub_data.py
+++ b/data_scrubbing/management/commands/scrub_data.py
@@ -1,3 +1,171 @@
+"""
+Management command to scrub the database of sensitive data from selected fields.
+
+See data scrubbing related ADRs for an overall description of the design of our
+data scrubbing capability. See DEVELOPERS.md for what developers need to know to
+obtain scrubbed and raw production database dumps for use in local development.
+See the automated tooling that calls this module as a management command to
+understand how this is used in production to generate the scrubbed dump.
+
+You may need to understand this module if you want to change how data scrubbing
+works.
+
+You may also be reading this because you are adding a new model or model field.
+In that case you need to add or update the DataScrubbing nested class on your
+model, as explained below.
+
+Data Scrubbing Configuration
+============================
+
+We identify which fields do and do not need scrubbing through the use of a
+special nested class attached to each Django model in our applications. This
+exhaustively lists all of the fields in that class and categorises them as
+needing scrubbing or allowed to be left unchanged. This is exhaustive to help
+ensure that we think about every field and make an active decision to allow it
+or scrub it. An example class looks like this:
+
+
+    class DataScrubbing:
+        fields_to_scrub = {
+            "email": fake_unique_email,
+            "user_name": "Fake user name",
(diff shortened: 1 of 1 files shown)


data_scrubbing/management/commands/scrub_data.py
```diff
@@ -1,3 +1,163 @@
...
+respectively. It appears after field definitions, manager definitions, and
+other "declarative" lines and just before the Meta nested class, or just before
+any model methods if there is no Meta.
+
+Classifying fields
+------------------
+
+The keys of fields_to_scrub correspond to the names of the model fields that
+contain sensitive values. The values associated define how to scrub that field.
+If the value is a generator object or a callable (such as a function)then it
```
tomodwyer:
```suggestion
If the value is a generator object or a callable (such as a function) then it
```

data_scrubbing/management/commands/scrub_data.py
```diff
@@ -1,3 +1,163 @@
...
+The primary key ID in Job Server models should always be defined as an opaque
+integer so should be allowed. Timestamps such as created_at are usually okay.
+
+Personal Data
+-------------
+
+Personal Data corresponds more or less to the definitions from UK GDPR and Data
+Protection Acts, which are binding on us when we process Personal Data; we must
+limit data processing to that which is necessary for the purposes we are
+processing it for (UK GDPR Article 5(1)(c). So when we create a dump for use
```
tomodwyer:
```suggestion
processing it for (UK GDPR Article 5(1)(c)). So when we create a dump for use
```

data_scrubbing/management/commands/scrub_data.py
```diff
@@ -1,3 +1,163 @@
...
+values are assigned to the model instances and saved to the database.
+
+The keys of allowed_fields correspond to the other fields of the model that do
+not need changing during data scrubbing because they do not contain sensitive
+data.
+
+You do need to include ForeignKeys and ManyToManyFields that your model has to
+other models. You do not need to include reverse relations that other models
+have to this one. That is because the underlying database table only has a
+column for the ForeignKey on the side that declared it, not in its target.
```
tomodwyer: I found this slightly tricky to parse, maybe something like this (assuming I've understood it correctly)?

```suggestion
You do need to include ForeignKeys. That is because the underlying database
table only has a column for the ForeignKey on the side that declared it,
not in its target. You do not need to include ManyToManyFields or reverse
relations, because they are not concrete fields stored in the model’s
database table.
```

mikerkelly: Thanks, that is clearer.

mikerkelly: I have tried splitting up further into two paragraphs, can you see if this is any good?

tomodwyer: Works for me, thanks

tomodwyer (approved):

PR #5978, opened 2026-07-16
Test `APPLICATIONS_TO_SCRUB` exhaustiveness

mikerkelly:

Test that all configured applications that have models are listed in either APPLICATIONS_TO_SCRUB or specifically allowed as hardcoded in this test.

Exceptions should be truncated in scrub_data, or verifed as not containing sensitive data.

This is intended to catch the case where a whole new app with models is added in future, either through a new third-party package, or an app that we define. In that case, we should decide how to scrub it, either through truncation or DataScrubbing.

### DIFF
--- a/tests/unit/test_data_scrubbing.py
+++ b/tests/unit/test_data_scrubbing.py
@@ -1,0 +1,45 @@
+from django.apps import apps
+
+from data_scrubbing.management.commands.scrub_data import APPLICATIONS_TO_SCRUB
+
+
+def test_applications_to_scrub():
+    """Test that all configured applications that have models are listed in
+    either APPLICATIONS_TO_SCRUB or specifically allowed as hardcoded in this
+    test.
+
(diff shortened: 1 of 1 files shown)


tests/unit/test_data_scrubbing.py
```diff
@@ -0,0 +1,45 @@
...
+
+from data_scrubbing.management.commands.scrub_data import APPLICATIONS_TO_SCRUB
+
+
+def test_applications_to_scrub():
+    """Test that all configured applications that have models are listed in
+    either APPLICATIONS_TO_SCRUB or specifically allowed as hardcoded in this
+    test.
+
+    Exceptions should be truncated in scrub_data, or verifed as not containing
```
tomodwyer:
```suggestion
    Exceptions should be truncated in scrub_data, or verified as not containing
```

tomodwyer (approved):

PR #5979, opened 2026-07-16
Document why OrgMembership is unscrubbed Personal Data

mikerkelly:

We should document the rationale for any unscrubbed Personal Data that we will
process locally for development purposes as this is a compliance requirement.

Slack thread confirming that decision:
https://bennettoxford.slack.com/archives/C069SADHP1Q/p1784203993625829

### DIFF
--- a/jobserver/models/org_membership.py
+++ b/jobserver/models/org_membership.py
@@ -30,6 +30,22 @@ class OrgMembership(models.Model):
 
     class DataScrubbing:
         fields_to_scrub = {}
+        # Institutional affiliation as represented by the link between an Org
+        # and a User is Personal Data. User.username is also Personal Data and
+        # is not scrubbed, so it can identify the user associated with a
+        # membership. However, usernames are not especially sensitive because
+        # they are already public on the production Job Server and probably
+        # elsewhere as well.
+        #
+        # We could remove all OrgMembership data or permute the relations, but
+        # as it is relational data this is more difficult, and this would make
+        # some kinds of testing harder. We could try to pseudonymise this data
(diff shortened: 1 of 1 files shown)


jobserver/models/org_membership.py
```diff
@@ -30,6 +30,21 @@ class OrgMembership(models.Model):
 
     class DataScrubbing:
         fields_to_scrub = {}
+        # Institutional affiliation as represented by the link between an Org
+        # and a User is Personal Data. User.username is not scrubbed, is
+        # Personal Data, and can be linked to this. However, it is not very
+        # sensitive Personal Data. It is public information on the production
+        # Job Server anyhow and probably in other places as well.
```
tomodwyer:
```suggestion
        # Institutional affiliation as represented by the link between an Org
        # and a User is Personal Data. User.username is also Personal Data and
        # is not scrubbed, so it can identify the user associated with a
        # membership. However, usernames are not especially sensitive because
        # they are already public on the production Job Server and probably
        # elsewhere as well.
```

I tried to rewrite this a little bit to remove the "this" and tweak some of the structure - I think it's clearer, but also happy to go with what you have.

jobserver/models/org_membership.py
```diff
@@ -30,6 +30,21 @@ class OrgMembership(models.Model):
...
+        #
+        # We could remove all OrgMembership Data or permute the relations, but
```
tomodwyer:
```suggestion
        # We could remove all OrgMembership data or permute the relations, but
```

tomodwyer (approved):
Two suggestions, but neither are required

PR #5986, opened 2026-07-20
Remove user name initials generation and display

### LINKED ISSUE #4724: IndexError: string index out of range
Sentry Issue: [JOB-SERVER-MF](https://ebm-datalab.sentry.io/issues/6041615623/?referrer=github_integration)

```
TypeError: 'User' object is not subscriptable
  File "django/template/base.py", line 883, in _resolve_lookup
    current = current[bit]
  File "django/utils/functional.py", line 253, in inner
    return func(_wrapped, *args)

IndexError: string index out of range
(24 additional frame(s) were not displayed)
...
  File "jobserver/models/user.py", line 218, in initials
    return "".join(w[0].upper() for w in self.name.split(" "))
  File "jobserver/models/user.py", line 218, in <genexpr>
    return "".join(w[0].upper() for w in self.name.split(" "))
```

tomodwyer:
Closes #4724

We generate initials to use for authenticated users on the home page in lieu of having avatars. Instead of trying to debug every last way the initials might break our site or design, lets just change the design to remove the need for that.

### DIFF
--- a/templates/index-authenticated.html
+++ b/templates/index-authenticated.html
@@ -13,7 +13,11 @@
     <div class="bg-white shadow-sm">
       <div class="grid gap-y-2 text-center text-slate-700 py-6 px-8">
         <h1 class="text-xl/6 font-semibold text-slate-900">Your OpenSAFELY Dashboard</h1>
-        <p class="text-slate-600 text-sm/tight">Signed in as: <strong class="text-slate-900">{{ user.fullname }}</strong></p>
+        <p class="text-slate-600 text-sm/tight">
+          Logged in as:
+          <strong class="text-slate-900">{{ user.fullname }} </strong>
+          ({{ user.username }} via GitHub)
+        </p>
         <p>Continue your research, view job requests, and see recent workspace activity.</p>
       </div>
 
--- a/templates/index-authenticated.html
+++ b/templates/index-authenticated.html
@@ -9,52 +9,27 @@
 {% endblock %}
 
 {% block content %}
-  {% #card container=True class="my-6 sm:mb-12 md:mt-0" %}
-    <h2 class="sr-only" id="profile-overview-title">Profile Overview</h2>
-    <div class="flex items-center justify-center sm:justify-between">
-      <div class="sm:flex sm:space-x-5 sm:items-center">
-        <div class="shrink-0">
-          <div
-            class="flex items-center justify-center w-16 h-16 bg-oxford-100 aspect-square rounded-2xl border border-oxford-500/25 overflow-hidden shrink-0 mx-auto sm:mx-0"
-          >
-            <span class="py-2 text-oxford-600 text-2xl font-bold" role="presentation">
-              {{ user.initials }}
-            </span>
-          </div>
-        </div>
-        <div class="mt-4 text-center sm:mt-0 sm:text-left">
-          <p class="text-sm font-medium text-slate-600">Welcome back,</p>
-          <p class="text-xl font-bold text-slate-900 sm:text-2xl">{{ user.fullname }}</p>
-        </div>
+  <section class="my-6 sm:mb-12 md:mt-0">
+    <div class="bg-white shadow-sm">
+      <div class="grid gap-y-2 text-center text-slate-700 py-6 px-8">
+        <h1 class="text-xl/6 font-semibold text-slate-900">Your OpenSAFELY Dashboard</h1>
+        <p class="text-slate-600 text-sm/tight">Signed in as: <strong class="text-slate-900">{{ user.fullname }}</strong></p>
+        <p>Continue your research, view job requests, and see recent workspace activity.</p>
       </div>
-    </div>
 
-    {% #card_footer class="py-0! px-0! -mb-3! md:-mb-5!" %}
-      <ul class="grid grid-cols-1 divide-y divide-slate-200 border-t border-slate-200 bg-slate-50 sm:grid-cols-3 sm:divide-y-0 sm:divide-x">
-        <li class="px-6 py-5 text-center text-sm font-medium text-slate-600">
-          <span class="text-slate-900">{{ counts.job_requests }}</span>
-          <span class="text-slate-600">job request{{ counts.job_requests|pluralize }}</span>
+      <ul class="grid divide-y divide-slate-200 border-t border-slate-200 bg-slate-50 sm:grid-cols-3 sm:divide-y-0 sm:divide-x">
+        <li class="px-6 py-4 text-center text-base/tight font-medium text-slate-600">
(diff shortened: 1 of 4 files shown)


tests/unit/jobserver/views/test_index.py
```diff
@@ -88,14 +88,16 @@ def test_index_authenticated_display_active_backend_job_requests(
...
 @pytest.mark.slow_test
 def test_index_authenticated_client(client, django_assert_num_queries):
-    user = UserFactory()
+    user = UserFactory(fullname="Test User")
     JobRequestFactory.create_batch(10)
 
     with django_assert_num_queries(41):
         client.force_login(user)
         response = client.get("/")
         assert response.status_code == 200
```
mikerkelly: Test bug: this is causing the `django_assert_num_queries` to fail as this introduces two new queries for SiteAlert:

```just test-ci -n0 --pdb -k test_index_authenticated_client -v

....

SELECT "jobserver_sitealert"."id", "jobserver_sitealert"."title", "jobserver_sitealert"."message", "jobserver_sitealert"."level", "jobserver_sitealert"."created_at", "jobserver_sitealert"."updated_at", "jobserver_sitealert"."created_by_id", "jobserver_sitealert"."updated_by_id" FROM "jobserver_sitealert" ORDER BY "jobserver_sitealert"."created_at" DESC

SELECT "jobserver_sitealert"."id", "jobserver_sitealert"."title", "jobserver_sitealert"."message", "jobserver_sitealert"."level", "jobserver_sitealert"."created_at", "jobserver_sitealert"."updated_at", "jobserver_sitealert"."created_by_id", "jobserver_sitealert"."updated_by_id" FROM "jobserver_sitealert" ORDER BY "jobserver_sitealert"."created_at" DESC

SELECT "jobserver_sitealert"."id", "jobserver_sitealert"."title", "jobserver_sitealert"."message", "jobserver_sitealert"."level", "jobserver_sitealert"."created_at", "jobserver_sitealert"."updated_at", "jobserver_sitealert"."created_by_id", "jobserver_sitealert"."updated_by_id" FROM "jobserver_sitealert" ORDER BY "jobserver_sitealert"."created_at" DESC
```

This does not happen in django-debug-toolbar locally on this branch on the main page, so just a test issue.

It must be that `response.rendered_content` is an `@property` method or similar that calls a function when accessed. So referencing it 3 times seems to be rendering the page 3 separate times. The following avoids this:

```python
        content = response.rendered_content
        assert "OpenSAFELY Jobs" in content
        assert "Signed in as:" in content
        assert user.fullname in content
```

I think this was missed before opening the PR because the test is marked as `slow_test` as it is using `client` and doing full template rendering. Using `just test-ci` can avoid this when developing such tests, or you can temporarily remove the mark to run the tests with `just test` locally. Or use `pytest` directly.

tests/unit/jobserver/views/test_index.py
```diff
@@ -88,14 +88,16 @@ def test_index_authenticated_display_active_backend_job_requests(
 
 @pytest.mark.slow_test
 def test_index_authenticated_client(client, django_assert_num_queries):
-    user = UserFactory()
+    user = UserFactory(fullname="Test User")
```
mikerkelly: Setting an explicit `fullname` here is not required as the `UserFactory` already supplies a `fullname` that we can assert on, derived from a sequence (so unique within one test). I think we should not set unnecessary things as that makes it harder to understand the point of the test and what is actually important about it.

templates/index-authenticated.html
```diff
@@ -9,52 +9,27 @@
...
-        </div>
-        <div class="mt-4 text-center sm:mt-0 sm:text-left">
-          <p class="text-sm font-medium text-slate-600">Welcome back,</p>
-          <p class="text-xl font-bold text-slate-900 sm:text-2xl">{{ user.fullname }}</p>
-        </div>
+  <section class="my-6 sm:mb-12 md:mt-0">
+    <div class="bg-white shadow-sm">
+      <div class="grid gap-y-2 text-center text-slate-700 py-6 px-8">
+        <h1 class="text-xl/6 font-semibold text-slate-900">Your OpenSAFELY Dashboard</h1>
+        <p class="text-slate-600 text-sm/tight">Signed in as: <strong class="text-slate-900">{{ user.fullname }}</strong></p>
```
mikerkelly: Suggestion: we could display "`fullname` (`username`)" here, as some users have more than one account, and this could make it clearer which account they are logged in with.

In fact, in the hamburger menu in the banner we use "Logged in as: `username` via GitHub", so probably we should be consistent with that. We talk about log-in, not sign-in, in the templates and URLs it seems. And we may as well contextualise it as being a GitHub username in case they forget/aren't familiiar?

```python
Logged in as: {{user.fullname}} ({{user.username}} via GitHub)
```

How about something like that?

mikerkelly: I suppose we don't want to mess around with their GitHub profile picture unless we have to.

mikerkelly:
```
mike@mike-Latitude-7450:~/job-server$ rg "log.in" jobserver staff/ templates/
jobserver/settings.py
22:from services.logging import logging_config_dict
272:# https://docs.djangoproject.com/en/5.2/topics/logging/
273:LOGGING = logging_config_dict

templates/job_request/detail.html
141:                  <li>Can be viewed by an approved researcher by logging into a highly secure environment</li>

jobserver/views/errors.py
66:            - you need to log in to view this page
68:            Please check the URL, try logging in, or contact us if you need further help.
103:            - you need to log in to view this page
106:            Try logging in, or contact us if you need to discuss permissions.

templates/emails/login.txt
4:We have received a request to log into OpenSAFELY for your account. If you think this was made in error please ignore this email.
6:Otherwise, please click the link below to log in, it is valid for {{ timeout_minutes }} minutes:

templates/emails/login.html
4:  <p>We have received a request to log into OpenSAFELY for your account. If you think this was made in error please ignore this email.</p>
6:  <p>Otherwise, please <a href="{{ url }}">click to log in.</a></p>

jobserver/actions/users.py
49:    # only github users can log in with token

templates/settings.html
78:        <p>You do not have access to log in to any OpenSAFELY backends</p>
```

```
mike@mike-Latitude-7450:~/job-server$ rg "sign.in"
tests/unit/jobserver/test_urls.py
73:            "/apply/sign-in",
```

mikerkelly (approved):
Nice. Full support for solving annoying code that does not achieve much by removing it. Page looks nicer now I feel.

CI is failing due to a test bug I think that I have diagnosed. And a couple of small comments, approving subject to your decision on these.

tests/unit/jobserver/views/test_index.py
tomodwyer: Thanks, I need to do a deep dive in to the test factories at some point to get a better understanding. I'll update that.

PR #5984, opened 2026-07-20
Send Slack when running `dump_raw_data` command

mikerkelly:
Fixes https://github.com/opensafely-core/security/issues/62.

There is a process to follow when using the `dump_raw_data`management command. This is documented in the module docstring and command help, but those could be missed. Let's add a Slack alert about it, as another way to remind developers to follow the process.

Note that these Slack messages ought to be reconcilable with the related log. But they do not constitute a comprehensive audit log of database dumps as they rely on manual action to update the log. There is nothing that guarantees that developers follow the process. Developers might act unilaterally or not create a record. More significantly, developers have access to create database dumps without running the `dump_raw_data` command by running the relevant tools directly. This Slack message is just a reminder hint.

Generated Slack messages look like this:

image

```
Someone ran `dump_raw_data` to generate a raw Job Server database dump. The <https://bennett.wiki/tech-group/policies/personal-data-copying-policy/|personal data copying policy> must be followed. The copying of raw personal data must be recorded in <https://docs.google.com/spreadsheets/d/1C1z3WV-WSL-H1keZZCVPm6hR_5ajjgRT5zGO5cLv2Aw|personal data copying decision log>. The raw dump must be removed from the server and developer machines as soon as it is no longer required.
```

PR #5987, opened 2026-07-20
Delete raw database dumps after their retention period (1 hour)

nishtha-kalra:
Closes https://github.com/opensafely-core/security/issues/56
Add an hourly job that checks the configured raw database dump path and deletes the dump once it has been available for at least one hour.

Use the file's modification time so newly created dumps remain available long enough to be downloaded while limiting how long production data is retained on the server.

Log whether the dump is missing, still within its retention period, or deleted.

I manually tested the job locally by renaming `jobserver_scrubbed.dump` to `jobserver.dump`, so no raw production data was involved. I manually tested the cases where the dump was older than the retention period, newer than the retention period, and did not exist. Each case behaved as expected.

### DIFF
--- a/jobserver/jobs/hourly/delete_raw_dump.py
+++ b/jobserver/jobs/hourly/delete_raw_dump.py
@@ -1,0 +1,72 @@
+"""
+Hourly job to delete raw JobServer database dumps after their retention period.
+
+Raw database dumps may contain production personal data and other sensitive
+information. They are created manually using the dump_raw_data management
+command when access has been agreed and recorded in accordance with the
+Personal Data Copying Policy.
+
+The raw dump needs to remain available for long enough for an authorised
+developer to download it, but it should not remain on the production server
+indefinitely. This job limits that exposure by deleting the dump once it has
+existed for at least one hour.
+
+The age of the dump is determined from its modification time. Checking its age,
+rather than deleting any dump that exists, prevents a newly created dump from
+being deleted by an hourly run before it can be downloaded.
+
+Because this job runs once an hour, a dump will normally be deleted between one
+and two hours after it is created. A missing dump is expected and is not treated
+as an error.
+"""
+
+from datetime import UTC, datetime, timedelta
+
+import structlog
+from django.conf import settings
+from django_extensions.management.jobs import HourlyJob, JobError
+from sentry_sdk.crons.decorator import monitor
+
+from services.sentry import monitor_config
+
+
+logger = structlog.get_logger(__name__)
+
+RAW_DUMP_RETENTION_PERIOD = timedelta(hours=1)
+"""Minimum period for which a raw dump remains available before deletion."""
+
+
+class Job(HourlyJob):
+    help = "Delete raw database dumps after the retention period to limit the availability of production data"
+
+    @monitor(monitor_slug="delete_raw_dump", monitor_config=monitor_config("0 * * * *"))
+    def execute(self):
+        raw_dump_path = settings.RAW_DATABASE_DUMP_PATH
+
+        if not raw_dump_path.exists():
+            logger.info("Raw database dump does not exist", path=raw_dump_path)
+            return
+
+        if not raw_dump_path.is_file():
+            raise JobError(f"Raw database dump path is not a file: {raw_dump_path}")
+
+        # Use the file's last-modified time as an approximation of when the dump
+        # finished being written, and convert it to UTC datetime.
+        modified_at = datetime.fromtimestamp(raw_dump_path.stat().st_mtime, tz=UTC)
+        # Calculate how long the raw dump has been available since it was last modified.
+        dump_age = datetime.now(UTC) - modified_at
+
+        if dump_age < RAW_DUMP_RETENTION_PERIOD:
+            logger.info(
+                "Raw database dump is still within its retention period",
+                age_seconds=dump_age.total_seconds(),
+                path=raw_dump_path,
+            )
+            return
+
+        raw_dump_path.unlink(missing_ok=True)
+        logger.info(
+            "Deleted raw database dump after its retention period",
+            age_seconds=dump_age.total_seconds(),
+            path=raw_dump_path,
+        )
--- a/jobserver/jobs/quarter_hourly/delete_raw_dump.py
+++ b/jobserver/jobs/quarter_hourly/delete_raw_dump.py
@@ -22,7 +22,7 @@ A raw database dump will not exist most of the time, so the dump not existing is
 error.
 """
 
-from datetime import UTC, datetime, timedelta
+import time
 
 import structlog
 from django.conf import settings
@@ -34,7 +34,7 @@ from services.sentry import monitor_config
 
 logger = structlog.get_logger(__name__)
 
-RAW_DUMP_AGE_LIMIT = timedelta(minutes=15)
+RAW_DUMP_AGE_LIMIT = 15 * 60
 """Delete files older than this when the job runs."""
 
 
@@ -57,17 +57,14 @@ class Job(QuarterHourlyJob):
         if not raw_dump_path.is_file():
             raise JobError(f"Raw database dump path is not a file: {raw_dump_path}")
 
-        # Use the file's last-modified time as an approximation of when the dump
-        # finished being written, and convert it to UTC datetime.
-        modified_at = datetime.fromtimestamp(raw_dump_path.stat().st_mtime, tz=UTC)
         # Calculate how long the raw dump has been available since it was last modified.
-        dump_age = datetime.now(UTC) - modified_at
+        dump_age = time.time() - raw_dump_path.stat().st_mtime
 
         if dump_age < RAW_DUMP_AGE_LIMIT:
             logger.info(
                 "Raw database dump not deleted as it had not reached deletion age",
-                age_seconds=dump_age.total_seconds(),
-                deletion_age=RAW_DUMP_AGE_LIMIT.total_seconds(),
+                age_seconds=dump_age,
+                deletion_age=RAW_DUMP_AGE_LIMIT,
                 path=raw_dump_path,
             )
             return
@@ -75,7 +72,7 @@ class Job(QuarterHourlyJob):
         raw_dump_path.unlink(missing_ok=True)
         logger.info(
             "Deleted raw database dump as it had reached deletion age",
-            age_seconds=dump_age.total_seconds(),
-            deletion_age=RAW_DUMP_AGE_LIMIT.total_seconds(),
+            age_seconds=dump_age,
+            deletion_age=RAW_DUMP_AGE_LIMIT,
             path=raw_dump_path,
         )
--- a/jobserver/jobs/quarter_hourly/delete_raw_dump.py
+++ b/jobserver/jobs/quarter_hourly/delete_raw_dump.py
@@ -1,5 +1,5 @@
 """
-Hourly job to delete raw JobServer database dumps after their retention period.
+QuarterHourly job to delete raw JobServer database dumps after their retention period.
 
 Raw database dumps may contain production personal data and other sensitive
 information. They are created manually using the dump_raw_data management
@@ -9,22 +9,22 @@ Personal Data Copying Policy.
 The raw dump needs to remain available for long enough for an authorised
 developer to download it, but it should not remain on the production server
 indefinitely. This job limits that exposure by deleting the dump once it has
-existed for at least one hour.
+existed for at least 15 minutes.
 
 The age of the dump is determined from its modification time. Checking its age,
 rather than deleting any dump that exists, prevents a newly created dump from
-being deleted by an hourly run before it can be downloaded.
+being deleted by an quarter hourly run before it can be downloaded.
 
-Because this job runs once an hour, a dump will normally be deleted between one
-and two hours after it is created. A missing dump is expected and is not treated
-as an error.
+Because this job runs once every 15 minutes, a dump will normally be deleted
+between 15-30 minutes after it is created. A missing dump is expected and is not
+treated as an error.
 """
 
 from datetime import UTC, datetime, timedelta
 
 import structlog
 from django.conf import settings
-from django_extensions.management.jobs import HourlyJob, JobError
+from django_extensions.management.jobs import JobError, QuarterHourlyJob
 from sentry_sdk.crons.decorator import monitor
 
 from services.sentry import monitor_config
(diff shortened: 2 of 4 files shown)


jobserver/jobs/hourly/delete_raw_dump.py
```diff
@@ -0,0 +1,72 @@
...
+
+Raw database dumps may contain production personal data and other sensitive
+information. They are created manually using the dump_raw_data management
+command when access has been agreed and recorded in accordance with the
+Personal Data Copying Policy.
+
+The raw dump needs to remain available for long enough for an authorised
+developer to download it, but it should not remain on the production server
+indefinitely. This job limits that exposure by deleting the dump once it has
+existed for at least one hour.
```
mikerkelly: Wording, optional: Instead of "should not remain indefinitely", I would rather say that it "should be deleted from the production server by the developer that created it as soon as it is no longer needed for that purpose."

As you note later this job *does not* delete the dump as soon as it ages to one hour as you state here, due to the lag in scheduled runs of this job. So instead of the last sentence, I would rather say "This job acts as a backstop to delete dumps within a certain period, in case of manual error". Then the detail of how long the file might live is only in the later discussion rather than needing to repeat it here.

jobserver/jobs/hourly/delete_raw_dump.py
```diff
@@ -0,0 +1,72 @@
+"""
+Hourly job to delete raw JobServer database dumps after their retention period.
```
mikerkelly: Terminology: A "retention period" without qualification could imply that something _should_ be retained for that long. We mean the opposite, it should be deleted as soon as no longer needed, by the developer, and in any case no longer than the period enforced in this job, as an automated backstop.

And as you note later how this job and its timing interact with the minimum and maximum age of the dump (assuming the developer did not delete it) needs some care to think about and is tricky to express really concisely.

Suggestion: "Delete raw JobServer database dumps that have reached a certain age." the docstring initial topic sentence just concisely and factually states what the script *does*. The context and the *why* are in the later discussion.

jobserver/jobs/quarter_hourly/delete_raw_dump.py
```diff
@@ -0,0 +1,72 @@
+"""
+Hourly job to delete raw JobServer database dumps after their retention period.
+
+Raw database dumps may contain production personal data and other sensitive
+information. They are created manually using the dump_raw_data management
+command when access has been agreed and recorded in accordance with the
+Personal Data Copying Policy.
```
mikerkelly: Wording suggestion: I think we should at least mention the existence of scrubbed dumps here, if we are setting the context for the raw dumps. That's the normal thing, raw dumps are the exception, and understanding that is important to understanding the reason this job exists.

Maybe "Usually developers should use the scrubbed database dump if they need near-realistic production data for use in their development process. Some investigations or tasks may need to use the raw database dump in exceptional circumstances. They are created manually using the dump_raw_data management
command when access has been agreed and recorded in accordance with the Personal Data Copying Policy."

jobserver/jobs/hourly/delete_raw_dump.py
```diff
@@ -0,0 +1,72 @@
...
+from django.conf import settings
+from django_extensions.management.jobs import HourlyJob, JobError
+from sentry_sdk.crons.decorator import monitor
+
+from services.sentry import monitor_config
+
+
+logger = structlog.get_logger(__name__)
+
+RAW_DUMP_RETENTION_PERIOD = timedelta(hours=1)
```
mikerkelly: Behaviour: As noted above, this timedelta could be shorter, 1h to download a ~150MB file is excessive.

Naming: as noted above "retention period" can be a misleading phrase, here. Suggestion: `DELETE_AFTER_AGE`.

jobserver/jobs/hourly/delete_raw_dump.py
```diff
@@ -0,0 +1,72 @@
...
+indefinitely. This job limits that exposure by deleting the dump once it has
+existed for at least one hour.
+
+The age of the dump is determined from its modification time. Checking its age,
+rather than deleting any dump that exists, prevents a newly created dump from
+being deleted by an hourly run before it can be downloaded.
+
+Because this job runs once an hour, a dump will normally be deleted between one
+and two hours after it is created. A missing dump is expected and is not treated
+as an error.
```
mikerkelly: Wording: Perhaps this will be clearer if we discuss both variables and their consequences together.

Speaking about missing dumps is a separate point that belongs in a separate paragraph. And it is not really "missing" if we don't necessarily expect it to be there.

> The age of the dump is determined from its modification time. Running every [script frequency] and deleting dumps older than [delete_after_age] allows at least [script frequency] to download the dump while ensuring the dump is deleted within [job frequency + delete_after_age].

> The dump will not exist most of the time, so the dump not existing is not an error.

jobserver/jobs/quarter_hourly/delete_raw_dump.py
```diff
@@ -0,0 +1,72 @@
...
+from django_extensions.management.jobs import HourlyJob, JobError
+from sentry_sdk.crons.decorator import monitor
+
+from services.sentry import monitor_config
+
+
+logger = structlog.get_logger(__name__)
+
+RAW_DUMP_RETENTION_PERIOD = timedelta(hours=1)
+"""Minimum period for which a raw dump remains available before deletion."""
```
mikerkelly: Wording: again, I think this will be simpler and more to the point if we just say what it does: "Delete files older than this when the job runs".

jobserver/jobs/quarter_hourly/delete_raw_dump.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class Job(HourlyJob):
+    help = "Delete raw database dumps after the retention period to limit the availability of production data"
```
mikerkelly: Wording: again, let's avoid talking about retention periods and exactly when the deletion happens. "Delete raw database dumps of a certain age" is to the point.

jobserver/jobs/quarter_hourly/delete_raw_dump.py
```diff
@@ -0,0 +1,72 @@
...
+            return
+
+        if not raw_dump_path.is_file():
+            raise JobError(f"Raw database dump path is not a file: {raw_dump_path}")
+
+        # Use the file's last-modified time as an approximation of when the dump
+        # finished being written, and convert it to UTC datetime.
+        modified_at = datetime.fromtimestamp(raw_dump_path.stat().st_mtime, tz=UTC)
+        # Calculate how long the raw dump has been available since it was last modified.
+        dump_age = datetime.now(UTC) - modified_at
```
mikerkelly: Question: do we need a `datetime`, `timedelta` and timezones here at all? They are for handling dates and timezone-aware times, which don't seem relevant here. We only care about the age, which is a difference in time so can be expressed as an int.

`mtime` and `time.time` are UTC epoch seconds regardless of location, so is `dump_age = time.time() - raw_dump_path.stat().st_mtime` sufficient to give an age in seconds that we can compare with something?

jobserver/jobs/quarter_hourly/delete_raw_dump.py
```diff
@@ -0,0 +1,72 @@
...
+
+        if dump_age < RAW_DUMP_RETENTION_PERIOD:
+            logger.info(
+                "Raw database dump is still within its retention period",
```
mikerkelly: Wording suggestion: "Raw database dump not deleted as it had not reached deletion age". And it could be helpful to include the deletion age in the log kwargs.

jobserver/jobs/quarter_hourly/delete_raw_dump.py
```diff
@@ -0,0 +1,72 @@
...
+                age_seconds=dump_age.total_seconds(),
+                path=raw_dump_path,
+            )
+            return
+
+        raw_dump_path.unlink(missing_ok=True)
+        logger.info(
+            "Deleted raw database dump after its retention period",
```
mikerkelly: Wording suggestion: "Deleted raw database dump as it had reached deletion age". And it could be helpful to include the deletion age in the log kwargs.

mikerkelly (commented):
Thanks, Nishtha. This seems like it should do the job, from inspection. But I don't know how you have tested it, could you please let me know? Could we write some automated tests? Will we rely on manual testing and how will that be done? How will you test it in production and its Sentry configuration?

We can observe https://github.com/opensafely-core/job-server/pull/5984 in production while testing this perhaps.

Timing suggestion: we have two variables here, the wait between script runs, and the minimum age for deletion. We will delete the dump within (age, age+wait). I think that we should pick these values based on what we think is a reasonable time to download the dump (use that as age), then run the script frequently (it will be very cheap to run, and this reduces the maximum life of a dump). They do not need to be the same. Right now they are 1 hour and 1 hour, so an undeleted dump will live for 1-2 hours, which seems excessive.

How about a 15m age, and running the job as a MinutelyJob or QuarterHourlyJob? That would mean the dump is cleaned up within 15-16 minutes (or 15-30). 15m seems ample to download a file of about 150MB size (10MB/m....).

I have a number of other smaller comments inline, mostly about wording, one on how we handle time.

mikerkelly:
Thought: if we had a background tasks framework (Celery, or database back Django 6.0 Tasks API), we could just schedule a task for this. This works, though.

nishtha-kalra:
> Thanks, Nishtha. This seems like it should do the job, from inspection. But I don't know how you have tested it, could you please let me know? Could we write some automated tests? Will we rely on manual testing and how will that be done? How will you test it in production and its Sentry configuration?
> 
> We can observe #5984 in production while testing this perhaps.
> 
> Timing suggestion: we have two variables here, the wait between script runs, and the minimum age for deletion. We will delete the dump within (age, age+wait). I think that we should pick these values based on what we think is a reasonable time to download the dump (use that as age), then run the script frequently (it will be very cheap to run, and this reduces the maximum life of a dump). They do not need to be the same. Right now they are 1 hour and 1 hour, so an undeleted dump will live for 1-2 hours, which seems excessive.
> 
> How about a 15m age, and running the job as a MinutelyJob or QuarterHourlyJob? That would mean the dump is cleaned up within 15-16 minutes (or 15-30). 15m seems ample to download a file of about 150MB size (10MB/m....).
> 
> I have a number of other smaller comments inline, mostly about wording, one on how we handle time.

Hi Mike, thanks for the suggestions; they all sound good.
I manually tested the job locally by renaming `jobserver_scrubbed.dump` to `jobserver.dump`, so no raw production data was involved. I manually tested the cases where the dump was older than the retention period, newer than the retention period, and did not exist. Each case behaved as expected.
I did not initially add automated tests because there are currently no tests for the scheduled jobs. However, I agree that the deletion and timing logic is worth testing, so I am happy to add unit tests for this particular job covering various scenarios.
I originally made this an hourly job because that was what we agreed during the [DPV](https://docs.google.com/document/d/1S9wrACzWM3Ia31E7_fe3tLKUfnObaX5iV4F1Ludxu-U/edit?tab=t.9vfaa2z6eiju#heading=h.hb2qud9cunho). However, I agree that a dump created just after an hourly run could remain available for almost two hours. I am therefore happy to change it to a 15-minute retention period and a quarter-hourly job, meaning the dump will be deleted within approximately 15–30 minutes. This will require a new quarter_hourly job directory, use of `QuarterHourlyJob`, and a new [app.json](url) entry running `python manage.py runjobs quarter_hourly` with the schedule '*/15 * * * *`. I will also update the Sentry monitor configuration to use the same schedule. Does this make sense?

mikerkelly:
Yes, please go ahead.

jobserver/jobs/hourly/delete_raw_dump.py
nishtha-kalra: With the above suggested changes in the module docstring, I am inclined towards the name `RAW_DUMP_AGE_LIMIT`. Thoughts?

nishtha-kalra:
> Thought: if we had a background tasks framework (Celery, or database back Django 6.0 Tasks API), we could just schedule a task for this. This works, though.

That’s a good idea. Since we don't currently have background-task infrastructure, perhaps we could add a data-scrubbing ticket for a spike to investigate it? (explore various options and if JobServer would benefit from it?)

mikerkelly:
> > Thought: if we had a background tasks framework (Celery, or database back Django 6.0 Tasks API), we could just schedule a task for this. This works, though.
> 
> That’s a good idea. Since we don't currently have background-task infrastructure, perhaps we could add a data-scrubbing ticket for a spike to investigate it? (explore various options and if JobServer would benefit from it?)

It would help with a lot of things, but I think we probably won't find time to look at it until we have done the Django 6.0+ upgrade, which is not imminent, I think. Doing the upgrade should be a trigger for the team to think about how we can use its new features.

jobserver/jobs/quarter_hourly/delete_raw_dump.py
```diff
@@ -0,0 +1,78 @@
+"""
+Quarter-hourly job to delete raw JobServer database dumps that have reached a certain
+age.
+
+Usually developers should use the scrubbed database dump if they need near-realistic
+production data for use in their development process. Some investigations or  tasks
```
mikerkelly: Typo: double space in "or  tasks".

jobserver/jobs/quarter_hourly/delete_raw_dump.py
```diff
@@ -0,0 +1,78 @@
...
+may need to use the raw database dump in exceptional circumstances. They are created
+manually using the dump_raw_data management command when access has been  agreed and
```
mikerkelly: Typo: double space in "been  agreed".

jobserver/jobs/hourly/delete_raw_dump.py
mikerkelly: Sounds good to me.

tests/unit/jobserver/jobs/quarter_hourly/test_delete_raw_dump.py
```diff
@@ -0,0 +1,69 @@
...
+
+    delete_raw_dump.Job().execute()
+
+    assert not raw_dump_path.exists()
+    assert (
+        log_output.entries[0]["event"]
+        == "Deleted raw database dump as it had reached deletion age"
+    )
+    assert log_output.entries[0]["dump_age"] == dump_age
+    assert log_output.entries[0]["path"] == raw_dump_path
```
mikerkelly, line 70: Completeness: in the previous test you test for `assert log_output.entries[0]["deletion_age"] == delete_raw_dump.RAW_DUMP_AGE_LIMIT`, which is also relevant here. So should we do that here too?

tests/unit/jobserver/jobs/quarter_hourly/test_delete_raw_dump.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def test_does_nothing_when_raw_dump_does_not_exist(raw_dump_path, log_output):
+    delete_raw_dump.Job().execute()
+
+    assert not raw_dump_path.exists()
+    assert len(log_output.entries) == 1
+    assert log_output.entries[0]["event"] == "Raw database dump does not exist"
+    assert log_output.entries[0]["path"] == raw_dump_path
+
```
mikerkelly, line 39: Completeness: I think that there is a test missing here that should exercise the following behaviour:

```
        if not raw_dump_path.is_file():
            raise JobError(f"Raw database dump path is not a file: {raw_dump_path}")
```

Can you add one please?

mikerkelly (approved):
Thanks, Nishtha. Looks good.

I spotted a couple of typos, and a couple missing things in the tests. Approving subject to those.

mikerkelly:
I tested this in production on 25 August 2026. See [Slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1787649799832629).
That included https://github.com/opensafely-core/security/issues/56 / https://github.com/opensafely-core/job-server/pull/5987 and https://github.com/opensafely-core/security/issues/62 / https://github.com/opensafely-core/job-server/pull/5984. All working, once an issue with the bot not being invited to the Slack channel was fixed.

PR #5988, opened 2026-07-21
Stop sending Slack messages for releases

### LINKED ISSUE #5972: Stop sending releases Slack notifications
# Why are we doing this?

When many release files are released at the same time, one message is sent to #opensafely-releases per file in the release. This can lead to rate-limiting and some messages being dropped (maybe including important unrelated messages).

We don't think that we need the releases messages or the channel where they were sent, #opensafely-releases, anymore, since Airlock #opensafely-outputs does all the required notification on the Airlock side, and all the discussion is in that channel (since 2 years ago(!)). So let's stop sending all such messages when the release file APIs are hit, and archive #opensafely-releases.

<details>
<summary>Details</summary>

Sentry Issue: [JOB-SERVER-W4](https://ebm-datalab.sentry.io/issues/7465381484/?referrer=github_integration)

```
SlackApiError: The request to the Slack API failed. (url: https://slack.com/api/chat.postMessage)
The server responded with: {'ok': False, 'error': 'ratelimited'}
(1 additional frame(s) were not displayed)
...
  File "services/slack.py", line 31, in post
    client.chat_postMessage(channel=channel, text=text)
```

There is more context in Slack threads [here](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1758644051320999) and [here](https://bennettoxford.slack.com/archives/C0A8AJA249K/p1778867319757569)

> This is the releases API notifying #opensafely-releases for every file it releases
> just released 117 files which must have hit the rate limit.
> Now that Airlock posts into #opensafely-outputs that it intends to upload files, and then that it's closed the release request issue when it's done, there's probably no need for it to try and notify for every file uploaded

Maybe this behaviour is no longer required and should be removed.

> [The API is hit once per file](https://github.com/opensafely-core/job-server/blob/cb098a67a518a0cdaf1644828a355e04dd7c8b5f/jobserver/api/releases.py#L358), and each hit synchronously triggers a Slack when the database has been updated.
> If managing all our requests to Slack became an issue in the future probably the solution is to bite the bullet and implement a task queue like Celery for background tasks like Slacks, emails, and potentially instead crons. That can handle load balancing retry and backoff well. But we may never need it.
> One scenario that seems problematic is, assuming this bot is shared between all the Job Server Slack service, it could get rate-limited for this reason while another Slack message is trying to send, such as for new project creation. Maybe that could lead to other important messages being dropped due to the large volume in this channel.

Asked for opinions in [Slack](https://bennettoxford.slack.com/archives/C03KD4LURV4/p1784039974936929), responses:

> I'm not aware of any real need for this channel anymore. It was useful when we released files manually as a way to check everything you expected to be released actually was, but I haven't worried about that since we've had Airlock
> I can't think of a reason not to get rid of this channel now

</details>

We no longer receive these errors due to large releases, either through stopping sending the updates, or some other fix.

- Remove the message sending from the API function?
- Batch or delay bursty use of the Slack API somehow? That could be a specific solution related to release files, a general solution in `services/slack.py` (simple internal queue in a database model?), or a very general solution to task management, like [Celery](https://docs.celeryq.dev/en/stable/getting-started/introduction.html) with a redis backend. Django 6.0 has a tasks API that can work with that and there is also a database-backed implementation.

# What are we doing?

✅ Confirming with the users of the #opensafely-releases channel whether these messages are still needed and removing or reducing them if not. Maybe we can send no message, only one per release. UPDATE: [Slack here](https://bennettoxford.slack.com/archives/C03KD4LURV4/p1784039974936929) -- no objections to stopping all of the releases messages and archiving the channel. DONE.

- Remove the slack messages from slacks.py
- Remove the code that called them from releases APIs
- Remove the associated tests and any documentation or other related stuff
- archive the channel, referencing this issue.

StevenMaude:
Fixes #5972.

The original problem was that these messages exceeded Slack's rate limit for the way we use Slack.

It turns out, after consultation, that the messages were not that useful. So a decision was made to remove the messages instead of addressing the rate limit.


StevenMaude:
A few thoughts:

* I opened https://github.com/opensafely-core/airlock/pull/1239 to tidy up [Airlock's repository entry for this Slack channel too](https://github.com/opensafely-core/airlock/blob/bcb1720adb522fbd7a8f4bca097c75fd9af8d77a/job-server/.env.jobserver.template#L13), after we merge this.
* I searched for `opensafely-releases` and `RELEASES_SLACK_CHANNEL` in our GitHub organizations, and the only other reference was in the copy of job-server in `bennettoxford` that referred to them.
  * I think it's *probably* OK to leave that, since it's an old copy. If it's resynced with this `main`, it would be updated.
* The `APPLICATIONS_SLACK_CHANNEL` setting actually references `RELEASES_SLACK_CHANNEL` in the environment for some reason that I'm not clear (mistaken copy-paste?).
  * ~We could tidy this up.~ *Now done.*
  * I don't think it's used any more since applications are no longer routed through job-server. There's also other code that could be removed following #5613, but that's presumably a deliberate choice to retain it.

mikerkelly:
Thanks, Steve. All looks good to me.

This removes an API, `ReleaseNotificationAPICreate` / `release-notifications`. Could you also search for references to that API and that URL fragment in case there are any to tidy up? I checked Slack [and the last related message](https://bennettoxford.slack.com/archives/C02K9CV63MX/p1644417004858009) was in 2022.


> I don't think it's used any more since applications are no longer routed through job-server. There's also other code that could be removed following #5613, but that's presumably a deliberate choice to retain it.

Yes, we did the minimum require to prepare applications in Job Server for non-COVID. There is a longer-term REX work item to archive the applications data and remove `applications` and all related code.

mikerkelly (approved):

StevenMaude:
> This removes an API, `ReleaseNotificationAPICreate` / `release-notifications`. Could you also search for references to that API and that URL fragment in case there are any to tidy up? I checked Slack [and the last related message](https://bennettoxford.slack.com/archives/C02K9CV63MX/p1644417004858009) was in 2022.

That's a good idea 👍🏻

I'll check for that too before merging this.

StevenMaude:
The only entries I could find relating to `ReleaseNotificationAPICreate` / `release-notifications` were in:

* `bennettoxford/os-schools` which is an old copy of this repository as mentioned
* `opensafely-core/osrelease` which is archived

StevenMaude:
Now merged opensafely-core/airlock#1239, so I think everything's done here.

PR #5994, opened 2026-07-23
Add `Project.Category` field

### LINKED ISSUE #5991: Add `category` field to `Project`, minimally
# Why are we doing this?

In #5681 we decided that projects should have categories and what the project categories should be. We should now add the field in a minimal way to the Project data model to expand on it later.

- Projects can be assigned a project category.
- All existing Projects in production have an empty category (for now).
- New Projects can still be created, and have an empty category.
- Existing Projects can still be edited and viewed (but category cannot be seen or changed in the UI, for now).

# What are we doing?

- Adding a `TextChoices` field to `Project` enumerating the possible categories. Provide an "empty" option.
- Adding a migration that assigns the "empty" category to all projects, or, optionally, assigns integer-identifier projects as "Legacy approved" and POS-identifier projects as "Approved".
  - Those without an identifier cannot be given a category until https://github.com/opensafely-core/job-server/issues/5702. We could wait until we are ready to do that to assign meaningful categories in another migration.
- Ensuring new projects can be created and have an appropriate category. Both data development and approved projects go through the same form, with INTERNAL- projects currently having no project identifier and approved projects having POS- identifier. If easy, auto assign category based on identifier. If not, okay to have NULL category (for now).
  - Changes to the project create actions (if required).
  - Check the project edit action (if required, don't allow changing category yet).
  - Check the project create form (don't allow setting the project category).

# What are we not doing in this issue

- #5702 
- #5681 
- Any frontend work apart from the above. That is #5989.
- ... and other issues blocked by those.

---

mikerkelly:
Fixes #5991.

Job Server projects may be created for both research and internal data and platform development purposes. The approval process for research projects has changed over the lifetime of the OpenSAFElY service. It is useful to represent the different categories of project explicitly in the model.

This can be merged before the service team approve the proposed set of categories and identifiers, because we know we want some kind of category, and tweaking them later is okay. This PR does not actually set any of them use category for anything yet, there are several follow on issues for that. See [proposal documentation](https://docs.google.com/document/d/1NYZCFARvXpGFCo1w8sx-7170hhMxPeRXV8YPhs-SYcg/edit?tab=t.xit0tt2q6ukg). Catherine confirmed ([Slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1785235967765319?thread_ts=1778498773.997489&cid=C069SADHP1Q))that it is okay to merge this PR if approved.

<details>
<summary>Details</summary>

We could use category to drive UI change, reporting and metrics, and potentially more. It also makes some logic based on the project identifier simpler. Once we have a category field, we need only set it when creating and editing Projects, then behaviour can be conditioned on `project.category`.

These are the project categories that we are agreeing with the service team. They can be tweaked or added to if required, so it is okay if they are not perfect yet.

Start with a the model change and a migration that sets all existing projects to have unknown category. This is simplest. We can use the category in later PRs, and fill in the semantically correct categories in a separate migration later.

Projects can still be created through the ProjectCreateForm, and will have the default category of "unknown". Projects can still be edited, but the category cannot be edited.

Call it "category" because that term is not used in the codebase and is less overloaded than "type". This makes it easier to search for and understand references to "category".
</details>

<details>
<summary>Screenshots - click to expand</summary>

### Adding a project - no change 

image

### Editing the project - no change 

image

### New and existing Projects in database shell have unknown category

```
>>> Project.objects.first()
<Project: 48 - Covid Vaccine Not Done>
>>> Project.objects.first().category
'unknown'
>>> Project.objects.last()
<Project: 132452 - new project with no category>
>>> Project.objects.last().category
'unknown'
>>> set(Project.objects.values_list('category', flat=True))
{'unknown'}
``` 
</details>


jobserver/models/project.py
```diff
@@ -39,6 +39,25 @@
...
+    LEGACY = (
+        "legacy",
+        "Legacy COVID research activity predating specific approval processes",
+    )
+    """Legacy COVID research activity that came before any approval process."""
+    LEGACY_APPROVED = "legacy_approved", "Legacy approved COVID research activity"
+    """Legacy COVID research activity that went through the Job Server-managed
+    approval process."""
+    APPROVED = "approved", "Approved non-COVID research activity"
+    """Research activity that went through the NHSE-managed approval process."""
```
KatieB5, line 56: For `legacy`, `approved` and `legacy_approved`, it seems like the values are trying to represent two different things: the research type (COVID vs non-COVID) and the approval process (no process/legacy, Job Server-managed, NHSE-managed). Is that your understanding too? If it is, is it OK for a single enum member to combine those concepts?

When I was reading through this I was thinking in future, what if we have a COVID-related research project approved via the NHSE-managed approval process, would we give this the category of `APPROVED`? In the [data development DPV](https://docs.google.com/document/d/1NYZCFARvXpGFCo1w8sx-7170hhMxPeRXV8YPhs-SYcg/edit?tab=t.xit0tt2q6ukg#heading=h.iyy1q7jngqr), we say that _There will be no future new legacy projects created_, so I think the answer is "yes". But if it is "yes", it doesn't feel like it aligns with the human-readable label `"Approved non-COVID research activity"`. 

Maybe the human-readable labels could match what we have in the docstrings? That way, the values underpinning the enum members only related to the approval process, not the research type. I'm not sure how the human-readable labels will be used in our UI, and I couldn't find any general examples in the [Django docs](https://docs.djangoproject.com/en/6.0/ref/models/fields/#enumeration-types). If it's important for them to represent the COVID/non-COVID distinction and the approval process, then we can leave them as they are.

mikerkelly: Thanks for the feedback, Katie.

> For legacy, approved and legacy_approved, it seems like the values are trying to represent two different things: the research type (COVID vs non-COVID) and the approval process (no process/legacy, Job Server-managed, NHSE-managed). Is that your understanding too? If it is, is it OK for a single enum to combine those concepts?

Correct, as described in the docstring of the enum these categories have two separate dimensions to them. I don't think that there is a general answer to whether an enum should reflect more than one dimension of variation. Or even more generally, what level of normalisation or denormalisation is most appropriate. It depends on the context and what you are going to use it for. We only have 4 categories and are not anticipating adding any more. This can represent the same information as two fields.

We could normalise this so that the internal/research dimension is a separate field but I am not sure what that will achieve right now, and it could be done later if we decide that is what we want, for example if we decided that we wanted to be able to distinguish types of internal project in more detail. That said, we should take care that the current design doesn't make it unnecessarily difficult to change things in future.

On the purpose of the field: I think categories are useful for driving UI display of information; downstream data analysis; (for example, in Grafana, excluding internal projects, analysis of only new or old approval process projects); defining and interpreting project identifiers (see the proposal in the DPV to normalise `category`, `number,` and `full identifier`); and possibly for future business rules based on category. I don't think normalising the two dimensions into one enum precludes us from any behaviour that would be possible with two fields. Some possible future uses will be more convenient, some less convenient. It is a judgement call where the balance lies, and I think we can change our minds later. I think most uses will test if a project is or is not in a specific category and that is not really easier or harder based on the normalisation level. It will be more annoying if there are subsets of project categories that we regularly want to reference together in a more complex way than that -- that would be a code smell that hints we should normalise futher.

To be more concrete, we know that the project identifier format has a relationship with the project category and it seems simpler to be able to represent that in code if category is one field rather than two, as in https://github.com/opensafely-core/job-server/pull/6003, for example, mapping the project categories to project identifier formats in a single data structure with projects as keys and regexes for the identifiers as values. That said, in future we could use a tuple-key with elements from separate "purpose" and "approval process" enums.

> When I was reading through this I was thinking in future, what if we have a COVID-related research project approved via the NHSE-managed approval process, would we give this the category of APPROVED? In the data development DPV, we say that There will be no future new legacy projects created, so I think the answer is "yes". And if it is "yes", it doesn't feel like it aligns with the human-readable label "Approved non-COVID research activity".

> Maybe the human-readable values could match what we have in the docstrings? That way, the values underpinning the ENUM members only related to the approval process, not the research type. I'm not sure how the human-readable labels will be used in our UI, and I couldn't find any general examples in the Django docs. If it's important for them to represent the COVID/non-COVID distinction, then we can leave them as they are.

I think the terminology is confusing here, but it is not something that REX have invented. Really we mean research authorised under [the NHS OpenSAFELY Data Analytics Service Pilot Directions 2025](https://digital.nhs.uk/about-nhs-digital/corporate-information-and-documents/directions-and-data-provision-notices/secretary-of-state-directions/nhs-opensafely-data-analytics-service-pilot-directions-2025), but that is a mouthful. Bennett has been referring to that as the "new directions" (maybe there is more than one?) and "non-COVID directions", but by that they mean that it is not restricted to COVID-research as [the 2020 directions](https://digital.nhs.uk/about-nhs-digital/corporate-information-and-documents/directions-and-data-provision-notices/secretary-of-state-directions/covid-19-public-health-directions-2020) were. That term does not intend that COVID research be excluded.

Referring to the NON-COVID taskforce workplan decision log number 2:

> All new projects will come through one application process whether COVID or Non-COVID.
> And all of these new projects will be expected to use only the three standard datasets.

I think you are right though that we should try to make this as easy to understand and precise as possible (those can be competing goals). How about:

```code
APPROVED = "approved", "Approved research activity under the 2025 or later Directions and the NHSE-managed approval process."""
"""Approved research activity under the 2025 or later Directions and the NHSE-managed approval process."""
```

The existing wording has already been approved by Catherine and is with the service team, but we can always change the wording to whatever comes out of that. We are not actually going to use category for anything other than #6003 until that approval. So I think it would be okay to merge this and amend the text or categories later. 

I could propose this wording to Catherine and update the proposal, also.

KatieB5: Thanks very much for coming back to me on this, M, and for providing the additional context from the wider discussions and work related to this PR. It was really helpful for understanding how the wording came to be as it is, how the category is expected to be used, and the thinking behind the current level of normalisation.

I think my slight preference would be to use the more comprehensive wording you suggest:
```
APPROVED = (
    "approved",
    "Approved research activity under the 2025 or later Directions and the NHSE-managed approval process",
)
```
To me, that's clearer than `"Approved non-COVID research activity"`, because as you've pointed out (and as decision log 2 says), the new approval process applies to both COVID and non-COVID research:

> "All new projects will come through one application process whether COVID or Non-COVID."

That said, I think the rationale you've given for representing the process and research-type dimensions as a single enum is reasonable. I can also see how keeping them together makes some of the expected use cases (such as mapping project categories to identifier formats) simpler, and as you say we can always normalise further later if we find ourselves regularly needing to reason about the two dimensions independently.

If the current wording has already been agreed with Catherine and the Service Team, I'm happy to go ahead with it as it is and revisit later if needed.

mikerkelly: Thanks, Katie. I asked about updating the wording to Catherine in https://bennettoxford.slack.com/archives/C069SADHP1Q/p1785845476695029?thread_ts=1778498773.997489&cid=C069SADHP1Q. The service team have said they want to defer a decision on this so I think it so not too late to change this. And quite likely it does need further updates after that anyhow.

I think I will update the wording in the PR and merge. If we have to update later, that is okay.

KatieB5 (approved):

jobserver/models/project.py
mikerkelly: Actually, that bit is not even part of the proposal, and as it is not visible in any part of the site (yet?) does not need approval from the service team.

mikerkelly: Ugh. I merged without rebasing these changes. I will add a new commit in https://github.com/opensafely-core/job-server/pull/6003.

PR #6022, opened 2026-07-30
Add new field to eventually allow more job request email notification options

eli-miriam:
This is the first step in #5771 - adding a new field that will eventually capture the notification setting for job requests.

The next phase will be wiring up this new field to handle all the logic for when to send email notifications, rendering the old field redundant; the last phase will be removing the old field.

I'm submitting these as separate PRs to avoid migration drift, and to keep the cognitive load of the middle (logic) PR as small as possible.

No testing done here apart from `just run` compiling the migrations successfully.


jobserver/models/job_request.py
```diff
@@ -129,6 +129,17 @@ class JobRequestStatus(models.TextChoices):
...
 
+class JobRequestNotifications(models.TextChoices):
+    """
+    The notification setting for this JobRequest, representing which emails will be sent to
+    the requesting researcher.
+    """
+
+    NONE = "none"
+    PER_JOB = "per_job"
+    PER_REQUEST = "per_request"
```
mikerkelly, line 140: Thought: nice, this is a lot more flexible than the single `will_notify` Boolean field that we have now. What happens though if a user wants both per-job and per-request notifications? We could add a "both" option, which would be okay if these were the only options. But I wonder how this scales if we wanted to add more options and allow multiple selection.

The user feedback [Slack thread](https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1776956468028129) and the issue mentioned further types of notification like "first job completes", "each failure" and we could imagine more along those lines. It would be reasonable to add extra choices for each of these, but if users wanted more than one of them at a time we now either have to add a category per allowed combination (and the possible combination grow rapidly), or rethink because this field type can only represent one choice from a list.

We might imagine eventually ending up with several checkboxes the user can choose between.

Given that we know the intended near-future direction on this, it could make sense to anticipate this and use a field type that can handle representing multiple choices cleanly. That might avoid having to deal with a migration to change field types. There are several options.

1. Have a `BooleanField` per type of notification that may be sent. This could work but potentially add quite a few fields that are only related through how we use them.
2. Have a whole other table representing notification types and a `ManyToMany` field to that. That could work, but is maybe more suitable for highly relational or dynamic data. We would have to find a way to get exactly the set of options in that table that we wanted (migrations or fixtures?).
3. Use a field that can represent a list. Django has [JSONField](https://docs.djangoproject.com/en/5.2/ref/models/fields/#jsonfield) which is usable, but also has support for the Postgres-specific [ArrayField](https://docs.djangoproject.com/en/6.0/ref/contrib/postgres/fields/#arrayfield) that is less general, specifically for lists, and has several useful functions for querying the elements in it, and I believe will play well wrapping a `TextField` with `choices`, and a `MultipleChoiceField` on the form.

jobserver/models/job_request.py
```diff
@@ -129,6 +129,17 @@ class JobRequestStatus(models.TextChoices):
     UNKNOWN_ERROR_CREATING_JOBS = "unknown_error_creating_jobs"
 
 
+class JobRequestNotifications(models.TextChoices):
```
mikerkelly, line 132: Suggestion: I find putting this class within `JobRequest` so it is reference by attribute as `JobRequest.Notifications` is a slightly cleaner pattern than baking in the model name to the class name. Sometimes the variable definition and import order forces you down one way or another.

We don't seem to be consistent on this in the codebase, but generally prefer a subclass compare `Job.Status`, `PublishRequest.decisions`, `ReleaseFileReview.Statuses`, `Shapshot.Statuses` et cetera with `JobRequestStatus`.

mikerkelly: Suggestion, naming: I think "notifications" is not quite the right part of speech here. This column does not record the notifications themselves, it records the users preference, setting, choice... You refer to "setting" in the docstring, so how about `JobRequest.NotificationSettings`? Or `JobRequest.NotifyOn`?

jobserver/models/job_request.py
```diff
@@ -151,6 +162,9 @@ class JobRequest(models.Model):
     sha = models.TextField()
     identifier = models.TextField(default=new_id, unique=True)
     will_notify = models.BooleanField(default=False)
+    notification = models.TextField(
+        default=JobRequestNotifications.NONE, choices=JobRequestNotifications
```
mikerkelly, line 166: Does this mean all existing job requests will have `NONE` for this setting? Will there be migration to update those? That matters for in-flight requests when we start doing something with this.

mikerkelly:
Thanks, Eli. Seems like a step in the right direction. I have some questions and suggestions about some of the specifices.

PR #6026, opened 2026-08-03
Fix react router navigation on browser forwards/back

tomodwyer:
In `FileList` the synchronisation effect runs only when the file query succeeds, so later when `location.pathname` changes from `history.back()` the `selectedFile` is not updated.

This PR adds a test to check for that issue, and then resolves the issue by changing the `useEffect` dependencies.


KatieB5 (approved):
Looks good, thanks for jumping on a call!

PR #6030, opened 2026-08-04
Update default `restore-db` arg

mikerkelly:
This recipe is expected to be used with this value most often, so we may as well make that the default. This is the scrubbed data that developers should be using locally most of the time, jobserver.dump is the default name for the rawdump that should only be used exceptionally.

I did not bother updating DEVELOPERS.md for this as the version specified there still works and illustrates how to use the parameter.

See ADRs on data scrubbing and `dump_scrubbed_db` for more information.


tomodwyer (approved):
Neat

Issue #6033, opened 2026-08-05
Enable newly enabled Ruff v0.16 rules that we currently ignore


StevenMaude:
# Why are we doing this?

Ruff v0.16 has new rules that we currently fail validation on.

# How will we know when it's done?

- [x] When update-dependencies action PRs can succeed
- [ ] We've enabled all the new Ruff rules.

# What are we doing?

* ~pin Ruff to v0.15 (probably least ideal)~
* set ruleset exclusions that make sense
* fix all the remaining issues with the v0.16 default ruleset

StevenMaude:
The pragmatic thing to do is exclude the failing rules, and then gradually fix them. There are quite a lot of small changes needed across the codebase, and this minimises dealing with merge conflicts.

StevenMaude:
Following #6034, it leaves:

* [`BLE001`](https://docs.astral.sh/ruff/rules/blind-except/); using catch-all exceptions
* [`TRY002`](https://docs.astral.sh/ruff/rules/raise-vanilla-class/); not raising custom exceptions
* [`RUF012`](https://docs.astral.sh/ruff/rules/mutable-class-default/); mutable class variables that Django uses extensively, which will need some path/file exemptions, and maybe some fixes to non-Django Python code

StevenMaude:
I've moved the PR to review, and moving this issue back to next as the last three rules could be tackled by someone else now. It could also go in the deck scrubbing pile; we'll get most of the benefit of the new rules once #6034 is merged.

PR #6034, opened 2026-08-05
Make code compliant with most new default Ruff v0.16 rules

StevenMaude:
See #6033.

This PR fixes our code to be compliant with most of the new default rules in Ruff v0.16.

The remaining new rules not yet enabled are listed in [a comment in #6033](https://github.com/opensafely-core/job-server/issues/6033#issuecomment-5207296503).

Many of these fixes are the autogenerated ones by Ruff. A few needed manual review and editing of the automated fixes, or more extensive intervention. In entirely automated cases, I reviewed the fixed diffs.

### DIFF
--- a/airlock/views.py
+++ b/airlock/views.py
@@ -36,7 +36,7 @@ class EventType(Enum):
     REQUEST_RETURNED = "request returned"
     REQUEST_RESUBMITTED = "request resubmitted"
     REQUEST_PARTIALLY_REVIEWED = "request reviewed"
-    REQUEST_REVIEWED = "request reviewed"
+    REQUEST_REVIEWED = "request reviewed"  # noqa: PIE796
 
     def status_label(self):
         """The GitHub Issue label that should be added for this request"""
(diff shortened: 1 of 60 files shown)


StevenMaude:
This doesn't cover quite all of the rules, but the last remaining three rules are a little trickier since they cover:

* mutable class variables, which needs a bit of review to see where we do or don't want to apply this
* exception creation and handling, which also needs a bit of thought throughout the code if we change these

Since these changes apply across in various parts of the source, it's preferable to this chunk earlier and avoid having to fix conflicts.

airlock/views.py
```diff
@@ -93,8 +93,6 @@ def from_payload(cls, data):
             org = ORG_OUTPUT_CHECKING_REPOS[lookup]["org"]
             repo = ORG_OUTPUT_CHECKING_REPOS[lookup]["repo"]
 
-        workspace.project
```
StevenMaude, line 96: question: I think this is a remnant or half-finished notion. But are there any possible side effects of this that we intentionally want?

tomodwyer: I can't see any reason this code exists, it was added here: https://github.com/opensafely-core/job-server/pull/4372/changes#diff-2e3ca96974f398f9a2bcc0e1fb38d5a5d3c36eb8a438a0a4697df6f1e98faaa1R76 and I can't see it ever being used.

tomodwyer (approved):
Reviewed commit-by-commit and everything makes sense. Tests and linting is passing so I'm happy to go ahead.

